PR #813, opened 2022-02-25
fix(admin): only compare against actual versions when they are present and fallback to using desired versions if not

machi1990:
## Description

Follows up https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/789 after accidentally deleting the branch, which caused that PR to be closed. Once I force pushed, the Re-Open PR button was grayed out so opening a new PR. 

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -706,17 +706,21 @@ func (k *kafkaService) VerifyAndUpdateKafkaAdmin(ctx context.Context, kafkaReque
 		return errors.New(errors.ErrorValidation, fmt.Sprintf("Unable to update kafka: %s with strimzi version: %s", kafkaRequest.ID, kafkaRequest.DesiredStrimziVersion))
 	}
 
-	if kafkaRequest.ActualKafkaIBPVersion != "" {
-		vCompOldNewIbp, eIbp := api.CompareBuildAwareSemanticVersions(kafkaRequest.ActualKafkaIBPVersion, kafkaRequest.DesiredKafkaIBPVersion)
+	currentIBPVersion := kafkaRequest.ActualKafkaIBPVersion
 
-		if eIbp != nil {
-			return errors.New(errors.ErrorValidation, fmt.Sprintf("Unable to compare actual ibp version: %s with desired ibp version: %s", kafkaRequest.ActualKafkaIBPVersion, kafkaRequest.DesiredKafkaVersion))
-		}
+	if currentIBPVersion == "" { // the ibp version could be empty is the Kafka have not yet been successful provisioned at first place
+		currentIBPVersion = kafkaRequest.DesiredKafkaIBPVersion // use the desired version for checks as eventually the actual version will be this one for new Kafkas
+	}
 
-		// actual ibp version cannot be greater than desired ibp version (no downgrade allowed)
-		if vCompOldNewIbp > 0 {
-			return errors.New(errors.ErrorValidation, fmt.Sprintf("Unable to downgrade kafka: %s ibp version: %s to a lower version: %s", kafkaRequest.ID, kafkaRequest.DesiredKafkaIBPVersion, kafkaRequest.ActualKafkaIBPVersion))
-		}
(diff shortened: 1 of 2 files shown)


machi1990:
@pb82 @miguelsorianod would you have sometime to look at the state of the PR? The subject about the problematic Kafka has come up on several occasions this week. It will be good if we can get this PR in and remove the Kafka in question.

internal/kafka/internal/services/kafka.go
```diff
@@ -706,15 +706,21 @@ func (k *kafkaService) VerifyAndUpdateKafkaAdmin(ctx context.Context, kafkaReque
...
 	}
 
-	vCompOldNewIbp, eIbp := api.CompareBuildAwareSemanticVersions(kafkaRequest.ActualKafkaIBPVersion, kafkaRequest.DesiredKafkaIBPVersion)
+	currentIBPVersion := kafkaRequest.ActualKafkaIBPVersion
+
+	if currentIBPVersion == "" { // the ibp version could be empty is the Kafka have not yet been successful provisioned at first place
+		currentIBPVersion = kafkaRequest.DesiredKafkaIBPVersion // use the desired version for checks as eventually the actual version will be this one for new Kafkas
+	}
+
+	vCompOldNewIbp, eIbp := api.CompareBuildAwareSemanticVersions(currentIBPVersion, kafkaRequest.DesiredKafkaIBPVersion)
```
JameelB, line 715: Do we really need to compare the versions if there is no actual version set? Can we skip the comparison instead?

JameelB (commented):
Looks good to me, just have one question :)

internal/kafka/internal/services/kafka.go
machi1990: Thanks for the review @JameelB 
The idea of falling back to using desired version was to ensure that at least the new version on the upgrade matched the previous desired one to reduce chances of it be a breaking change.

internal/kafka/internal/services/kafka.go
```diff
@@ -728,15 +734,20 @@ func (k *kafkaService) VerifyAndUpdateKafkaAdmin(ctx context.Context, kafkaReque
 		return errors.New(errors.ErrorValidation, fmt.Sprintf("Unable to update kafka: %s ibp version: %s with kafka version: %s", kafkaRequest.ID, kafkaRequest.DesiredKafkaIBPVersion, kafkaRequest.DesiredKafkaVersion))
 	}
 
-	vCompKafka, ek := api.CompareSemanticVersionsMajorAndMinor(kafkaRequest.ActualKafkaVersion, kafkaRequest.DesiredKafkaVersion)
+	currentKafkaVersion := kafkaRequest.ActualKafkaVersion
+	if currentKafkaVersion == "" { // the kafka version could be empty is the Kafka that have not yet been successful provisioned at first place
+		currentKafkaVersion = kafkaRequest.DesiredKafkaVersion // use the desired version for checks as eventually the actual version will be this one for new Kafkas
+	}
```
ziccardi, line 740: We have code like this often in the code.
WDYT about creating an utility function like this:
```go
func OneOf(values ...string) (string, error) {
	for _, val := range values {
		if val != "" {
			return val, nil
		}
	}
	return "", fmt.Errorf("No value found")
}
```

so that previous code would become:
```go
currentKafkaVersion, _ := utils.OneOf(kafkaRequest.ActualKafkaVersion, kafkaRequest.DesiredKafkaVersion)
```

internal/kafka/internal/services/kafka.go
JameelB: ah that makes sense, thanks!

internal/kafka/internal/services/kafka.go
machi1990: That sounds good, reminds me of [lodash#get](https://lodash.com/docs/#get)

It is something I won't consider adding this to this PR but it's worth a discussion in the technical discussion call and see if it is something that can be widely adopted and added as a pattern. WDYT?

ziccardi: @machi1990 agree

JameelB (approved):

PR #829, opened 2022-03-04
feat: separate ingress and egress

pb82:
## Description
Provide separate fields for ingress and egress.

## Verification Steps

* CI tests should pass


config/kafka-sizes-configuration.yaml
```diff
@@ -0,0 +1,41 @@
+# This configuration defines a list of supported Kafka instance profile and their sizes for KAS Fleet Manager.
```
pawelpaszki, line 1: I think this file should be removed after recent naming changes

pawelpaszki: looks, like proper rebase is required

PR #837, opened 2022-03-08
implements kafka plan param with capacity validation

pawelpaszki:
## Description
- Accept plan parameter in the create Kafka endpoint
- Accept size parameter in the create Kafka endpoint
- Add capacity information to Kafka requests database table
- Update capacity check when creating a new cluster

- [ ] Code Review completed


internal/kafka/internal/handlers/cloud_providers.go
```diff
@@ -78,6 +80,8 @@ func (h cloudProvidersHandler) ListCloudProviderRegions(w http.ResponseWriter, r
 						maxCapacityReached := true
 						kafka.InstanceType = instType
 						kafka.Region = cloudRegion.Id
+						size, _ := h.kafkaConfig.GetFirstAvailableSize(instType)
```
pb82, line 83: why ignore the error here?

internal/kafka/internal/handlers/kafka.go
```diff
@@ -42,10 +47,34 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
...
 			ValidateKafkaClaims(ctx, &kafkaRequest, convKafka),
 			ValidateCloudProvider(&h.service, convKafka, h.providerConfig, "creating kafka requests"),
 			handlers.ValidateMultiAZEnabled(&kafkaRequest.MultiAz, "creating kafka requests"),
+			func() *errors.ServiceError { // Validate plan
+				instanceType, err := h.service.DetectInstanceType(convKafka)
+				if err != nil {
+					return err
+				}
+				if !stringNotSet(&kafkaRequest.Plan) {
+					p := strings.Split(kafkaRequest.Plan, ".")
```
pb82, line 61: we can also `Trim` it, just in case

internal/kafka/internal/handlers/kafka.go
```diff
@@ -42,10 +47,34 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
...
+					_, e := h.kafkaConfig.GetKafkaInstanceSize(instanceType.String(), p[1])
```
pb82, line 62: there might not be a `p[1]` leading to a nil pointer exception. Better check for array length before this.

internal/kafka/internal/handlers/cloud_providers.go
pawelpaszki: in theory an error should never occur. not sure what to do if it did in this case, but I will try to add error handing here

internal/kafka/internal/handlers/kafka.go
```diff
@@ -42,10 +47,34 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
...
 			ValidateKafkaClusterNameIsUnique(&kafkaRequest.Name, h.service, r.Context()),
 			ValidateKafkaClaims(ctx, &kafkaRequest, convKafka),
 			ValidateCloudProvider(&h.service, convKafka, h.providerConfig, "creating kafka requests"),
 			handlers.ValidateMultiAZEnabled(&kafkaRequest.MultiAz, "creating kafka requests"),
+			func() *errors.ServiceError { // Validate plan
+				instanceType, err := h.service.DetectInstanceType(convKafka)
+				if err != nil {
+					return err
+				}
+				if !stringNotSet(&kafkaRequest.Plan) {
```
pb82, line 60: double negation. Maybe better `stringSet(&kafkaRequest.Plan)` or `kafkaRequest.Plan != ""`?

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -89,11 +92,16 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
 		return nil, errf
 	}
 
+	reqSize, e := strconv.Atoi(strings.TrimPrefix(kafka.SizeId, "x"))
```
pb82, line 95: you could also do something like this:
```
type KafkaSize string

func (s KafkaSize) ToInt() {
  return strconv.Atoi(strings.TrimPrefix(s, "x"))
}
```

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -89,11 +92,16 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
...
+	if e != nil {
```
pb82, line 96: should we at least write a log message here?

internal/kafka/internal/services/util.go
```diff
@@ -72,3 +73,11 @@ func BuildKeycloakClientNameIdentifier(kafkaRequestID string) string {
 func BuildCustomClaimCheck(kafkaRequest *dbapi.KafkaRequest) string {
 	return fmt.Sprintf("@.rh-org-id == '%s'|| @.org_id == '%s'", kafkaRequest.OrganisationId, kafkaRequest.OrganisationId)
 }
+
+func ParseSize(sizeId string) (int64, error) {
+	size, err := strconv.ParseInt(strings.TrimPrefix(sizeId, "x"), 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	return size, nil
```
JameelB, line 82: Same as comment above, instead of parsing the size from the id let's use the `CapacityConsumed` property in the config.

internal/kafka/internal/handlers/kafka.go
```diff
@@ -42,10 +47,34 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
 			handlers.ValidateAsyncEnabled(r, "creating kafka requests"),
 			handlers.ValidateLength(&kafkaRequest.Name, "name", &handlers.MinRequiredFieldLength, &MaxKafkaNameLength),
 			ValidKafkaClusterName(&kafkaRequest.Name, "name"),
+			// ValidatePlan(h.service, h.kafkaConfig, &kafkaRequest.Plan, convKafka),
```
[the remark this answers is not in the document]
pawelpaszki: no. this line should be removed. good spot!

internal/kafka/internal/config/kafka.go
```diff
@@ -86,3 +86,23 @@ func (c *KafkaConfig) ReadFiles() error {
 func (c *KafkaConfig) Validate(env *environments.Env) error {
 	return c.SupportedInstanceTypes.Configuration.validate()
 }
+
+func (c *KafkaConfig) GetFirstAvailableSize(instanceType string) (string, error) {
+	kafkaInstanceType, err := c.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(instanceType)
+	if err != nil {
+		return "", err
+	}
+	return kafkaInstanceType.Sizes[0].Id, nil
```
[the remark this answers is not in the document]
pawelpaszki: @pb82 - do you have any opinion here?

internal/kafka/internal/handlers/kafka.go
ziccardi: If our standard will be that `plan` will always be in form of `instanceType.sizeid`, I believe it would be better to create a `Plan` type with two methods: `GetInstanceType` and `GetSizeId`

```go
type Plan string

func(p Plan) GetInstanceType() {
	...
}

func (p Plan) GetSizeID() {
	...
}
```

internal/kafka/internal/config/kafka.go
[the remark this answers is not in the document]
ziccardi: I believe they must be ordered. See here: https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/42/files#diff-4a09f9937fd7e06026f7cfe0e7d4143d202388410486747c30750de03aca4a26R180

internal/kafka/internal/config/kafka.go
```diff
@@ -86,3 +86,23 @@ func (c *KafkaConfig) ReadFiles() error {
...
+}
+
+func (c *KafkaConfig) GetKafkaInstanceSize(instanceType, sizeId string) (KafkaInstanceSize, error) {
```
ziccardi, line 98: I believe this should return a `*KafkaInstanceSize` to avoid copying the object too many times

internal/kafka/internal/config/kafka.go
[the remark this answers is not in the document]
JameelB: Thanks @ziccardi! didn't see that one there :) We'll just need to make a note of this in the SOP and docs :+1:

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -128,3 +128,21 @@ func NewKafkaSupportedInstanceTypesConfig() *KafkaSupportedInstanceTypesConfig {
...
+}
+
+func (kp *KafkaInstanceType) GetKafkaInstanceSizeByID(instanceType, sizeId string) (KafkaInstanceSize, error) {
+	for _, size := range kp.Sizes {
+		if size.Id == sizeId {
+			return size, nil
+		}
+	}
+	return KafkaInstanceSize{}, fmt.Errorf("Kafka instance size id: '%s' not found for '%s' instance type", sizeId, instanceType)
+}
```
ziccardi, line 148: This would be better as:
```go
func (kp *KafkaInstanceType) GetKafkaInstanceSizeByID(instanceType, sizeId string) (*KafkaInstanceSize, error) {
	for _, size := range kp.Sizes {
		if size.Id == sizeId {
                         ret := size // copy the object to avoid exposing mutable internals
			return &ret, nil
		}
	}
	return nil, fmt.Errorf("Kafka instance size id: '%s' not found for '%s' instance type", sizeId, instanceType)
}
```

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -128,3 +128,21 @@ func NewKafkaSupportedInstanceTypesConfig() *KafkaSupportedInstanceTypesConfig {
...
 }
+
+func (s *SupportedKafkaInstanceTypesConfig) GetKafkaInstanceTypeByID(instanceType string) (KafkaInstanceType, error) {
+	for _, t := range s.SupportedKafkaInstanceTypes {
+		if t.Id == instanceType {
+			return t, nil
+		}
+	}
+	return KafkaInstanceType{}, fmt.Errorf("Unable to find kafka instance type for '%s'", instanceType)
+}
```
ziccardi, line 139: Although we are returning a copy of the object, since `KafkaInstanceType` contains an array, only the array address is copied (shallow copy), thus exposing mutable internals to the caller. A better approach would be to make a deep copy.

Finally, would be better to return a pointer to avoid instantiating useless objects and reduce the number of copy operations:

```go
func (s *SupportedKafkaInstanceTypesConfig) GetKafkaInstanceTypeByID(instanceType string) (*KafkaInstanceType, error) {
	for _, t := range s.SupportedKafkaInstanceTypes {
		if t.Id == instanceType {
			ret := t // or, better, make a deep copy
			return &ret, nil
		}
	}
	return nil, fmt.Errorf("Unable to find kafka instance type for '%s'", instanceType)
}
```

internal/kafka/internal/services/kafka.go
```diff
@@ -214,13 +235,6 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
...
 	kafkaRequest.ID = api.NewID()
 
-	instanceType, err := k.DetectInstanceType(kafkaRequest)
-	if err != nil {
-		return err
-	}
-
-	kafkaRequest.InstanceType = instanceType.String()
-
 	hasCapacity, err := k.HasAvailableCapacityInRegion(kafkaRequest)
```
ziccardi, line 238: I'm confused. Why this has been deleted?

pawelpaszki: because instance type is assigned in the handler

ziccardi: Got it. However if this method is not used to detect the instance type anymore, it should be renamed

pawelpaszki: renamed

internal/kafka/internal/handlers/validation.go
```diff
@@ -95,18 +95,18 @@ func ValidateCloudProvider(kafkaService *services.KafkaService, kafkaRequest *db
...
 			return errors.FieldValidationError("Failed to update Kafka Request. Expecting at least one of the following fields: strimzi_version, kafka_version, kafka_ibp_version or kafka_storage_size to be provided")
 		}
 		return nil
 	}
 }
 
-func stringNotSet(value *string) bool {
-	return value == nil || len(*value) < 1
+func stringSet(value *string) bool {
+	return value != nil && len(*value) > 0
```
ziccardi, line 109: A string composed of just blanks is considered 'set'? If not, I believe we should trim the string before checking the length

pawelpaszki: fixed

internal/kafka/internal/config/kafka.go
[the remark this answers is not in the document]
pb82: Yes, they need to be ordered. The default size always needs to be x1.

internal/kafka/internal/config/kafka.go
```diff
@@ -86,3 +89,22 @@ func (c *KafkaConfig) ReadFiles() error {
...
+	}
+	return kafkaInstanceType.Sizes[0].Id, nil
+}
+
+func (c *KafkaConfig) GetKafkaInstanceSize(instanceType, sizeId string) (*KafkaInstanceSize, error) {
+	kafkaInstanceType, err := c.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(instanceType)
+	if err != nil {
+		return nil, err
+	}
+	return kafkaInstanceType.GetKafkaInstanceSizeByID(instanceType, sizeId)
```
ziccardi, line 109:
```suggestion
	return kafkaInstanceType.GetKafkaInstanceSizeByID(sizeId)
```

We are calling a method of `KafkaInstanceType`, we don't need to pass the `instanceType`

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -128,3 +130,45 @@ func NewKafkaSupportedInstanceTypesConfig() *KafkaSupportedInstanceTypesConfig {
...
+
+func (kp *KafkaInstanceType) GetKafkaInstanceSizeByID(instanceType, sizeId string) (*KafkaInstanceSize, error) {
+	for _, size := range kp.Sizes {
+		if size.Id == sizeId {
+			ret := size
+			return &ret, nil
+		}
+	}
+	return nil, fmt.Errorf("Kafka instance size id: '%s' not found for '%s' instance type", sizeId, instanceType)
+}
```
ziccardi, line 152:
```suggestion
func (kp *KafkaInstanceType) GetKafkaInstanceSizeByID(sizeId string) (*KafkaInstanceSize, error) {
	for _, size := range kp.Sizes {
		if size.Id == sizeId {
			ret := size
			return &ret, nil
		}
	}
	return nil, fmt.Errorf("Kafka instance size id: '%s' not found for '%s' instance type", sizeId, kp.Id)
}
```

internal/kafka/internal/handlers/kafka.go
```diff
@@ -46,6 +50,37 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
...
+					_, e2 := h.kafkaConfig.GetKafkaInstanceSize(instTypeFromPlan, size)
+
+					if e2 != nil {
+						return errors.NewWithCause(errors.ErrorGeneral, e2, fmt.Sprintf("Unsupported plan provided: '%s'", kafkaRequest.Plan))
+					}
+					convKafka.SizeId = size
+				} else {
+					rSize, err := h.kafkaConfig.GetFirstAvailableSize(instanceType.String())
+					if err != nil {
+						return errors.NewWithCause(errors.ErrorGeneral, err, "Unsupported kafka instance type: '%s' provided in the plan parameter: '%s'", instanceType.String(), kafkaRequest.Plan)
```
ziccardi, line 77: here `kafakRequest.Plan` is guaranteed to be empty

internal/kafka/internal/services/cluster_placement_strategy.go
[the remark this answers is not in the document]
pb82: It is fine to assume this now. Although keep in mind that at some point we might offer discounts where capacity consumed != size.

ziccardi (changes requested):

internal/kafka/internal/services/clusters.go
```diff
@@ -501,36 +503,47 @@ func (c clusterService) GetExternalID(clusterID string) (string, *apiErrors.Serv
...
 	if err := query.Error; err != nil {
 		return nil, apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to query by cluster info")
 	}
+
+	clusterIdCountMap := map[string]int{}
+
+	var res []ResKafkaInstanceCount
+
+	for _, k := range kafkas {
+		reqSize, e := strconv.Atoi(strings.TrimPrefix(k.SizeId, "x"))
```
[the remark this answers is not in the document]
pb82: see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/837#discussion_r823533137

internal/kafka/internal/config/kafka_supported_instance_types.go
pb82: @ziccardi what internal do you mean with `copy the object to avoid exposing mutable internals`?

internal/kafka/internal/handlers/kafka.go
pawelpaszki: good catch!

internal/kafka/internal/config/kafka_supported_instance_types.go
ziccardi: The `KafkaInstanceType` contains a list of `KafkaInstanceSize`. The list is not deeply copied: only the address is copied. That means that if the caller changes some object inside the received `KafkaInstanceType.Sizes` array, the called object gets the changes too.

That could lead to errors hard to root since the content of an internal is changed somewhere else in the code.

internal/kafka/internal/handlers/kafka.go
pawelpaszki: @ziccardi - I have addressed your most recent comments. can you have another look please?

internal/kafka/internal/handlers/validation.go
```diff
@@ -95,18 +96,18 @@ func ValidateCloudProvider(kafkaService *services.KafkaService, kafkaRequest *db
...
 func ValidateKafkaUpdateFields(kafkaUpdateRequest *private.KafkaUpdateRequest) handlers.Validate {
 	return func() *errors.ServiceError {
-		if stringNotSet(&kafkaUpdateRequest.StrimziVersion) &&
-			stringNotSet(&kafkaUpdateRequest.KafkaVersion) &&
-			stringNotSet(&kafkaUpdateRequest.KafkaIbpVersion) &&
-			stringNotSet(&kafkaUpdateRequest.KafkaStorageSize) {
+		if !stringSet(&kafkaUpdateRequest.StrimziVersion) &&
+			!stringSet(&kafkaUpdateRequest.KafkaVersion) &&
+			!stringSet(&kafkaUpdateRequest.KafkaIbpVersion) &&
+			!stringSet(&kafkaUpdateRequest.KafkaStorageSize) {
```
ziccardi, line 102: this would be more readable as:
```go

if !(stringSet(&kafkaUpdateRequest.StrimziVersion) ||
	 stringSet(&kafkaUpdateRequest.KafkaStorageSize) {
```

ziccardi (approved):
Just a small comment. LGTM overall.

Great job!

internal/kafka/internal/handlers/validation.go
pb82: hwo about using a variadic function, e.g. `func multiStringSet(strings ...string)`?

pawelpaszki: @pb82 - it would probably be nice, but there is an overhead of constructing the slice of kafkaUpdateRequest params. are you ok to go with Massimo's suggestion?

pb82: Yes, that's fine too.

internal/kafka/internal/config/kafka_supported_instance_types.go
ziccardi: @pb82 In the comment you are referring to, I mean that we need to create a copy and return the address of that copy.
If we would return `&size` directly, we would return the address of the object inside the `kp.Sizes` array, thus giving the chance to the caller to change the internal state of the object.

ziccardi: However I believe that it is not really needed to make that copy here since `range` already returns a shallow copy of the object in the array

pb82: Good spot, thanks @ziccardi

internal/kafka/internal/handlers/validation.go
pawelpaszki: I just successfully ran the tests locally with the latest changes. Merging this PR

PR #838, opened 2022-03-08
feat: return browser_url for kafka admin UI

JasonPowr:
## Description
Managed Kafka Control plane API should return URL to the data plane UI for Kafka in order to improve user experience.
Users should be able to navigate from various integrations like CLI, Ansible directly to UI where they can see metrics/monitoring etc.

jira: https://issues.redhat.com/browse/MGDSTRM-5706

## Verification Steps
1. Create Kafka instance.
2. In the response body there should be a url that links to admin UI

- [ ] ~~Documentation added for the feature
- [ ] Code Review completed


pawelpaszki:
@JasonPowr - is this still a WiP? I just ran it and it works as expected

JasonPowr:
> @JasonPowr - is this still a WiP? I just ran it and it works as expected

@pawelpaszki I just need to add some changes to tests and it should be OK.

JasonPowr:
> @JasonPowr - is this still a WiP? I just ran it and it works as expected

@pawelpaszki should be good now. :)

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1783,6 +1784,8 @@ components:
           type: boolean
         kafka_storage_size:
           type: string
+        browser_url:
+          type: string
```
machi1990, line 1788: I am wondering if there is a better way of sending these UI metadata back. 
/cc @miguelsorianod @pb82 this is like the discussion we had in the last backlog refinement about the content schema

JameelB:
@JasonPowr can you rebase this PR when you get a chance please :)

openapi/kas-fleet-manager.yaml
```diff
@@ -892,6 +892,8 @@ components:
               type: boolean
```
machi1990, line 892: We should also bump the version https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/a8cbfdd30d7c48eceb33f68a625e4980577aaa08/openapi/kas-fleet-manager.yaml#L5 to 1.4.0

internal/kafka/internal/api/public/api/openapi.yaml
JameelB: any follow up on this @machi1990? I wonder if we need to discuss this further before considering merging this PR?

internal/kafka/internal/config/kafka.go
```diff
@@ -38,6 +39,7 @@ func NewKafkaConfig() *KafkaConfig {
 		KafkaCapacityConfigFile:        "config/kafka-capacity-config.yaml",
 		KafkaLifespan:                  NewKafkaLifespanConfig(),
 		Quota:                          NewKafkaQuotaConfig(),
+		BrowserUrl:                     "https://console.redhat.com/",
```
JameelB, line 42: I wonder if we can use `http://localhost:8080/` as default? I believe the [UI](https://github.com/bf2fc6cc711aee1a0c2a/kafka-ui) can be run locally.

internal/kafka/internal/presenters/kafka.go
```diff
@@ -52,9 +53,17 @@ func PresentKafkaRequest(kafkaRequest *dbapi.KafkaRequest) public.KafkaRequest {
 		InstanceType:            kafkaRequest.InstanceType,
 		ReauthenticationEnabled: kafkaRequest.ReauthenticationEnabled,
 		KafkaStorageSize:        kafkaRequest.KafkaStorageSize,
+		BrowserUrl:              fmt.Sprintf("%sapplication-services/streams/kafkas/%s/dashboard", validateBrowserUrl(browserUrl), reference.Id),
```
JameelB, line 56: I think this bit will never change: `application-services/streams/kafkas/`. We can move this to the saas template instead. 

```suggestion
		BrowserUrl:              fmt.Sprintf("%s%s/dashboard", validateBrowserUrl(browserUrl), reference.Id),
```

JameelB, line 56: Wdyt of adding the `/` after the `%s`? It may make it easier to read. 
```suggestion
		BrowserUrl:              fmt.Sprintf("%s/application-services/streams/kafkas/%s/dashboard", validateBrowserUrl(browserUrl), reference.Id),
```

In this case, you don't need to check if it has "/" at the end. It can be removed instead, if the browser url was defined with it. 

There's a function called [strings.TrimSuffix()](https://pkg.go.dev/strings#TrimPrefix) that you can use for this.

internal/kafka/internal/config/kafka.go
JameelB: @machi1990 have you done this by any chance? :thinking:

internal/kafka/internal/presenters/kafka.go
JasonPowr: I agree it would make it easier to read. :)

internal/kafka/internal/config/kafka.go
machi1990: No, this is not something I’ve done but I am sure that’s how the UI team is doing their devs. Maybe @wtrocki would have some input

internal/kafka/internal/api/public/api/openapi.yaml
machi1990: There has not been followups on this. 
In my opinion, the answer to the comment I raised above about the other issue maybe applicable here.

internal/kafka/internal/presenters/kafka.go
JasonPowr: @JameelB Could I just remove it from the default values and add it here then remove the validation on it altogether ?

wtrocki (approved):

internal/kafka/internal/presenters/kafka.go
```diff
@@ -52,9 +53,17 @@ func PresentKafkaRequest(kafkaRequest *dbapi.KafkaRequest) public.KafkaRequest {
 		InstanceType:            kafkaRequest.InstanceType,
 		ReauthenticationEnabled: kafkaRequest.ReauthenticationEnabled,
 		KafkaStorageSize:        kafkaRequest.KafkaStorageSize,
+		BrowserUrl:              fmt.Sprintf("%s/%s/dashboard", validateBrowserUrl(browserUrl), reference.Id),
```
JameelB, line 56: don't need to check if it has a "/" suffix here, the `TrimSuffix()` will return the same string if it doesn't. 

```suggestion
		BrowserUrl:              fmt.Sprintf("%s/%s/dashboard", strings.TrimSuffix(browserUrl), reference.Id),
```

internal/kafka/internal/presenters/kafka.go
```diff
@@ -52,9 +53,17 @@ func PresentKafkaRequest(kafkaRequest *dbapi.KafkaRequest) public.KafkaRequest {
...
 	}
 }
 
+func validateBrowserUrl(BrowserUrl string) string {
+	if strings.HasSuffix(BrowserUrl, "/") {
+		return strings.TrimSuffix(BrowserUrl, "/")
+	}
+	return BrowserUrl
+}
+
```
JameelB, line 66: We can remove this, calling `TrimSuffix()` should be enough :)

internal/kafka/internal/cmd/kafka/list.go
```diff
@@ -48,6 +49,9 @@ func runList(env *environments.Env, cmd *cobra.Command, _ []string) {
 	var kafkaService services.KafkaService
 	env.MustResolveAll(&kafkaService)
 
+	var kafkaConfig config.KafkaConfig
+	env.MustResolveAll(&kafkaConfig)
```
JameelB, line 53: Was testing this and found that the kafkaConfig type needs to be a pointer otherwise it can't find it in the di container. 

Also found that `MustResolveAll` can take multiple args so we don't need to call it twice :) 

```suggestion
	var kafkaService services.KafkaService
	var kafkaConfig *config.KafkaConfig
	env.MustResolveAll(&kafkaService, &kafkaConfig)
```

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1106,7 +1106,7 @@ components:
         updated_at: 2020-10-05T12:56:36.362208Z
         version: 2.6.0
         instance_type: standard
-        browser_url: https://console.redhat.com/
+        browser_url: http://localhost:8080/
```
wtrocki, line 1109: Why this ?

JasonPowr: I believe it was because of this discussion https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/838#discussion_r826056113

wtrocki: Yes! Good suggestion by @JameelB 
Using default value for development purposes is ok, but OpenAPI is user facing so if possible can we leave examples and others with still console.redhat.com while default in the code will be localhost (and we can override that in deployment)

wtrocki: This is just suggestion. Ignore me completely if that doesn't make sense :D

internal/kafka/internal/api/public/api/openapi.yaml
machi1990: Resolving this comment in favor of letting the discussion to be progressed on the other issue.

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1106,6 +1106,7 @@ components:
         updated_at: 2020-10-05T12:56:36.362208Z
         version: 2.6.0
         instance_type: standard
+        browser_url: https://console.redhat.com/
```
JameelB, line 1109: Changed in wrong file. This one is generated by `make openapi/generate`, it should be changed in the openapi/kas-fleet-manager.yaml instead.

openapi/kas-fleet-manager.yaml
```diff
@@ -1418,6 +1420,7 @@ components:
         updated_at: "2020-10-05T12:56:36.362208Z"
         version: "2.6.0"
         instance_type: standard
+        browser_url: "http://localhost:8080/"
```
JameelB, line 1423: When changing this, let's add in the full url since it's an example of the response.  

```suggestion
        browser_url: "https://console.redhat.com/application-services/streams/kafkas/1isy6rq3jki8q0otmjqfd3ocfrg/dashboard"
```

JameelB (approved):
Looks good to me, verified locally and works as expected :+1: 

Would be great to also have @wtrocki approval here too :)

templates/service-template.yml
```diff
@@ -446,6 +446,10 @@ parameters:
   description: The domain name to use for Kafka instances
   value: kafka.bf2.dev
 
+- name: BROWSER_URL
+  description: browser url pointing to qaprodauth.console.redhat.com
```
wtrocki, line 450: Description or value might be invalid here

wtrocki:
This looks amazing! 
@JasonPowr World class job!

internal/kafka/internal/cmd/kafka/list.go
```diff
@@ -46,7 +47,9 @@ func runList(env *environments.Env, cmd *cobra.Command, _ []string) {
 	page := flags.MustGetString(FlagPage, cmd.Flags())
 	size := flags.MustGetString(FlagSize, cmd.Flags())
 	var kafkaService services.KafkaService
-	env.MustResolveAll(&kafkaService)
+
+	var kafkaConfig *config.KafkaConfig
+	env.MustResolveAll(&kafkaService, &kafkaConfig)
```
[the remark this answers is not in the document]
JasonPowr: Sorry for the late reply @ziccardi yes that is correct.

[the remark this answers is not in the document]
JameelB: @ziccardi looks like it has to be a pointer to a pointer. Was getting this error otherwise: 
```
service container di failure: /home/jbriones/projects/gitlab.com/service-delivery/managed-services-api/pkg/environments/environment.go:223: type config.KafkaConfig not exists in the container
```

It's because the `NewKafkaConfig` returns a pointer, so di can't find it unless it's that type. We do a similar thing here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/integration/data_plane_cluster_test.go#L960

JasonPowr:
> Reverified with latest changes and its working as expected :)
> 
> Can you regenerate the openapi spec again? I'm getting a different bindata.go than the one in this PR.

Done now :) @JameelB

PR #840, opened 2022-03-08
feat: adding script for creating local mas-sso instance

akoserwal:
Adding a script to create a local mas-sso using keycloak container with required realm config.

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has created for changes required on the client side

PR #841, opened 2022-03-08
feat(connector): Add multitenancy support using namespaces.

dhirajsb:
## Description
Adds support for multiple connector tenants for individual users and organizations. 

## Verification Steps
Multiple extensive integration tests are included. They should also pass in CI. 

- [X] Documentation added for the feature
- [X] CI and all relevant tests are passing
- [X] Code Review completed
- [X] Verified independently by reviewer


dhirajsb:
Had to create a new PR since old fork got detached after bf2 repo was made public.

dhirajsb:
@lburgazzoli I've added namespace status. The agent should post namespace status updates after they are created in the data plane.

openapi/connector_mgmt-private.yaml
```diff
@@ -383,6 +387,24 @@ components:
           description: the version of the operator
           type: string
 
+    ConnectorNamespaceStatus:
```
lburgazzoli, line 390: can you add some detail about how to use such object ?

i.e:
- I recall we discussed about adding the count of connectors but you said it was not needed
- how do you expect we use the phase/version/conditions ?

openapi/connector_mgmt.yaml
```diff
@@ -957,17 +1299,16 @@ components:
...
         - name
         - connector_type_id
-        - deployment_location
         - desired_state
       properties:
         name:
           type: string
         connector_type_id:
           type: string
+        namespace_id:
```
lburgazzoli, line 1308: This should be listed among the required fields

openapi/connector_mgmt.yaml
```diff
@@ -957,17 +1299,16 @@ components:
       required:
         - name
```
lburgazzoli, line 1300: As a side note, we require `name` to be provided, but we do not validate for uniqueness

openapi/connector_mgmt.yaml
```diff
@@ -1080,6 +1421,134 @@ components:
...
 
+    #
+    # Connector Namespaces
+    #
+    ConnectorNamespaceRequestMeta:
+      type: object
+      properties:
+        name:
+          type: string
+        annotations:
```
lburgazzoli, line 1432: on the [kas](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c2e0d76158fe955b227e200491bbaf06b33cd8a9/openapi/kas-fleet-manager-private.yaml#L240) side, annotations are defined as a key/value as you would expect on kubernetes. They have a fixed set of properties and we don't but I think we should use the same syntax as kubernetes.

internal/connector/test/integration/features/connector-old-path.feature
```diff
@@ -21,10 +21,6 @@ Feature: the old connectors path are still valid
           "name": "example 1",
           "kafka_id":"mykafka"
         },
-        "deployment_location": {
```
lburgazzoli, line 24: I think we can delete this feature at all as we don't have such paths anymore (if so, we should remove them)

internal/connector/test/integration/features/connector-multitenancy-api.feature
```diff
@@ -0,0 +1,721 @@
+Feature: connector namespaces API
+  In order to deploy connectors to an addon OSD cluster
+  As a regular user, I need to be able to create and manage namespaces in both evaluation and non-evaluation clusters
```
lburgazzoli, line 3: is this sentence correct ? 
I would expect that only the  connector cluster owner/admin can create namespaces on a cluster at this stage.

internal/connector/internal/migrations/202202070000_add_connector_namespace_tables.go
```diff
@@ -0,0 +1,91 @@
...
+		db.ExecAction(
+			"ALTER TABLE connector_namespaces DROP CONSTRAINT IF EXISTS fk_connector_namespaces_cluster_id",
+			""),
+		db.ExecAction(
+			"ALTER TABLE connector_namespaces ADD CONSTRAINT fk_connector_namespaces_cluster_id "+
+				"FOREIGN KEY (cluster_id) REFERENCES connector_clusters(id)",
+			"ALTER TABLE connector_namespaces DROP CONSTRAINT IF EXISTS fk_connector_namespaces_cluster_id"),
+
+		// only one of tenantUser or tenantOrg in ConnectorNamespace can be not null
+		db.ExecAction(
```
valdar, line 83: same here for `fk_connector_namespaces_cluster_id`

internal/connector/internal/migrations/202203030000_add_connector_namespace_lease.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+	"time"
+)
+
+func addConnectorNamespaceLease(migrationId string) *gormigrate.Migration {
+
+	type LeaderLease struct {
```
valdar, line 18: What is the purpose of this `LeaderLease` entity/table?

valdar (changes requested):

.github/workflows/ci.yaml
```diff
@@ -111,7 +111,7 @@ jobs:
           export GOPATH=$(go env GOPATH)
           export PATH=${PATH}:$GOPATH/bin
           make verify lint binary test test/integration
-        timeout-minutes: 14
+        timeout-minutes: 20
```
machi1990, line 114: Why 20?

dhirajsb: Just a guess because 14 failed :) 

It looks like there are a lot more tests now like for kafka auth, it's timing out in 20 mins while just running kafka tests. It hasn't gotten to connector tests yet, see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/runs/5475757655?check_suite_focus=true

internal/connector/internal/migrations/202203030000_add_connector_namespace_lease.go
dhirajsb: It's the framework bits required for background workers and reconcile threads.

internal/connector/test/integration/features/connector-multitenancy-api.feature
dhirajsb: Any user is allowed to create an eval namespace, it automatically gets an expiration time. That's the opt-in flow described in the eval brief document.

internal/connector/test/integration/features/connector-old-path.feature
dhirajsb: Agreed, if nobody is using the old paths we should get rid of this feature test.

openapi/connector_mgmt.yaml
dhirajsb: It's optional.

openapi/connector_mgmt.yaml
dhirajsb: That'll take me another day or so to fix all the tests that'll break in connector-api.feature that don't use a namespace_id at this point. Would you like me to do that on this branch or on main?

internal/connector/test/integration/features/connector-multitenancy-api.feature
lburgazzoli: yeah but the wording here seems to suggest that a user can create a non eval namespace  
do we have a test that verify that only the owner of a cluster can create non eval namespaces ?

.github/workflows/ci.yaml
chirino: @machi1990 FYI - maybe someone should look into figuring out how run more kafka tests in parallel to reduce total test time.

openapi/connector_mgmt.yaml
lburgazzoli: it is optional but it leaks into the api so people may want to sue it
if it make sense, we should extend this to any request I guess.

openapi/connector_mgmt.yaml
lburgazzoli: there is one already

openapi/connector_mgmt-private.yaml
lburgazzoli: clarified

openapi/connector_mgmt.yaml
dhirajsb: Yes, will do that before going to staging.

openapi/connector_mgmt.yaml
dhirajsb: Actually, I did create a constraint in the DB, so user will see an error, although about a DB constraint violation. :)

internal/connector/test/integration/features/connector-multitenancy-api.feature
dhirajsb: Right, that's part of the access control requirements I'm going to sort out next after this PR, which will require adding checks for org admin claim.

.github/workflows/ci.yaml
machi1990: > It looks like there are a lot more tests now like for kafka auth

@dhirajsb Hmm I think the test were timing out for a different reason. And this has since been fixed https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/840, so I'd advise rebasing on main if you haven't done so.

 > @machi1990 FYI - maybe someone should look into figuring out how run more kafka tests in parallel to reduce total test time.

+1 true, but you cannot really run parallel integration tests with a shared database that's get's cleanup / reset after each test. You are only opening room for some racey conditions. As I said above the the test timeout issue was more to do with the CI secrets setup (fixed in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/840) rather than the tests being many as the problem would have occurred in either way the tests are run.

internal/connector/internal/migrations/202202070000_add_connector_namespace_tables.go
```diff
@@ -0,0 +1,91 @@
...
+		db.ExecAction(`
+			CREATE OR REPLACE FUNCTION connector_namespaces_version_trigger() RETURNS TRIGGER LANGUAGE plpgsql AS '
+			BEGIN
+			NEW.version := nextval(''connector_namespaces_version_seq'');
+			RETURN NEW;
+			END;'
+		`, `
+			DROP FUNCTION IF EXISTS connector_namespaces_version_trigger
+		`),
+		db.ExecAction(`DROP TRIGGER IF EXISTS connector_namespaces_version_trigger ON connector_namespaces`, ``),
```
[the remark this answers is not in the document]
dhirajsb: It's for getting rid of the unused `version` field and trigger in the namespace table that was added in a previous migration.

openapi/connector_mgmt.yaml
dhirajsb: I can also just change the API for now without requiring it in the implementation which will break a whole feature. We don't validate the incoming JSON against the API schema.

.github/workflows/ci.yaml
dhirajsb: That makes sense. I looked at older PR that this is a copy of, and tests were taking 8mins earlier. I'll revert this change, rebase and push again.

dhirajsb: The CI issue is resolved, integration tests are working again.

internal/connector/internal/migrations/202202070000_add_connector_namespace_tables.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return db.CreateMigrationFromActions(migrationId,
+		db.CreateTableAction(&ConnectorTenantUser{}),
+		db.CreateTableAction(&ConnectorTenantOrganisation{}),
+		db.CreateTableAction(&ConnectorNamespaceAnnotation{}),
+		db.CreateTableAction(&ConnectorNamespace{}),
+
+		// connector namespace version
+		db.ExecAction(`
```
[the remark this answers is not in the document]
dhirajsb: It was adding a version sequence number, which could have been used for detecting phase changes from the agent. But it was later removed in another migration.

internal/connector/test/integration/features/connector-old-path.feature
dhirajsb: @valdar do you see any issues in deleting this old path test feature?

internal/connector/internal/migrations/202202070000_add_connector_namespace_tables.go
```diff
@@ -0,0 +1,91 @@
...
+		db.ExecAction(`DROP TRIGGER IF EXISTS connector_namespaces_version_trigger ON connector_namespaces`, ``),
+		db.ExecAction(`
+			CREATE TRIGGER connector_namespaces_version_trigger BEFORE INSERT OR UPDATE ON connector_namespaces
+			FOR EACH ROW EXECUTE PROCEDURE connector_namespaces_version_trigger();
+		`, `
+			DROP TRIGGER IF EXISTS connector_namespaces_version_trigger ON connector_namespaces
+		`),
+
+		// foreign key relationship between namespaces and clusters
+		db.ExecAction(
```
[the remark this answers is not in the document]
dhirajsb: Yes, the extra DROP is required because there isn't a CREATE IF NOT EXISTS, see https://stackoverflow.com/questions/6801919/postgres-add-constraint-if-it-doesnt-already-exist. 

I'm not exactly sure how the DB could get into that situation, maybe a failed partial migration or something. In theory it shouldn't be present, but the extra DROP guards against a situation where it could be present already.

.github/workflows/ci.yaml
machi1990: Thanks @dhirajsb

openapi/connector_mgmt.yaml
dhirajsb: I changed it to comply with k8s style as you suggested. Can you elaborate what you mean by "extend this to any request?"

internal/connector/test/integration/features/connector-multitenancy-api.feature
dhirajsb: This has been addresed by recent commits to fix authz checks for clusters and namespaces. Only org admins are allowed group wide create and delete access.

openapi/connector_mgmt.yaml
lburgazzoli: we could extend the support for annotations to any resource which could be handy for clients to attach metadata to the resource they create. i.e. you can attach some attributes to a `ConnectorRequest`

dhirajsb: Yeah, it's good architectural design and complies with k8s resource model. I'll create a JIRA ticket for it. We can do it after preview release if we don't have time to get to it before that.

dhirajsb: Created issue https://issues.redhat.com/browse/MGDCTRS-632

internal/connector/test/integration/features/connector-old-path.feature
valdar: @dhirajsb  since the tests are passing, it means that path is served by some part of the codebase, we should get rid of that as well. I see no evident issues in doing so.

dhirajsb: I am getting rid of all old paths with `-` in them, so that only paths with `_` in the path prefix are supported in compliance with the latest connector openapi yaml files. 
Also, I found routes for this path in the route loader, which is also going to be removed:
```
/api/connector_mgmt/v1/kafka_connectors_of/{connector_type_id}
```

valdar (approved):

dhirajsb:
@lburgazzoli , @valdar are we good to this into main?

internal/connector/internal/handlers/connector_admin.go
```diff
@@ -197,3 +176,205 @@ func (h *ConnectorAdminHandler) UpgradeConnectorsByOperator(writer http.Response
...
+
+func (h *ConnectorAdminHandler) DeleteConnector(writer http.ResponseWriter, request *http.Request) {
+	connectorId := mux.Vars(request)["connector_id"]
+	cfg := handlers.HandlerConfig{
+		Validate: []handlers.Validate{
+			handlers.Validation("connector_id", &connectorId, handlers.MinLen(1), handlers.MaxLen(maxConnectorIdLength)),
+		},
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+
+			// TODO add flag to force deletion of connector and deployments
```
valdar, line 373: Is this been implemented yet?

dhirajsb: flag to force deletion is a TODO

valdar (approved):

internal/connector/internal/handlers/connector_admin.go
dhirajsb: Working on adding the force flag to nuke connector bits in the manager.

dhirajsb: It's not a TODO anymore!

PR #843, opened 2022-03-09
feat: expose supported Kafka sizes as public endpoint

vmanley:
## Description

Purpose of this change is to expose supported Kafka instance types through a public endpoint. 
This endpoint returns a list of supported Kafka instance types based on cloud provider and cloud region. Each returned instance type contains a list of supported sizes.
The endpoint requires two parameters:
- `cloud_provider` - Name of the supported cloud provider eg. "aws"
- `cloud_region` - Name of the supported cloud region eg. "us-east-1"
## Verification Steps

1. Run kas-fleet-manager locally with this change.
2. Run the `make run/docs` command to run Swagger. The endpoint requires use of Bearer token.
3. The endpoint for returning the list of supported Kafka instance types is listed as:
   `/api/kafkas_mgmt/v1/instance_types/{cloud_provider}/{cloud_region}`
4. The endpoint requires a supported cloud provider name and supported cloud region name. The following parameters can be used:
   cloud_provider: aws
   cloud_region:  us-east-1
5. Endpoint should returned the list of supported Kafka instance types and sizes.
   Example return:
```
{
  "instance_types": [
    {
      "id": "developer",
      "display_name: "Trial"
      "sizes": [
        {
          "id": "x1",
          "ingress_throughput_per_sec": {
            "bytes": 31457280
          },
          "egress_throughput_per_sec": {
            "bytes": 31457280
          },
          "total_max_connections": 3000,
          "max_data_retention_size": {
            "bytes": 1073741800000
          },
          "max_partitions": 1000,
          "max_data_retention_period": "P14D",
          "max_connection_attempts_per_sec": 100,
          "quota_consumed": 1,
          "quota_type": "rhosak",
          "capacity_consumed": 1
        }
      ]
    },
    {
      "id": "standard",
      "display_name: "Standard"
      "sizes": [
        {
          "id": "x1",
          "ingress_throughput_per_sec": {
            "bytes": 52428800
          },
          "egress_throughput_per_sec": {
            "bytes": 104857600
          },
          "total_max_connections": 3000,
          "max_data_retention_size": {
            "bytes": 1073741800000
          },
          "max_partitions": 1500,
          "max_data_retention_period": "P14D",
          "max_connection_attempts_per_sec": 100,
          "quota_consumed": 1,
          "quota_type": "rhosak",
          "capacity_consumed": 1
        }
      ]
    }
  ]
}
```

- [ ] Code Review completed


openapi/kas-fleet-manager.yaml
```diff
@@ -951,6 +985,77 @@ components:
...
+          description: "Maximium connection attempts per second available to this Kafka instance size."
+          type: int
+        quotaConsumed:
+          description: "Quota consumed by this Kafka instance size."
+          type: int
+        quotaType:
+          description: "Quota type used by this Kafka instance size."
+          type: string
+        capacityConsumed:
+          description: "Data plane cluster capacity consumed by this Kafka instance size"
```
machi1990, line 1058:
```suggestion
        capacityConsumed:
          description: "Data plane cluster capacity consumed by this Kafka instance size"
          type: int
```
adding a missing `type` field.

openapi/kas-fleet-manager.yaml
```diff
@@ -951,6 +985,77 @@ components:
...
+      description: "Supported Kafka Size"
+      type: object
+      properties:
+        kind:
+          description: "Indicates the type of this object. Will be 'SupportedKafkaSize' link."
+          type: string
+        id:
+          description: "Unique identifier of this Kafka instance size."
+          type: string
+        ingressThroughputPerSec:
```
machi1990, line 1030: I think the naming strategy should be `snake_case` i.e `ingress_throughput_per_sec` instead of `ingressThroughputPerSec`. 
This has to be applied throughout.

openapi/kas-fleet-manager.yaml
```diff
@@ -1436,6 +1541,26 @@ components:
...
+              ingressThroughputPerSec: "30Mi",
+              egressThroughputPerSec: "30Mi",
+              totalMaxConnections: 3000,
+              maxDataRetentionSize: "1000Gi",
+              maxPartitions: 1000,
+              maxDataRetentionPeriod: "P14D",
+              maxConnectionAttemptsPerSec: 100,
+              quotaConsumed: 1,
+              quotaType: "rhosak",
+              capacityConsumed: 1,
```
machi1990, line 1561: api naming convention is snake_case.
I think we even have validations.

machi1990 (changes requested):
left some suggestions on the openapi spec file

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1570,6 +1640,74 @@ components:
...
+            size.
+          type: string
+        egressThroughputPerSec:
+          description: Egress throughput per second available to this Kafka instance
+            size.
+          type: string
+        totalMaxConnections:
+          description: Total amount of connections available to this Kafka instance
+            size.
+          type: int
```
akoserwal, line 1685: Formate will be  `format: int64` ?

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -521,6 +521,60 @@ paths:
       security:
       - Bearer: []
       summary: Creates a Kafka request
+  /api/kafkas_mgmt/v1/supported_instance_types:
+    get:
+      operationId: getSupportedInstanceTypes
+      parameters:
```
pb82, line 527: do those parameters belong to the endpoint? In ADR-81 cloud provider(e.g. `aws`) and region (e.g. `us-east-1`) path parameters are required. So we'd have to cross check the response with the supported cloud providers array.

pb82: The way I can see it working is: you use [cloud provider and region](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/managed-services/cicd/saas/saas-kas-fleet-manager.yaml#L439) to get a list of supported instance types. Then you return the matching instance types from [supported instance types](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/managed-services/cicd/saas/saas-kas-fleet-manager.yaml#L126).

openapi/kas-fleet-manager.yaml
vmanley: Thanks @machi1990

openapi/kas-fleet-manager.yaml
vmanley: Thanks @machi1990

internal/kafka/internal/api/public/api/openapi.yaml
vmanley: Thanks for clarification @pb82

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1133,6 +1205,22 @@ components:
...
+          ingressThroughputPerSec: 30Mi
+          egressThroughputPerSec: 30Mi
+          totalMaxConnections: 3000
+          maxDataRetentionSize: 1000Gi
+          maxPartitions: 1000
+          maxDataRetentionPeriod: P14D
+          maxConnectionAttemptsPerSec: 100
+          quotaConsumed: 1
+          quotaType: rhosak
+          capacityConsumed: 1
```
machi1990, line 1223:
```suggestion
          ingress_throughput_per_sec: 30Mi
          egress_throughput_per_sec: 30Mi
          total_max_connections: 3000
          max_data_retention_size: 1000Gi
          max_partitions: 1000
          max_data_retention_period: P14D
          max_connection_attempts_per_sec: 100
          quota_consumed: 1
          quota_rype: rhosak
          capacity_consumed: 1
```

openapi/kas-fleet-manager.yaml
```diff
@@ -1450,6 +1568,26 @@ components:
...
```
machi1990, line 1588:
```suggestion
          ingress_throughput_per_sec: 30Mi,
          egress_throughput_per_sec: 30Mi,
          total_max_connections: 3000,
          max_data_retention_size: 1000Gi,
          max_partitions: 1000,
          max_data_retention_period: P14D,
          max_connection_attempts_per_sec: 100,
          quota_consumed: 1,
          quota_rype: rhosak,
          capacity_consumed: 1
```

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -652,6 +652,78 @@ paths:
       security:
       - Bearer: []
       summary: Returns the list of supported regions of the supported cloud provider
+  /api/kafkas_mgmt/v1/instance_types/{cloud_id}/{cloud_region}:
```
pb82, line 655: We call it cloud_provider everywhere else, let's stick to that instead of `cloud_id`

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -652,6 +652,78 @@ paths:
...
+        style: simple
+      - description: Name of the supported cloud provider region
+        explode: false
+        in: path
+        name: cloud_region
+        required: true
+        schema:
+          type: string
+        style: simple
+      - description: Page index
```
wtrocki, line 675: Do we need pagination for this? I think many integrations will rely on complete list

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1133,6 +1205,22 @@ components:
         failed_reason: a reason the Kafka request creation failed
         instance_type: standard
         reauthentication_enabled: true
+    SupportedKafkaInstanceTypeExample:
+      value:
+        kind: SupportedKafkaInstanceType
+        id: eval
+        sizes:
+        - id: x1
+          ingress_throughput_per_sec: 30Mi
```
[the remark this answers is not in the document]
wtrocki: Ignore. Clarified by @pb82 that we not going to follow ADR spec.

openapi/kas-fleet-manager.yaml
```diff
@@ -64,51 +64,51 @@ paths:
...
                 KafkaRequestGetResponseWithFailedCreationStatusExample:
-                  $ref: '#/components/examples/KafkaRequestFailedCreationStatusExample'
+                  $ref: "#/components/examples/KafkaRequestFailedCreationStatusExample"
           description: Kafka request found by ID
         "401":
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Error'
+                $ref: "#/components/schemas/Error"
```
wtrocki, line 78: Can we avoid changing this format or do it in separate PR. It does produce number of conflicts as well as prevent us from generating valid changelog in SDKs

internal/kafka/internal/api/public/api/openapi.yaml
vmanley: Thanks @pb82 I'll make that change

internal/kafka/internal/api/public/api/openapi.yaml
vmanley: Thanks @pb82 @wtrocki I'll remove pagination

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1583,6 +1671,75 @@ components:
...
+            size.
+          type: string
+        egress_throughput_per_sec:
+          description: Egress throughput per second available to this Kafka instance
+            size.
+          type: string
+        total_max_connections:
+          description: Total amount of connections available to this Kafka instance
+            size.
+          type: int
```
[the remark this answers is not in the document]
vmanley: Thanks @wtrocki I'll correct that

openapi/kas-fleet-manager.yaml
vmanley: Sorry about that. I think my IDE made those changes. I'll reverse it. Thanks @wtrocki

config/kafka-instance-types-configuration.yaml
```diff
@@ -0,0 +1,47 @@
...
+#   - maxPartitions: Maximum partitions for all topics combined.
+#   - quotaConsumed: Quota consumed for selecting this size for a Kafka instance.
+#   - quotaType: Quota type that will be consumed when this size is selected.
+#   - capacityConsumed: Data plane cluster capacity consumed by this Kafka instance size (only used for manual scaling)
+---
+supported_instance_types:
+  - id: standard
+    sizes:
+    - id: x1
+      ingressThroughputPerSec: "50Mi"
```
ajaypratap003, line 25: It would be better if we can have separate value and unit so that UI can handle translations and custom unit if required. For UI point of view we are not going to use backend units, we are using custom units which defined in UI translation files which will helpful for translate units.
 
For now UI showing as given below
```
Ingress
up to 30 MiB/second each

Egress
up to 30 MiB/second each
```
Proposal is separate value and unit 
`ingressThroughputPerSec: {value:50, unit:"Mi"}`

pb82: @ajaypratap003 is that proposal documented somewhere? In ADR-81 there are actually no units on the response: https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/42/files#diff-4a09f9937fd7e06026f7cfe0e7d4143d202388410486747c30750de03aca4a26R125

pb82: ping @wtrocki

ajaypratap003: @pb82 this proposal is from my side for UI point of view. Even UI don't need units, need value only. Avoid units change.

openapi/kas-fleet-manager.yaml
```diff
@@ -970,6 +1014,68 @@ components:
...
+          type: integer
+        quota_consumed:
+          description: 'Quota consumed by this Kafka instance size.'
+          type: integer
+        quota_type:
+          description: 'Quota type used by this Kafka instance size.'
+          type: string
+        capacity_consumed:
+          description: 'Data plane cluster capacity consumed by this Kafka instance size.'
+          type: integer
```
JameelB, line 1078: Do we need to return `capacity_consumed`? This is only used by KAS Fleet Manager and clients shouldn't need to know about it. 

Not too sure about `quota_consumed` and `quota_type` too.

pkg/api/kafka_instance_types.go
```diff
@@ -0,0 +1,22 @@
+package api
+
+type SupportedKafkaInstanceType struct {
+	Id                  string               `json:"id"`
+	SupportedKafkaSizes []SupportedKafkaSize `json:"sizes"`
+}
+
+type SupportedKafkaSize struct {
```
JameelB, line 8: Do we need this? We already have these types in the kafka supported instance types config: [KafkaInstanceType](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/larger-kafka-instances/internal/kafka/internal/config/kafka_supported_instance_types.go#L14) and [KafkaInstanceSize](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/larger-kafka-instances/internal/kafka/internal/config/kafka_supported_instance_types.go#L47)

openapi/kas-fleet-manager.yaml
```diff
@@ -970,6 +1014,68 @@ components:
...
+              - $ref: '#/components/schemas/SupportedKafkaSize'
+    SupportedKafkaSize:
+      description: 'Supported Kafka Size'
+      type: object
+      properties:
+        id:
+          description: 'Unique identifier of this Kafka instance size.'
+          type: string
+        ingress_throughput_per_sec:
+          description: 'Ingress throughput per second available to this Kafka instance size.'
```
JameelB, line 1050: Let's make it clear that this is the maximum limit. 

```suggestion
          description: 'Maximum Ingress throughput per second available to this Kafka instance size.'
```

openapi/kas-fleet-manager.yaml
```diff
@@ -970,6 +1014,68 @@ components:
...
+          type: string
+        egress_throughput_per_sec:
+          description: 'Egress throughput per second available to this Kafka instance size.'
```
JameelB, line 1053:
```suggestion
          description: 'Maximum Egress throughput per second available to this Kafka instance size.'
```

openapi/kas-fleet-manager.yaml
```diff
@@ -970,6 +1014,68 @@ components:
...
+          type: string
+        total_max_connections:
+          description: 'Total amount of connections available to this Kafka instance size.'
```
JameelB, line 1056:
```suggestion
          description: 'Maximum amount of total connections available to this Kafka instance size.'
```

openapi/kas-fleet-manager.yaml
```diff
@@ -970,6 +1014,68 @@ components:
...
+          type: integer
+        max_data_retention_size:
+          description: 'Maximum data storage available to this Kafka instance size.'
+          type: string
+        max_partitions:
+          description: 'Total amount of partitions available to this Kafka instance size.'
```
JameelB, line 1062:
```suggestion
          description: 'Maximum amount of total partitions available to this Kafka instance size.'
```

internal/kafka/internal/services/kafka_instance_types.go
```diff
@@ -0,0 +1,75 @@
...
+
+	region, regionFound := provider.Regions.GetByName(regionId)
+	if !regionFound {
+		return nil, errors.RegionNotSupported(fmt.Sprintf("cloud region '%s' is unsupported", regionId))
+	}
+
+	for k := range region.SupportedInstanceTypes {
+		instanceType, err := t.kafkaConfig.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(k)
+		if err != nil {
+			return nil, errors.InstanceTypeNotSupported(fmt.Sprintf("instance type '%s' is unsupported", err.Error()))
```
JameelB, line 44:
```suggestion
			return nil, errors.InstanceTypeNotSupported(fmt.Sprintf("instance type '%s' is unsupported", k))
```

internal/kafka/internal/handlers/kafka_instance_types.go
```diff
@@ -0,0 +1,53 @@
...
+			handlers.ValidateLength(&cloudRegion, "cloud_region", &handlers.MinRequiredFieldLength, nil),
+		},
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+			supportedKafkaInstanceTypeList := public.SupportedKafkaInstanceTypesList{
+				InstanceTypes: []public.SupportedKafkaInstanceType{},
+			}
+
+			regionInstanceTypeList, err := h.service.GetSupportedKafkaInstanceTypesByRegion(cloudProvider, cloudRegion)
+			if err != nil {
+				return nil, err
```
JameelB, line 41: Not sure if we want to expose the InstanceTypeUnsupported error back to the client since this is something out of their control.

This error occurs when we have a mismatch in our configuration. We probably want to return a 500 error to the client and log/capture the InstanceTypeUnsupported error in sentry. wdyt?

internal/kafka/internal/cmd/supportedkafkainstancetypes/cmd.go
```diff
@@ -0,0 +1,26 @@
...
+func NewSupportedKafkaInstanceTypesCommand(env *environments.Env) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "supported_kafka_instance_types",
+		Short: "Perform supported Kafka instance types actions directly",
+		Long:  "Perform supported Kafka instance types actions directly",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			err := env.CreateServices()
+			if err != nil {
+				glog.Fatalf("Unable to initialize environment: %s", err.Error())
+			}
```
ziccardi, line 18:
```suggestion
			if err := env.CreateServices(); err != nil {
				glog.Fatalf("Unable to initialize environment: %s", err.Error())
			}
```

internal/kafka/internal/handlers/kafka_instance_types.go
ziccardi: +1 on logging in sentry.
About the error code, maybe a 503 would be more appropriate?

internal/kafka/internal/services/kafka_instance_type_test.go
```diff
@@ -0,0 +1,110 @@
...
+				providerConfig: tt.fields.providerConfig,
+				kafkaConfig:    tt.fields.kafkaConfig,
+			}
+			got, err := k.GetSupportedKafkaInstanceTypesByRegion(tt.args.cloudProvider, tt.args.cloudRegion)
+			if err != nil && !tt.wantErr {
+				t.Errorf("unexpected error %v", err)
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("supported Kafka instance type lists dont match. want: %v got: %v", tt.want, got)
+			}
```
ziccardi, line 107: Use
```go
Expect(got).To(Equal(tt.want))
```

instead. Equal will use `DeepEqual`. See https://onsi.github.io/gomega/#equalexpected-interface

internal/kafka/internal/services/kafka_instance_type_test.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			k := supportedKafkaInstanceTypesService{
+				providerConfig: tt.fields.providerConfig,
+				kafkaConfig:    tt.fields.kafkaConfig,
+			}
+			got, err := k.GetSupportedKafkaInstanceTypesByRegion(tt.args.cloudProvider, tt.args.cloudRegion)
+			if err != nil && !tt.wantErr {
+				t.Errorf("unexpected error %v", err)
+			}
```
ziccardi, line 104:
```suggestion
			Expect(err != nil).To(Equal(tt.wantErr))
```

internal/kafka/internal/services/kafka_instance_types.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		sizes := instanceType.Sizes
+		alreadyCollectedSize := map[string]bool{}
+		supportedSizesList := []api.SupportedKafkaSize{}
+		for _, size := range sizes {
+			_, sizeCollected := alreadyCollectedSize[size.Id]
+			if sizeCollected {
+				continue
+			}
```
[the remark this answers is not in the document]
ziccardi: Adding to Jameel comment: sizes are guaranteed to be sorted, so we don't need a map to see if the size has been already collected. We can simply check if it is the same as the previous one.

pkg/api/kafka_instance_types.go
JameelB: I don't think this is used anywhere apart from the `GetSupportedKafkaInstanceTypesByRegion()` and unit test which are both in the internal package so they shouldn't need to be visible at this level.

internal/kafka/internal/routes/route_loader.go
```diff
@@ -212,6 +214,15 @@ func (s *options) buildApiBaseRouter(mainRouter *mux.Router, basePath string, op
...
 	apiV1Router.HandleFunc("", v1Metadata.ServeHTTP).Methods(http.MethodGet)
 
+	// /api/kafkas_mgmt/v1/instance_types/{cloud_provider}/{cloud_region}
+	apiV1SupportedKafkaInstanceTypesRouter := apiV1Router.PathPrefix("/instance_types/{cloud_provider}/{cloud_region}").Subrouter()
+	apiV1SupportedKafkaInstanceTypesRouter.HandleFunc("", supportedKafkaInstanceTypesHandler.ListSupportedKafkaInstanceTypes).
+		Name(logger.NewLogEvent("list-supported-kafka-instance-types-by cloud-region", "list supported kafka instance types by cloud region").ToString()).
+		Methods(http.MethodGet)
+	apiV1SupportedKafkaInstanceTypesRouter.Use(requireIssuer)
+	apiV1SupportedKafkaInstanceTypesRouter.Use(requireOrgID)
+	apiV1SupportedKafkaInstanceTypesRouter.Use(authorizeMiddleware)
```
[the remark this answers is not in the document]
JameelB: Hmm, since the cloud provider endpoints for getting the cloud provider id and regions also require authentication, we may need to have it here as well. Can ignore this comment :)

internal/kafka/internal/handlers/kafka_instance_types.go
machi1990: > About the error code, maybe a 503 would be more appropriate?

Why say `503 - Service Unavailable`, I was thinking more in line with what @JameelB suggested, `500 Internal Server Error` in this case.  Because the service is available, but an internal error due to config mismatch has occurred. Or we are saying, the error is to take the service down, my understanding is non due to configuration validation happening at app startup and technically this error will likely not occur during runtime.

ziccardi: I'm unsure between 500 and 503, but I'm ok for 500 as well.

I was thinking 503 because the service, for that instance type is not available, but you can retry later.
While error 500 is about any "unexpected condition that prevented the server from fullfilling the request".

I'm interpreting 500 as something like 'network error', 'disk full', 'out of memory' or similar error that could happen at any time.

wtrocki: > Not sure if we want to expose the InstanceTypeUnsupported error back to the client since this is something out of their control.

Would that affect also old clients that still want to use some unsupported instance types?
While we do not care about HTTP error codes, defining this error code might be beneficial for client to understand that they requesting some old type.

internal/kafka/internal/generated/bindata.go
```diff
@@ -78,7 +78,7 @@ func (fi bindataFileInfo) Sys() interface{} {
...
...96\x09\x3b\x67\x5f\x1d\xa9\x10\xdf\x73\x6e\x7c\xc5\x73\x6f\x6b\x2f\x31\x29\xd5\x37\x18\x6d\x84\xbf\xb8\x39\xdd\x2f\x1a\xc9\x7e\x61\xb5\xb4\xc9\xfb\x56\xe9\xdb\x96\x0d\x86\x13\xef\xa5\xb2\xdd\xf9\x3a\x9b\x4f\xdf\xbf\x98\xbd\x3d\xe4\xbf\x5c\xec\x7d\xc9\xa8\x6c\xbd\xd4\xde\x01\xad\xc6\x1e\xb3\xa3\xb3\x8f\xd2\xe7\xcf\x40\x6e\x16\x38\x16\x63\xf8\xf0\xea\x7d\x6f\xef\x8f\xde\x8b\x71\x72\xf8\xa4\xdf\x2b\x93\x34\xe5\x6d\xf0\x95\xe8\x15\x0e\xe3\xae\x9c\x2d\x3f\xf4\xfc\xe0\xdc\x39\x9f\xba\xcf\x38\x11\x68\x9b\xfb\xdf\x2e\x9e\x9b\xbb\x5a\xe9\xfe\xa6\xaa\x25\x19\x30\x98\x6d\x7b\xcf\x9f\x9f\x3b\x3e\x73\xbd\x8b\xd1\xec\x19\xf2\x4f\x9f\x71\x7f\x3a\x0b\xbf\x6d\x79\xf3\x53\xfe\xed\xbf\xfe\xcf\x7f\xef\xfd\x71\x7c\xb8\x0b\x4f\x35\xb5\x7d\xc5\x9e\x9f\xf3\x3a\x6f\x06\x6c\xc2\xf5\x03\xaf\x1b\x8a\x0f\xea\xe3\xab\x77\x9f\x8f\x8e\xf7\x0e\xd3\xa5\xc4\x19\x75\x55\x3a\x53\x26\x51\xb3\x60\x9c\x6c\x3f\x98\x6d\x53\xb6\xed\x5c\x90\xd8\x79\x46\xb1\x94\xd7\x9c\x9d\xb9\xc3\x1d\x6f\x36\x15\xdf\x06\xc8\x2d\x3c\x8d\x9a\x96\xb5\xea\x2e\x23\xc2\x70\x54\xfe\xa7\x69\x3d\x3e\xe6\x5f\xd8\x62\x27\xe4\xe7\xa7\x43\x7e\x10\xbc\xf9\xb6\x7d\xfa\x47\xf4\xfa\xd9\x2b\xd4\x79\xf2\xbf\x01\x00\x00\xff\xff\x16\xc3\xb8\xbd\x05\xe2\x00\x00")
```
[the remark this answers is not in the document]
vmanley: Sure will do

internal/kafka/internal/services/kafka_instance_types.go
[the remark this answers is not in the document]
vmanley: I'll leave it out so. Thanks

openapi/kas-fleet-manager.yaml
vmanley: They're included in ADR-81 https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/42/files#diff-4a09f9937fd7e06026f7cfe0e7d4143d202388410486747c30750de03aca4a26R119-R121. Seems to be needed by UI

pkg/api/kafka_instance_types.go
vmanley: I'll try using the config types. Thanks

openapi/kas-fleet-manager.yaml
vmanley: Thanks I'll update those descriptions

internal/kafka/internal/services/kafka_instance_types.go
vmanley: Thanks I'll make that change

internal/kafka/internal/handlers/kafka_instance_types.go
vmanley: I'll change the error reporting here. Thanks

internal/kafka/internal/routes/route_loader.go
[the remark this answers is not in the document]
vmanley: Yeah there's a comment in the Jira confirming this

internal/kafka/internal/handlers/kafka_instance_types.go
JameelB: > > Not sure if we want to expose the InstanceTypeUnsupported error back to the client since this is something out of their control.
> 
> Would that affect also old clients that still want to use some unsupported instance types? While we do not care about HTTP error codes, defining this error code might be beneficial for client to understand that they requesting some old type.

Clients don't send a specific instance type in the request though, they're meant to get all of the instance types that KAS Fleet Manager supports. The only thing they can specify is the cloud provider id and region. If they requested an invalid cloud provider id/region then they'll get a service error code in relation to those.

This particular error is caused by misconfiguration on KAS Fleet Manager side. It will occur in this particular scenario: 
```
cloud provider config
---
supported_providers:
  - name: aws
    default: true
    regions:
      - name: us-east-1
        default: true
        supported_instance_type:
          standard: {}
          eval: {}
---

supported instance types config
---
supported_instance_types:
  - id: standard
    sizes:
       ...
---
```

`eval` is listed in `supported_instance_type` in us-east-1 but it's not available in the `supported_instance_types` configuration.

For this sort of error, maybe we could send back an `ErrorGeneral` service error (code: 9, http code: 500)? It's not something that clients can do anything about.

openapi/kas-fleet-manager.yaml
JameelB: thanks for confirming! :)

internal/kafka/internal/cmd/supportedkafkainstancetypes/flags.go
```diff
@@ -0,0 +1,8 @@
+package supportedkafkainstancetypes
+
+const (
+	// FlagCloudId is a flag representing cloud provider Id
```
JameelB, line 4:
```suggestion
	// FlagCloudProvider is a flag representing cloud provider Id
```

internal/kafka/internal/presenters/kafka_instance_types.go
```diff
@@ -0,0 +1,22 @@
...
+	return public.SupportedKafkaInstanceType{
+		Id:    reference.Id,
+		Sizes: GetSupportedKafkaSizes(supportedInstanceType.SupportedKafkaSizes),
+	}
+}
+
+func GetSupportedKafkaSizes(supportedKafkaSizes []api.SupportedKafkaSize) []public.SupportedKafkaSize {
+	sizes := make([]public.SupportedKafkaSize, 0)
+	for _, c := range supportedKafkaSizes {
+		sizes = append(sizes, public.SupportedKafkaSize{Id: c.Id, IngressThroughputPerSec: public.SupportedKafkaSizeBytesValueItem(c.IngressThroughputPerSec), EgressThroughputPerSec: public.SupportedKafkaSizeBytesValueItem(c.EgressThroughputPerSec), TotalMaxConnections: c.TotalMaxConnections, MaxDataRetentionSize: public.SupportedKafkaSizeBytesValueItem(c.MaxDataRetentionSize), MaxPartitions: c.MaxPartitions, MaxDataRetentionPeriod: c.MaxDataRetentionPeriod, MaxConnectionAttemptsPerSec: c.MaxConnectionAttemptsPerSec, QuotaConsumed: c.QuotaConsumed, QuotaType: c.QuotaType, CapacityConsumed: c.CapacityConsumed})
```
JameelB, line 19: [Optional] It would make it easier to read this if it's split in multiple lines :)

vmanley: Good point. I'll split it up. Thanks

internal/kafka/internal/cmd/supportedkafkainstancetypes/flags.go
vmanley: Thanks Jameel. Missed that

internal/kafka/internal/services/kafka_instance_types.go
```diff
@@ -0,0 +1,89 @@
...
+	sort.Slice(instanceTypeList, func(i, j int) bool {
+		return instanceTypeList[i].Id < instanceTypeList[j].Id
+	})
+
+	return instanceTypeList, nil
+}
+
+//convertStringToValue converts string value used to create a resource.Quantity type to float32
+//input string values have already been validated
+func convertStringToValue(input string) float32 {
```
JameelB, line 85: I think this is generic enough, this could probably go into the [shared](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/tree/main/pkg/shared) pkg? 

Let's change the function name to be specific to what it does, maybe something like `ConvertStrQuantityToFloat()`?

JameelB: If float32 isn't required, maybe we could use the util function from k8s for converting this to float64 [AsApproximateFloat64](https://pkg.go.dev/k8s.io/apimachinery@v0.21.2/pkg/api/resource#Quantity.AsApproximateFloat64)?

pkg/api/kafka_instance_types.go
JameelB: Ah didn't realize that those are not string values any longer. I think these can be moved into the internal package though since it's not used by other services like connectors. Maybe we can move this into the `internal/kafka/internal/api`?

internal/kafka/internal/services/kafka_instance_type_test.go
```diff
@@ -0,0 +1,119 @@
...
+			fields: fields{
+				providerConfig: buildProviderConfiguration(testKafkaRequestRegion, MaxClusterCapacity, MaxClusterCapacity, false),
+				kafkaConfig:    &defaultKafkaConf,
+			},
+			args: args{
+				cloudProvider: "aws",
+				cloudRegion:   "us-east-2",
+			},
+			wantErr: true,
+		},
```
JameelB, line 105: Could add another test scenario here where instance type isn't supported.

internal/kafka/internal/services/kafka_instance_types.go
ziccardi: What about creating a 'Quantity' type with a `ToFloat32` method?
That way we could use `Quantity` instead of `string` and code would be more readable
WDYT?

internal/kafka/internal/services/kafka_instance_type_test.go
vmanley: Sure will do .Thanks

internal/kafka/internal/services/kafka_instance_type_test.go
```diff
@@ -0,0 +1,119 @@
+package services
+
+import (
+	"testing"
+
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/onsi/gomega"
```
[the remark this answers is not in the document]
vmanley: Of course. More readable. Thanks

internal/kafka/internal/services/kafka_instance_types.go
vmanley: Good idea about placing in the shared pkg. I'll use the AsApproximateFloat64() for the shared function. Thanks

vmanley: @ziccardi missed your comment earlier. Sounds good

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -174,3 +174,19 @@ func (p Plan) GetSizeID() (string, error) {
 	}
 	return t[1], nil
 }
+
+type Quantity string
+
+func (q Quantity) String() string {
```
ziccardi, line 180: Method declaration in `go` is just syntactic sugar: your method will be translated to:
```go
func String(q Quantity) string {
```

thus, everytime you call `String` your string gets copied.

That would be solved by using a pointer receiver:
```suggestion
func (q *Quantity) String() string {
```

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -174,3 +174,19 @@ func (p Plan) GetSizeID() (string, error) {
...
+	return string(q)
+}
+
+func (q Quantity) ToFloat32() (float32, error) {
```
ziccardi, line 184: See previous comment

internal/kafka/internal/presenters/kafka_instance_types.go
```diff
@@ -0,0 +1,23 @@
...
+func PresentSupportedKafkaInstanceType(supportedInstanceType *public.SupportedKafkaInstanceType) public.SupportedKafkaInstanceType {
+	reference := PresentReference(supportedInstanceType.Id, supportedInstanceType)
+	return public.SupportedKafkaInstanceType{
+		Id:    reference.Id,
+		Sizes: GetSupportedKafkaSizes(supportedInstanceType.Sizes),
+	}
+}
+
+func GetSupportedKafkaSizes(supportedKafkaSizes []public.SupportedKafkaSize) []public.SupportedKafkaSize {
+	sizes := make([]public.SupportedKafkaSize, 0)
```
miguelsorianod, line 16: minor optimization:

We know the size we need for this array so it can be initialized with:
```
 sizes := make([]public.SupportedKafkaSize, len(supportedKafkaSizes))
```

In this way the array won't need to be dynamically re-allocated when iterating

internal/kafka/internal/config/kafka_supported_instance_types.go
vmanley: Thanks @ziccardi. I'll update

internal/kafka/internal/presenters/kafka_instance_types.go
vmanley: Thanks @miguelsorianod. I'll include that

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -174,3 +174,19 @@ func (p Plan) GetSizeID() (string, error) {
...
+
+type Quantity string
+
+func (q *Quantity) String() string {
+	return string(*q)
+}
+
+func (q *Quantity) ToFloat32() (float32, error) {
+	var output float32
+	p, err := resource.ParseQuantity(q.String())
```
ziccardi, line 186: Invoking `q.String()` here generates a copy of the string. The string is then copied again when passed to `ParseQuantity`.
We can avoid one copy by simply casting `*q` to `string` here

vmanley: Thanks @ziccardi I'll make this change

internal/kafka/internal/services/kafka_instance_types.go
```diff
@@ -0,0 +1,84 @@
...
+				MaxDataRetentionSize: public.SupportedKafkaSizeBytesValueItem{
+					Bytes: retentionSizeBytes,
+				},
+				MaxPartitions:               int32(size.MaxPartitions),
+				MaxDataRetentionPeriod:      size.MaxDataRetentionPeriod,
+				MaxConnectionAttemptsPerSec: int32(size.MaxConnectionAttemptsPerSec),
+				QuotaConsumed:               int32(size.QuotaConsumed),
+				QuotaType:                   size.QuotaType,
+				CapacityConsumed:            int32(size.CapacityConsumed),
+			})
```
JameelB, line 71: The service shouldn't really have to deal with any of the public api types. In case another internal component other than the handler (i.e. another service/worker) calls this function, they will get the public api type which should only be used for the API responses. Instead of having this here, there should be a presenter that handles this part, converting the internal obj to a public api obj.

wtrocki: Sounds like old good DTO pattern

https://nurcahyaari.medium.com/how-i-implement-clean-code-architecture-on-golang-projects-68be58830621#:~:text=DTO%20use%20as%20a%20struct,the%20repository%2C%20and%20vice%20versa.

vmanley: Thanks @JameelB  and @wtrocki thanks for the feedback. I'll put presenter in

openapi/kas-fleet-manager.yaml
```diff
@@ -463,6 +463,50 @@ paths:
         - $ref: '#/components/parameters/page'
         - $ref: '#/components/parameters/size'
         - $ref: '#/components/parameters/instance_type'
+  /api/kafkas_mgmt/v1/instance_types/{cloud_provider}/{cloud_region}:
+    get:
+      summary: Returns the list of supported Kafka instance types and sizes filtered by cloud provider and region
+      operationId: getInstanceTypesByCloudProviderAndRegion
+      security:
+        - Bearer: [ ]
+      responses:
```
JameelB, line 472: Missing status 400 here, this endpoint could also return provider/region not supported

vmanley: Thanks @JameelB I'll add 400 status

JameelB (approved):
Looks good to me! Verified locally, works as expected. :+1: Approving as all comments seem to have been addressed.

@machi1990 I see you still have requested changes here. Do you mind giving this another look when you get a chance? 

Great job on this @vmanley!

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -66,22 +66,22 @@ type KafkaInstanceSize struct {
 // - any non-id string values must be parseable
 // - any int values must not be less than or equal to zero
 func (k *KafkaInstanceSize) validate(instanceTypeId string) error {
-	if k.EgressThroughputPerSec == "" || k.IngressThroughputPerSec == "" ||
-		k.MaxDataRetentionPeriod == "" || k.MaxDataRetentionSize == "" || k.Id == "" || k.QuotaType == "" {
+	if k.EgressThroughputPerSec.String() == "" || k.IngressThroughputPerSec.String() == "" ||
```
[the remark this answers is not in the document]
vmanley: Good idea thanks @ziccardi

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -66,22 +66,22 @@ type KafkaInstanceSize struct {
...
+		k.MaxDataRetentionPeriod == "" || k.MaxDataRetentionSize.String() == "" || k.Id == "" || k.QuotaType == "" {
 		return fmt.Errorf("Kafka instance size '%s' for instance type '%s' is missing required parameters.", k.Id, instanceTypeId)
 	}
 
-	egressThroughputQuantity, err := resource.ParseQuantity(k.EgressThroughputPerSec)
+	egressThroughputQuantity, err := resource.ParseQuantity(k.EgressThroughputPerSec.String())
 	if err != nil {
 		return fmt.Errorf("egressThroughputPerSec for Kafka instance type '%s', size '%s' is invalid: %s", k.Id, instanceTypeId, err.Error())
 	}
```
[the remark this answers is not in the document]
vmanley: I can include that. Thanks @ziccardi

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -174,3 +174,19 @@ func (p Plan) GetSizeID() (string, error) {
...
+
+func (q *Quantity) ToFloat32() (float32, error) {
+	var output float32
+	p, err := resource.ParseQuantity(string(*q))
+	if err != nil {
+		return 0, err
+	}
+	output = float32(p.Value())
+	return output, nil
+}
```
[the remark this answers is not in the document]
vmanley: Thanks @ziccardi

internal/kafka/internal/handlers/kafka_instance_types.go
```diff
@@ -0,0 +1,56 @@
...
+	service services.SupportedKafkaInstanceTypesService
+}
+
+func NewSupportedKafkaInstanceTypesHandler(service services.SupportedKafkaInstanceTypesService) *supportedKafkaInstanceTypesHandler {
+	return &supportedKafkaInstanceTypesHandler{
+		service: service,
+	}
+}
+
+func (h supportedKafkaInstanceTypesHandler) ListSupportedKafkaInstanceTypes(w http.ResponseWriter, r *http.Request) {
```
ziccardi, line 25: Can we change this to a pointer to avoid copying the `supportedKafkaInstanceTypesHandler` at every call to `ListSupportedKafkaInstanceTypes`?
```suggestion
func (h *supportedKafkaInstanceTypesHandler) ListSupportedKafkaInstanceTypes(w http.ResponseWriter, r *http.Request) {
```

internal/kafka/internal/services/kafka_instance_types.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func NewSupportedKafkaInstanceTypesService(providerConfig *config.ProviderConfig, kafkaConfig *config.KafkaConfig) SupportedKafkaInstanceTypesService {
+	return &supportedKafkaInstanceTypesService{
+		providerConfig: providerConfig,
+		kafkaConfig:    kafkaConfig,
+	}
+}
+
+func (t supportedKafkaInstanceTypesService) GetSupportedKafkaInstanceTypesByRegion(providerId string, regionId string) ([]config.KafkaInstanceType, *errors.ServiceError) {
```
ziccardi, line 28:
```suggestion
func (t *supportedKafkaInstanceTypesService) GetSupportedKafkaInstanceTypesByRegion(providerId string, regionId string) ([]config.KafkaInstanceType, *errors.ServiceError) {
```

See previous comment.

pkg/errors/errors.go
```diff
@@ -446,6 +446,10 @@ func (e *ServiceError) IsFailedToCheckQuota() bool {
 	return e.Code == FailedToCheckQuota("").Code
 }
 
+func (e *ServiceError) IsInstanceTypeNotSupported() bool {
+	return e.Code == InstanceTypeNotSupported("").Code
```
ziccardi, line 450: Instead of instantiating InstanceTypeNotSupported, I think we can check the code directly:
```suggestion
	return e.Code == ErrorInstanceTypeNotSupported
```

internal/kafka/internal/handlers/kafka_instance_types.go
vmanley: Thanks I'll update this

vmanley:
@ziccardi think I've addressed the last couple of comments. Could you have a look?

ziccardi (approved):
LGTM

wtrocki:
LGTM

JameelB (approved):

PR #849, opened 2022-03-10
feat: update capacity reporting in cloud providers region

JameelB:
## Description
We currently report capacity information in the /cloud_providers/{id}/region endpoint. For each region, there is a capacity object that includes information on the instance type and if the max capacity for that instance type in the region has been reached. 

e.g.
```
{
     "kind": "CloudRegion",
      "id": "eu-west-1",
      ...
      "capacity": [
         {
              "instance_type": "standard",
              "max_capacity_reached": true
         }
      ]
},
```

This will no longer work with multiple sizes i.e. there could be capacity for a size 'x1' but may not have capacity for 'x2'. Instead of having this boolean value represent this, we can return a list of sizes that can still be created on that region. 

The new response should look like the following:
```
{
     "kind": "CloudRegion",
      "id": "eu-west-1",
      ...
      "capacity": [
         {
              "instance_type": "standard",
              "max_capacity_reached": true // to be deprecated
              "available_sizes": [ "x1", "x2", ...] // it'll be an empty array if no capacity is left
         }
      ]
},
```

Example response: 
```
{
    "kind": "CloudRegionList",
    "page": 1,
    "size": 22,
    "total": 22,
    "items": [
          ...
          {
              "kind": "CloudRegion",
              "id": "sa-east-1",
              "display_name": "South America, São Paulo",
              "enabled": false,
              "supported_instance_types": [],
              "capacity": []
          },
          {
              "kind": "CloudRegion",
              "id": "us-east-1",
              "enabled": true,
              "supported_instance_types": [
                  "standard",
                  "eval"
              ],
              "capacity": [
                  {
                      "instance_type": "standard",
                      "max_capacity_reached": false,
                      "available_sizes": [
                          "x1",
                          "x2"
                      ]
                  },
                  {
                      "instance_type": "eval",
                      "max_capacity_reached": false,
                      "available_sizes": [
                          "x1"
                      ]
                  }
              ]
          },
          ...
     ]
}
```

## Verification Steps
- Review test changes and ensure they are passing
- [Optional] Manual verification:
  - Run KAS Fleet Manager
  - Create a dummy cluster on 'us-east-1'.
     - It should be in a `ready` state and supports `standard,eval` instance types.
     - Set the kafka instance limit to 0 
  - Call the cloud_provider/aws/regions endpoint.
  - Look for `us-east-1` in the list.
     - capacity should have 2 items: one for `standard` and one for `eval`. Each should have the following fields:
        - instance_type, max_capacity_reached and available_sizes. 
        - available_sizes should be an empty array and max_capacity_reached should be 'true'
   - Modify the dummy cluster and set its kafka instance limit to 1. Restart KAS Fleet Manager
     - Call the cloud_provider/aws/regions endpoint again. Look for `us-east-1` in the list. available_sizes for each instance types should now have "x1"
   - Create a `standard` Kafka instance.
     ```
     {
       "multi_az": true,
       "name": "test",
       "region": "us-east-1",
       "plan": "standard.x1"
     }
     ```
    - Call the cloud_provider/aws/regions endpoint again. Look for `us-east-1` in the list. available_sizes for each instance types should now be an empty array since capacity is now used up.
 

- [ ] Code Review completed


machi1990:
Nice. @JameelB the draft PR will now only need a rebase on main to get the latest CI changes to skip draft PR.
The same will have to be done for the larger-kafka-instances branches.

internal/kafka/internal/services/kafka.go
```diff
@@ -179,6 +181,56 @@ func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapa
...
+			MultiAZ:       criteria.MultiAZ,
+			SizeId:        it.Sizes[i].Id,
+		}
+
+		// Check against region limits
+		hasCapacity, err := k.HasAvailableCapacityInRegion(kafka)
+		if err != nil {
+			errMsg := fmt.Sprintf("unable to get available sizes in region: failed to check region capacity: %s", err.Error())
+			glog.Infoln(errMsg)
+			return []string{}, errors.New(errors.ErrorGeneral, errMsg)
```
ziccardi, line 210: no need to create an empty array: `nil` can be safely returned here

internal/kafka/internal/services/kafka.go
```diff
@@ -179,6 +181,56 @@ func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapa
...
+		}
+
+		if hasCapacity {
+			// Check if there is an available cluster in the region that can fit this Kafka instance type and size
+			cluster, err := k.clusterPlacementStrategy.FindCluster(kafka)
+			if err != nil {
+				errMsg := fmt.Sprintf("unable to get available sizes in region: failed to find available cluster: %s", err.Error())
+				glog.Infoln(errMsg)
+				return []string{}, errors.New(errors.ErrorGeneral, errMsg)
```
ziccardi, line 219: no need to create an empty array: nil can be safely returned here

internal/kafka/internal/services/kafka.go
```diff
@@ -179,6 +181,56 @@ func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapa
 	return instTypeRegCapacity == nil || count <= int64(*instTypeRegCapacity), nil
 }
 
+func (k *kafkaService) GetAvailableSizesInRegion(criteria *FindClusterCriteria) ([]string, *errors.ServiceError) {
+	supportedInstanceTypes := k.kafkaConfig.SupportedInstanceTypes.Configuration
+	it, err := supportedInstanceTypes.GetKafkaInstanceTypeByID(criteria.SupportedInstanceType)
+	if it == nil {
+		errMsg := fmt.Sprintf("unable to get available sizes in region: failed to find instance type: %s", err.Error())
+		glog.Infoln(errMsg)
+		return []string{}, errors.New(errors.ErrorGeneral, errMsg)
```
ziccardi, line 190: no need to create an empty array: nil can be safely returned here

internal/kafka/internal/services/kafka.go
```diff
@@ -179,6 +181,56 @@ func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapa
 	return instTypeRegCapacity == nil || count <= int64(*instTypeRegCapacity), nil
 }
 
+func (k *kafkaService) GetAvailableSizesInRegion(criteria *FindClusterCriteria) ([]string, *errors.ServiceError) {
+	supportedInstanceTypes := k.kafkaConfig.SupportedInstanceTypes.Configuration
+	it, err := supportedInstanceTypes.GetKafkaInstanceTypeByID(criteria.SupportedInstanceType)
+	if it == nil {
```
ziccardi, line 187: I think this should be 
`if err != nil`

JameelB: yep you're right, good spot! :smiley:

pawelpaszki:
I've verified your changes. The code works as expected. Will you be adding some tests @JameelB ?

internal/kafka/internal/services/kafka.go
JameelB: I tried this at the start but in the response I get the following: 
```
{
    "instance_type": "standard",
    "max_capacity_reached": true,
    "available_sizes": null
},
```

I could be missing something from the openapi side of things though. In the openapi docs, since this property is required, I can't set a default value. `nullable` is also false, but maybe cause it has the type array it can be? :thinking: 

I could set the value of the `available_sizes` to an empty slice outside of this function though if it returns nil. wdyt?

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1674,7 +1677,15 @@ components:
           description: flag indicating whether the capacity for the instance type
             in the region is reached
           type: boolean
+        available_sizes:
+          description: list of available Kafka instance sizes that can be created
```
pawelpaszki, line 1681: maybe add some note about current limits, e.g. `list of available Kafka instance sizes that can be created when taking into account current capacity and limits`. WDYT?

JameelB:
> I've verified your changes. The code works as expected. Will you be adding some tests @JameelB ?

Thanks for verifying @pawelpaszki! Yep, i'm working on adding some tests right now. Also need to check if existing tests needs to be modified.

internal/kafka/internal/services/kafka.go
machi1990: +1 on creating an empty array to avoid the `null` issue. 
I think we are doing that already in some parts of the codebase to avoid not sending the field at all or setting it to `null`.

ziccardi: But here we are returning an error. If this null arrives to the response, means the error is ignored somewhere...

JameelB: I actually have a question around the error handling here. So an error will be returned in `GetAvailableSizesInRegion()` with the following conditions:
- provider/region/instance type/size is not supported
   - if any are unsupported, we should just return an empty array.
- error occurred when trying to retrieve Kafkas from the db.
   - should we return a 500 in this case?

It's hard to differentiate between the two atm. We will probably need to refactor [FindCluster](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/internal/services/cluster_placement_strategy.go) to return a serviceError so we can differentiate between the two. 

what do you guys think? If it's the case that we don't really need to return an err to the client then i'm ok with removing the returned error here and just log it.

ziccardi: This is an old and very discussed topic.

I'm not too expert in `go` best practice, so I will try to make a generic reasoning: when I invoke a `Find` function, the function is expected to search for an object. The result of search could be a `object found` or `object not found`.

Not finding the object is a normal behaviour for a `find` function, so no error should be returned.
In Java or C++ I would use an `optional` object to avoid using a null return.

The best I can thing of with `go` could be to change the `FindCluster` so that it returns 3 values:

`func FindCluster(kafka *dbapi.KafkaRequest) (*api.Cluster, bool, error)`

so when we call it we do:

```go
cluster, found, error := FindCluster(req)
```

JameelB: It makes sense to refactor it that way. There's still a question of do we return an error back to the user/client for db errors though? If that's the case, that refactor needs to be done along with this PR. 

If not, that could be an enhancement in a follow on PR. Since we don't need to return the error back to the user, we can remove the returned err in `GetAvailableSizesInRegion`.

JameelB: Thinking about this again, I think we should probably return the 500 for the db errors. It could be misleading to report there's no capacity left even though that may not be true and there is actually a problem on the server side retrieving this list.

ziccardi: Totally agree

JameelB: @ziccardi updated this now. 

FindCluster should now return a ServiceError. Had to do it this way than what was described above because we already kind of have the `found` when the cluster returned is nil. What we really need is a way to differentiate between the different type of errors returned by FindCluster since in the GET cloud provider regions handler, we need to ignore any non General errors. 

The GetAvailableSizes() should now return nil instead of creating an empty array on return.

internal/kafka/internal/api/public/api/openapi.yaml
JameelB: Description has now been updated to note the limits

internal/kafka/internal/services/kafka.go
ziccardi: > we already kind of have the found when the cluster returned is nil

It can certainly be done that way. The only drawback is that if for any reason the caller won't check for nil, he will get a panic. On the other side, if you get a boolean you intentionally have to ignore it.

wtrocki (approved):

internal/kafka/internal/handlers/cloud_providers.go
```diff
@@ -93,7 +94,23 @@ func (h cloudProvidersHandler) ListCloudProviderRegions(w http.ResponseWriter, r
...
+
+						availableSizes, err := h.kafkaService.GetAvailableSizesInRegion(&services.FindClusterCriteria{
+							Provider:              cloudRegion.CloudProvider,
+							Region:                cloudRegion.Id,
+							SupportedInstanceType: instType,
+							MultiAZ:               true,
+						})
+						if err != nil && err.Code == errors.ErrorGeneral {
+							return nil, errors.GeneralError("unable to list cloud provider regions at this time")
+						}
```
ziccardi, line 107: Why do we ignore all the errors but `ErrorGeneral`?

JameelB:
> @JameelB do you have some example of json returned from API handy? I want to update mock
> 
> https://github.com/redhat-developer/app-services-sdk-js/pull/320/files#diff-4a99b679a901affa1c429a9806464f90c372cbddd889385f1d471dd0b4591322

@wtrocki I've update the description of this PR with an example response.

pawelpaszki:
@JameelB - I think your PR needs a rebase. I can have a final look after conflicts are resolved

JameelB:
thanks @pawelpaszki, should be ready now :)

pawelpaszki:
> thanks @pawelpaszki, should be ready now :)

I have looked through the code and verified it again. Looks good to me. @ziccardi @machi1990 - can you take a final look as well please?

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -92,7 +92,7 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
 
 	kafkaInstanceSize, e := f.KafkaConfig.GetKafkaInstanceSize(kafka.InstanceType, kafka.SizeId)
 	if e != nil {
-		return nil, e
+		return nil, errors.NewWithCause(errors.ErrorInstancePlanNotSupported, err, "failed to find cluster with criteria '%v'", criteria)
 	}
```
ziccardi, line 96: Not strictly related to this PR, but I believe this is the least expensive check. Wouldn't be better to make this as first check to avoid doing the more expensive ones if the plan is not supported?

internal/kafka/internal/services/kafka.go
```diff
@@ -164,21 +166,75 @@ func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapa
...
 
+func (k *kafkaService) GetAvailableSizesInRegion(criteria *FindClusterCriteria) ([]string, *errors.ServiceError) {
+	if criteria == nil {
+		err := errors.GeneralError("unable to get available sizes in region: criteria was not specified")
+		logger.Logger.Error(err)
+		return nil, err
+	}
+
+	supportedInstanceTypes := k.kafkaConfig.SupportedInstanceTypes.Configuration
+	it, err := supportedInstanceTypes.GetKafkaInstanceTypeByID(criteria.SupportedInstanceType)
```
ziccardi, line 192: to me, variable name `it` looks always like an `iterator`. Mind changing this to something more meaningful like `instanceType` or so?

ziccardi (commented):
Just added a couple of minor comments.

Awesome work!

internal/kafka/internal/services/cluster_placement_strategy.go
JameelB: good spot, agreed this can be done first before doing anything else.

pb82 (approved):

PR #853, opened 2022-03-11
feat: MGDSTRM-7563 Report capacity limits in GET Kafka endpoint

avi5kdonrh:

- [ ] Code Review completed


internal/kafka/internal/presenters/kafka.go
```diff
@@ -31,28 +33,48 @@ func ConvertKafkaRequest(kafkaRequestPayload public.KafkaRequestPayload, dbKafka
...
-func PresentKafkaRequest(kafkaRequest *dbapi.KafkaRequest) public.KafkaRequest {
+func PresentKafkaRequest(kafkaRequest *dbapi.KafkaRequest, config *config.KafkaConfig) public.KafkaRequest {
 	reference := PresentReference(kafkaRequest.ID, kafkaRequest)
-
+	var ingressThroughputPerSec, egressThroughputPerSec, maxDataRetentionPeriod string
+	var totalMaxConnections, maxPartitions, maxConnectionAttemptsPerSec int
+	if config != nil {
+		kafkaConfig, err := config.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)
+		if err != nil {
+			glog.Warning(err)
```
ziccardi, line 43: If this happens, that means that in our database we have a plan (instanceType+SizeId) that doesn't exists. I believe this should be logged in sentry too. WDYT?

internal/kafka/internal/presenters/kafka.go
```diff
@@ -31,28 +33,48 @@ func ConvertKafkaRequest(kafkaRequestPayload public.KafkaRequestPayload, dbKafka
...
+		} else {
+			ingressThroughputPerSec = kafkaConfig.IngressThroughputPerSec
+			egressThroughputPerSec = kafkaConfig.EgressThroughputPerSec
+			totalMaxConnections = kafkaConfig.TotalMaxConnections
+			maxPartitions = kafkaConfig.MaxPartitions
+			maxDataRetentionPeriod = kafkaConfig.MaxDataRetentionPeriod
+			maxConnectionAttemptsPerSec = kafkaConfig.MaxConnectionAttemptsPerSec
+		}
```
ziccardi, line 51: Not strictly related to the code per-se, but If the configuration has been changed since when the cluster has been created, returned infos will be wrong

internal/kafka/internal/presenters/kafka.go
avi5kdonrh: @ziccardi Do you mean something like this? sentry.CaptureException(err)

ziccardi: Yes, since I think it is something we should react to. @machi1990 @JameelB WDYT?

internal/kafka/internal/presenters/kafka.go
avi5kdonrh: @ziccardi Yeah, but how can we bet the latest values here? Shouldn't the fleet-manager pod be restarted once the configuration is changed?

ziccardi: I don't think the kafka instance will be destroyed if we change the configuration. This is something that can happen. With the current implementation, there is no way to return the actual infos (unless I miss something). My comment was just to document a known, potential, issue.

avi5kdonrh: Yeah, I've tested with this and it does happen. Once you create the Kafka with say x1 and standard and then you remove the x1 from the standard array but it's still there in the kafka_requests table, the API will not return these parameters. How should we deal with this? @pb82, @JameelB

internal/kafka/internal/presenters/kafka.go
machi1990: > @ziccardi Do you mean something like this? sentry.CaptureException(err)

we could use our logger.log.Error() or logger.log.Warning() which logs the log to sentry too.

machi1990: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/36e09f13dc46c4fbd97b54f233ca2a2b4c8e68a4/docs/error-handing.md#errors-should-be-logged-and-forward-to-sentry-in-central-places

internal/kafka/internal/presenters/kafka.go
JameelB: that's a really good point. That doesn't just affect this information but the way capacity is calculated as well. 

We could check that there's no Kafka with the specific size/instance type before it can be removed from the config. But even then, while the change is deployed someone else could create a Kafka of that size/instance type. There isn't a way for us to disable creation of certain sizes atm without removing them from the config.

@pb82 we may need to go back to that conversation of adding these to the kafka_requests table :sweat_smile:

JameelB: After recent discussions on this, for now we've agreed that once an instance type/size has been added to the configuration, it should not be removed. If needed, we can scope out this work in a separate follow on PR/task. 

Is this ok with you @ziccardi @pb82?

pawelpaszki:
Closing this PR in favour of https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/866

PR #855, opened 2022-03-11
Mgdstrm 7519

pb82:
## Description
Separate ingress and egerss

## Verification Steps
n.a.


openapi/kas-fleet-manager-private.yaml
```diff
@@ -400,7 +405,10 @@ components:
         total:
           type: object
           properties:
-            ingressEgressThroughputPerSec:
+            ingressThroughputPerSec:
+              type: string #an example is "40Gi", a k8s quantity. Should we have some special format at openapi level?" this example also happens on other fields
```
pawelpaszki, line 409: I guess we should change the description (applies to all `Should we...`)?

templates/service-template.yml
```diff
@@ -373,11 +373,6 @@ parameters:
   description: The public HTTP host URL of the service
   value: "https://api.openshift.com"
 
-- name: KAFKA_CAPACITY_INGRESS_THROUGHPUT
```
akoserwal, line 376: Is this environment is not required? have we also added this in app-interface?

pb82: We want to remove it, because those values will come from the new supported instance types array. However, i'm not sure if we should do the removal in the same PR.

@JameelB what do you think? Should we keep the existing env vars and default values at first?

config/kafka-capacity-config.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 # see https://docs.google.com/spreadsheets/d/1W4H1IgDXKv24Sf7aWeqeAqxjKVj8EJB9giYN9obqPgU/edit#gid=493008265
 ingressEgressThroughputPerSec: "2Mi"
```
pawelpaszki, line 3: should this change to two separate values?

templates/service-template.yml
JameelB: I think we can keep the existing env vars and default values for now and remove them in a separate PR since it's not only this one that will be removed. We won't need any of the `KAFKA_CAPACITY_*` once we start using the limits in the instance type/size configuration.

internal/kafka/internal/services/kafka.go
```diff
@@ -962,12 +962,13 @@ func BuildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
...
 		Spec: managedkafka.ManagedKafkaSpec{
 			Capacity: managedkafka.Capacity{
-				IngressEgressThroughputPerSec: k.IngressThroughputPerSec,
-				TotalMaxConnections:           k.TotalMaxConnections,
-				MaxDataRetentionSize:          kafkaRequest.KafkaStorageSize,
-				MaxPartitions:                 k.MaxPartitions,
-				MaxDataRetentionPeriod:        k.MaxDataRetentionPeriod,
-				MaxConnectionAttemptsPerSec:   k.MaxConnectionAttemptsPerSec,
+				IngressThroughputPerSec:     k.IngressThroughputPerSec,
+				EgressThroughputPerSec:      k.EgressThroughputPerSec,
```
JameelB, line 966: Is there a version of fleetshard right now that accepts these two fields?

config/kafka-capacity-config.yaml
```diff
@@ -1,6 +1,7 @@
...
+maxCapacity: 1000
```
JameelB, line 4: This was only used for the /status endpoint which has now been [removed](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/824/files). Was this added intentionally? :thinking:

internal/kafka/internal/config/kafka.go
```diff
@@ -11,13 +11,14 @@ import (
...
-	MaxConnectionAttemptsPerSec   int    `yaml:"maxConnectionAttemptsPerSec"`
-	MaxCapacity                   int64  `yaml:"maxCapacity"`
+	IngressThroughputPerSec     string `json:"ingressThroughputPerSec"`
+	EgressThroughputPerSec      string `json:"egressThroughputPerSec"`
+	TotalMaxConnections         int    `json:"totalMaxConnections"`
+	MaxDataRetentionSize        string `json:"maxDataRetentionSize"`
+	MaxPartitions               int    `json:"maxPartitions"`
+	MaxDataRetentionPeriod      string `json:"maxDataRetentionPeriod"`
+	MaxConnectionAttemptsPerSec int    `json:"maxConnectionAttemptsPerSec"`
+	MaxCapacity                 int64  `json:"maxCapacity"`
```
JameelB, line 21: just a note, this MaxCapacity isn't actually used anymore, will create a PR against main to remove it.

JameelB, line 21: just a note, this MaxCapacity isn't actually used anymore, will create a PR against main to remove it.

internal/kafka/internal/handlers/data_plane_cluster.go
```diff
@@ -139,9 +139,13 @@ func (h *dataPlaneClusterHandler) validateResizeInfo(request *private.DataPlaneC
...
 			return errors.FieldValidationError("resizeInfo delta data retention size must be set")
 		}
-		if resizeInfo.Delta.IngressEgressThroughputPerSec == nil {
-			return errors.FieldValidationError("resizeInfo delta ingressegress throughput per second must be set")
+		if resizeInfo.Delta.IngressThroughputPerSec == nil {
+			return errors.FieldValidationError("resizeInfo delta ingress throughput per second must be set")
 		}
+		if resizeInfo.Delta.EgressThroughputPerSec == nil {
+			return errors.FieldValidationError("resizeInfo delta egress throughput per second must be set")
+		}
```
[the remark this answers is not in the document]
machi1990: Looking at the latest epic brief on capacity reporting, the resizeInfo field will be refactored and likely that `.Delta` field won't be there.
I am still swimming around the subject (I am yet to confirm it with fleetshard team)but chances are this will change to something differently.

internal/kafka/internal/api/dbapi/data_plane_cluster_status.go
```diff
@@ -30,10 +30,11 @@ type DataPlaneClusterStatusResizeInfo struct {
 }
 
 type DataPlaneClusterStatusCapacity struct {
-	IngressEgressThroughputPerSec string
-	Connections                   int
-	DataRetentionSize             string
-	Partitions                    int
+	IngressThroughputPerSec string
+	EgressThroughputPerSec  string
```
machi1990, line 34: Looking at the change https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/663 on the FSO side, only the `.Spec` part is changed not the `.Status`

machi1990 (changes requested):
@pb82 looking at the change in the OpenAPI spec file and compairing them with the change on the FSO side https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/663

I feel like only this schema should be changed: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/08797b29ba1735aa0cdc9f977252599219290498/openapi/kas-fleet-manager-private.yaml#L199

and not the rest e.g https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/08797b29ba1735aa0cdc9f977252599219290498/openapi/kas-fleet-manager-private.yaml#L358. 

Because the rest of the schemas are what will be returned in the `.Status` and that has not been changed by FSO. Only the `.Spec` part (the part provided by the kas-fleet-manager) has changed. /cc @shawkins is my line of thinking correct?

internal/kafka/internal/config/kafka.go
JameelB: looks like this is not in main but in the larger-kafka-instances branch. Created PR to remove this in the feature branch: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/863

config/kafka-capacity-config.yaml
pb82: I think we said to leave the config files for now. But on the other hand, that will prevent the fleet manager from running locally right? So maybe better to update it in this PR.

pb82:
@machi1990 I've addressed your change request about not updating the ManagedKafkaStatus for now. Please have another look when you get a chance.

internal/kafka/test/mocks/kasfleetshardsync/kas-fleetshard-sync.go
```diff
@@ -279,18 +280,20 @@ func SampleDataPlaneclusterStatusRequestWithAvailableCapacity() *private.DataPla
 			},
 		},
 		Remaining: private.DataPlaneClusterUpdateStatusRequestTotal{
-			Connections:                   &[]int32{1000000}[0], // TODO set the values taking the scale-up value if possible or a deterministic way to know we'll pass it
-			Partitions:                    &[]int32{1000000}[0],
-			IngressEgressThroughputPerSec: &[]string{"test"}[0],
-			DataRetentionSize:             &[]string{"test"}[0],
+			Connections:             &[]int32{1000000}[0], // TODO set the values taking the scale-up value if possible or a deterministic way to know we'll pass it
+			Partitions:              &[]int32{1000000}[0],
+			IngressThroughputPerSec: &[]string{"test"}[0],
```
machi1990, line 285: The CI is complaining because this change (along side other changes in the test folders) will need to still reference `IngressEgress...` since we won't be changing the Status structs.

machi1990 (commented):
The openapi spec change lgtm. Thanks @pb82 
I've added a general comment about updating some files so that they continue to reference `IngressEgress...`

internal/kafka/test/mocks/kasfleetshardsync/kas-fleetshard-sync.go
pb82: Yep, fixed

internal/kafka/internal/config/kafka.go
```diff
@@ -11,12 +11,13 @@ import (
...
 
 type KafkaCapacityConfig struct {
-	IngressEgressThroughputPerSec string `yaml:"ingressEgressThroughputPerSec"`
-	TotalMaxConnections           int    `yaml:"totalMaxConnections"`
-	MaxDataRetentionSize          string `yaml:"maxDataRetentionSize"`
-	MaxPartitions                 int    `yaml:"maxPartitions"`
-	MaxDataRetentionPeriod        string `yaml:"maxDataRetentionPeriod"`
-	MaxConnectionAttemptsPerSec   int    `yaml:"maxConnectionAttemptsPerSec"`
+	IngressThroughputPerSec     string `json:"ingressThroughputPerSec"`
+	EgressThroughputPerSec      string `json:"egressThroughputPerSec"`
```
machi1990, line 15:
```suggestion
	IngressThroughputPerSec     string `yaml:"ingressThroughputPerSec"`
	EgressThroughputPerSec      string `yaml:"egressThroughputPerSec"`
```

internal/kafka/test/mocks/kasfleetshardsync/kas-fleetshard-sync.go
machi1990: Awesome, thanks.

machi1990 (approved):
Added one suggestion, otherwise the change lgtm.

internal/kafka/internal/config/kafka.go
machi1990: I think it should be `yaml` and not `json` here.

pawelpaszki (approved):

config/kafka-capacity-config.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 # see https://docs.google.com/spreadsheets/d/1W4H1IgDXKv24Sf7aWeqeAqxjKVj8EJB9giYN9obqPgU/edit#gid=493008265
-ingressEgressThroughputPerSec: "2Mi"
+ingressThroughputPerSec: "2Mi"
+egressThroughputPerSec: "2Mi"
```
JameelB, line 4: We don't really need to change this right now since it's going to be removed later right? Depending on when this and the removal of this config lands on main, we may need to update the saas template twice?

PR #859, opened 2022-03-14
feat: Deprecate instanceType for cloud type region endpoint

JasonPowr:
## Description
We now have instanceType argument in the cloud provider region but the API also returns RegionCapacityListItem which has instance_types. Currently capacity field is returning all instance types despite the argument being passed So our logic can check if the region is enabled, Check if instance type is supported and then check if we have capacity in that region Since the capacity field already returns the instance types, we no longer need to return the `supported_instance_types` as well. The `instance_type` filter can be removed since it's not being used.

Jira: https://issues.redhat.com/browse/MGDSTRM-7493

## Verification Steps
1. Run make run/docs
2. Go to /regions endpoint
3. Can see that instance_type filter is deprecated 
4. Scroll to bottom can see that in the CloudRegion schema that supported_instance_types is deprecated

- [ ] Code Review completed


internal/kafka/internal/presenters/cloudproviders.go
```diff
@@ -19,12 +19,11 @@ func PresentCloudProvider(cloudProvider *api.CloudProvider) public.CloudProvider
 func PresentCloudRegion(cloudRegion *api.CloudRegion) public.CloudRegion {
 	reference := PresentReference(cloudRegion.Id, cloudRegion)
 	return public.CloudRegion{
-		Id:                     reference.Id,
-		Kind:                   reference.Kind,
-		DisplayName:            cloudRegion.DisplayName,
-		Enabled:                cloudRegion.Enabled,
-		SupportedInstanceTypes: getSupportedInstanceTypes(cloudRegion.SupportedInstanceTypes),
```
JameelB, line 26: We need to keep this in for now. 

Marking the field as "deprecated" is to notify clients/users using this API that this field will be removed. The field shouldn't be removed until the deprecation period is over.

openapi/kas-fleet-manager.yaml
```diff
@@ -1022,7 +1022,8 @@ components:
           type: boolean
           default: false
         supported_instance_types:
-          description: 'The Kafka instance types supported by this region.'
+          deprecated: true  
+          description: DEPRECATION NOTICE - instance_type will be deprecated
```
JameelB, line 1026: Since the `deprecated` field is attached to this property, I don't think we need to change the description here. Let's keep this the same as before. 

Instead, we can update the description of the [endpoint](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/openapi/kas-fleet-manager.yaml#L429) if needed. Do you think this would be useful @wtrocki, @machi1990? Or is adding the deprecated field enough?

machi1990: Since we are not deprecating the endpoint but a field in the response returned by the endpoint, I'd go with the approach of modifying the field's specific description as done in the PR. 
My reasons for this are:
- The change on deprecation notice is localised. 
- That was when a field is removed in the future we won't have to think about updating the endpoint description

JameelB: I'm reluctant to completely replace the previous description in case someone looks at our api docs and wonder what these were used for since it's not completely removed just yet. 

Maybe we can have a multi-line description here with both the current description and the deprecation notice?

wtrocki: There are no rules per se. If deprecated flag is set and field is simply going to be removed then there is no need to change description/do much else. The only exception is when we replacing one field with another. Then description might mention what is the new field to use

machi1990: > Maybe we can have a multi-line description here with both the current description and the deprecation notice?

+1

JameelB:
Do we need to increase the API version for this as well? cc @machi1990. Other than that, changes here looks good to me :+1: 

Same with the removal of /status endpoint, @wtrocki is 1 month deprecation period long enough?

JameelB (approved):

pb82 (approved):

PR #861, opened 2022-03-16
feat: reserve quota based on instance size

pawelpaszki:
## Description
feat: reserve quota based on instance size

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -89,7 +89,7 @@ type KafkaService interface {
 	ListKafkasWithRoutesNotCreated() ([]*dbapi.KafkaRequest, *errors.ServiceError)
 	VerifyAndUpdateKafkaAdmin(ctx context.Context, kafkaRequest *dbapi.KafkaRequest) *errors.ServiceError
 	ListComponentVersions() ([]KafkaComponentVersions, error)
-	HasAvailableCapacityInRegion(kafkaRequest *dbapi.KafkaRequest, sizeRequired int) (bool, *errors.ServiceError)
+	HasAvailableCapacityInRegion(kafkaRequest *dbapi.KafkaRequest) (bool, *errors.ServiceError)
 }
 
 var _ KafkaService = &kafkaService{}
@@ -125,7 +125,7 @@ func NewKafkaService(connectionFactory *db.ConnectionFactory, clusterService Clu
 	}
 }
 
-func (k *kafkaService) HasAvailableCapacityInRegion(kafkaRequest *dbapi.KafkaRequest, sizeRequired int) (bool, *errors.ServiceError) {
+func (k *kafkaService) HasAvailableCapacityInRegion(kafkaRequest *dbapi.KafkaRequest) (bool, *errors.ServiceError) {
 	// get region limit for instance type
 	regInstTypeLimit, e := k.providerConfig.GetInstanceLimit(kafkaRequest.Region, kafkaRequest.CloudProvider, kafkaRequest.InstanceType)
 	if e != nil {
@@ -141,10 +141,10 @@ func (k *kafkaService) HasAvailableCapacityInRegion(kafkaRequest *dbapi.KafkaReq
 		return true, nil
 	}
 	// check capacity
-	return k.capacityAvailableForRegionAndInstanceType(regInstTypeLimit, sizeRequired, kafkaRequest)
+	return k.capacityAvailableForRegionAndInstanceType(regInstTypeLimit, kafkaRequest)
 }
 
-func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapacity *int, sizeRequired int, kafkaRequest *dbapi.KafkaRequest) (bool, *errors.ServiceError) {
+func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapacity *int, kafkaRequest *dbapi.KafkaRequest) (bool, *errors.ServiceError) {
 	errMessage := fmt.Sprintf("Failed to check kafka capacity for region '%s' and instance type '%s'", kafkaRequest.Region, kafkaRequest.InstanceType)
 
 	dbConn := k.connectionFactory.New()
@@ -169,7 +169,12 @@ func (k *kafkaService) capacityAvailableForRegionAndInstanceType(instTypeRegCapa
 		count += int64(kafkaInstanceSize.CapacityConsumed)
 	}
 
-	count += int64(sizeRequired)
+	kafkaInstanceSize, e := k.kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)
+	if e != nil {
+		return false, errors.NewWithCause(errors.ErrorGeneral, e, errMessage)
+	}
+
+	count += int64(kafkaInstanceSize.CapacityConsumed)
 
 	return instTypeRegCapacity == nil || count <= int64(*instTypeRegCapacity), nil
 }
@@ -219,7 +224,7 @@ func (k *kafkaService) reserveQuota(kafkaRequest *dbapi.KafkaRequest, sizeRequir
 	if factoryErr != nil {
 		return "", errors.NewWithCause(errors.ErrorGeneral, factoryErr, "unable to check quota")
 	}
-	subscriptionId, err = quotaService.ReserveQuota(kafkaRequest, types.KafkaInstanceType(kafkaRequest.InstanceType), sizeRequired, k.kafkaConfig)
+	subscriptionId, err = quotaService.ReserveQuota(kafkaRequest, types.KafkaInstanceType(kafkaRequest.InstanceType))
 	return subscriptionId, err
 }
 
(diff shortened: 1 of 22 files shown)


internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -112,12 +113,12 @@ func (q amsQuotaService) getAvailableBillingModelFromKafkaInstanceType(externalI
...
 }
 
-func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (string, *errors.ServiceError) {
+func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType, sizeRequired int, kafkaConfig *config.KafkaConfig) (string, *errors.ServiceError) {
 	kafkaId := kafka.ID
 
 	rr := newBaseQuotaReservedResourceResourceBuilder()
 
-	bm, err := q.getAvailableBillingModelFromKafkaInstanceType(kafka.OrganisationId, instanceType)
+	bm, err := q.getAvailableBillingModelFromKafkaInstanceType(kafka.OrganisationId, instanceType, sizeRequired)
```
akoserwal, line 121: Is this change verified at the AMS api?

internal/kafka/internal/services/quota.go
```diff
@@ -11,7 +12,7 @@ type QuotaService interface {
 	// CheckIfQuotaIsDefinedForInstanceType checks if quota is defined for the given instance type
 	CheckIfQuotaIsDefinedForInstanceType(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (bool, *errors.ServiceError)
 	// ReserveQuota reserves a quota for a user and return the reservation id or an error in case of failure
-	ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (string, *errors.ServiceError)
+	ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType, sizeRequired int, kafkaConfig *config.KafkaConfig) (string, *errors.ServiceError)
```
machi1990, line 15: The `kafkaConfig *config.KafkaConfig` parameter can be passed on during service instantiation time as a constructor argument. There is no need to pass it here.

machi1990, line 15: IIRC, the `kafka *dbapi.KafkaRequest` has the `SizeId`, isn't that enough to infer the `sizeRequired int` info which are passing on as a parameter here?

pawelpaszki: the size has to be processed to get the int value of it

pawelpaszki: that was my first thought, but this requires a lot of refactoring of the quota related stuff

internal/kafka/internal/services/quota/ams_quota_service.go
pawelpaszki: no

internal/kafka/internal/services/kafka.go
```diff
@@ -235,7 +230,12 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
 	// we need to pre-populate the ID to be able to reserve the quota
 	kafkaRequest.ID = api.NewID()
 
-	hasCapacity, err := k.HasAvailableCapacityInRegion(kafkaRequest)
+	kafkaInstanceSize, e := k.kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)
+	if e != nil {
+		return errors.NewWithCause(errors.ErrorGeneral, e, fmt.Sprintf("Failed to check kafka capacity for region '%s' and instance type '%s'", kafkaRequest.Region, kafkaRequest.InstanceType))
+	}
+
+	hasCapacity, err := k.HasAvailableCapacityInRegion(kafkaRequest, kafkaInstanceSize.CapacityConsumed)
```
ziccardi, line 238: No need to add this parameter here: `HasAvailableCapacityInRegion` can retrieve the `KafkaInstanceSize` with `k.kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)`

internal/kafka/internal/services/kafka.go
```diff
@@ -257,7 +257,7 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
 	}
 
 	kafkaRequest.ClusterID = cluster.ClusterID
-	subscriptionId, err := k.reserveQuota(kafkaRequest)
+	subscriptionId, err := k.reserveQuota(kafkaRequest, kafkaInstanceSize.CapacityConsumed)
```
ziccardi, line 260: No need to add this parameter (see previous comment)

internal/kafka/internal/services/quota.go
ziccardi: Can't you retrieve it from the kafkaConfig?

```go
kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)
```

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -112,12 +113,12 @@ func (q amsQuotaService) getAvailableBillingModelFromKafkaInstanceType(externalI
 	return billingModel, nil
 }
 
-func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (string, *errors.ServiceError) {
+func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType, sizeRequired int, kafkaConfig *config.KafkaConfig) (string, *errors.ServiceError) {
```
ziccardi, line 116: Why need for `sizeRequred` parameter?
It can be retrieved from 
```go
kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)
```

ziccardi:
There a re a number of methods that receives both a `KafkaRequest,KafkaConfig` pair and a `requiredSize` parameter.
I believe the latter could be inferred from the former:
```go
kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)
```

internal/kafka/internal/services/quota.go
pawelpaszki: see this comment: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/861#discussion_r827778661

internal/kafka/internal/services/kafka.go
pawelpaszki: > There a re a number of methods that receives both a `KafkaRequest,KafkaConfig` pair and a `requiredSize` parameter. I believe the latter could be inferred from the former:
> 
> ```go
> kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, kafkaRequest.SizeId)
> ```

yes. this requires a lot of refactoring to the quota related stuff

internal/kafka/internal/services/quota.go
machi1990: > but this requires a lot of refactoring of the quota related

Hm the only thing I see happening is passing the default config here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f404864d106496c3c405cad4735b59b7da0640e/internal/kafka/internal/services/quota/default_quota_service_factory.go#L17 as a constructor argument and use it when instantiating the different quota services in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f404864d106496c3c405cad4735b59b7da0640e/internal/kafka/internal/services/quota/default_quota_service_factory.go#L23-L24

of course the config will have to be declared in the different structs but it is something you are already doing at: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f404864d106496c3c405cad4735b59b7da0640e/internal/kafka/internal/services/quota/quota_management_list_service.go#L18. And you'll need to do the same for https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f404864d106496c3c405cad4735b59b7da0640e/internal/kafka/internal/services/quota/ams_quota_service.go#L14

pawelpaszki:
@machi1990 @ziccardi - I think I have addressed your comments. @akoserwal - I will try to verify my changes against AMS today with Massimo

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: Verified:

1. Checked quota available: it was `consumed:1, allowed:3`
2. Reserved a resource with count 2
3. Checked quota available: it was `consumed:3, allowed:3`
4. Deleted the resource 
3. Checked quota available: it was `consumed:1, allowed:3`

pawelpaszki: verification of the same logic was verified by Massimo

internal/kafka/internal/services/quota.go
pawelpaszki: removed the params and used kafkaConfig instead

machi1990: Thank @pawelpaszki

ziccardi (approved):
LGTM

machi1990:
I've more of a general question regarding the `quotaType` field in https://github.com/pawelpaszki/kas-fleet-manager/blob/11acbbafb87bfe2e9b23cebd061350464ad58b91/config/kafka-instance-types-configuration.yaml#L18, is this something that we intend to keep for now? I'd expect its value to be used somehow as part of this change but I do not think it is the case and we have gone with a different approach (i.e the original approach of using the same quota).
does that mean the field can be removed or it is something we see being used eventually in the near future? @pawelpaszki @ziccardi @pb82 @JameelB any thoughts?

ziccardi:
> is this something that we intend to keep for now? 

Good point. Here my 2 cents:

1.  I believe the values in  https://github.com/pawelpaszki/kas-fleet-manager/blob/11acbbafb87bfe2e9b23cebd061350464ad58b91/config/kafka-instance-types-configuration.yaml#L18 should be `RHOSAK` for standard and `RHOSAKTrial` for eval
2. Currently in AMS we have only RHOSAK and RHOSAKTrial. In my understanding in the future we can have different licensing types, so we are open for that event.

@pb82 WDYT?

PR #864, opened 2022-03-18
fix(connectors): MGDCTRS-494 fix Missing deployment inconsistency by recreating them during reconcile if they are missing.

valdar:
## Description

If a connector deployment is missing for  whatever reason during the lifecycle of a connector, that connector become stuck and can not be acted upon. If we detect such  scenario during reconciliation, we create the deployments again.

- [x] ~~Documentation added for the feature~~
- [x] Code Review completed
- [x] ~~Required metrics/dashboards/alerts have been added (or PR created).~~
- [x] ~~Required Standard Operating Procedure (SOP) is added.~~
- [x] ~~JIRA has created for changes required on the client side~~

### DIFF
--- a/internal/connector/internal/workers/connector_mgr.go
+++ b/internal/connector/internal/workers/connector_mgr.go
@@ -138,12 +138,14 @@ func (k *ConnectorManager) Reconcile() []error {
 	serviceErr = k.connectorService.ForEach(func(connector *dbapi.Connector) *serviceError.ServiceError {
 		return InDBTransaction(k.ctx, func(ctx context.Context) error {
 			if err := k.reconcileConnectorUpdate(ctx, connector); err != nil {
-				return serviceError.GeneralError("failed to reconcile assigned connector %s: %v", connector.ID, err.Error())
+				errs = append(errs, err)
+				glog.Errorf("failed to reconcile assigned connector %s: %v", connector.ID, err.Error())
 			}
 			if err := db.AddPostCommitAction(ctx, func() {
 				k.lastVersion = connector.Version
 			}); err != nil {
-				return err
+				errs = append(errs, err)
+				glog.Errorf("failed to AddPostCommitAction to save lastVersion %d: %v", connector.Version, err.Error())
 			}
 			return nil
 		})
@@ -251,6 +253,9 @@ func (k *ConnectorManager) reconcileConnectorUpdate(ctx context.Context, connect
 	// Get the deployment for the connector...
 	deployment, serr := k.connectorClusterService.GetDeploymentByConnectorId(ctx, connector.ID)
 	if serr != nil {
+		if derr := k.recoverMissingDeployment(ctx, connector); derr != nil {
+			return derr
+		}
 		return serr
 	}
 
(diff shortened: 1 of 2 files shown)


lburgazzoli (approved):

dhirajsb:
@valdar did you figure out why this is happening? AFAICT, deployments are only deleted in two places:
* on startup to clean any deployments that don't have connectors
* when agent sends back a `deleted` status for a connector deployment status update
I get that the integration test deletes the deployment directly in the DB to create this condition. But what's causing a deployment to disappear in a running live fleet manager? 
Without that root cause analysis, I suspect the fleet manager is just going to go in a cycle deleting and then restoring deployments.

valdar:
> @valdar did you figure out why this is happening? AFAICT, deployments are only deleted in two places:
> 
>     * on startup to clean any deployments that don't have connectors
> 
>     * when agent sends back a `deleted` status for a connector deployment status update
> 
> 
> I get that the integration test deletes the deployment directly in the DB to create this condition. But what's causing a deployment to disappear in a running live fleet manager? Without that root cause analysis, I suspect the fleet manager is just going to go in a cycle deleting and then restoring deployments.

I admit that the analysis was all theoretical since we observed the connectors in those states but we have yet no mean to better investigate how they get there since read only query service is still not in place. A possible explanation is that between lines https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/services/connector_cluster.go#L499 and https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/services/connector_cluster.go#L505 something happen preventing line 505 to be executed, for instance:

- a concurrent change in desired state 
- k8s decided to bounce the pod
- we bounced it by issuing a deployment 

in that case the connectors could end up in the state they are stuck in.

dhirajsb:
That's a good hypothesis. I am assuming you will be able to try this change before the multitenancy PR is merged. Can you also change the reconcile loop at:
```
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/workers/connector_mgr.go#L141 
to accumulate errors like in line:
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/workers/connector_mgr.go#L119
```
That would make it so that one failing connector doesn't stop other connectors from reconciling.

valdar:
> That's a good hypothesis. I am assuming you will be able to try this change before the multitenancy PR is merged. 

That is the idea yes, to first merge this one and release. Then review and merge multitenancy to be released next Thursday.

> Can you also change the reconcile loop at:
> ```
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/workers/connector_mgr.go#L141 
> to accumulate errors like in line:
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/workers/connector_mgr.go#L119
> ```
> 
> That would make it so that one failing connector doesn't stop other connectors from reconciling.

I think I have done it.

PR #865, opened 2022-03-21
 feat: adding ssoprovider selector to switch between mas-sso & sso.

akoserwal:
…d token handler for service account end-points

Added SSO provider selector
Remove redhatsso config and using keycloak config
Add token handler for service account exposed handlers in the token proxy layer

- [ ] Code Review completed


pkg/providers/core.go
```diff
@@ -81,7 +81,16 @@ func ServiceProviders() di.Option {
 		di.Provide(acl.NewAccessControlListMiddleware),
 		di.Provide(handlers.NewErrorsHandler),
 		di.Provide(func(c *keycloak.KeycloakConfig) sso.KafkaKeycloakService {
-			return sso.NewKeycloakService(c, c.KafkaRealm)
+			var keycloakService sso.KeycloakService
+			if c.SelectSSOProvider == keycloak.MAS_SSO {
+				keycloakService = sso.NewKeycloakService(c, c.KafkaRealm)
```
ziccardi, line 86:
```suggestion
keycloakService = sso.NewKeycloakServiceBuilder().
	ForKeycloak(c).
	WithConfiguration(c, c.KafkaRealm).
 	Build()

File: pkg/services/sso/keycloak_service.go [hunk 2]
```diff
@@ -495,14 +495,14 @@ func (kc *masService) DeRegisterKasFleetshardOperatorServiceAccount(accessToken
 
 func (kc *masService) RegisterConnectorFleetshardOperatorServiceAccount(accessToken string, agentClusterId string) (*api.ServiceAccount, *errors.ServiceError) { // (agentClusterId string, roleName string) (*api.ServiceAccount, *errors.ServiceError) {
```
ziccardi, line 496: `RegisterConnectorFleetshardOperatorServiceAccount` and `RegisterKasFleetshardOperatorServiceAccount` are identical

pkg/services/sso/keycloak_service_proxy.go
```diff
@@ -26,6 +27,29 @@ func (r *keycloakServiceProxy) retrieveToken() (string, *errors.ServiceError) {
...
+		token, err = retrieveUserToken(ctx)
+	} else {
+		token, err = r.retrieveToken()
+	}
+	if err != nil {
+		return "", err
+	}
+	return token, nil
+}
+
```
ziccardi, line 52: The `tokenProvider` interface should be used to provide the token.
Both RedhatSSO Service implementation and Keycloak Service implementation should provide a tokenProvider that returns the token as needed.

This way we will avoid the `if r.GetConfig().SelectSSOProvider == keycloak.REDHAT_SSO` and it will be more extensible/maintainable in the event we will need to add/remove an implementation in the future.

pkg/services/sso/keycloak_service_proxy.go
```diff
@@ -67,35 +91,35 @@ func (r *keycloakServiceProxy) IsKafkaClientExist(clientId string) *errors.Servi
 }
 
 func (r *keycloakServiceProxy) CreateServiceAccount(serviceAccountRequest *api.ServiceAccountRequest, ctx context.Context) (*api.ServiceAccount, *errors.ServiceError) {
-	if token, err := r.retrieveToken(); err != nil {
+	if token, err := tokenForServiceAPIHandler(ctx, r); err!= nil {
```
ziccardi, line 94: token should be retrieved from the provider.

ziccardi (changes requested):

pkg/services/sso/keycloak_service_proxy.go
akoserwal: Agreed, but this is a different type of token. the "user" token needs to be retrieved from the request header and pass it to sso api.

pkg/providers/core.go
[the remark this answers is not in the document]
akoserwal: Realm is required, as even when sso provider is enabled, another implementation will still point to sre realm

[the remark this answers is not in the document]
akoserwal: > Not sure I get it.The realm is part of the KafkaConfig. You mean we can have 2 RedHatSSO client with teh same KafkaConfig and a different Realm?

I didn't get it. So, we have two versions of mas-sso provider one with `c.KafkaRealm` & `c.OSDClusterIDPRealm`

akoserwal: I believe both does the same thing?

pkg/client/keycloak/config.go
```diff
@@ -103,7 +127,18 @@ func (kc *KeycloakConfig) ReadFiles() error {
...
+		return err
+	}
+	err = shared.ReadFileValueString(kc.RedhatSSORealm.ClientIDFile, &kc.RedhatSSORealm.ClientID)
+	if err != nil {
+		return err
+	}
+	err = shared.ReadFileValueString(kc.RedhatSSORealm.ClientSecretFile, &kc.RedhatSSORealm.ClientSecret)
+	if err != nil {
+		return err
+	}
```
machi1990, line 141: I think we should be only be reading RH sso secrets if the provider is set to redhat_sso.

akoserwal: Good point

pkg/providers/core.go
akoserwal: Good point!

pkg/services/sso/redhatsso_service.go
```diff
@@ -148,13 +147,17 @@ func (r *redhatssoService) ListServiceAcc(accessToken string, ctx context.Contex
...
+
+func (r *redhatssoService) registerAgentServiceAccount(accessToken string, agentClusterId string) (*api.ServiceAccount, *errors.ServiceError) {
 	svcData, err := r.client.CreateServiceAccount(accessToken, agentClusterId, fmt.Sprintf("service account for agent on cluster %s", agentClusterId))
 	if err != nil {
-		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to update agent service account")
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to create agent service account")
 	}
-
 	return convertServiceAccountDataToAPIServiceAccount(&svcData), nil
 }
```
ziccardi, line 159: This method is called by both `RegisterKasFleetshardOperatorServiceAccount` and `RegisterConnectorFleetshardOperatorServiceAccount` but those two methods doesn't do anything else than calling it and returning its return value.

I think it would be better to:

1. put the code of `registerAgentServiceAccount` into `RegisterKasFleetshardOperatorServiceAccount` 
2. make `RegisterConnectorFleetshardOperatorServiceAccount` call `RegisterKasFleetshardOperatorServiceAccount`
3. delete `registerAgentServiceAccount`

I see two main benefits in doing that:

1. By reading the code it will be easy to understand they both do the same thing
2. We will avoid an useless call in half of the cases

ziccardi (commented):
Added a couple of comments

pkg/services/sso/redhatsso_service.go
akoserwal: For now, I will keep it. because we need to do some rework here once we test it with sso.rh.

akoserwal:
@ziccardi @valdar Please review and approve. 
This feature is still turned off with the feature flag. This work will require some more work after MGDSTRM-7910 & testing with sso

dhirajsb:
Other than my questions about clearer descriptions, LGTM

pkg/client/keycloak/config.go
machi1990: The same could be done for the mas_sso provider: https://github.com/akoserwal/kas-fleet-manager-1/blob/3a44f1500d408a7163b7f6b0967234e779e7c0f3/pkg/client/keycloak/config.go#L114-L133

WDYT? it could be something for future improvement

akoserwal: yes, agreed !!

PR #868, opened 2022-03-22
feat: renamed eval instance type to developer

pawelpaszki:
## Description
feat: renamed eval instance type to developer

## Verification Steps
1. Carefully inspect the code, clone the PR locally and confirm that there are no `eval` occurrences unless expected to be there. 

2. Check the kafka reconcilers and confirm that they will continue to operate as expected after this PR is merged

- [ ] Code Review completed


internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1800,6 +1800,8 @@ components:
           type: string
         instance_type:
           type: string
+        display_name:
+          type: string
```
machi1990, line 1804:
```suggestion
        instance_type_name:
          type: string
```
possible renaming suggestion to avoid confusion between the relationship of `name` vs `display_name` in the API model.

rareddy:
Should we be using the "trial" nomenclature instead of "developer"? BU mentioned in the future we will have a "developer plan" based "trial" instance later in the development cycle. I was thinking this may be confusing then.

internal/kafka/internal/api/public/api/openapi.yaml
machi1990: @pb82 what do you think of this renaming proposal?

internal/kafka/internal/presenters/kafka.go
```diff
@@ -86,3 +88,14 @@ func setBootstrapServerHost(bootstrapServerHost string) string {
...
+func getDisplayName(instanceType string) string {
+	switch instanceType {
+	case types.DEVELOPER.String():
+		return types.DEVELOPER_DISPLAY_NAME.String()
+	case types.STANDARD.String():
+		return types.STANDARD_DISPLAY_NAME.String()
+	default:
+		return ""
+	}
+}
```
ziccardi, line 101: I believe this method should be part of `KafkaInstanceType`

internal/kafka/internal/services/kafka.go
```diff
@@ -193,30 +194,30 @@ func (k *kafkaService) AssignInstanceType(kafkaRequest *dbapi.KafkaRequest) (typ
...
 		var count int64
 		if err := dbConn.Model(&dbapi.KafkaRequest{}).
-			Where("instance_type = ?", types.EVAL).
+			Where("instance_type = ?", types.DEVELOPER).
 			Where("owner = ?", kafkaRequest.Owner).
 			Where("organisation_id = ?", kafkaRequest.OrganisationId).
 			Count(&count).
 			Error; err != nil {
-			return "", errors.NewWithCause(errors.ErrorGeneral, err, "failed to count kafka eval instances")
+			return "", errors.NewWithCause(errors.ErrorGeneral, err, "failed to count kafka developer instances")
```
ziccardi, line 216: If these errors are returned to the caller (maybe in the UI), I believe we should use the `display name` instead of `developer` here.

internal/kafka/internal/services/kafka.go
```diff
@@ -193,30 +194,30 @@ func (k *kafkaService) AssignInstanceType(kafkaRequest *dbapi.KafkaRequest) (typ
...
 		}
 
 		if count > 0 {
-			return "", errors.TooManyKafkaInstancesReached("only one eval instance is allowed")
+			return "", errors.TooManyKafkaInstancesReached("only one developer instance is allowed")
```
ziccardi, line 220: See previous comment

openapi/kas-fleet-manager.yaml
```diff
@@ -890,6 +890,8 @@ components:
               type: string
             instance_type:
               type: string
+            display_name:
+              type: string
```
machi1990, line 894: The suggestion in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/868#discussion_r832277564 of changing the `display_name` to `instance_type_name` or `instance_type_display_name` may have to be applied here if accepted.

internal/kafka/internal/api/public/api/openapi.yaml
pawelpaszki: @pb82 - any opinion?

internal/kafka/internal/api/public/model_kafka_request.go
```diff
@@ -33,6 +33,7 @@ type KafkaRequest struct {
 	FailedReason                string    `json:"failed_reason,omitempty"`
 	Version                     string    `json:"version,omitempty"`
 	InstanceType                string    `json:"instance_type,omitempty"`
+	DisplayName                 string    `json:"display_name,omitempty"`
```
pb82, line 36: Why is `display_name` part of the request?

internal/kafka/internal/api/public/api/openapi.yaml
pb82: But do `name` and `display_name` appear in the same APIs?

machi1990: Yes. 
When you do a `GET kafkas` or `GET kafkas by ID`  name and display_name will appear on the returned Kafkas.

internal/kafka/internal/kafkas/types/types.go
```diff
@@ -5,22 +5,33 @@ import "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/ocm"
...
+	return string(t)
+}
+
 func (t KafkaInstanceType) GetQuotaType() ocm.KafkaQuotaType {
 	if t == STANDARD {
 		return ocm.StandardQuota
 	}
-	return ocm.EvalQuota
+	return ocm.DeveloperQuota
 }
```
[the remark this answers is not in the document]
pb82: But why keep the `EvalQuota` when there will be no more Eval in the future. I think we discussed this and decided to also clean up the naming.

internal/kafka/internal/api/public/model_kafka_request.go
pawelpaszki: its a part of the response actually. the naming isn't the best in this case

internal/kafka/internal/api/public/api/openapi.yaml
pb82: I see. Can we make it so that `instance_type_name` is what we return from the endpoint, while in app-interface it's still `display_name`?

pawelpaszki: `I see. Can we make it so that instance_type_name is what we return from the endpoint, while in app-interface it's still display_name?` - yes

internal/kafka/internal/services/kafka.go
pawelpaszki: done

internal/kafka/internal/services/kafka.go
pawelpaszki: done

internal/kafka/internal/presenters/kafka.go
pawelpaszki: this function is now removed (values are obtained from the kafkaConfig)

pawelpaszki:
> Should we be using the "trial" nomenclature instead of "developer"? BU mentioned in the future we will have a "developer plan" based "trial" instance later in the development cycle. I was thinking this may be confusing then.

@pb82 - can you reply please?

internal/kafka/internal/api/public/api/openapi.yaml
pawelpaszki: changed to `instance_type_name`

pawelpaszki:
@pb82 @machi1990 @ziccardi - I think I addressed your comments

internal/kafka/internal/api/public/api/openapi.yaml
machi1990: Thanks.

machi1990:
> @pb82 @machi1990 @ziccardi - I think I addressed your comments

Perfect, thanks. My feedback has been addressed, I'll let others do another round of reviews on this.

JameelB:
Should we migrate existing eval Kafkas to developer?

ziccardi:
> Should we migrate existing eval Kafkas to developer?

I don't think so. I believe we will let them expire.

machi1990:
> > Should we migrate existing eval Kafkas to developer?
> 
> I don't think so. I believe we will let them expire.

Unless we temporarily (for 48 hours) stop the creation of new eval instances.

pb82:
@JameelB @machi1990 @ziccardi the idea is to let them expire while not blocking the creation of new instances during that time. The reconciler should ignore unknown instance types. This way the existing eval instances keep running but are no longer reconciled.

machi1990:
> @JameelB @machi1990 @ziccardi the idea is to let them expire while not blocking the creation of new instances during that time. The reconciler should ignore unknown instance types. This way the existing eval instances keep running but are no longer reconciled.

Thanks @pb82 

My concern will be this change: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/868/files#diff-857088f5fd8df40a086e707a60a4cb3bbde2af4983617de894c5cffb9a51dabdR36. 
If we do not perform a database migration of instance type from `eval` to `developer`, logic like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f5b97f86ba41994489acf14f4a9cb96719c4319/internal/kafka/internal/services/kafka.go#L1008 will break if the eval `id` is not in the supported instance sizes list.

ziccardi:
@machi1990 
> My concern will be this change: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/868/files#diff-857088f5fd8df40a086e707a60a4cb3bbde2af4983617de894c5cffb9a51dabdR36. If we do not perform a database migration of instance type from `eval` to `developer`, logic like
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f5b97f86ba41994489acf14f4a9cb96719c4319/internal/kafka/internal/services/kafka.go#L1008
> 
> will break if the eval `id` is not in the supported instance sizes list.

Good point. I think we could either ignore the error there (just log it) and return an empty config (not ideal) or add a 'deprecated' attribute into the `instance_type` section. If an instance type is deprecated, we will refuse to create them, but we will still have their configuration available.

machi1990:
> @machi1990
> 
> > My concern will be this change: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/868/files#diff-857088f5fd8df40a086e707a60a4cb3bbde2af4983617de894c5cffb9a51dabdR36. If we do not perform a database migration of instance type from `eval` to `developer`, logic like
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f5b97f86ba41994489acf14f4a9cb96719c4319/internal/kafka/internal/services/kafka.go#L1008
> > 
> > will break if the eval `id` is not in the supported instance sizes list.
> 
> Good point. I think we could either ignore the error there (just log it) and return an empty config (not ideal) or add a 'deprecated' attribute into the `instance_type` section. If an instance type is deprecated, we will refuse to create them, but we will still have their configuration available.

So, the link above is post the creation of a Kafka: that's the part where the Kafka is sent to fleetshard for provisioning. 
We cannot just ignore it as that might cause issues on the dataplane. We'll have to have a proper migration plan. And.... (**_continued down_**)
  
> Or migrate them: probably it will be just a database update? 

And a database migration seems the easiest to me.

Another option is to make sure that `eval` is not removed from that list. Any creation of new kafkas that references `eval` will be changed to `developer`. This has a lot of code complexities.
Do we use the instance_type anywhere else? I think however that soon or later we should however think about a deprecation policy.

pb82:
How would we migrate the database? Since this is not about adding a new field, can gorm migrations also deal with updating the value of an existing field if that field has a certain value?

machi1990:
> How would we migrate the database? Since this is not about adding a new field, can gorm migrations also deal with updating the value of an existing field if that field has a certain value?

Yes gorm can handle it. It will running the UPDATE query similar to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f5b97f86ba41994489acf14f4a9cb96719c4319/internal/kafka/internal/migrations/20210928110000_reset_canary_service_account_with_two_dashes.go#L17 changing `instance_type` to `developer` if type is `eval`.

pb82:
@machi1990 that seems like a good option then. I hadn't considered that there would not be a way to delete the expired instances.

machi1990:
> @machi1990 that seems like a good option then. 

+100 Yes. @pawelpaszki @ziccardi @JameelB how does that sound to you? 

> I hadn't considered that there would not be a way to delete the expired instances.

For an instance to be fully deleted, we need to send the deletion info to the data plane for it to be removed there.
So if we ignore `eval` instances here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f5b97f86ba41994489acf14f4a9cb96719c4319/internal/kafka/internal/services/kafka.go#L1008, those instances won't be sent to the data plane via the ManagedKafkaCR, and they'll stay there forever and in consequence in our database with a `deprovision` state.

pawelpaszki:
> > @machi1990 that seems like a good option then.
> 
> +100 Yes. @pawelpaszki @ziccardi @JameelB how does that sound to you?
> 
> > I hadn't considered that there would not be a way to delete the expired instances.
> 
> For an instance to be fully deleted, we need to send the deletion info to the data plane for it to be removed there. So if we ignore `eval` instances here:
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5f5b97f86ba41994489acf14f4a9cb96719c4319/internal/kafka/internal/services/kafka.go#L1008
> 
> , those instances won't be sent to the data plane via the ManagedKafkaCR, and they'll stay there forever and in consequence in our database with a `deprovision` state.

makes sense to migrate in this case

ziccardi:
> > @machi1990 that seems like a good option then.
> 
> +100 Yes. @pawelpaszki @ziccardi @JameelB how does that sound to you?
> 

100% agree on the migration approach

pb82:
When we migrate from `eval` to `developer` then we also don't need to modify the reconciler.

internal/kafka/internal/migrations/20220325111000_rename_eval_kafka_instance_type.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func renameEvalKafkaInstanceType() *gormigrate.Migration {
+	return &gormigrate.Migration{
+		ID: "20220325111000",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.Unscoped().Exec("UPDATE kafka_requests SET instance_type='developer' where instance_type='eval';").Error
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return nil
```
machi1990, line 15:
```suggestion
			return tx.Unscoped().Exec("UPDATE kafka_requests SET instance_type='eval' where instance_type='developer';").Error
```
Maybe the rollback will be the opposite.

pawelpaszki: I thought the same. Btw. I think this for some reason doesn't work, e.g. when I have an `eval` kafka, its `instance_type` is not changed. do you know why it might be?

machi1990: The migration worked for me locally. How are you running the migration?

pawelpaszki: I just started a service with `eval` kafka in the database. how did you run yours?

machi1990: I see. 
If you start the service with `make run`, it will execute the migrations for you if they weren't executed: there is not entry for `20220325111000` in the `migrations` table.
Alternative, if you run the binary, you could do:
```shell
./kas-fleet-manager migrate #run the migration
./kas-fleet-manager serve #run the service
```

pawelpaszki: got it. thanks!

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 type KafkaInstanceType struct {
-	Id    string              `yaml:"id"`
-	Sizes []KafkaInstanceSize `yaml:"sizes"`
+	Id          string              `yaml:"id"`
+	DisplayName string              `yaml:"display_name"`
```
JameelB, line 16: Is `display_name` a required parameter? Do we want to add validation for this to ensure it's specified?

config/kafka-instance-types-configuration.yaml
```diff
@@ -1,6 +1,6 @@
-# This configuration defines a list of supported Kafka instance types and their available sizes for KAS Fleet Manager.
+# This configuration defines a list of supported Kafka instance types and their available sizes for KAS Fleet Manager. 
 # Modify this file to add additional Kafka instance types and sizes or change the default values listed below.
-#
+# 
 # The following properties must be defined for each Kafka instance type:
 #   - id: Identifier for the Kafka instance type. Each instance type name should be unique.
```
JameelB, line 5: Let's add the `display_name` description in the docs here.

config/kafka-instance-types-configuration.yaml
```diff
@@ -32,7 +33,8 @@ supported_instance_types:
       quotaConsumed: 1
       quotaType: "rhosak"
       capacityConsumed: 1
-  - id: eval
+  - id: developer
+    display_name: Trial
```
ziccardi, line 37: Why putting the 'display name' into the configuration? IMO the display name is decided by the consumer, here we should have only the `id`

ziccardi: ignore this.

internal/kafka/internal/config/kafka_supported_instance_types.go
pawelpaszki: done. let me know if that's sufficient

config/kafka-instance-types-configuration.yaml
pawelpaszki: done

internal/kafka/internal/config/kafka_supported_instance_types.go
JameelB: looks good, thanks!

JameelB (approved):
Verified locally. 
* Migration was successful
* Able to create developer instead of eval instances

PR #885, opened 2022-03-30
Saving client secret into database

ziccardi:
## Description
With the new SSO, client secret needs to be saved into the database because, after creation, it can't be retrieved anymore.

PR contains 2 commits:

- one contains changes generated by `make generate`
- other one contains changes for this PR

## Verification Steps
Check that when a cluster is created, the secret is saved into the database and doesn't gets recreated many times.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kas_fleetshard_operator_addon.go
+++ b/internal/kafka/internal/services/kas_fleetshard_operator_addon.go
@@ -20,7 +20,7 @@ const (
 	//parameter names for the kas-fleetshard-operator service account
 	kasFleetshardOperatorParamMasSSOBaseUrl        = "sso-auth-server-url"
 	KasFleetshardOperatorParamServiceAccountId     = "sso-client-id"
-	kasFleetshardOperatorParamServiceAccountSecret = "sso-secret"
+	KasFleetshardOperatorParamServiceAccountSecret = "sso-secret"
 	// parameter names for the cluster id
 	kasFleetshardOperatorParamClusterId = "cluster-id"
 	// parameter names for the control plane url
@@ -70,7 +70,7 @@ type kasFleetshardOperatorAddon struct {
 
 func (o *kasFleetshardOperatorAddon) Provision(cluster api.Cluster) (bool, ParameterList, *errors.ServiceError) {
 	kasFleetshardAddonID := o.OCMConfig.KasFleetshardAddonID
-	params, paramsErr := o.getAddonParams(cluster)
+	params, paramsErr := o.getAddonParams(&cluster)
 	if paramsErr != nil {
 		return false, nil, paramsErr
 	}
@@ -95,7 +95,7 @@ func (o *kasFleetshardOperatorAddon) Provision(cluster api.Cluster) (bool, Param
 
 func (o *kasFleetshardOperatorAddon) ReconcileParameters(cluster api.Cluster) (ParameterList, *errors.ServiceError) {
 	kasFleetshardAddonID := o.OCMConfig.KasFleetshardAddonID
-	params, paramsErr := o.getAddonParams(cluster)
+	params, paramsErr := o.getAddonParams(&cluster)
 	if paramsErr != nil {
 		return nil, paramsErr
 	}
@@ -122,12 +122,16 @@ func (o *kasFleetshardOperatorAddon) ReconcileParameters(cluster api.Cluster) (P
 	}
 }
 
-func (o *kasFleetshardOperatorAddon) getAddonParams(cluster api.Cluster) ([]types.Parameter, *errors.ServiceError) {
-	acc, pErr := o.provisionServiceAccount(cluster.ClusterID)
-	if pErr != nil {
-		return nil, errors.GeneralError("failed to create service account for cluster %s due to error: %v", cluster.ClusterID, pErr)
+func (o *kasFleetshardOperatorAddon) getAddonParams(cluster *api.Cluster) ([]types.Parameter, *errors.ServiceError) {
+	var acc *api.ServiceAccount
+	if cluster.ClientID == "" || cluster.ClientSecret == "" {
+		var pErr *errors.ServiceError
+		acc, pErr = o.provisionServiceAccount(cluster.ClusterID)
+		if pErr != nil {
+			return nil, errors.GeneralError("failed to create service account for cluster %s due to error: %v", cluster.ClusterID, pErr)
+		}
 	}
-	params := o.buildAddonParams(acc, cluster.ClusterID)
+	params := o.buildAddonParams(cluster, acc)
 	return params, nil
 }
 
@@ -136,7 +140,19 @@ func (o *kasFleetshardOperatorAddon) provisionServiceAccount(clusterId string) (
 	return o.SsoService.RegisterKasFleetshardOperatorServiceAccount(clusterId)
 }
 
-func (o *kasFleetshardOperatorAddon) buildAddonParams(serviceAccount *api.ServiceAccount, clusterId string) []types.Parameter {
+func (o *kasFleetshardOperatorAddon) buildAddonParams(cluster *api.Cluster, serviceAccount *api.ServiceAccount) []types.Parameter {
+
+	var clientId string
+	var clientSecret string
+
+	if cluster.ClientID != "" && cluster.ClientSecret != "" {
+		clientId = cluster.ClientID
+		clientSecret = cluster.ClientSecret
+	} else {
+		clientId = serviceAccount.ClientID
+		clientSecret = serviceAccount.ClientSecret
+	}
+
 	p := []types.Parameter{
 
 		{
@@ -145,11 +161,11 @@ func (o *kasFleetshardOperatorAddon) buildAddonParams(serviceAccount *api.Servic
 		},
 		{
 			Id:    KasFleetshardOperatorParamServiceAccountId,
-			Value: serviceAccount.ClientID,
+			Value: clientId,
 		},
 		{
-			Id:    kasFleetshardOperatorParamServiceAccountSecret,
-			Value: serviceAccount.ClientSecret,
+			Id:    KasFleetshardOperatorParamServiceAccountSecret,
+			Value: clientSecret,
 		},
 		{
 			Id:    kasFleetshardOperatorParamControlPlaneBaseURL,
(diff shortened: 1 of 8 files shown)


machi1990:
Thanks @ziccardi for this.

The two commits do not seem related. Is it okay to open a separate PR for the first commit so that we can get that one in? I think that is something that is worth getting in irrespective of the changes and the review time it might take on this current PR one.

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -659,7 +659,11 @@ func (c *ClusterManager) reconcileAddonOperator(provisionedCluster api.Cluster)
...
 	if strimziOperatorIsReady && kasFleetshardOperatorIsReady && (clusterLoggingOperatorIsReady || c.OCMConfig.ClusterLoggingOperatorAddonID == "") {
 		glog.V(5).Infof("Set cluster status to %s for cluster %s", api.ClusterWaitingForKasFleetShardOperator, provisionedCluster.ClusterID)
-		if err := c.ClusterService.UpdateStatusAndClientId(provisionedCluster, api.ClusterWaitingForKasFleetShardOperator, params.GetParam(services.KasFleetshardOperatorParamServiceAccountId)); err != nil {
+		if err := c.ClusterService.
+			UpdateStatusAndClient(provisionedCluster,
+				api.ClusterWaitingForKasFleetShardOperator,
+				params.GetParam(services.KasFleetshardOperatorParamServiceAccountId),
+				params.GetParam(services.KasFleetshardOperatorParamServiceAccountSecret)); err != nil {
 			return errors.Wrapf(err, "failed to update local cluster %s status: %s", provisionedCluster.ClusterID, err.Error())
 		}
```
machi1990, line 668: Let's also make sure that `ready` and `waiting_for_kas_fleetshard_operator` clusters are also updated:
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/43ba6d92a9af9fcffedde181b37c66a0847e76cc/internal/kafka/internal/workers/clusters_mgr.go#L596-L601
That's where it is happening.

machi1990 (changes requested):
i think we should also update the reconciler method that handles `ready` and `waiting_for_kas_fleetshard_operator` clusters.

internal/kafka/internal/workers/clusters_mgr.go
ziccardi: Good catch! Should be fixed now

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -595,7 +595,7 @@ func (c *ClusterManager) reconcileKasFleetshardOperator(cluster api.Cluster) err
 	} else {
 		if cluster.ClientID == "" {
```
machi1990, line 596:
```suggestion
		if cluster.ClientID == "" || cluster.ClientSecret == "" {
```

I think this should be checking the two as we need to update both.

internal/kafka/internal/services/kas_fleetshard_operator_addon.go
```diff
@@ -136,7 +144,19 @@ func (o *kasFleetshardOperatorAddon) provisionServiceAccount(clusterId string) (
 	return o.SsoService.RegisterKasFleetshardOperatorServiceAccount(clusterId)
 }
 
-func (o *kasFleetshardOperatorAddon) buildAddonParams(serviceAccount *api.ServiceAccount, clusterId string) []types.Parameter {
+func (o *kasFleetshardOperatorAddon) buildAddonParams(cluster *api.Cluster, serviceAccount *api.ServiceAccount) []types.Parameter {
+
+	var clientId string
+	var clientSecret string
+
+	if cluster.ClientID != "" {
```
machi1990, line 152:
```suggestion
	if cluster.ClientID != "" && cluster.ClientSecret {
```
We need to also check that the client_id and client_secret are there.

internal/kafka/internal/services/kas_fleetshard_operator_addon.go
```diff
@@ -122,12 +126,16 @@ func (o *kasFleetshardOperatorAddon) ReconcileParameters(cluster api.Cluster) (P
 	}
 }
 
-func (o *kasFleetshardOperatorAddon) getAddonParams(cluster api.Cluster) ([]types.Parameter, *errors.ServiceError) {
-	acc, pErr := o.provisionServiceAccount(cluster.ClusterID)
-	if pErr != nil {
-		return nil, errors.GeneralError("failed to create service account for cluster %s due to error: %v", cluster.ClusterID, pErr)
+func (o *kasFleetshardOperatorAddon) getAddonParams(cluster *api.Cluster) ([]types.Parameter, *errors.ServiceError) {
+	var acc *api.ServiceAccount
+	if cluster.ClientID == "" {
```
machi1990, line 131:
```suggestion
	if cluster.ClientID == "" || cluster.ClientSecret == "" {
```
let's check for the two of them.

pkg/api/cluster_types.go
```diff
@@ -143,6 +143,7 @@ type Cluster struct {
 	IdentityProviderID string        `json:"identity_provider_id"`
 	ClusterDNS         string        `json:"cluster_dns"`
 	ClientID           string        `json:"client_id"`
+	ClientSecret       string        `json:"client_secret"`
```
machi1990, line 146: I am wondering if these two should be named like:
```suggestion
	KasFleetshardClientID           string        `json:"kas_fleetshard_client_id"`
	KasFleetshardClientSecret       string        `json:"kas_fleetshard_client_secret"`
```

I am suggesting that because client_id and client_secret are two generic and also in the long run, I feel like the same approach will need to be done for the cluster identity provider client id and client secret - https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8ac8f3584c4f86f712a0833449df7976cd3dbf43/internal/kafka/internal/workers/clusters_mgr.go#L1089 so if we keep the client_id, client_secret naming it might be confusing, it might lead to questions like "which client id, or which client secret"
@ziccardi @akoserwal 


this might require a database migration for the `client_id` as it exists already.

internal/kafka/internal/services/kas_fleetshard_operator_addon.go
```diff
@@ -68,9 +68,13 @@ type kasFleetshardOperatorAddon struct {
 	KeycloakConfig      *keycloak.KeycloakConfig
 }
 
+func (o *kasFleetshardOperatorAddon) ProvisionServiceAccount(cluster api.Cluster) (*api.ServiceAccount, *errors.ServiceError) {
+	return o.SsoService.RegisterKasFleetshardOperatorServiceAccount(cluster.ClusterID)
+}
```
machi1990, line 73: This is not needed, correct?
```suggestion
```

machi1990 (changes requested):
Thanks @ziccardi left a couple of other considerations

machi1990:
Is there any change that is needed for the connector side, or they are saving the client id and secrets already and won't be impacted by the eventual constraints of moving to sso.redhat? @ziccardi @akoserwal @lburgazzoli @dhirajsb @valdar

valdar:
@dhirajsb correct me if I am wrong but I fear we need to store the secret too...

dhirajsb:
Yes, we currently only store the client id. It needs to be updated to store both. Also, the service account is currently lazily created in the cluster handler when the add-on parameters call happens. We could move that to cluster service create call to save them with the new cluster in the DB.

valdar:
> Yes, we currently only store the client id. It needs to be updated to store both. Also, the service account is currently lazily created in the cluster handler when the add-on parameters call happens. We could move that to cluster service create call to save them with the new cluster in the DB.

should we open an issue and better coordinate there? linking here?

dhirajsb:
> won't be impacted by the eventual constraints of moving to sso.redhat?

@machi1990 is it the case that secrets are only available at account creation time in sso.redhat? Looks like it from the sso.redhat Keycloak client but I want to confirm.

machi1990:
> > won't be impacted by the eventual constraints of moving to sso.redhat?
> 
> @machi1990 is it the case that secrets are only available at account creation time in sso.redhat? Looks like it from the sso.redhat Keycloak client but I want to confirm.

Yes, this is correct. Secrets are only available at creation time.

internal/kafka/internal/services/kas_fleetshard_operator_addon.go
ziccardi: Correct. It is a leftover.

dhirajsb:
Created JIRA issue for Connectors change https://issues.redhat.com/browse/MGDCTRS-717
and pull request to add client_secret column https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/891

dhirajsb:
@ziccardi can you also run `make generate`? There are some compilation errors in `internal/kafka` generated code.

akoserwal:
> https://issues.redhat.com/browse/MGDCTRS-717

yes, with sso we will only get the secret when the account is created.

machi1990:
> Thanks @ziccardi for this.
> 
> The two commits do not seem related. Is it okay to open a separate PR for the first commit so that we can get that one in? I think that is something that is worth getting in irrespective of the changes and the review time it might take on this current PR one.

I noticed that the first commit was dropped without the PR being created - I created one here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/892 by running `make generate code/fix`.

akoserwal:
> Yes, we currently only store the client id. It needs to be updated to store both. Also, the service account is currently lazily created in the cluster handler when the add-on parameters call happens. We could move that to cluster service create call to save them with the new cluster in the DB.

yes, I agree

ziccardi:
> @ziccardi can you also run `make generate`? There are some compilation errors in `internal/kafka` generated code.

I did, but we preferred to move that to its own PR here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/892

pkg/api/cluster_types.go
machi1990: Sure. We can do it when we start storing the kafka sre client id/secrets. We just need to remember that.

machi1990 (approved):
lgtm, thanks for addressing the feedback @ziccardi 

leaving it to @akoserwal for final review.

akoserwal (approved):

PR #886, opened 2022-03-30
feat: added array utilities

ziccardi:
## Description
Added the following array utilities:

- **func FindFirst(predicate func(x interface{}) bool, values ...interface{}) (int, interface{})**: find the first occurrence matching the passed in predicate into the list of received values. Won't be needed anymore when we will have _generics_.
This function accepts a `...interface{}` instead of an `[]interface{}` because this way it can be used seamlessly with strings, int, int64, objects, etc. The drawbacks are:
    - the values are converted to a []interface
    - returned value is an interface{}
- **func FindFirstString(values []string, predicate func(x string) bool) (int, string)**: returns the index of the first string matching the passed in predicate and the string itself. `-1` and empty string if not found
- **func FilterStringSlice(values []string, predicate func(x string) bool) []string**: returns a slice of all the string into the passed in array matching the given predicate
- **func FirstNonEmpty(values ...string) (string, error)**: returns the first non empty string between the passed in values. If not found returns an error and the empty string
- **func FirstNonEmptyOrDefault(defaultValue string, values ...string) string**: returns the first non empty string between the passed in values. If not found returns `defaultValue`

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services/sso"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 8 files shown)


machi1990 (commented):
It will be good to use some of these in some places in the codebase e.g the parts discussed in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/813#discussion_r832075772, wdyt?

ziccardi:
> It will be good to use some of these in some places in the codebase e.g the parts discussed in [#813 (comment)](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/813#discussion_r832075772), wdyt?

Yes, but I was thinking about another PR for that. Do you think I should add some here as an example usage?

machi1990:
> > It will be good to use some of these in some places in the codebase e.g the parts discussed in [#813 (comment)](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/813#discussion_r832075772), wdyt?
> 
> Yes, but I was thinking about another PR for that. Do you think I should add some here as an example usage?

Yes. I think having some will be good. But if you have another PR lined up for this as well, that's also perfect. Thank you so much, this is really great work.

ziccardi:
> Yes. I think having some will be good. But if you have another PR lined up for this as well, that's also perfect. Thank you so much, this is really great work.

Added a commit with some code refactored using the new utility function

machi1990 (approved):
thanks @ziccardi 

lgtm.

The docs will be added afterwards, correct?

PR #891, opened 2022-03-31
feat(connectors): fixes MGDCTRS-717, save agent serviceaccount secret with connector cluster

dhirajsb:
## Description
Saves serviceaccount client secret in connector_cluster on cluster creation to avoid looking it up when providing addon parameters.

## Verification Steps
Cluster creation tests in integration tests MUST continue to work as before. 

- [ ] Code Review completed

### DIFF
--- a/internal/connector/internal/handlers/connector_cluster.go
+++ b/internal/connector/internal/handlers/connector_cluster.go
@@ -4,7 +4,6 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/services/authz"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/keycloak"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/ocm"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/server"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services/sso"
@@ -39,7 +38,6 @@ type ConnectorClusterHandler struct {
 	ConnectorTypes     services.ConnectorTypesService
 	ConnectorNamespace services.ConnectorNamespaceService
 	Vault              vault.VaultService
-	KeycloakConfig     *keycloak.KeycloakConfig
 	ServerConfig       *server.ServerConfig
 	AuthZ              authz.AuthZService
 	QuotaConfig        *config.ConnectorsQuotaConfig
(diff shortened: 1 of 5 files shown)


dhirajsb:
@valdar please review and let me know what you think.

internal/connector/internal/handlers/connector_cluster.go
```diff
@@ -244,22 +245,22 @@ func (o *ConnectorClusterHandler) buildAddonParams(serviceAccount *api.ServiceAc
...
-			Value: serviceAccount.ClientSecret,
+			Value: cluster.ClientSecret,
 		},
 	}
 	return p
 }
 
-func (o *ConnectorClusterHandler) buildTokenURL(serviceAccount *api.ServiceAccount) (string, error) {
+func (o *ConnectorClusterHandler) buildTokenURL(cluster dbapi.ConnectorCluster) (string, error) {
 	u, err := url.Parse(o.KeycloakConfig.KafkaRealm.TokenEndpointURI)
```
akoserwal, line 259: "KeycloakConfig.KafkaRealm.TokenEndpointURI" this needs to be set on basis of which sso provider is used. `https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/client/keycloak/config.go#L89`

if the provider is MAS_SSO use the KafkaRealm.
If the provider is REDHAT_SSO use the RedhatSSORealm

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/services/sso/keycloak.go#L26 will return the set sso service/provider realm

akoserwal, line 259: `KeycloakConfig.KafkaRealm.TokenEndpointURI`should be set on the basis of SelectSSOProvider
* if MAS_SSO then `KeycloakConfig.KafkaRealm.TokenEndpointURI`
* if REDHAT_SSO then `KeycloakConfig.RedhatSSORealm.TokenEndpointUR`

or use this which will return the set sso provider
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/services/sso/keycloak.go#L26

akoserwal: Something like: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/894

valdar (changes requested):
add issue for missing tests and address @akoserwal comments, other than that it lgtm.

dhirajsb:
There is already an integration test for addon parameters. The agent APIs in `connector-agent-api.feature` won't work without being able to retrieve the client id and secret from the new column in the DB.

dhirajsb:
@lburgazzoli ideally we should also rename the addon-parameters for base url and realm:
```
mas-sso-base-url - > sso-base-url
mas-sso-realm - > sso-realm
```
I'm updating this PR to use the sso service agnostic values. We can update these param names later to resolve https://issues.redhat.com/browse/MGDCTRS-721

dhirajsb:
@valdar I updated the new `client_secret` column to have a default value of '' to allow existing connector clusters in `dev` to continue working. Their addon parameters have already been generated, so no need for restoring the client secret in the DB. The clusters in `staging` are going to be deleted for a previous multi-tenancy migration so this won't be an issue.

valdar (approved):

PR #894, opened 2022-04-01
chore: adding method get selected sso provider realm config

akoserwal:
All instance like 
keycloakConfig.KafkaRealm.JwksEndpointURI
can be replace 
keycloakConfig.GetSSOProviderRealm().JwksEndpointURI

- [ ] Code Review completed

### DIFF
--- a/pkg/client/keycloak/config.go
+++ b/pkg/client/keycloak/config.go
@@ -57,9 +57,21 @@ func (c *KeycloakRealmConfig) setDefaultURIs(baseURL string) {
 	c.TokenEndpointURI = baseURL + "/auth/realms/" + c.Realm + "/protocol/openid-connect/token"
 }
 
+func (kc *KeycloakConfig) GetSSOProviderRealm() *KeycloakRealmConfig {
+	provider := kc.SelectSSOProvider
+	switch provider {
+	case MAS_SSO:
+		return kc.KafkaRealm
+	case REDHAT_SSO:
+		return kc.RedhatSSORealm
+	default:
+		return kc.KafkaRealm
+	}
+}
+
 func NewKeycloakConfig() *KeycloakConfig {
 	kc := &KeycloakConfig{
-		SsoBaseUrl:"https://sso.redhat.com",
+		SsoBaseUrl:                  "https://sso.redhat.com",
 		EnableAuthenticationOnKafka: true,
 		KafkaRealm: &KeycloakRealmConfig{
 			ClientIDFile:     "secrets/keycloak-service.clientId",


pkg/client/keycloak/config.go
```diff
@@ -57,9 +57,21 @@ func (c *KeycloakRealmConfig) setDefaultURIs(baseURL string) {
...
+	provider := kc.SelectSSOProvider
+	switch provider {
+	case MAS_SSO:
+		return kc.KafkaRealm
+	case REDHAT_SSO:
+		return kc.RedhatSSORealm
+	default:
+		return kc.KafkaRealm
+	}
+}
```
ziccardi, line 70: Why this method? We already have this: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c99eb533406db7e1e7e6b2c437eb617087895df5/pkg/services/sso/keycloak.go#L27

akoserwal: In some places, we have keycloakconfig direct references to get URLs current. For now, use this method & later refactor them to use GetRealmConfig() *keycloak.KeycloakRealmConfig  via keycloak service.

ziccardi: I believe we are accumulating too many of these 'for now'. Can we check if in those places where we have keycloakconfig we can't use the keycloakservice directly?

akoserwal: I'll create a jira to identify all the places & try to replace them with a keycloak service.

akoserwal: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/handlers/authentication.go#L29
Here don't make sense to inject service

ziccardi: It make sense [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/handlers/authentication.go#L14)
Then [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/handlers/authentication.go#L29) you can use:
```
KeysURL(keycloakService.GetRealmConfig().JwksEndpointURI) 
```

akoserwal: *authentication.HandlerBuilder: type sso.KeycloakService does not exist in the container

It will require a lot of changes to implement for injecting it with di

ziccardi: I believe you have to use `sso.KafkaKeyKloakService`.
Dunno why, but it has been remapper: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/7f22d08f523b76f072147a06ab028ed07f68c014/pkg/providers/core.go#L84

akoserwal: Even if you provide `sso.KafkaKeyKloakService.` di expect the authentication handler to implement the keycloak service. Replacing with Keycloak config with the keycloak service is also going to provide the same config object. I don't see any issues. One benefit would be we get the config of the selected provider. but there could be a chance we need to set multiple jwks URLs configured in keycloak config irrespective of the provider configured.

ziccardi: It's just a matter of duplicate code. We already have logic to switch between KeycloakService and RedhatSSOService and here we are duplicating the same logic.
If someone will need to fix a bug, add a provider, etc, it will be natural to update the builder, while with high probability no-one will check here unless they intentionally search for it. 
Anyway, mine was just a suggestion. I'm ok if you don't agree.

akoserwal: your suggestion is valid, I was thinking about some other use-cases.

machi1990:
> > > @dhirajsb @machi1990 @ziccardi: this will be required. especially around the integration test.
> > 
> > 
> > At the moment the `.GetSSOProviderRealm` method that has been added is not used anywhere, is that correct?
> 
> yes, it not been used any where

Okay thanks. Are the plans to use it somewhere? If so, I'd say we can do it in this same PR.

akoserwal:
make sense. I'll close this one.

PR #895, opened 2022-04-01
refactor: refactored the keycloak service builder

ziccardi:
## Description
The builder now just takes the configuration and returns the appropriate KeycloakService instance (Keycloak or RedhatSSO)

- [ ] Code Review completed

### DIFF
--- a/pkg/services/sso/keycloak_service_builder.go
+++ b/pkg/services/sso/keycloak_service_builder.go
@@ -1,0 +1,64 @@
+package sso
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/keycloak"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/redhatsso"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+)
+
+var _ KeycloakServiceBuilderSelector = &keycloakServiceBuilderSelector{}
+var _ KeycloakServiceBuilder = &keycloakServiceBuilder{}
+
+type KeycloakServiceBuilder interface {
+	WithRealmConfig(realmConfig *keycloak.KeycloakRealmConfig) KeycloakServiceBuilder
+	Build() KeycloakService
+}
+
+type keycloakServiceBuilder struct {
+	config      *keycloak.KeycloakConfig
+	realmConfig *keycloak.KeycloakRealmConfig
+}
+
+// Build returns an instance of KeycloakService ready to be used.
+// If a custom realm is configured (WithRealmConfig called), then always Keycloak provider is used
+// irrespective of the `builder.config.SelectSSOProvider` value
+func (builder *keycloakServiceBuilder) Build() KeycloakService {
+	notNilPredicate := func(x interface{}) bool {
+		return x.(*keycloak.KeycloakRealmConfig) != nil
+	}
+
+	// Temporary: if a realm configuration different from the one into the config is specified
+	// we always instantiate MAS_SSO irrespective of the selected provider
+	if builder.config.SelectSSOProvider == keycloak.MAS_SSO ||
+		builder.realmConfig != nil {
+		_, realmConfig := arrays.FindFirst(notNilPredicate, builder.realmConfig, builder.config.KafkaRealm)
+		return newKeycloakService(builder.config, realmConfig.(*keycloak.KeycloakRealmConfig))
+	} else {
+		_, realmConfig := arrays.FindFirst(notNilPredicate, builder.realmConfig, builder.config.RedhatSSORealm)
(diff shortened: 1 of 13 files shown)


pkg/services/sso/keycloak_service_builder.go
```diff
@@ -0,0 +1,55 @@
...
+	if builder.config.SelectSSOProvider == keycloak.MAS_SSO {
+		realmConfig := builder.realmConfig
+		if realmConfig == nil {
+			realmConfig = builder.config.KafkaRealm
+		}
+		return newKeycloakService(builder.config, realmConfig)
+	} else {
+		client := redhatsso.NewSSOClient(builder.config)
+		return &keycloakServiceProxy{
+			accessTokenProvider: client,
```
akoserwal, line 31: ``` realmConfig  = builder.config.RedhatSSORealm ```
do we need to define it here as well?

ziccardi: Currently redhatsso supports only one realconfig from the KetCloakConfig object.
If we need the ability to support a custom realm in redhatsso as we did in mas-sso, then I can make the change (it will be trivial)

akoserwal: make sense

ziccardi: It doesn't. Currently the RedhatSSO takes the realm directly from the config

ziccardi: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3773316adce8f9d17b62fcf90b97a72eaf59b716/pkg/client/redhatsso/client.go#L141

akoserwal (approved):

pkg/services/sso/keycloak_service_builder.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (builder *keycloakServiceBuilder) Build() KeycloakService {
+	if builder.config.SelectSSOProvider == keycloak.MAS_SSO {
+		realmConfig := builder.realmConfig
+		if realmConfig == nil {
+			realmConfig = builder.config.KafkaRealm
+		}
+		return newKeycloakService(builder.config, realmConfig)
+	} else {
```
akoserwal, line 28: can we add it here
if `c.OSDClusterIDPRealm` is passed. it should always initialise `newKeycloakService(builder.config, realmConfig)`

not the redhat sso client.

ziccardi: Not sure I get it. You mean that even if `SelectSSOProvider` is `Redhat_SSO`, if `c.OSDClusterIDRealm` is passed we create a MAS_SSO service anyway?

pkg/services/sso/keycloak_service_builder.go
```diff
@@ -0,0 +1,59 @@
...
+	config      *keycloak.KeycloakConfig
+	realmConfig *keycloak.KeycloakRealmConfig
+}
+
+func (builder *keycloakServiceBuilder) Build() KeycloakService {
+	if builder.config.SelectSSOProvider == keycloak.MAS_SSO {
+		realmConfig := builder.realmConfig
+		if realmConfig == nil {
+			realmConfig = builder.config.KafkaRealm
+		}
```
miguelsorianod, line 26: This part is duplicated. It can be de-duplicated and moved before the conditions

ziccardi: Was a bug. Second should have been `RedhatSSORealm`. Fixed and changed to use the new array utils

pkg/services/sso/keycloak_service_builder.go
akoserwal: yes, irrespective of what is set for "SelectSSOProvider". if `c.OSDClusterIDRealm` passed as realm config, keycloak service should be intiated

pkg/services/sso/keycloak_service_builder.go
```diff
@@ -0,0 +1,61 @@
...
+func (builder *keycloakServiceBuilder) Build() KeycloakService {
+	notNilPredicate := func(x interface{}) bool {
+		return x.(*keycloak.KeycloakRealmConfig) != nil
+	}
+
+	// Temporary: if a realm configuration different from the one into the config is specified
+	// we always instantiate MAS_SSO irrespective of the selected provider
+	if builder.config.SelectSSOProvider == keycloak.MAS_SSO ||
+		builder.realmConfig != nil {
+		_, realmConfig := arrays.FindFirst(notNilPredicate, builder.realmConfig, builder.config.KafkaRealm)
```
akoserwal, line 31: In this case, realm config will be `builder.config.KafkaRealm`?

ziccardi: It will be `builder.config.KafkaRealm` if `builder.realmConfig` is nil, otherwise it will be `builder.realmConfig`

PR #896, opened 2022-04-01
poc: support eval and developer poc

JameelB:
## Description
This is a POC for supporting both eval and developer instances with the larger kafka instances feature. It should not be merged.

Review is welcome though :)


pawelpaszki:
@JameelB - what about [deprovisioning expired kafkas](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/1c79ebacb3003824bc6fd064dfe2a12142874a4e/internal/kafka/internal/services/kafka.go#L558)? right now I think only developer instance is considered to be valid instance type to delete if expired. see [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/1c79ebacb3003824bc6fd064dfe2a12142874a4e/internal/kafka/internal/services/kafka.go#L41)

JameelB:
@pawelpaszki Thanks for pointing that out, didn't realize that it was only valid for Developer. We need to convert this to a list that includes both 'developer' and 'eval'.

It was also agreed upon that we only need to return 'eval' in the cloud provider region endpoint. We don't need to support the creation of 'eval' Kafkas. Will close this poc PR.

PR #898, opened 2022-04-05
fix: remove max_capacity_reached as required

wtrocki:
Remove max_capacity_reached - having it as required might not be valid as field is deprecated (so it should not be used)

Screenshot 2022-04-05 at 15 50 46


wtrocki:
CC @vmanley @JameelB

JameelB (commented):
Can you generate the openapi here as well? `make openapi/generate`

wtrocki:
Done

internal/kafka/internal/api/public/model_region_capacity_list_item.go
```diff
@@ -15,7 +15,7 @@ type RegionCapacityListItem struct {
 	InstanceType string `json:"instance_type"`
 	// flag indicating whether the capacity for the instance type in the region is reached
 	// Deprecated
-	DeprecatedMaxCapacityReached bool `json:"max_capacity_reached"`
+	DeprecatedMaxCapacityReached bool `json:"max_capacity_reached,omitempty"`
```
JameelB, line 18: Just note that because it's `omitempty` here, if `max_capacity_reached` was *false*, it won't show up in the response. 

Example:
```
"capacity": [
                {
                    "instance_type": "standard",
                    "available_sizes": [
                        "x1"
                    ]
                },
                {
                    "instance_type": "developer",
                    "available_sizes": [
                        "x1"
                    ]
                }
            ]
```

Is that ok with you?

wtrocki: Yes. No one will use it and sdks will filter out deprecated fields. On backend we still need those as users might choose to not update instantly to latest SDKs

wtrocki:
Because of how booleans are initialised I do not think this will cause any issue on parsing for golang and other langs:

Example to confim
```
package main

import (
	"encoding/json"
	"fmt"
)

// We'll use these two structs to demonstrate encoding and
// decoding of custom types below.
type Response1 struct {
	hasPage bool `json:"hasPage,omitempty"`
}

func main() {

	resp := Response1{false}
	res1B, _ := json.Marshal(resp)
	//fmt.Println(string(res1B))

	resp2 := Response1{false}
	err := json.Unmarshal(res1B, resp2)
	fmt.Println(err)
	fmt.Println(resp2.hasPage)
}

```

wtrocki:
I think we good to merge that

JameelB (approved):

PR #900, opened 2022-04-06
fix(connectors): fixes MGDCTRS-737, added flag unresolved_secrets to deployment API to indicate secrets could not be retrieved from vault service

dhirajsb:
## Description
Added flag unresolved_secrets to agent API. 

## Verification Steps
Integration test added to look for new flag for missing secrets. 

- [ ] Code Review completed


valdar (dismissed):
I am a little bit hesitant to merge this for a number of reasons: 

- not sure it is appropriate given the feature freeze we namely are in
- is that flag needed at all? We can just present the connectors without the spec part or with the connector part empty?

@lburgazzoli wdyt?

lburgazzoli:
I think this need a little bit more thinking as it is not clear to what is the expectation from the fleet shard side like: 
- should the fleetshard report this condition an an error ? 
- what should we do with running connectors ?

I think this fall into the discussion about states and conditions we started some time ago.

dhirajsb:
@valdar , @lburgazzoli I've added descriptions for this issue in the Epic and tasks in https://issues.redhat.com/browse/MGDCTRS-736
The descriptions in the manager and agent tasks explain the logic behind how this can work. Let's continue the discussion in JIRA and we can come back to this PR.

dhirajsb:
@lburgazzoli @valdar I updated the fix to use the status phase `secrets-unavailable` and not including the connector spec if the secrets cannot be retrieved from the vault.

internal/connector/internal/handlers/connector_agent.go
```diff
@@ -222,9 +222,34 @@ func (h *ConnectorClusterHandler) presentDeployment(r *http.Request, resource db
...
+	ctx := r.Context()
+	apiSpec, invalidSecrets, err := h.Service.GetConnectorWithBase64Secrets(ctx, resource)
 	if err != nil {
-		return private.ConnectorDeployment{}, err
+		if invalidSecrets {
+			// log error in getting secrets and signal that connector spec doesn't have secrets
+			glog.Errorf("Error getting connector %s with base64 secrets: %s", apiSpec.ID, err)
+
+			apiSpec.ConnectorSpec = []byte("{}")
+			connectorPhase = dbapi.ConnectorStatusPhaseSecretsUnavailable
```
lburgazzoli, line 240: this mean that the "missing-secrets" is ephemeral right ? si it is not persisted in the database

openapi/connector_mgmt.yaml
```diff
@@ -1175,6 +1175,7 @@ components:
         - deleted
         - provisioning
         - deprovisioning
+        - missing-secrets
```
lburgazzoli, line 1178: Is it correct to have `missing-secrets` as part of the public  APIs ?
nitpick, it should use snake case `missing_secrrets`

internal/connector/internal/handlers/connector_agent.go
```diff
@@ -222,9 +222,34 @@ func (h *ConnectorClusterHandler) presentDeployment(r *http.Request, resource db
...
+		} else {
+			return private.ConnectorDeployment{}, err
+		}
+	}
+
+	// check for phase change/reset due to secrets handling
+	if connectorPhase != resource.Status.Phase {
+		converted.Status.Phase = private.ConnectorState(connectorPhase)
+		resource.Status.Phase = connectorPhase
+		if err := h.Service.UpdateConnectorDeploymentStatus(ctx, resource.Status); err != nil {
```
dhirajsb, line 250: Phase updates to `secrets-missing` and reset back to `assigned` are persisted in the DB. Users will be able to see it in the UI.

internal/connector/internal/handlers/connector_agent.go
dhirajsb: See line 250 where it's persisted to DB.

internal/connector/test/integration/features/connector-agent-api.feature
```diff
@@ -584,11 +584,70 @@ Feature: connector agent API
...
+    #-----------------------------------------------------------------------------------------------------------------
+    # In this part of the Scenario we test whether agent gets deployments even when missing secrets in the vault
+    #-----------------------------------------------------------------------------------------------------------------
+    # save the service account secret in the client id and set the secret to an invalid value
+    Given I run SQL "UPDATE connectors SET service_account_client_id=service_account_client_secret, service_account_client_secret='missing-secret' WHERE id = '${connector_id}';" expect 1 row to be affected.
+    # check that the deployment update is still sent to the agent without a spec and phase set to 'secrets-unavailable'
+    Given I am logged in as "Shard"
+    Given I set the "Authorization" header to "Bearer ${shard_token}"
+    And I wait up to "5" seconds for a response event
+    Then the ".object.spec.connector_spec" selection from the response should match "null"
```
dhirajsb, line 596: @lburgazzoli note that the agent api is setting the spec to `{}`. But that gets marshaled to `nil` in it's json handling so `connector_spec` will be missing in the deployment.

openapi/connector_mgmt.yaml
valdar: In my opinion no, it should not leak into the api. It is more  like an implementation detail.

dhirajsb: As a user, why wouldn't you want to know that there is something wrong with your connector? What's the argument for hiding this condition from the user?

lburgazzoli: IMHO, I don't think this is a phase is more an error condition

internal/connector/internal/handlers/connector_agent.go
lburgazzoli: Since the status is owned by the fleet shard, would this be overridden by a status update from the sync ? 
It feel wrong to me that the fleet manager manipulate the deployment status.

openapi/connector_mgmt.yaml
dhirajsb: Ok, so when the agent sees this, will it respond with a `failed` phase and an error message in a status condition? I just want to make sure it's visible to the user in case it happens when the agent needs the spec for starting/updating. 
If that's the case, the fleetmanager can just pass this value along without updating the DB expecting the agent to circle back with an error. 

BTW, if the value is not part of the enum, won't it break API generated clients that might be validating the string?

lburgazzoli: > Ok, so when the agent sees this, will it respond with a `failed` phase and an error message in a status condition? I just want to make sure it's visible to the user in case it happens when the agent needs the spec for starting/updating. If that's the case, the fleetmanager can just pass this value along without updating the DB, expecting the agent to circle back with an error.
> 

Yes so, we can think about something like:
- if transitioning the connector to a stop/delete phase, then the presence of secrets may not be relevant hence, the action can be performed.
- if reify is required, then the process will fail and an error will be reported

In bot cases, we can add a condition that signal the lack of spec, like:

```yaml
- type: ConnectorSpecAvailable 
  status: "False"
  reason: "ConnectorSpecNotAvailable"
  message: "..."
```
And in case of failure, we can use the `Ready` condition to report the error.

> BTW, if the value is not part of the enum, won't it break API generated clients that might be validating the string?

If `missing-secrets` is not a phase, then I don't think we have such issue as the user should use the `failed` phase and eventually check the error.

dhirajsb:
```
If missing-secrets is not a phase, then I don't think we have such issue as the user should use the failed phase and eventually check the error.
```
Deployment status phase references `ConnectorState` enum. Is the client in the agent using Java enums or String for it?

lburgazzoli: enum

dhirajsb: If we don't want to treat `secrets_unavailable` as a phase, we should go back to what I did at first for this PR, i.e. make it a separate flag. That way we don't pollute the phase enum with a value that's used to convey something else. 

So, in case secrets are not available:
* connector spec will be empty
* deployment will include a flag `secrets_unavailable=true`

internal/connector/internal/handlers/connector_agent.go
dhirajsb: Yeah, I agree. It's violating the producer/consumer roles. I'm going to remove that.

dhirajsb: Removed fleetmanager DB update.

openapi/connector_mgmt.yaml
dhirajsb: Added required flag `resolved_secrets` as discussed.

internal/connector/internal/handlers/connector_agent.go
dhirajsb: Removed fleetmanager DB updates.

lburgazzoli (approved):

PR #901, opened 2022-04-06
fix: get the realm config from the service when building the CR

ziccardi:
## Description
We have 2 realms configured into the KeycloakService config. Only the service knows which one is the one to be used, so the realm config have to be requested to the service.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -12,7 +12,6 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/kafkas/types"
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/keycloak"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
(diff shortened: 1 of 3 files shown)


akoserwal (approved):

akoserwal:
@ziccardi Please can you also update this: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/internal/services/kas_fleetshard_operator_addon.go#L160

akoserwal (approved):

internal/kafka/internal/services/kafka.go
```diff
@@ -940,13 +939,16 @@ func BuildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 		Status: managedkafka.ManagedKafkaStatus{},
 	}
 
+	keycloakConfig := keycloakService.GetConfig()
+	keycloakRealmConfig := keycloakService.GetRealmConfig()
```
machi1990, line 943: The `GetRealmConfig` method might be too generic. Which realm? The service accounts realm or the sre realm?

ziccardi: Not sure I get it.
Currently in the `KafkaServiceConfiguration` we have 3 realms configurations:

- KafkaRealm
- OSDClusterIDPRealm
- RedhatSSORealm

However a single KeycloackService instance will use only one of them at a time. The `GetRealmConfig` method of the `KeycloakService` object returns the `RealmConfig` that the current `KeyclocakService` instance is using.

machi1990: > However a single KeycloakService instance will use only one of them. 

You are write. I thought that was changed but I was wrong.

> The GetRealmConfig method of the KeycloakService object returns the RealmConfig that the current KeyclocakService instance is using.

I agree. I see where I my reasoning was wrong. Thanks for the clarification.

PR #909, opened 2022-04-08
feat: allow single-az trial requests

vmanley:

- [ ] Code Review completed


internal/kafka/internal/handlers/validation.go
```diff
@@ -94,6 +94,19 @@ func ValidateCloudProvider(kafkaService *services.KafkaService, kafkaRequest *db
 	}
 }
 
+//ValidateAZOption returns a validator that returns error if incorrect multiAZ value is used
+func ValidateAZOption(kafkaRequest *dbapi.KafkaRequest) handlers.Validate {
+	return func() *errors.ServiceError {
+		if kafkaRequest.InstanceType == "standard" && !kafkaRequest.MultiAZ {
```
machi1990, line 100: We could use the defined constants in `"internal/kafka/internal/kafkas/types/types.go"` instead of hardcoded strings e.g  "standard", "eval"

internal/kafka/internal/handlers/validation.go
```diff
@@ -94,6 +94,19 @@ func ValidateCloudProvider(kafkaService *services.KafkaService, kafkaRequest *db
...
+			return errors.NotMultiAzActionNotSupported()
+		}
+		if kafkaRequest.InstanceType == "developer" && kafkaRequest.MultiAZ {
```
machi1990, line 103: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/909/files#r846152919

internal/kafka/internal/handlers/validation.go
vmanley: Thanks @machi1990 I'll use the constants

machi1990: perfect. I forgot to mention the added tests too

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -318,27 +318,110 @@ func Test_Validation_validateCloudProvider(t *testing.T) {
...
+	tests := []struct {
+		name string
+		args args
+		want result
+	}{
+		{
+			name: "success when standard type and mulit-AZ",
+			args: args{
+				kafkaRequest: dbapi.KafkaRequest{
+					InstanceType: "standard",
```
machi1990, line 360: same suggestion as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/909#discussion_r846152919 to be applied throughout the added test

internal/kafka/internal/handlers/validation.go
vmanley: Thanks I forgot to add them there

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -79,10 +79,10 @@ func Test_Validation_validateKafkaClusterNameIsUnique(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			gomega.RegisterTestingT(t)
+			RegisterTestingT(t)
```
ziccardi, line 82: This doesn't needs to be executed for each iteration. We can move it outside of the loop.

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -79,10 +79,10 @@ func Test_Validation_validateKafkaClusterNameIsUnique(t *testing.T) {
...
 			validateFn := ValidateKafkaClusterNameIsUnique(&tt.arg.name, tt.arg.kafkaService, tt.arg.context)
 			err := validateFn()
-			gomega.Expect(tt.want).To(gomega.Equal(err))
+			Expect(tt.want).To(Equal(err))
```
ziccardi, line 85: Parameter of `Expect` is the **actual** value, not the *expected one*.
This should be:
```suggestion
			Expect(err).To(Equal(tt.want))
```

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -127,13 +127,13 @@ func Test_Validations_validateKafkaClusterNames(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
-			gomega.RegisterTestingT(t)
+			RegisterTestingT(t)
```
ziccardi, line 130: See previous comment about this

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -318,27 +318,110 @@ func Test_Validation_validateCloudProvider(t *testing.T) {
...
+			want: result{
+				wantErr: true,
+				reason:  "Only multiAZ Kafkas are supported, use multi_az=true",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			RegisterTestingT(t)
```
ziccardi, line 410: See previous comment about this

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -318,27 +318,110 @@ func Test_Validation_validateCloudProvider(t *testing.T) {
...
+			validateFn := ValidateAZOption(&tt.args.kafkaRequest)
+			err := validateFn()
+			Expect(tt.want.wantErr).To(Equal(err != nil))
```
ziccardi, line 413:
```suggestion
			Expect(err != nil).To(Equal(tt.want.wantErr))
```

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -318,27 +318,110 @@ func Test_Validation_validateCloudProvider(t *testing.T) {
...
+			if !tt.want.wantErr && err != nil {
+				t.Errorf("ValidateAZOption() expected not to throw error but threw %v", err)
```
ziccardi, line 415: This is already checked with the previous 'Expect'.

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -499,14 +582,14 @@ func Test_Validation_ValidateKafkaUserFacingUpdateFields(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			gomega.RegisterTestingT(t)
+			RegisterTestingT(t)
```
ziccardi, line 585: See other comments about this

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -499,14 +582,14 @@ func Test_Validation_ValidateKafkaUserFacingUpdateFields(t *testing.T) {
...
 			validateFn := ValidateKafkaUserFacingUpdateFields(tt.arg.ctx, tt.arg.authService, tt.arg.kafka, &tt.arg.kafkaUpdateRequest)
 			err := validateFn()
-			gomega.Expect(tt.want.wantErr).To(gomega.Equal(err != nil))
+			Expect(tt.want.wantErr).To(Equal(err != nil))
 			if !tt.want.wantErr && err != nil {
 				t.Errorf("ValidateKafkaUserFacingUpdateFields() expected not to throw error but threw %v", err)
```
ziccardi, line 590: This is already checked in the previous `Expect`

internal/kafka/internal/handlers/validation_test.go
vmanley: Thanks @ziccardi I'll tidy that up in the tests in this file

internal/kafka/internal/handlers/validation_test.go
vmanley: I'll reverse where needed. Thanks @ziccardi

internal/kafka/internal/handlers/validation_test.go
vmanley: I didn't introduce this test so I'd rather not make a change here unless you feel strongly about removing this check. Is that ok @ziccardi ?

ziccardi: if `tt.want.wantErr` is false and err is not nil,
```go
Expect(tt.want.wantErr).To(Equal(err != nil))
```
will fail, so that test is useless. I believe it adds confusion.

vmanley: Would it make more sense if `Expect(tt.want.wantErr).To(Equal(err != nil))` was included after the if block? I understand that the check would show up unexpected errors which could be useful

ziccardi: The detail of the error can be logged with gomega too: https://onsi.github.io/gomega/#annotating-assertions

ziccardi: If you prefer to leave it, I'm ok with that: it won't harm. Maybe we could just track that it needs to be cleaned up somewhere and do it later

vmanley: Thanks @ziccardi I'll use annotations to give error details here

vmanley:
@ziccardi @pb82 could you have a look at changes made. Thanks

ziccardi (approved):
Awesome job!

PR #911, opened 2022-04-09
refactor: moved the state machine and string scanner into the utilities

ziccardi:
## Description

In Kas Fleet Manager, we have a parser we use to parse query string passed to the list endpoint.
The parser is composed by different component:
* The `State Machine`: this is used to validate the token composing the query string so that we can validate the query syntax
* The `String Scanner`: this is used to split the string into tokens, so that each token is managed by a single state in the StateMachine
* `The Parser`: this is the topmost object. This configure the StateMachine and the StringScanner to validate the grammar we are interested it (in our case, SQL Where statements).

Both the State Machine and the String Scanner are general purpose and can be used to perform other tasks (different parsers, validate flows, etc): for this reason, I moved them to the utility folder.

One example of an alternative usage can be seen in the `statemachine_test.go` file.

- [ ] Code Review completed


pkg/shared/utils/state_machine_parser/statemachine_builder.go
```diff
@@ -49,26 +49,24 @@ func (smb *stateMachineBuilder) Build() State {
...
+			Build()
 	}
 
 	// add all the transitions
 	for _, transition := range smb.grammar.Transitions {
 		token := tokenMap[transition.TokenName]
 		for _, targetTokenName := range transition.ValidTransitions {
 			targetToken := tokenMap[targetTokenName]
-			token.addNextState(targetToken)
+			token.(*state).addNextState(targetToken)
```
machi1990, line 69: Why the need to `.(*state).`?

machi1990:
Is there a reason why this is still draft?

ziccardi:
> Is there a reason why this is still draft?

Yes. Was still separating all the reusable utilities we have into the parser: the state machine and the scanner.

pkg/shared/utils/state_machine_parser/statemachine_builder.go
machi1990: So it is safe to say that not all the implementation of the `State` interface will have the `addNextState` method?
I wonder what happens then with this cast.

ziccardi: Nope. We will have only one implementation of the State interface, however, the State interface exposes only readonly methods.
There is no way to replace the internal representation of the State with another implementation.

machi1990: Okay. It make sense. To remove the temptation, I'd really remove that method from the *State* implementation and pass whatever that needs to be set to the State as a constructor argument i.e construction/instanciation only  argument. Because, this seems like what the code is doing here.

machi1990 (approved):
lgtm

I left a comment on *really* hiding the internal details thus avoid the cast. 
This is very minor and the the state machine builder already hides this somehow and its the only place with the cast.

PR #914, opened 2022-04-11
fix(connectors): fixes MGDCTRS-753, use optimistic locking in connector updates

dhirajsb:
## Description
Added optimistic locking using connector version in where condition to ensure the connector didn't change just before the update. 

## Verification Steps
CI tests should pass. 

- [ ] Code Review completed

### DIFF
--- a/internal/connector/internal/services/connectors.go
+++ b/internal/connector/internal/services/connectors.go
@@ -286,18 +286,17 @@ func selectConnectorWithConditions(dbConn *gorm.DB) *gorm.DB {
 func (k connectorsService) Update(ctx context.Context, resource *dbapi.Connector) *errors.ServiceError {
 
 	// If the version is set, then lets verify that the version has not changed...
-	if resource.Version != 0 {
-		dbConn := k.connectionFactory.New()
-		dbConn = dbConn.Where("id = ? AND version = ?", resource.ID, resource.Version)
-		t := dbapi.Connector{}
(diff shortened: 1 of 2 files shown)


internal/connector/internal/services/connectors.go
```diff
@@ -286,19 +286,18 @@ func selectConnectorWithConditions(dbConn *gorm.DB) *gorm.DB {
...
 	}
 
 	dbConn := k.connectionFactory.New()
-	if err := dbConn.Model(resource).Updates(resource).Error; err != nil {
+	update := dbConn.Model(resource).Where("id = ? AND version = ?", resource.ID, resource.Version).Updates(resource)
+	if err := update.Error; err != nil {
 		return errors.GeneralError("failed to update: %s", err.Error())
 	}
+	if update.RowsAffected == 0 {
+		return errors.BadRequest("resource version changed")
```
lburgazzoli, line 299: should this be 409 (Conflict) ?

valdar (changes requested):
Can we add a basic test with 2 request with the same version and check that the second one fails?

internal/connector/internal/services/connectors.go
dhirajsb: agreed, changed to conflict.

valdar:
> Can we add a basic test with 2 request with the same version and check that the second one fails?

we can not in the easy way I described, unfortunately.

valdar (approved):

PR #920, opened 2022-04-12
Add speical mgmt org to keycloak config. Add canary client ID in custom claim check in Kafka CR

Rajagopalan-Ranganathan:
JIRA: https://issues.redhat.com/browse/MGDSTRM-7754

## Description
1) Add the special management org in the keycloak config. though not used now, this is
intended to be used later post migration - for the verification of fleetshard communication
back to KFM. Ex: Fleet shard client will be created in speical management org, and we need this
information to verify
2) Add Canary client ID in the custom claim check for Kafka CR. This improves the security that
only the canary created for this Kafka can authenticate against the broker.

## Verification Steps

1. Create a new KFM image/run KFM locally against a OSD cluster with the changes
2. Create a Kafka cluster
3. Verify the Kafka CR - custom claim check has the associated canary clientID 

Note: @akoserwal  and myself verified the changes over a call.

- [ ]  ~~Documentation added for the feature~~ 
- [ ] Code Review completed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -717,6 +717,7 @@ deploy/service: MAS_SSO_ENABLE_AUTH ?= "true"
 deploy/service: MAS_SSO_INSECURE ?= "false"
 deploy/service: MAS_SSO_BASE_URL ?= "https://identity.api.stage.openshift.com"
 deploy/service: MAS_SSO_REALM ?= "rhoas"
+deploy/service: SSO_SPECIAL_MANAGEMENT_ORG_ID ?= "13640203"
 deploy/service: USER_NAME_CLAIM ?= "clientId"
 deploy/service: FALL_BACK_USER_NAME_CLAIM ?= "preferred_username"
 deploy/service: MAX_ALLOWED_SERVICE_ACCOUNTS ?= "2"
@@ -768,6 +769,7 @@ deploy/service: deploy/envoy deploy/route
 		-p MAS_SSO_ENABLE_AUTH="${MAS_SSO_ENABLE_AUTH}" \
 		-p MAS_SSO_BASE_URL="$(MAS_SSO_BASE_URL)" \
 		-p MAS_SSO_REALM="$(MAS_SSO_REALM)" \
+		-p SSO_SPECIAL_MANAGEMENT_ORG_ID="${SSO_SPECIAL_MANAGEMENT_ORG_ID}" \
 		-p USER_NAME_CLAIM="$(USER_NAME_CLAIM)" \
 		-p FALL_BACK_USER_NAME_CLAIM="$(FALL_BACK_USER_NAME_CLAIM)" \
 		-p MAX_ALLOWED_SERVICE_ACCOUNTS="${MAX_ALLOWED_SERVICE_ACCOUNTS}" \
--- a/docs/deploying-kas-fleet-manager-to-openshift.md
+++ b/docs/deploying-kas-fleet-manager-to-openshift.md
@@ -146,6 +146,7 @@ make deploy/service IMAGE_TAG=<your-image-tag-here> <OPTIONAL_PARAMETERS>
 - `MAS_SSO_ENABLE_AUTH`: Enables MAS SSO authentication for the Data Plane. Defaults to `true`.
 - `MAS_SSO_BASE_URL`: MAS SSO base url. Defaults to `https://identity.api.stage.openshift.com`.
 - `MAS_SSO_REALM`: MAS SSO realm url. Defaults to `rhoas`.
+- `SSO_SPECIAL_MANAGEMENT_ORG_ID`: Special Management Organization ID used for creating internal Service accounts. Defaults to `13640203` which is the special management organisation id  organisation id for Stage environment.
 - `MAX_ALLOWED_SERVICE_ACCOUNTS`: The default value of maximum number of service accounts that can be created by users. Defaults to `2`.
 - `MAX_LIMIT_FOR_SSO_GET_CLIENTS`: The default value of maximum number of clients fetch from mas-sso. Defaults to `100`.
 - `OSD_IDP_MAS_SSO_REALM`: MAS SSO realm for configuring OpenShift Cluster Identity Provider Clients. Defaults to `rhoas-kafka-sre`.
--- a/internal/kafka/internal/services/util_test.go
+++ b/internal/kafka/internal/services/util_test.go
@@ -6,8 +6,10 @@ import (
 	"testing"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/keycloak"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
+	"github.com/google/uuid"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	serviceError "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
@@ -22,6 +24,46 @@ const (
 	mockKafkaRequestID     = "9bsv0s3fd06g02i2be9g" // sample kafka request ID
 	mockIDWithInvalidChars = "vp&xG^nl9MStC@SI*#c$6V^TKq0"
 )
+
+func buildKafkaDBApiRequest() dbapi.KafkaRequest {
+	route := []byte("unitesthost.com/v1/test-route")
+	return dbapi.KafkaRequest{
+		Region:                           "us-east-1",
+		ClusterID:                        uuid.NewString(),
+		CloudProvider:                    "AWS",
+		MultiAZ:                          false,
+		Name:                             "test-cluster",
+		Status:                           "Creating",
+		SsoClientID:                      uuid.NewString(),
+		SsoClientSecret:                  uuid.NewString(),
+		CanaryServiceAccountClientID:     uuid.NewString(),
+		CanaryServiceAccountClientSecret: uuid.NewString(),
+		SubscriptionId:                   "test",
+		Owner:                            "unit=test-user",
+		OwnerAccountId:                   uuid.NewString(),
+		BootstrapServerHost:              "unittesthost.com",
+		OrganisationId:                   "unit-test-org",
+		FailedReason:                     "",
+		PlacementId:                      "unit-test",
+		DesiredKafkaVersion:              "3",
+		ActualKafkaVersion:               "3",
+		DesiredStrimziVersion:            "0.27",
+		ActualStrimziVersion:             "0.27",
+		DesiredKafkaIBPVersion:           "v1",
+		ActualKafkaIBPVersion:            "v1",
+		KafkaUpgrading:                   false,
+		StrimziUpgrading:                 false,
+		KafkaIBPUpgrading:                false,
+		KafkaStorageSize:                 "5gi",
+		InstanceType:                     "developer",
+		QuotaType:                        "ams",
+		Routes:                           route,
+		RoutesCreated:                    true,
+		Namespace:                        "unit-test",
+		ReauthenticationEnabled:          true,
+		RoutesCreationId:                 uuid.NewString(),
+	}
+}
 
 func Test_HandleGetError(t *testing.T) {
 	cause := pkgErr.WithStack(gorm.ErrInvalidData)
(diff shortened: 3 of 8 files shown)


Rajagopalan-Ranganathan:
@akoserwal @ziccardi - take a look when you find time

pkg/client/keycloak/config.go
```diff
@@ -110,6 +113,7 @@ func (kc *KeycloakConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&kc.RedhatSSORealm.ClientIDFile, "redhat-sso-client-id-file", kc.RedhatSSORealm.ClientIDFile, "File containing Keycloak privileged account client-id that has access to the OSD Cluster IDP realm")
 	fs.StringVar(&kc.RedhatSSORealm.ClientSecretFile, "redhat-sso-client-secret-file", kc.RedhatSSORealm.ClientSecretFile, "File containing Keycloak privileged account client-secret that has access to the OSD Cluster IDP realm")
 	fs.StringVar(&kc.SsoBaseUrl, "redhat-sso-base-url", kc.SsoBaseUrl, "The base URL of the mas-sso, integration by default")
+	fs.StringVar(&kc.SpecialManagementOrgID, "special-management-org-id", SPEICAL_MGMT_ORG_ID_STAGE, "The Special Management Organization ID used for creating internal Service accounts")
```
machi1990, line 116:
```suggestion
	fs.StringVar(&kc.SpecialManagementOrgID, "sso-special-management-org-id", SPEICAL_MGMT_ORG_ID_STAGE, "The Special Management Organization ID used for creating internal Service accounts")
```
the flag `special-management-org-id` might be too generic, I propose prefixing it with some thing that says it is a sso related management id. 

Questions:
1. This is meant to be used for the redhatsso_service.go right?
2. What do you think about start using it right away i.e all the internal service accounts are created with this information?

internal/kafka/internal/services/util_test.go
```diff
@@ -353,3 +395,41 @@ func Test_contains(t *testing.T) {
...
+			},
+			expectedCustomClaim: fmt.Sprintf("@.rh-org-id == '%s'|| @.org_id == '%s' || @.clientId == '%s'", kafkaRequest.OrganisationId, kafkaRequest.OrganisationId, kafkaRequest.CanaryServiceAccountClientID),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			receivedBuiltCustomClaimCheck := BuildCustomClaimCheck(tt.args.kafkaRequest, tt.args.ssoconfigProvider)
+			if receivedBuiltCustomClaimCheck != tt.expectedCustomClaim {
+				t.Errorf("BuildCustomClaimCheck() = %v, want %v", receivedBuiltCustomClaimCheck, tt.expectedCustomClaim)
+			}
```
machi1990, line 432: Let's use gomega to perform the assertions.

internal/kafka/internal/services/util_test.go
```diff
@@ -23,6 +25,46 @@ const (
...
+		DesiredStrimziVersion:            "0.27",
+		ActualStrimziVersion:             "0.27",
+		DesiredKafkaIBPVersion:           "v1",
+		ActualKafkaIBPVersion:            "v1",
+		KafkaUpgrading:                   false,
+		StrimziUpgrading:                 false,
+		KafkaIBPUpgrading:                false,
+		KafkaStorageSize:                 "5gi",
+		InstanceType:                     "developer",
+		QuotaType:                        "developer",
```
machi1990, line 59:
```suggestion
		QuotaType:                        "ams",
```
Quota type is either `ams` or `quota-management-list`

docs/deploying-kas-fleet-manager-to-openshift.md
```diff
@@ -146,6 +146,7 @@ make deploy/service IMAGE_TAG=<your-image-tag-here> <OPTIONAL_PARAMETERS>
 - `MAS_SSO_ENABLE_AUTH`: Enables MAS SSO authentication for the Data Plane. Defaults to `true`.
 - `MAS_SSO_BASE_URL`: MAS SSO base url. Defaults to `https://identity.api.stage.openshift.com`.
 - `MAS_SSO_REALM`: MAS SSO realm url. Defaults to `rhoas`.
+- `SPECIAL_MANAGEMENT_ORG_ID`: Special Management Organization ID used for creating internal Service accounts. Defaults to the Org ID of stage.
```
machi1990, line 149:
```suggestion
- `SPECIAL_MANAGEMENT_ORG_ID`: Special Management Organization ID used for creating internal Service accounts. Defaults to `13640203` which is the special management organisation id for Stage environment.
```

Makefile
```diff
@@ -768,6 +768,7 @@ deploy/service: deploy/envoy deploy/route
 		-p MAS_SSO_ENABLE_AUTH="${MAS_SSO_ENABLE_AUTH}" \
 		-p MAS_SSO_BASE_URL="$(MAS_SSO_BASE_URL)" \
 		-p MAS_SSO_REALM="$(MAS_SSO_REALM)" \
+		-p SPECIAL_MANAGEMENT_ORG_ID="${SPECIAL_MANAGEMENT_ORG_ID}" \
```
machi1990, line 771: Can we provide a default for just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/2c19e350d9b8e70ead439167ce69ad1b86f69886/Makefile#L707 or the rest of the variables below?

docs/deploying-kas-fleet-manager-to-openshift.md
Rajagopalan-Ranganathan: Cannot expose the Org ID in a public repo. this is the Org ID that will have all the internal service accounts, so better not to specify the same in ReadME. @akoserwal i forgot that this is a public repo, so should we think about defaulting to stage OrgID - const declared in the code? Probably we can get away with an empty string. Would prefer not to have this in the public domain

internal/kafka/internal/services/util_test.go
Rajagopalan-Ranganathan: Thanks! changed to ams

internal/kafka/internal/services/util_test.go
Rajagopalan-Ranganathan: Thanks! should i update for other tests  as well in the file? a Few are not usng the gomega format, I will update for the other tests as well

internal/kafka/internal/services/util_test.go
machi1990: Thanks

internal/kafka/internal/services/util_test.go
machi1990: Thanks. That will be better. But if it starts to become too complex, maybe just focus on this one and the rest will be updated as we go.

pkg/client/keycloak/config.go
Rajagopalan-Ranganathan: 1) Yes, post migration this information - is not needed to be sent in the request when creating service accounts, reason: we will be using a "ROOT" service account created in the special managment org (similar to the KFM client in MAS SSO). Since the root account is associated to the org, SSO takes care of associating the newly created internal service accounts to the appropriate org. This is needed as a future enhancement, for extra security for fleetshard communication back to KFM.
2) This can be done, but we deferred it since, that would cause an inconsistency for the already created service accounts. So migration tool has been updated to link to the correct orgID during the migration of internal service accounts from MAS SSO to sso.redhat.com

Rajagopalan-Ranganathan: For the Naming comment, Yes it makes sense, if we agree, the changes should be reflected in app-interface MR as well.

machi1990: > 1. Yes, post migration this information - is not needed to be sent in the request when creating service accounts, reason: we will be using a "ROOT" service account created in the special managment org (similar to the KFM client in MAS SSO). Since the root account is associated to the org, SSO takes care of associating the newly created internal service accounts to the appropriate org. This is needed as a future enhancement, for extra security for fleetshard communication back to KFM.

Why post migration? Won't we find ourselves with some internal accounts created without the special managed org id attribute until this work is completed? 

> 2. This can be done, but we deferred it since, that would cause an inconsistency for the already created service accounts. So migration tool has been updated to link to the correct orgID during the migration of internal service accounts from MAS SSO to sso.redhat.com

Thanks. This future work is tracked somewhere, correct?

internal/kafka/internal/services/util_test.go
Rajagopalan-Ranganathan: Done, updated the entire file to goemga for assertions

pkg/client/keycloak/config.go
Rajagopalan-Ranganathan: 1) Currently in MAS SSO that is how it is. The reason for this in sso.redhat.com is that every service account (user or internal) should have a orgID associated so that it can be "MANAGED" via the API. So this is only required in sso.redhat.com env. 
2) This is already in place in the migration tool and we have verified it. The associated issue: https://issues.redhat.com/browse/MGDSTRM-7672

machi1990: I was asking for the special management organisation id to be used in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/14ae2b4dc33e4c3c7127dec35e703456b7e17241/pkg/services/sso/redhatsso_service.go or a issue/jira to track it and we are sure that it will be done before we start using that service file. 

But this is deferred for now until post migration, correct? So my question was: if we move to sso.redhat without making sure that the newly created internal service accounts are created under the special management id (using the redhatsso_service.go), won't we find ourselves with a gap there where some of the internal service accounts created post migration will be missing the special management org id value?

Makefile
Rajagopalan-Ranganathan: I am not sure we can specify defaults for this. as this is a part of the oc process command. Also, There are no defaults for other parameters , they take env value or are hard coded. WDYT?

docs/deploying-kas-fleet-manager-to-openshift.md
Rajagopalan-Ranganathan: agreed. code changed

pkg/client/keycloak/config.go
Rajagopalan-Ranganathan: No, that wont happen: When doing the migration: we will start using a new service account credential in vault (ssoredhat.com speical mgmt org root account). So there is no need to do this mapping. to compare: this root account is similar to "privileged KFM service account in MAS SSO" which can create/manage other SAs in this org. So when we switch to sso.redhat.com, we will switch to use a new SA (root SA for special mgmt org) to create service accounts. Service accounts created using this are mapped to the special mgmt org automatically/internally at sso.redhat.com and there is no need for adding that in the client side/SA representation as it is done on the server side (sso.redhat.com side).

Rajagopalan-Ranganathan: with the variable name, i have updated the code and every where to prefix with sso and updated the MR in app-interface as well

Makefile
machi1990: Hey, so basically if you do something like what's done for these variables: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/2c19e350d9b8e70ead439167ce69ad1b86f69886/Makefile#L710-L748, it should default to the value you'll give it. If you do not supply the default value, it'll be an empty string and it is a bad dev experience IMO.

> Also, There are no defaults for other parameters , they take env value or are hard coded

Yes, the hardcode values are the default values. See the linked make target above. So, let's hardcode the default `SPECIAL_MANAGEMENT_ORG_ID `value in the make file like:
```
deploy/service: SPECIAL_MANAGEMENT_ORG_ID ?= "13640203"
```

pkg/client/keycloak/config.go
machi1990: Thanks for updating the flag name. 

I'll follow up on the other discussion privately. I feel like there is a piece of information that I am lacking and I want to be sure that we are not leaving off gapes.

machi1990:
> > 
> 
> This needs to be updated after the migration. Since the custom claim checks will differ when provider is SSO. Would it better to do this one off post migration? WDYT?

+1, let's just keep it in mind.

Rajagopalan-Ranganathan:
> 

Yes, Let me create a JIRA to track this and update here in the comment.

Makefile
Rajagopalan-Ranganathan: ah sorry, understood you comment now, i got confused since it was pointing to line inside oc process -f , will update the same now

Rajagopalan-Ranganathan: fixed.

pkg/client/keycloak/config.go
Rajagopalan-Ranganathan: Sure Manyanda, lets discuss tomorrow

machi1990: Thanks @Rajagopalan-Ranganathan for the explanation.

> I feel like there is a piece of information that I am lacking and I want to be sure that we are not leaving off gapes.

We are all good here. What is going to happen is that the with sso.redhat.com, user facing service account creation will be using the user access token and for internal ones we'll be using the credentials. With the info on token, sso.redhat will be able to create the users service accounts into the user org and the internal ones in the internal org. This is the info I was missing.

machi1990 (approved):
lgtm

leaving it to @akoserwal and @ziccardi for the full functional review

akoserwal (approved):
LGTM

PR #935, opened 2022-04-19
feat: dont add label to managed kafka CR if instance type is eval

pawelpaszki:
## Description
feat: dont add label to managed kafka CR if instance type is eval

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/kafkas/types/types.go
+++ b/internal/kafka/internal/kafkas/types/types.go
@@ -7,6 +7,7 @@ type KafkaInstanceType string
 const (
 	DEVELOPER KafkaInstanceType = "developer"
 	STANDARD  KafkaInstanceType = "standard"
+	EVAL      KafkaInstanceType = "eval"
 )
 
 var ValidKafkaInstanceTypes = []string{
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -1015,6 +1015,12 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 	if err != nil {
 		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "unable to list kafka request")
 	}
+	labels := map[string]string{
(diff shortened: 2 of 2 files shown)


internal/kafka/internal/services/kafka.go
```diff
@@ -1015,6 +1015,12 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 	if err != nil {
 		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "unable to list kafka request")
 	}
+	labels := map[string]string{
+		"bf2.org/kafkaInstanceProfileQuotaConsumed": strconv.Itoa(k.QuotaConsumed),
+	}
+	if strings.Contains(api.AllInstanceTypeSupport.String(), kafkaRequest.InstanceType) {
```
miguelsorianod, line 1021: This is not totally accurate because `AllInstanceTypeSupport` is only used if the data plane config file does not explicitly set the values for supported_instance_types.
If we assume we will never run the new code with "eval" in any manually configured data plane cluster then I guess this might be enough, as long as we delete this code later.
An alternative to just compare with "eval" (or a constant that contains it).

machi1990: > An alternative to just compare with "eval" (or a constant that contains it).

+1, I think it is better if the comparison is done against a know instance type i.e `eval`.

pb82: +1 to compare against `eval`, however you will have to add the constant back (types.EVAL)

internal/kafka/internal/kafkas/types/types.go
```diff
@@ -7,6 +7,7 @@ type KafkaInstanceType string
 const (
 	DEVELOPER KafkaInstanceType = "developer"
 	STANDARD  KafkaInstanceType = "standard"
+	EVAL      KafkaInstanceType = "EVAL"
```
miguelsorianod, line 10: I think the value for this should be lowercase isn't it?

pawelpaszki: yes. fixed. thanks

pb82 (approved):

PR #942, opened 2022-04-20
feat: delete expired kafka eval instances

miguelsorianod:
## Description
This covers issues https://issues.redhat.com/browse/MGDSTRM-8193 and https://issues.redhat.com/browse/MGDSTRM-8188

With the changes in here the behavior we (wan to) achieve is:
* Automatic deletion after the expiration time of existing `eval` instances is supported
* Creation of kafka intsances of type `eval` is prevented. An error is returned if 
* Reconciliation of already existing `eval` instances is supported
* Creation and reconciliation of `standard` and `developer` instances is supported
* The cloud_providers endpoints do not provide `eval` as a supported type anymore

This behavior will allow us to ensure that we end up without any kafka `eval` instance after they expire, users can't create new ones so we can finish the migration to `developer` instances. Once that is done then all code and workarounds that deal with `eval` instances should be removed in a future PR.

The following tests have been done, using quota-management-list quota type:

* Created a kafka eval instance in main branch, then re-run kas-fleet-manager in this PR's branch and it is still there as ready and there haven't been any error logs identified in kas fleet manager logs
* The eval kafka instance is shown when listing kafka instances and no error logs are shown in kas fleet manager logs
* The eval kafka instance is shown when getting that specific kafka instance and no error logs are shown in kas fleet manager logs
* Calling the cloud_providers endpoint works and no error logs are generated in kas fleet manager
* Calling the cloud_providers/aws/regions endpoint returns all the regions correctly and the us-east-1 region contains standard and developer but not eval, as expected
* When creating a new kafka instance without explicitly specifying a plan it is created as developer plan and no error logs are shown in kas fleet manager
* When creating a new kafka explicitly defining "eval.x1" plan it an error is returned with the following message: KAFKAS-MGMT-21","reason":"Unable to detect instance type in plan provided: 'eval.x1'
* Creating a kafka standard instance is supported and can coexist with the other types
* Updating the creation time of an 'eval' kafka instance to older than 48 hours triggers the deletion process, no kas fleet manager error logs are shown and the deleted_at time is updated
* Updating the creation time of a 'developer' kafka instance to older than 48 hours triggers the deletion process, no kas fleet manager error logs are shown and the deleted_at time is updated
* Manually deleting a 'developer' instances works as expected
* Manually deleting a 'standard' instance works as expected

## Verification Steps

To test this the following can be done:

1. Create kafka eval instance by running kas fleet manager from the 'main' branch and wait until it is in ready state
1. Move to the PR branch and recompile kas-fleet-manager binary. Then run make db/migrate. Then modify the manual clusters config to set supported instance types to "standard,developer" (removing eval and adding developer)
1. Rerun kas-fleet-manager again
1. Verify that the cluster info in the db has been updated to contain supported types "standard,developer" (eval has been removed and developer has been added)
1. Verify that the existing kafka instance still is in ready state and there are no errors in the kas fleet manager logs. Also verify it by calling the kafka get and list endpoints
1. Modify the creation date of the kafka instance to 3 days before to simulate an expired eval instance
1. Verify that kas fleet manager deletes the eval instance and there are no errors on the kas fleet manager logs
1. Verify that the cloud provider endpoint does not include "eval" as the result
1. Verify that newly created kafka instances are developer type
1. Verify that you cannot create a kafka instance of eval type (explicitely specifying the plan or implicitely)

- [ ] Code Review completed


pawelpaszki:
looks good to me. I'll let others have a look too

config/kafka-instance-types-configuration.yaml
```diff
@@ -48,3 +48,17 @@ supported_instance_types:
...
+    sizes:
+    - id: x1
+      ingressThroughputPerSec: "30Mi"
+      egressThroughputPerSec: "30Mi"
+      totalMaxConnections: 3000
+      maxConnectionAttemptsPerSec: 100
+      maxDataRetentionSize: "1000Gi"
+      maxDataRetentionPeriod: "P14D"
+      maxPartitions: 1000
+      quotaConsumed: 1
```
ziccardi, line 62: `quotaConsumed` should be 0

config/kafka-instance-types-configuration.yaml
```diff
@@ -48,3 +48,17 @@ supported_instance_types:
...
+      quotaType: "rhosak"
```
ziccardi, line 63: `quotaType` should be "RHOSAKTrial"

config/kafka-instance-types-configuration.yaml
miguelsorianod: I think it does not really matter at this point as no kafka eval instances can be created right? or would existing instances be updated?

config/kafka-instance-types-configuration.yaml
pb82: we're also assigning `rhosak` to developer instances. It should be the same here.

miguelsorianod: The contents of the "eval" instance type are a copy of the contents I found in the "developer" type with the exception of the displayName and the id.
The reason for that is that in the larger-kafka-instances branch there was "eval" and it was renamed to "developer". If that's not accurate let me know as I don't have the context about this

config/kafka-instance-types-configuration.yaml
pb82: this is used in `ReserveQuota` which is only called when creating an instance. @ziccardi can you confirm?

config/kafka-instance-types-configuration.yaml
ziccardi: @pb82 @miguelsorianod : it should be:
```yaml
quotaConsumed: 0
quotaType: "RHOSAKTrial"
```

for both `eval` and `developer`

pb82: Ok, got you. `quotaType` should refer to the Quoty (RHOSAK and RHOSAKTrial) and not to the resource (rhosak).

config/kafka-instance-types-configuration.yaml
pb82: @ziccardi quotyConsumed is used: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/larger-kafka-instances/internal/kafka/internal/services/quota/ams_quota_service.go#L136

pb82: so we need to use the correct values.

pb82 (approved):

miguelsorianod:
I also updated the service-template file to update the default SUPPORTED_INSTANCE_TYPES value

config/kafka-instance-types-configuration.yaml
miguelsorianod: Values updated. For details see the new comments in the issue

config/kafka-instance-types-configuration.yaml
miguelsorianod: Updated. For details see the new comments in the issue

ziccardi (approved):

PR #951, opened 2022-04-21
Remove unused config files

pb82:
## Description

Remove the old `kafka-capacity-config` template. It is replaced by the supported instance types config. Some of the dynamic scaling tests were relying on this config. They are skipped for the moment as the dynamic scaling logic itself is going to change.

The service template is also updated and the values should be removed from app-interface too.

## Verification Steps

Passing tests should be sufficient.

### DIFF
--- a/internal/kafka/internal/services/data_plane_cluster.go
+++ b/internal/kafka/internal/services/data_plane_cluster.go
@@ -432,7 +432,9 @@ func (d *dataPlaneClusterService) scaleUpThresholds(status *dbapi.DataPlaneClust
 // to consider that a kafka cluster has capacity available
 func (d *dataPlaneClusterService) minimumKafkaCapacity() *dataPlaneComputeNodesKafkaCapacityAttributes {
 	return &dataPlaneComputeNodesKafkaCapacityAttributes{
-		Connections: d.KafkaConfig.KafkaCapacity.TotalMaxConnections,
-		Partitions:  d.KafkaConfig.KafkaCapacity.MaxPartitions,
+		// using hardcoded values because the dynamic scaling implementation will be revised in the future and
+		// does not support streaming units in it's current state
+		Connections: 100,
+		Partitions:  100,
 	}
 }
--- a/internal/kafka/internal/services/data_plane_cluster_test.go
+++ b/internal/kafka/internal/services/data_plane_cluster_test.go
@@ -104,6 +104,7 @@ func Test_DataPlaneCluster_updateDataPlaneClusterNodes(t *testing.T) {
 		inputFactory   func() *input
 		expectedResult int
 		wantErr        bool
+		skip           bool
 	}{
 		{
 			name: "when scale-up thresholds are crossed number of compute nodes is increased",
@@ -197,9 +198,9 @@ func Test_DataPlaneCluster_updateDataPlaneClusterNodes(t *testing.T) {
 			wantErr:        false,
 		},
 		{
+			skip: true,
 			name: "when all scale-down threshold is crossed number of compute nodes is decreased",
 			inputFactory: func() *input {
-				kafkaConfig := sampleValidApplicationConfigForDataPlaneClusterTest(nil).KafkaConfig
 				testStatus := sampleValidBaseDataPlaneClusterStatusRequest()
 				testStatus.NodeInfo.Current = 6
 				testStatus.NodeInfo.Ceiling = 10000
@@ -207,10 +208,6 @@ func Test_DataPlaneCluster_updateDataPlaneClusterNodes(t *testing.T) {
 				// We set remaining to a value much higher than resizeInfo.value which to
 				// simulate a scale-down is needed, as scale-down thresholds are
 				// calculated from resizeInfo.Delta value
-				testStatus.ResizeInfo.Delta.Connections = kafkaConfig.KafkaCapacity.TotalMaxConnections * 10
-				testStatus.ResizeInfo.Delta.Partitions = kafkaConfig.KafkaCapacity.MaxPartitions * 10
-				testStatus.Remaining.Connections = kafkaConfig.KafkaCapacity.TotalMaxConnections * 1000
-				testStatus.Remaining.Partitions = kafkaConfig.KafkaCapacity.MaxPartitions * 1000
 				apiCluster := &api.Cluster{
 					ClusterID: testClusterID,
 					MultiAZ:   true,
@@ -236,19 +233,16 @@ func Test_DataPlaneCluster_updateDataPlaneClusterNodes(t *testing.T) {
 			wantErr:        false,
 		},
 		{
+			skip: true,
 			name: "when not all scale-down threshold are crossed number of compute nodes is not decreased",
 			inputFactory: func() *input {
-				kafkaConfig := sampleValidApplicationConfigForDataPlaneClusterTest(nil).KafkaConfig
 				testStatus := sampleValidBaseDataPlaneClusterStatusRequest()
 				testStatus.NodeInfo.Current = 6
 				testStatus.NodeInfo.Ceiling = 10000
 				testStatus.NodeInfo.CurrentWorkLoadMinimum = 3
-				testStatus.ResizeInfo.Delta.Connections = kafkaConfig.KafkaCapacity.TotalMaxConnections * 10
-				testStatus.ResizeInfo.Delta.Partitions = kafkaConfig.KafkaCapacity.MaxPartitions * 10
 				// We simulate connections scale-down threshold not being crossed
 				// and partitions scale-down threshold being crossed
 				testStatus.Remaining.Connections = testStatus.ResizeInfo.Delta.Connections - 1
-				testStatus.Remaining.Partitions = kafkaConfig.KafkaCapacity.MaxPartitions * 1000
 				apiCluster := &api.Cluster{
 					ClusterID: testClusterID,
 					MultiAZ:   true,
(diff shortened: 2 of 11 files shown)


internal/kafka/internal/services/kafka.go
```diff
@@ -332,8 +332,19 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
...
+	if instanceTypeErr != nil {
+		return errors.InstanceTypeNotSupported(instanceTypeErr.Error())
+	}
+
+	size, sizeErr := instanceType.GetKafkaInstanceSizeByID(kafkaRequest.SizeId)
+	if sizeErr != nil {
+		return errors.InstancePlanNotSupported(sizeErr.Error())
+	}
+
+	kafkaRequest.KafkaStorageSize = size.MaxDataRetentionSize.String()
```
machi1990, line 347: good catch!

templates/service-template.yml
```diff
@@ -590,20 +590,6 @@ objects:
...
-      annotations:
-        qontract.recycle: "true"
-    data:
-      kafka-capacity-config.yaml: |-
-        ingressEgressThroughputPerSec: ${KAFKA_CAPACITY_INGRESS_THROUGHPUT}
-        totalMaxConnections: ${KAFKA_CAPACITY_TOTAL_MAX_CONNECTIONS}
-        maxDataRetentionSize: ${KAFKA_CAPACITY_MAX_DATA_RETENTION_SIZE}
-        maxPartitions: ${KAFKA_CAPACITY_MAX_PARTITIONS}
-        maxDataRetentionPeriod: ${KAFKA_CAPACITY_MAX_DATA_RETENTION_PERIOD}
-        maxConnectionAttemptsPerSec: ${KAFKA_CAPACITY_MAX_CONNECTION_ATTEMPTS_PER_SEC}
```
machi1990, line 606: Do we want to remove the `KAFKA_CAPACITY_XX` https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/2bbd6264ea829806205681c7359a5f00a24934a3/templates/service-template.yml#L421-L449 parameters as well?

internal/kafka/internal/services/data_plane_cluster.go
```diff
@@ -432,7 +432,9 @@ func (d *dataPlaneClusterService) scaleUpThresholds(status *dbapi.DataPlaneClust
 // to consider that a kafka cluster has capacity available
 func (d *dataPlaneClusterService) minimumKafkaCapacity() *dataPlaneComputeNodesKafkaCapacityAttributes {
 	return &dataPlaneComputeNodesKafkaCapacityAttributes{
-		Connections: d.KafkaConfig.KafkaCapacity.TotalMaxConnections,
-		Partitions:  d.KafkaConfig.KafkaCapacity.MaxPartitions,
+		// using hardcoded values because the dynamic scaling implementation will be revised in the future and
```
ziccardi, line 435: I'd suggest to prefix these kind of comments with `FIXME` or `TODO` so that it will be easy to find them
```suggestion
		// FIXME: using hardcoded values because the dynamic scaling implementation will be revised in the future and
```

machi1990 (approved):
lgtm

internal/kafka/internal/services/data_plane_cluster_test.go
```diff
@@ -388,6 +370,10 @@ func Test_DataPlaneCluster_updateDataPlaneClusterNodes(t *testing.T) {
 
 	for _, tt := range cases {
 		t.Run(tt.name, func(t *testing.T) {
+			if tt.skip {
+				t.Skip("skipping dynamic scaling related tests")
+			}
+
 			input := tt.inputFactory()
 			if input == nil {
 				t.Fatalf("invalid input")
```
ziccardi, line 379: I know it is not part of this PR, but this should be changed to
```go
Expect(input).ToNot(BeNil())
```

internal/kafka/internal/services/data_plane_cluster_test.go
```diff
@@ -960,6 +945,10 @@ func Test_DataPlaneCluster_setClusterStatus(t *testing.T) {
 
 	for _, tt := range cases {
 		t.Run(tt.name, func(t *testing.T) {
+			if tt.skip {
+				t.Skip("skipping dynamic scaling related tests")
+			}
+
 			f, spyReceivedStatus := tt.inputFactory()
 			if f == nil {
 				t.Fatalf("dataPlaneClusterService is nil")
```
ziccardi, line 954: Same as previous comment. This should be:
```suggestion
			Expect(f).ToNot(BeNil(), "dataPlaneClusterService is nil")
```

internal/kafka/test/integration/kafkas_test.go
```diff
@@ -143,7 +143,13 @@ func TestKafkaCreate_Success(t *testing.T) {
 	// this is set by the mockKasfFleetshardSync
 	Expect(kafkaRequest.DesiredStrimziVersion).To(Equal("strimzi-cluster-operator.v0.23.0-0"))
 	// default kafka_storage_size should be set on creation
-	Expect(kafkaRequest.KafkaStorageSize).To(Equal(test.TestServices.KafkaConfig.KafkaCapacity.MaxDataRetentionSize))
+	instanceType, err := test.TestServices.KafkaConfig.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(kafka.InstanceType)
+	Expect(err).To(BeNil())
```
ziccardi, line 147:
```suggestion
	Expect(err).ToNot(HaveOccurred())
```

internal/kafka/test/integration/kafkas_test.go
```diff
@@ -143,7 +143,13 @@ func TestKafkaCreate_Success(t *testing.T) {
...
+
+	instanceSize, err := instanceType.GetKafkaInstanceSizeByID(kafka.SizeId)
+	Expect(err).To(BeNil())
```
ziccardi, line 150:
```suggestion
	Expect(err).ToNot(HaveOccurred())
```

internal/kafka/test/integration/kafkas_test.go
```diff
@@ -1424,7 +1430,15 @@ func TestKafkaGet(t *testing.T) {
 	Expect(kafka.Name).To(Equal(mockKafkaName))
 	Expect(kafka.Status).To(Equal(constants2.KafkaRequestStatusAccepted.String()))
 	Expect(kafka.ReauthenticationEnabled).To(BeFalse())
-	Expect(kafka.KafkaStorageSize).To(Equal(test.TestServices.KafkaConfig.KafkaCapacity.MaxDataRetentionSize))
+
+	instanceType, err := test.TestServices.KafkaConfig.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(kafka.InstanceType)
+	Expect(err).To(BeNil())
```
ziccardi, line 1435:
```suggestion
	Expect(err).ToNot(HaveOccurred())
```

internal/kafka/test/integration/kafkas_test.go
```diff
@@ -1424,7 +1430,15 @@ func TestKafkaGet(t *testing.T) {
...
+
+	instanceSize, err := instanceType.GetKafkaInstanceSizeByID(kafka.SizeId)
+	Expect(err).To(BeNil())
```
ziccardi, line 1438:
```suggestion
	Expect(err).ToNot(HaveOccurred())
```

internal/kafka/internal/services/data_plane_cluster.go
Rajagopalan-Ranganathan: I personally think a TODO or FIXME is always good, anyone who stumbles across such prefixes can kick start a conversation if we had missed anything.

machi1990: The thing here is that, dynamic scaling implementation as it is implemented now will be gone. It'll be a complete rewrite . 
This part of the code is not used by anyone at the moment, other than the integration tests. Someone could just remove the code and it'll be the same. 

> .. anyone who stumbles across such prefixes can kick start a conversation if we had missed anything.

If we have JIRAs/issues for the situation at handy for it, then nothing is missed.

To be clear, I agree on having TODOs comment with associated JIRAs for something that needs improving. But adding them to something that we know for sure is going to be re-written (whole of it) in a couple of sprint time (issue to track this are there) isn't really worth it.

Rajagopalan-Ranganathan: If we create JIRAs then and there no issues, often we miss that (or like we discussed, when the fix is simple compared to time to create JIRA, we dont create them) , thats is where TODOs come in handy :) .

ziccardi: It is a new comment added in this PR. Since there is no harm in having the TODO/FIXME prefix, I'd prefer to have it anyway.

pb82: I agree with Manyanda, there not really anything to do or to fix about this. In the future it will be removed and replaced by something different. Adding  `TODO` or `FIXME` to those lines would suggest that they have to be updated which is not the case.

ziccardi: That is exactly what the comment is about: what do we have TO DO? Remove it in the future. So it is a TODO. The rest of the comment explains when it has to be done and how. 

However, I will agree with the majority here, so I'm fine with not adding it.

ziccardi (approved):

PR #953, opened 2022-04-21
feature: set kafka multiaz attribute depending on instance type

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-8264.

When registering a Kafka instance, set the MultiAZ attribute for that instance depending on the instance type.
The explained reason found on the related issue on why we need to do this is:
```
It has been found that incoming requests to control plane are set to multi-AZ by default regardless of instance types. UI plan to not send AZ value. Control plane needs to assign the correct AZ value based on instance type.
```

There were two possible places on where to implement this logic:
1. In the RegisterKafkaJob method, which is where this PR has implemented it
1. In the Kafka create handler

The first place has been chosen due to in the case of the `kafka create` CLI subcommand the RegisterKafkaJob method is called without going through the Kafka create handler.

The reason on why to do this

## Verification Steps

* Create a kafka standard instance and verify that multiaz is set to true
* Create a kafka eval instance and verify that multiaz is set to false
* Try to create a kafka standard instance with multiaz set to false and verify that an error is returned
* Try to create a kafka eval instance with multiaz set to true and verify that an error is returned

1. Go to `XX >> YY >> SS`
3. Create a new item `N` with the info `X`
4. Try to edit this item 
5. Check if in the left menu the feature X is not so long present.

If manual verifications required, please provide an environment where the reviewers can easily validate the changes if possible to speed up the review process. 
-->

- [ ] Code Review completed


miguelsorianod:
This PR has been implemented based on the original request and the original justification on why this needs to be done.

However I think we should discuss how do we want to go forward with this, as implementing this leads to some implications that create some inconsistencies:

With this changes we now set the MultiAZ value based on the instance type. However, multi_az attribute is a field that we have at OpenAPI level for the kafka_request. Why having that field at all if it is totally ignored now?

Why can't the UI set the MultiAZ field from their side?

Do we really want to tie the MultiAZ value to the instance type? This means that an instance type will only support either MultiAZ or non MultiAZ but not both at the same time.

machi1990:
> With this changes we now set the MultiAZ value based on the instance type. However, multi_az attribute is a field that we have at OpenAPI level for the kafka_request. Why having that field at all if it is totally ignored now?

+1. Furthermore I think with what we have here, the check https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e0da31b2ffe016dd6d677af0d05fdaa43afe1f77/pkg/handlers/validation.go#L78 will also go away.

> Do we really want to tie the MultiAZ value to the instance type? This means that an instance type will only support either MultiAZ or non MultiAZ but not both at the same time.

+1. Better to get clarification on this. Maybe @pb82 might have some additional context.

pb82:
@machi1990 @miguelsorianod at present it's Ok and confirmed by BU to tie the value of multi-az to the instance type. There are currently no plans to allow multi-az dev instances (or other combinations). This might get re-evaluated in the future, but likely not soon. And then we have a few options, like adding multi-az to the plan definition (or re-introducing the field).

As for deprecating instead of removing: I think that's just following the deprecation policy. After some time (3 months?) we can also remove it.

machi1990:
> @machi1990 @miguelsorianod at present it's Ok and confirmed by BU to tie the value of multi-az to the instance type. There are currently no plans to allow multi-az dev instances (or other combinations). This might get re-evaluated in the future, but likely not soon. And then we have a few options, like adding multi-az to the plan definition (or re-introducing the field).
> 
> As for deprecating instead of removing: I think that's just following the deprecation policy. After some time (3 months?) we can also remove it.

Thanks @pb82 

Then +1 on deprecating the multi_az from the Kafka Request payload during creation. 
This field should still be returned on response body and shouldn't be deprecated there. We also do not need to validate for it in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e0da31b2ffe016dd6d677af0d05fdaa43afe1f77/pkg/handlers/validation.go#L78

pb82:
yes agree @machi1990

miguelsorianod:
> We also do not need to validate for it in

Are you sure we don't need to? even if it is deprecated users might still be setting it so I think it is a good safeguard

machi1990:
> > We also do not need to validate for it in
> 
> Are you sure we don't need to? 

Yes

> even if it is deprecated users might still be setting it 

Setting shouldn't have any effect as the value passed in the payload shouldn't be used at all

> ...so I think it is a good safeguard

I do not think it is a safeguard worth keeeping if the value is not used anymore going forward. 

Also imagine a user setting it for a developer plan:
- they first try to set multi_az to false, the check fails saying that multi_az is not supported
- the change their request this time setting mutli_az to true, we validate against it (at this point the user expects that the multi_az to be `true`) but when their kafka request has been accepted, the `multi_az` is returned as false due to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/953/files#diff-f876d2c0a69be46c37a7a608a4d1847036dfe692b7648b4cfef180936dcdbc18R356. At this point as a user I'll be lost. 

Thus why I think it is best, to deprecate the field in the Kafka Request creation payload. And add a description saying that `"...field deprecated and if passed it won't be considered. The API automatically determines this value from the `plan` attribute"`

miguelsorianod:
> Also imagine a user setting it for a developer plan:
> 
>     * they first try to set multi_az to false, the check fails saying that multi_az is not supported
> 
>     * the change their request this time setting mutli_az to true, we validate against it (at this point the user expects that the multi_az to be `true`) but when their kafka request has been accepted, the `multi_az` is returned as false due to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/953/files#diff-f876d2c0a69be46c37a7a608a4d1847036dfe692b7648b4cfef180936dcdbc18R356. At this point as a user I'll be lost.

The first point is not correct though right? If a user tries to create a kafka with developer plan and multi_az set to false that will not fail, as developer instances are non multi-az.
And in the second case the request would fail, as we would want

miguelsorianod:
@pb82 @machi1990 ready for review

internal/kafka/internal/presenters/kafka.go
```diff
@@ -23,7 +23,7 @@ func ConvertKafkaRequest(kafkaRequestPayload public.KafkaRequestPayload, dbKafka
 	kafka.Region = kafkaRequestPayload.Region
 	kafka.Name = kafkaRequestPayload.Name
 	kafka.CloudProvider = kafkaRequestPayload.CloudProvider
-	kafka.MultiAZ = kafkaRequestPayload.MultiAz
+	kafka.MultiAZ = kafkaRequestPayload.DeprecatedMultiAz
```
machi1990, line 26: This can be removed now but let's keep it and it will be removed when the multi_az field is removed from the request payload.

machi1990 (approved):
lgtm, thanks @miguelsorianod 

leaving it to @pb82 to get an extra eye on this.

pb82 (approved):

PR #959, opened 2022-04-24
feat: allow developer instances to be scheduled on multi az clusters

pb82:
Allow single AZ developer instances to be scheduled on multi AZ clusters, but not multi AZ instances on single AZ clusters.

### DIFF
--- a/internal/kafka/internal/services/cluster_placement_strategy.go
+++ b/internal/kafka/internal/services/cluster_placement_strategy.go
@@ -3,6 +3,7 @@ package services
 import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/kafkas/types"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -58,11 +59,15 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
...
 		Region:                kafka.Region,
-		MultiAZ:               kafka.MultiAZ,
 		Status:                api.ClusterReady,
 		SupportedInstanceType: kafka.InstanceType,
 	}
 
+	// check the multi-az value of a cluster only for standard instances, so that developer instances can be scheduled to multi-az clusters
+	if kafka.InstanceType == string(types.STANDARD) {
+		criteria.MultiAZ = kafka.MultiAZ
+	}
```
machi1990, line 69: I think this works out of the box already and the change is not needed. 

I say it works out of the box because `gorm` the orm we are using, ignores the `multi_az` parameter when it is `false` during the construction of the SQL query. So automatically, when a developer instance is being assigned, it'll land to either a single az or a multi_az cluster.

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -58,11 +59,15 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
 	criteria := FindClusterCriteria{
 		Provider:              kafka.CloudProvider,
 		Region:                kafka.Region,
-		MultiAZ:               kafka.MultiAZ,
```
machi1990, line 61: This might not have the effect we want as well, since the default value when a boolean isn't set is `false`.

internal/kafka/internal/services/cluster_placement_strategy.go
machi1990: This test: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/2e4f0cd716186573d4a94f447496c968f8347a66/internal/kafka/test/integration/kafkas_test.go#L205-L219 kinds of validate that developers instance are correctly assigned.

pb82: Good point. So before it was more of a bug but now it's behaviour that we actually want.

machi1990: > So before it was more of a bug but now it's behaviour that we actually want.

Yes :-)

PR #962, opened 2022-04-26
fix: update supported instance type limits

JameelB:
## Description
Update the supported instance type configuration to have the correct limits. 

## Verification Steps
* Code Review

- [ ] Code Review completed

### DIFF
--- a/config/kafka-instance-types-configuration.yaml
+++ b/config/kafka-instance-types-configuration.yaml
@@ -34,17 +34,28 @@ supported_instance_types:
       quotaConsumed: 1
       quotaType: "RHOSAK"
       capacityConsumed: 1
+    - id: x2
+      ingressThroughputPerSec: "100Mi"
+      egressThroughputPerSec: "200Mi"
+      totalMaxConnections: 6000
+      maxDataRetentionSize: "2000Gi"
+      maxPartitions: 3000
+      maxDataRetentionPeriod: "P14D"
+      maxConnectionAttemptsPerSec: 200
+      quotaConsumed: 2
+      quotaType: "RHOSAK"
+      capacityConsumed: 2
   - id: developer
     display_name: Trial
     sizes:
     - id: x1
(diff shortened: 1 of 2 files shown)


JameelB:
@miguelsorianod @pb82

pb82 (approved):
LGTM, checked against spreadsheet values

config/kafka-instance-types-configuration.yaml
```diff
@@ -34,17 +34,28 @@ supported_instance_types:
...
-      maxConnectionAttemptsPerSec: 100
-      maxDataRetentionSize: "1000Gi"
+      ingressThroughputPerSec: "1Mi"
+      egressThroughputPerSec: "1Mi"
+      totalMaxConnections: 100
+      maxConnectionAttemptsPerSec: 50
+      maxDataRetentionSize: "10Gi"
       maxDataRetentionPeriod: "P14D"
-      maxPartitions: 1000
+      maxPartitions: 100
```
ziccardi, line 58: Shouldn't we make the same updates for `eval` as well?

JameelB: don't think so. `eval` and `developer` have different limits. `eval` will have the same limits as it does right now, which is closer to what a `standard.x1` size is.  Once we no longer have to support it, we can remove it from here.

JameelB: I believe we can no longer request for an `eval`, an error is returned to the user if it was specified explicitly in the request payload. 

These limits will only be returned for the existing eval instances on migration since that is their actual size. We don't want to set the limits down to what `developer` has since developer is a single broker and eval will remain with 3 brokers. If we decrease the limits, we'll end up with some sort of hybrid instance. 

If the user doesn't specify the plan, it'll be based on their quota. So for users with no quota, the instance created will be `developer` with actual developer limits. There's no longer any reference to `eval` in our API.

miguelsorianod (approved):

PR #965, opened 2022-04-26
Larger kafka instances and single AZ kafka instances

miguelsorianod:
## Description
Related to https://issues.redhat.com/browse/MGDSTRM-7810, https://issues.redhat.com/browse/MGDSTRM-7494

* Introduces `developer` instance types. `developer` instance types are single-az and consume less resources. Users that do not have RHOSAK quota are allowed to create one `developer` instance.
* Removes the ability to create `eval` instance types. Already existing `eval` instance types are still supported. In the future the support for already existing `eval` instances will be removed
* The concept of kafka `plan`s is introduced: A plan is a combination of a kafka instance type and a kafka size. A given instance type can have multiple sizes.
* Now AZ'ness of a kafka instance is determined by the kafka instance type. Thus, the multi_az attribute in kafka creation request has been deprecated and it has no effect anymore
* Service limits for `standard` instances have been redefined according to new values requested by BU
* The configuration of the service limits for the different kafka plans is defined in the `config/kafka-instance-types-configuration.yaml`. This also replaces the original `config/capacity-configuration.yaml` file which had a different format and dealt with different concepts that don't apply anymore.
* The communication of the ingress/egress network throughput desired for the managedkafka instances, originally specified in the `IngressEgressThroughputPerSec` has now been split into the `IngressPerSec` and `EgressPerSec` attributes to allow differentiation between inbound and outbound network throughput
* K8s labels for `developer` and `standard` instances are now set into the ManagedKafka instances API resource containing the kafka instance type and the quota consumed
* New API endpoints have been added that report information about the instance types available in a given cloud provider region. The cloud provider endpoints have also been updated to provide information about the instance types and its capacity information.

- [ ] Code Review completed


machi1990:
Great to see this work coming up. No reviews from my side as I'll leave that out to @JameelB @pb82 only here to send some love. Let's :ship: this :-)

valdar (approved):

internal/kafka/internal/cmd/kafka/list.go
```diff
@@ -79,7 +79,10 @@ func runList(env *environments.Env, cmd *cobra.Command, _ []string) {
 	}
 
 	for _, kafkaRequest := range kafkaList {
-		converted := presenters.PresentKafkaRequest(kafkaRequest, kafkaConfig.BrowserUrl)
+		converted, err := presenters.PresentKafkaRequest(kafkaRequest, kafkaConfig)
+		if err != nil {
+			glog.Fatalf("Failed to format kafka request: %s", err.Error())
```
Rajagopalan-Ranganathan, line 84: Should this be a fatal error when iterating through the list? would it be better to log all the presenter issues for kakfa request items and then exit based on error or a bool value? Reason: we might exit after first error and would not unmask further errors down

internal/kafka/internal/cmd/supportedkafkainstancetypes/flags.go
```diff
@@ -0,0 +1,8 @@
+package supportedkafkainstancetypes
+
+const (
+	// FlagCloudProvider is a flag representing cloud provider Id
+	FlagCloudProvider = "cloud-provider"
```
Rajagopalan-Ranganathan, line 5: should this be cloud-provder-id  to add more clarity?

internal/kafka/internal/config/kafka.go
```diff
@@ -65,13 +68,33 @@ func (c *KafkaConfig) ReadFiles() error {
 	if err != nil {
 		return err
 	}
-	content, err := shared.ReadFile(c.KafkaCapacityConfigFile)
+
+	SupportedInstanceTypesFileContent, err := shared.ReadFile(c.SupportedInstanceTypes.ConfigurationFile)
```
Rajagopalan-Ranganathan, line 72: there is a new method added to shared lib that can read yaml files directly, https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/75ee96dc51150d2ae51a4a8eecc8ae036b1e2135/pkg/shared/config.go#L84 better to use that

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return fmt.Errorf("ingressThroughputPerSec for Kafka instance type '%s', size '%s' is invalid: %s", k.Id, instanceTypeId, err.Error())
+	}
+
+	maxDataRetentionSize, err := k.MaxDataRetentionSize.ToK8Quantity()
+	if err != nil {
+		return fmt.Errorf("maxDataRetentionSize for Kafka instance type '%s', size '%s' is invalid: %s", k.Id, instanceTypeId, err.Error())
+	}
+
+	maxDataRetentionPeriod, err := duration.ParseISO8601(k.MaxDataRetentionPeriod)
```
Rajagopalan-Ranganathan, line 89: Assuming this parse function only validates the format, should there be a check to determine if the provided time is less than current time? meaning a past time than current time cannot be valid.

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,202 @@
...
+		ingressThroughputQuantity.CmpInt64(1) < 0 || maxDataRetentionSize.CmpInt64(1) < 0 ||
+		k.TotalMaxConnections <= 0 || k.MaxPartitions <= 0 || k.MaxConnectionAttemptsPerSec <= 0 ||
+		k.QuotaConsumed < 1 || k.CapacityConsumed < 1 {
+		return fmt.Errorf("Kafka instance size '%s' for instance type '%s' specifies a property value less than or equals to Zero.", k.Id, instanceTypeId)
+	}
+
+	return nil
+}
+
+type SupportedKafkaInstanceTypesConfig struct {
```
Rajagopalan-Ranganathan, line 104: struct definitions are better to be placed at the start of the file along with other definitions

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,202 @@
...
+
+		if err := KafkaInstanceType.validate(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+type KafkaSupportedInstanceTypesConfig struct {
```
Rajagopalan-Ranganathan, line 125: same as above, can struct definitions be placed at the top

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,202 @@
...
+func NewKafkaSupportedInstanceTypesConfig() *KafkaSupportedInstanceTypesConfig {
+	return &KafkaSupportedInstanceTypesConfig{
+		ConfigurationFile: "config/kafka-instance-types-configuration.yaml",
+	}
+}
+
+func (s *SupportedKafkaInstanceTypesConfig) GetKafkaInstanceTypeByID(instanceType string) (*KafkaInstanceType, error) {
+	for _, t := range s.SupportedKafkaInstanceTypes {
+		if t.Id == instanceType {
+			ret := t
```
Rajagopalan-Ranganathan, line 139: trying to understand why we need to copy/assign this to another variable ? can we use the index and return appropriately rather than making a copy?

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,202 @@
...
+			return &ret, nil
+		}
+	}
+	return nil, fmt.Errorf("Unable to find kafka instance type for '%s'", instanceType)
+}
+
+func (kp *KafkaInstanceType) GetKafkaInstanceSizeByID(sizeId string) (*KafkaInstanceSize, error) {
+	for _, size := range kp.Sizes {
+		if size.Id == sizeId {
+			ret := size
```
Rajagopalan-Ranganathan, line 149: same as above, using the index would solve the need for a local copy

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,202 @@
...
+
+type Plan string
+
+func (p Plan) String() string {
+	return string(p)
+}
+
+func (p Plan) GetInstanceType() (string, error) {
+	t := strings.Split(strings.TrimSpace(p.String()), ".")
+	if len(t) != 2 {
```
Rajagopalan-Ranganathan, line 164: if this size of 2 is number of type of plans, better to have it as a constant.

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,202 @@
...
+		return "", errors.New(errors.ErrorGeneral, fmt.Sprintf("Unsupported plan provided: '%s'", p))
+	}
+	return t[0], nil
+}
+
+func (p Plan) GetSizeID() (string, error) {
+	t := strings.Split(strings.TrimSpace(p.String()), ".")
+	if len(t) != 2 {
```
Rajagopalan-Ranganathan, line 172: have number of plans as a constant, to avoid future maintenance work when plans increase/decrease

internal/kafka/internal/handlers/cloud_providers.go
```diff
@@ -75,9 +77,15 @@ func (h cloudProvidersHandler) ListCloudProviderRegions(w http.ResponseWriter, r
 					cloudRegion.Enabled = true
 					cloudRegion.SupportedInstanceTypes = region.SupportedInstanceTypes.AsSlice()
 					for _, instType := range cloudRegion.SupportedInstanceTypes {
+						// --- to be removed once MaxCapacityReached has been removed. ---
 						maxCapacityReached := true
 						kafka.InstanceType = instType
 						kafka.Region = cloudRegion.Id
+						size, e := h.kafkaConfig.GetFirstAvailableSize(instType)
+						if e != nil {
+							return nil, errors.NewWithCause(errors.ErrorGeneral, err, "Unable to list cloud provider regions")
```
Rajagopalan-Ranganathan, line 86: the error message doesnt seem correct when , err is used instead of e

internal/kafka/internal/services/kafka.go
```diff
@@ -147,20 +151,95 @@ func (k *kafkaService) HasAvailableCapacityInRegion(kafkaRequest *dbapi.KafkaReq
...
+		}
+
+		// Check against region limits
+		hasCapacity, err := k.HasAvailableCapacityInRegion(kafka)
+		if err != nil {
+			logger.Logger.Error(err)
+			return nil, err
+		}
+
+		if hasCapacity {
```
Rajagopalan-Ranganathan, line 222: should there be a error returned if there is no capacity in the region

internal/kafka/internal/services/kafka.go
```diff
@@ -174,30 +253,36 @@ func (k *kafkaService) DetectInstanceType(kafkaRequest *dbapi.KafkaRequest) (typ
...
 		var count int64
 		if err := dbConn.Model(&dbapi.KafkaRequest{}).
-			Where("instance_type = ?", types.EVAL).
+			Where("instance_type = ?", types.DEVELOPER).
 			Where("owner = ?", kafkaRequest.Owner).
 			Where("organisation_id = ?", kafkaRequest.OrganisationId).
 			Count(&count).
 			Error; err != nil {
-			return "", errors.NewWithCause(errors.ErrorGeneral, err, "failed to count kafka eval instances")
+			return "", errors.NewWithCause(errors.ErrorGeneral, err, "failed to count kafka %s instances", instType.DisplayName)
```
Rajagopalan-Ranganathan, line 281: would returning an DB error expose anything that shouldnt be exposed/returned. not sure where this error is consumed and what is done, just curious

internal/kafka/internal/config/kafka_supported_instance_types.go
pb82: @Rajagopalan-Ranganathan this is a duration (e.g. `P14D` for 14 days), not a timestamp. So this can't lay in the past. Or did you mean something different?

internal/kafka/internal/config/kafka_supported_instance_types.go
pb82: I believe we return a copy to avoid exposing the original pointer (which would allow modification).

internal/kafka/internal/config/kafka_supported_instance_types.go
pb82: this refers to the format of the plan, it has to be `<plan-name>.<size-id>`, so two components. Not sure if this really needs to be a constant.

internal/kafka/internal/cmd/supportedkafkainstancetypes/flags.go
miguelsorianod: I agree. I will be renaming this

internal/kafka/internal/services/kafka.go
miguelsorianod: Looking at the code I don't think that's the case. If there is no capacity then the result (the available sizes in the region) is a nil slice/without elements.

cc @JameelB might help clarify this but I think that was the original intention looking at the code.

internal/kafka/internal/services/kafka.go
miguelsorianod: That totally depends on the error returned by the DB. I don't think in this case there should be anything sensible. It is performing a count on the database with information that is known when the kafka instance is created.

internal/kafka/internal/handlers/cloud_providers.go
miguelsorianod: :+1: that's correct. I added a new commit with the fix

internal/kafka/internal/config/kafka.go
miguelsorianod: :+1:  I added a new commit with the change

internal/kafka/internal/cmd/supportedkafkainstancetypes/flags.go
miguelsorianod: I added a new commit with the change

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: I think this is a personal preference matter.

For example I tend to find useful to have the struct type definition and then all its methods following it instead of placing the struct definitions at the start of the file.

In the Go source code itself you can observe examples on how they follow the "struct type definition and then all its methods" approach instead of placing all struct definitions at the top. See for example https://go.dev/src/image/geom.go

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/965#discussion_r859546733

internal/kafka/internal/config/kafka_supported_instance_types.go
pb82: see above, this refers to the format of the plan.

internal/kafka/internal/config/kafka_supported_instance_types.go
Rajagopalan-Ranganathan: Then its perfect, i assumed it was a timestamp - perfect!

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/965#discussion_r859506885 comment

internal/kafka/internal/config/kafka_supported_instance_types.go
Rajagopalan-Ranganathan: Agreed its a personal preference, but i saw in KFM in other places that was followed, thought that was a practice :)

internal/kafka/internal/config/kafka_supported_instance_types.go
Rajagopalan-Ranganathan: Ah , good catch i missed that, fine then

internal/kafka/internal/services/kafka.go
pb82: no error should be raised here as `no size can be placed` is a valid return.

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: It is not the number of plans but the number of components that there are in the format of a plan.

I don't see it strictly necessary but it does not harm anyway so I added a new commit with a constant for it.

internal/kafka/internal/cmd/kafka/list.go
miguelsorianod: This is used in a CLI command so returning an error at the first occurrence seems reasonable to me

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -0,0 +1,205 @@
...
+	if p, err := resource.ParseQuantity(string(*q)); err != nil {
+		return nil, err
+	} else {
+		return &p, nil
+	}
+}
+
+func (q *Quantity) IsEmpty() bool {
+	return q == nil
+}
```
ziccardi, line 205: This file is defining the following objects:

- KafkaInstanceType
- SupportedKafkaInstanceTypesConfig
- KafkaSupportedInstanceTypesConfig
- Plan
- Quantity

Together with their methods. IMO, this makes the code hard do read/maintains.
Wouldn't be better to place each object in its own file?

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: Not that I am aware. Marking this thread as resolved as it is a design preference

internal/kafka/internal/services/kafka.go
JameelB: Yes that is the intention, and as what Peter mentioned a region not having capacity is a valid return. 

This is used for the cloud_provider/{id}/regions endpoint. If an instance type for a particular region doesn't have any capacity, the `available_sizes` should just show up as an empty list. i.e.

```
{
      "kind": "CloudRegion",
      "id": "us-east-1",
      "enabled": true,
      "supported_instance_types": [
        "standard"
      ],
      "capacity": [
        {
          "instance_type": "standard",
          "max_capacity_reached": true,
          "available_sizes": []
        }
      ]
    },
```

internal/kafka/internal/config/kafka_supported_instance_types.go
pb82: It's a good suggestion @ziccardi , but can be done in a separate PR.

internal/kafka/internal/config/kafka_supported_instance_types.go
Rajagopalan-Ranganathan: thanks Miguel!

internal/kafka/internal/cmd/supportedkafkainstancetypes/flags.go
Rajagopalan-Ranganathan: thanks Migurel

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: :+1

Makefile
```diff
@@ -732,7 +732,7 @@ deploy/service: TOKEN_ISSUER_URL ?= "https://sso.redhat.com/auth/realms/redhat-e
 deploy/service: SERVICE_PUBLIC_HOST_URL ?= "https://api.openshift.com"
 deploy/service: ENABLE_TERMS_ACCEPTANCE ?= "false"
 deploy/service: ENABLE_DENY_LIST ?= "false"
-deploy/service: ALLOW_EVALUATOR_INSTANCE ?= "true"
+deploy/service: ALLOW_DEVELOPER_INSTANCE ?= "true"
 deploy/service: QUOTA_TYPE ?= "quota-management-list"
 deploy/service: STRIMZI_OLM_INDEX_IMAGE ?= "quay.io/osd-addons/managed-kafka:production-82b42db"
 deploy/service: KAS_FLEETSHARD_OLM_INDEX_IMAGE ?= "quay.io/osd-addons/kas-fleetshard-operator:production-82b42db"
```
JameelB, line 738: We forgot to change `eval` -> `developer` on [L#753](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/965/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R753)

docs/deploying-kas-fleet-manager-to-openshift.md
```diff
@@ -163,7 +163,7 @@ make deploy/service IMAGE_TAG=<your-image-tag-here> <OPTIONAL_PARAMETERS>
 - `OBSERVATORIUM_RHSSO_REALM`: Realm of RHSSO used for Observatorium authentication. Defaults to `''`.
 - `OBSERVABILITY_CONFIG_REPO`: URL of the configuration repository used by the Observability stack. Defaults to `https://api.github.com/repos/bf2fc6cc711aee1a0c2a/observability-resources-mk/contents`.
 - `ENABLE_TERMS_ACCEPTANCE`: Enables terms acceptance through AMS. Defaults to `false`.
-- `ALLOW_EVALUATOR_INSTANCE`: Enables creation of evaluator Kafka instances. Defaults to `true`.
+- `ALLOW_DEVELOPER_INSTANCE`: Enables creation of developer Kafka instances. Defaults to `true`.
 - `QUOTA_TYPE`: Quota management service to be used. Options: `quota-management-list` and `ams`, Default: `quota-management-list`.
 - `KAS_FLEETSHARD_OLM_INDEX_IMAGE`: KAS Fleetshard operator OLM index image. Defaults to `quay.io/osd-addons/kas-fleetshard-operator:production-82b42db`.
 - `STRIMZI_OLM_INDEX_IMAGE`: Strimzi operator OLM index image. Defaults to `quay.io/osd-addons/managed-kafka:production-82b42db`.
```
JameelB, line 169: Same as above, `eval` needs to be changed to `developer` in [L#175](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/965/files#diff-7fed3a215608ed06db2187e2c796e30c3332b2c7bb4223214647d6309adf0498R175)

Makefile
pb82: good catch! Is that only used when running locally?

docs/deploying-kas-fleet-manager-to-openshift.md
miguelsorianod: fixed

Makefile
miguelsorianod: fixed

docs/deploying-kas-fleet-manager-to-openshift.md
JameelB: thanks!

JameelB (approved):

PR #910, opened 2022-04-08
docs: added best practices

ziccardi:
## Description
Added some best practices and utilities documentation

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has created for changes required on the client side

PR #967, opened 2022-04-26
fix: added no-op flow for PR that doesn't change code

ziccardi:

- [ ] Code Review completed

### DIFF
--- a/.github/workflows/ci-noop.yaml
+++ b/.github/workflows/ci-noop.yaml
@@ -1,0 +1,42 @@
+# This is a noop flow that runs only when only non-code files are changed.
+name: Lint & Test
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '*.md'
+      - '*.sh'
+      - '.github/*.md'
+      - '.github/workflows/openapi_update.yaml'
+      - '.github/CODEOWNERS'
+      - 'templates/**'
+      - '.openapi-generator-ignore'
+      - 'openapi/**'
+      - 'docs/**'
+      - 'pkg/api/openapi/docs/**'
+      - 'pkg/api/openapi/.openapi-generator-ignore'
+
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - '*.md'
+      - '*.sh'
+      - '.github/*.md'
+      - '.github/workflows/openapi_update.yaml'
+      - '.github/CODEOWNERS'
+      - 'templates/**'
+      - '.openapi-generator-ignore'
+      - 'openapi/**'
+      - 'docs/**'
+      - 'pkg/api/openapi/docs/**'
+      - 'pkg/api/openapi/.openapi-generator-ignore'
+
+jobs:
+  lint-test:
+    name: "Lint & Test"
+    runs-on: ubuntu-latest
+
+    steps:
+      - run: 'echo "No build required"'


machi1990:
What's the motivation for this? As these files are already ignored in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c35cfd7f3bb2d76c31fc7b4ad05605702e7267a2/.github/workflows/ci.yaml#L8 and https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c35cfd7f3bb2d76c31fc7b4ad05605702e7267a2/.github/workflows/ci.yaml#L23

machi1990 (changes requested):
See comment above

ziccardi:
> What's the motivation for this? As these files are already ignored in
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c35cfd7f3bb2d76c31fc7b4ad05605702e7267a2/.github/workflows/ci.yaml#L8
> 
> and
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c35cfd7f3bb2d76c31fc7b4ad05605702e7267a2/.github/workflows/ci.yaml#L23

This PR does exactly the opposite and runs when the other one does not run.
It is to solve the issue we are seeing [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/910) where the branch can't be merged because the required `Lint & Test` is not running.

The solution is the one documented here: https://github.com/github/docs/pull/9126/files#diff-3052509b0e9a9ca90d0b911cc12c333506e652dbac7c7d5bd41d462be5c07dfdR87

machi1990:
> > What's the motivation for this? As these files are already ignored in
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c35cfd7f3bb2d76c31fc7b4ad05605702e7267a2/.github/workflows/ci.yaml#L8
> > 
> > and
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c35cfd7f3bb2d76c31fc7b4ad05605702e7267a2/.github/workflows/ci.yaml#L23
> 
> This PR does exactly the opposite and runs when the other one does not run. It is to solve the issue we are seeing [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/910) where the branch can't be merged because the required `Lint & Test` is not running.
> 
> The solution is the one documented here: https://github.com/github/docs/pull/9126/files#diff-3052509b0e9a9ca90d0b911cc12c333506e652dbac7c7d5bd41d462be5c07dfdR87

Ah perfect, thanks for the link. This seems like a workaround for a Github bug IMO. Github should take into account ignored files. The reason why I do not see this as a potential solution is that we are duplicating the "ignored" files, and now someone has to update the files in ci.yaml file in two places - one being the opposite of others. I think we are better off deactivating required checks on continue to rely on the PR template and code reviewers making sure that the CI is passing.

ziccardi:
> Ah perfect, thanks for the link. This seems like a workaround for a Github bug IMO. Github should take into account ignored files. The reason why I do not see this as a potential solution is that we are duplicating the "ignored" files, and now someone has to update the files in ci.yaml file in two places - one being the opposite of others. I think we are better off deactivating required checks on continue to rely on the PR template and code reviewers making sure that the CI is passing.


While I see your point, I don't agree with your conclusion for 2 main reasons:

1. I don't think the list of files to be ignored will change often
2. Even if someone should ever forget to update the `noop` workflow, the worst that will happen will be that it will 'run' too. But since it is a noop, it won't cause any harm.

On the other hand, disabling the `required check` could result in non working PR being merged.

While I agree it is a workaround, it is the currently official solution: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#example

machi1990:
> On the other hand, disabling the required check could result in non working PR being merged.

I do not think we've ever merged a non working PR. Moreover, the PR template checklist, I copied it below requires that CI and tests are passing. It is part of the review process.
 
> ## Checklist (Definition of Done)
> * [ ]  All acceptance criteria specified in JIRA have been completed
> * [ ]  Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
> * [ ]  Documentation added for the feature
> * [ ]  CI and all relevant tests are passing
> * [ ]  Code Review completed
> * [ ]  Verified independently by reviewer
> * [ ]  Required metrics/dashboards/alerts have been added (or PR created).
> * [ ]  Required Standard Operating Procedure (SOP) is added.
> * [ ]  JIRA has created for changes required on the client side

> While I see your point, I don't agree with your conclusion for 2 main reasons:
> I don't think the list of files to be ignored will change often

True, which makes it even easier to cause inconsistency because of having two modify multiple files that people are not so familiar in changing (this can be documented). 

> Even if someone should ever forget to update the noop workflow, the worst that will happen will be that it will 'run' too. But since it is a noop, it won't cause any harm.

Yes. But, someone can as well forget to update the other file which can cause undesired triggers.

> While I agree it is a workaround, it is the currently official solution: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#example

Having to maintain two CI files (one just workaround a bug in an external system) is not ideal. 

This is my opinion, I'll let others give their thoughts/review on this as well.

Rajagopalan-Ranganathan:
I would suggest instead of noop/dummy job, for Docs, we can add a markdown linter job. This would also help to validate/lint the markdown files: examples: https://christinavhastenrath.medium.com/github-actions-markdown-lint-setup-86d7f6ec934d , https://github.com/actionshub/markdownlint, https://github.com/ruzickap/action-my-markdown-linter . WDYT @ziccardi @machi1990 ?

ziccardi:
> I do not think we've ever merged a non working PR. Moreover, the PR template checklist, I copied it below requires that CI and tests are passing. It is part of the review process. And also, why merge a non working PR to begin with?

Of course that wouldn't be intentional. Sometimes happens that the tests are working locally and failing in CI. It could be overlooked.

> True, which makes it even easier to cause inconsistency because of having to modify multiple files that people are not so familiar in changing (this can be documented).

Yep. Could be added to the PR template and be part of the review as well.

> 
> > Even if someone should ever forget to update the noop workflow, the worst that will happen will be that it will 'run' too. But since it is a noop, it won't cause any harm.
> 
> Yes. But, someone can as well forget to update the other file which can cause undesired triggers.

Again, that would be no harm: it would just happen that when merging a `non-code` PR (very rare), a compilation would trigger.

> Having to maintain two CI files (one just workaround a bug in an external system) is not ideal.

True. But how often do you think we will need to update the list of ignored files? I don't think we will need to do anytime soon.

ziccardi:
> I would suggest instead of noop/dummy job, for Docs, we can add a markdown linter job. This would also help to validate/lint the markdown files: examples: https://christinavhastenrath.medium.com/github-actions-markdown-lint-setup-86d7f6ec934d , https://github.com/actionshub/markdownlint, https://github.com/ruzickap/action-my-markdown-linter . WDYT @ziccardi @machi1990 ?

That wouldn't solve the issue, because the required `Lint & Test` wouldn't run anyway unless we give it the same name `Lint & Test`, but that would mean that when a PR contains both code and doc changes we would see 2 `Lint & Test` flows running.

PR #973, opened 2022-04-28
fix(connectors): fixes MGDCTRS-881, allow admin force delete to delete dangling connector deployments

dhirajsb:
## Description
Allow admin force delete to delete dangling connector deployments

## Verification Steps
Integration tests to verify fix included. 

- [ ] Code Review completed

### DIFF
--- a/internal/connector/internal/services/connectors.go
+++ b/internal/connector/internal/services/connectors.go
@@ -381,12 +381,14 @@ func (k *connectorsService) ForceDelete(ctx context.Context, id string) *errors.
 				return err
 			}
 		}
-		if err := k.Delete(ctx, id); err != nil {
-			return err
-		}
 		return nil
 	}); err != nil {
 		return services.HandleDeleteError("Connector", "id", id, err)
+	}
+
+	// delete connector in a separate transaction to allow deleting dangling deployments
+	if err := k.Delete(ctx, id); err != nil {
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


internal/connector/internal/services/connectors.go
```diff
@@ -381,12 +381,14 @@ func (k *connectorsService) ForceDelete(ctx context.Context, id string) *errors.
...
-		}
 		return nil
 	}); err != nil {
 		return services.HandleDeleteError("Connector", "id", id, err)
 	}
+
+	// delete connector in a separate transaction to allow deleting dangling deployments
+	if err := k.Delete(ctx, id); err != nil {
+		return err
+	}
```
lburgazzoli, line 392: Isn't there a risk here that then the deployment gets recreated in case the connector delete fails ? 
I recall there were a reconcile loop that recreates missing deployment: @valdar am I correct ?

dhirajsb: Well, the assumption is that the admin API soft delete should be used first. The last resort force delete should only be used for connectors already in `deleting` state, or the dangling deployments scenario. We can add a check to make sure that either connector is already in this state or we have the dangling deployment situation.

lburgazzoli (approved):

internal/connector/internal/services/connectors.go
valdar: If the assumption is that connectors are in `deleting` state once you use the force delete then what @lburgazzoli refers to should not happen.

PR #974, opened 2022-04-28
fix: remove unused availability zone errors

pb82:
the single/multi AZ errors are no longer used since the request param has been deprecated.

### DIFF
--- a/internal/connector/test/integration/features/connector-api.feature
+++ b/internal/connector/test/integration/features/connector-api.feature
@@ -1635,13 +1635,6 @@ Feature: create a connector
             "reason": "Maximum field length has been depassed"
           },
           {
-            "code": "CONNECTOR-MGMT-35",
-            "href": "/api/connector_mgmt/v1/errors/35",
-            "id": "35",
-            "kind": "Error",
-            "reason": "Only multiAZ Kafkas are supported, use multi_az=true"
-          },
-          {
             "code": "CONNECTOR-MGMT-36",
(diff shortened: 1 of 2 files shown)


machi1990 (changes requested):
Good catch!

We can also safely removed:
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6171483febc0016cec5a824d7fde9b9267cc9c91/pkg/errors/errors.go#L196-L197
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6171483febc0016cec5a824d7fde9b9267cc9c91/pkg/errors/errors.go#L156-L157

pb82:
@machi1990 the other references are also removed now

machi1990 (approved):
lgtm

machi1990:
I think one of the connectors test might fail as they reference the removed errors. The test will need to  be updated.

machi1990:
> I think one of the connectors test might fail as they reference the removed errors. The test will need to be updated.

This is the test in question: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/47360744395aff45f165f79261135895659d445d/internal/connector/test/integration/features/connector-api.feature#L1637 

and https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/47360744395aff45f165f79261135895659d445d/internal/connector/test/integration/features/connector-api.feature#L1693

JameelB (approved):
lgtm too! :)

PR #977, opened 2022-04-28
fix: skip dynamic scaling related tests

pb82:
skip dynamic scaling related tests, they are failing on the e2e pipeline because the old kafka-capacity-config is no longer available. Not worth porting them to the new supported instance types since dynamic scaling will change.


machi1990 (approved):
lgtm

Thanks.

There is a rebase issue and there is a lot more commits than what's desired for the change I think, mind having a look?

JameelB (approved):

PR #982, opened 2022-04-28
Update image pull secret type in dataplane cluster config

daveconde:
Updates the type of the image pull secret to SecretTypeDockerConfigJson from SecretTypeDockercfg.


MikeEdgar:
I think one more small change is needed for this. `DockerConfigKey` should be changed to `DockerConfigJsonKey` the same method, `buildImagePullSecret`.

machi1990:
Thanks @MikeEdgar fantastic catch.

PR #985, opened 2022-04-28
fix: remove supported instance types

wtrocki:
Two problems:

- Any required field cannot be deprecated. If you consider deprecating field it should be removed from required fields.
- API no longer returns that object field which is very confusing - deprecation doesn't give permission to remove stuff


machi1990:
Very good points. My 2 cents below.
> Any required field cannot be deprecated. If we consider deprecating field it should be removed from required fields.

If the field "requiredness" was removed, that would also mean a breaking change i.e from being required to not being required. What I mean here is that, if the array for example is empty, the field wouldn't be returned at all and that would break existing clients/contract. The require here guarantee that the field will always be present. I think that shouldn't be changed when the field is marked as deprecated, more generally I think "a deprecated field being required should be removed as part of removing the deprecated field in the next version of the API not when the first time when the field becomes deprecated".

This takes me to the second point..

> API no longer returns that object field which is very confusing - deprecation should not mean we can stop returning that data.

 +1.  When a field is deprecated, we should always continue to honor the behavior: to a maximum degree of the possibility of it being done. I did a quick check in stage and the data is set for the `supported_instance_types`.
```
ocm get /api/kafkas_mgmt/v1/cloud_providers/aws/regions/
{
  "kind": "CloudRegionList",
  "page": 1,
  "size": 22,
  "total": 22,
  "items": [
    {
      "kind": "CloudRegion",
      "id": "af-south-1",
      "display_name": "Africa (Cape Town)",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-east-1",
      "display_name": "Asia Pacific (Hong Kong)",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-northeast-1",
      "display_name": "Asia Pacific, Tokyo",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-northeast-2",
      "display_name": "Asia Pacific, Seoul",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-northeast-3",
      "display_name": "Asia Pacific (Osaka)",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-south-1",
      "display_name": "Asia Pacific, Mumbai",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-southeast-1",
      "display_name": "Asia Pacific, Singapore",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-southeast-2",
      "display_name": "Asia Pacific, Sydney",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ap-southeast-3",
      "display_name": "Asia Pacific, Jakarta",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "ca-central-1",
      "display_name": "Canada, Central",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "eu-central-1",
      "display_name": "EU, Frankfurt",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "eu-north-1",
      "display_name": "EU, Stockholm",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "eu-south-1",
      "display_name": "Europe (Milan)",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "eu-west-1",
      "display_name": "EU, Ireland",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "eu-west-2",
      "display_name": "EU, London",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "eu-west-3",
      "display_name": "EU, Paris",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "me-south-1",
      "display_name": "Middle East, Bahrain",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "sa-east-1",
      "display_name": "South America, São Paulo",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "us-east-1",
      "enabled": true,
      "supported_instance_types": [
         "standard",
         "developer"
      ],
      "capacity": [
        {
          "instance_type": "standard",
          "available_sizes": [
             "x1",
             "x2"
          ]
        },
        {
          "instance_type": "developer",
          "available_sizes": [
             "x1"
          ]
        }
      ]
    },
    {
      "kind": "CloudRegion",
      "id": "us-east-2",
      "display_name": "US East, Ohio",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "us-west-1",
      "display_name": "US West, N. California",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    },
    {
      "kind": "CloudRegion",
      "id": "us-west-2",
      "display_name": "US West, Oregon",
      "enabled": false,
      "supported_instance_types": [],
      "capacity": []
    }
  ]
}
``` 

I can see two things which have always been the case:
- when a cloud region is disabled, the supported_instance_types is an empty array. See the `"eu-west-2"` region for example.
- when the cloud region is enabled, at least one supported instance type is there (a non empty array is returned). See the `"us-east-1"` region for example.

Where are you seeing this? Or Am I looking at the wrong endpoint?

wtrocki:
UI and CLI do not get that field. I think it could be based on the fact that is deprecated. 
SDKs do not compile if field is deprecated and required at the same time. 

Since this is only return type would it be possible to remove required field? Would it be breaking. change if part of the input?

Screenshot 2022-04-28 at 17 04 08

wtrocki:
FYI. There is no concern of backwards compatibility for clients here. Those will be using old SDK etc. 
 What we want is to have new SDK compiling. This is little bit time sensitive change.
Screenshot 2022-04-29 at 09 29 26

wtrocki:
We will watch releases on https://github.com/patternfly/patternfly-quickstarts/releases

PR #988, opened 2022-04-29
fix: removed validators side effects

ziccardi:
## Description
Validators for Kafka create handler had multiple side effects:

- one was converting the KafkaPayload to a database KafkaRequest object
- one was retrieving user and owner from the claims and setting them into the KafkaRequest
- one was retrieving the instance type and size id and setting them into the KafkaRequest
- etc

Instead of fragmenting the Validators into a number of small, almost meaningless functions just to retrieve the same values from both the handler and the Validator, this PR adds the concept of `ValidationContext`.

This way the flow is changed like this:

1. The validator checks if some of the values it needs are in the Validation Context. If yes, it uses them, otherwise it will retrieve them by itself
2. The validator performs the validation and eventually saves produced data into the Validation Context
3. The handler retrieve the data produced by the validators from the validation context and uses that data to populate the KafkaRequest object

## Verification Steps

- Try to create a kafka and check that everything is correctly populated into the database.
- Try with invalid requests (invalid size, region, etc) and verify they are correctly detected

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -3,10 +3,10 @@ package handlers
 import (
 	"context"
 	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+	"github.com/golang-jwt/jwt/v4"
 	"regexp"
 	"strings"
-
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/presenters"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/admin/private"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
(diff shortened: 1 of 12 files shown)


miguelsorianod:
> The validator performs the validation and eventually saves produced data into the Validation Context

This is a side effect. Validators should only perform validation. Filling a context object increases the complexity considerably as with what was before because now you have to know where the context is filled and with what keys. Validators should only perform validation.

If there is a need to convert a KafkaPayload to a database KafkaRequest object (this is, initializing a kafkarequest object), that should be done separately.
We can either validate the kafkapayload object and then initialize the kafkarequest object with what we know are valid values, or initialize the kafkarequest object (with potentially
invalid values) and then perform validation on the kafkarequest object.

ziccardi:
> This is a side effect. Validators should only perform validation. 

I though about that too. I pretty much agree

> Filling a context object increases the complexity considerably as with what was before because now you have to know where the context is filled and with what keys. 

Not really. The validator use the values in the context only if they are there. If they are not, then the validator extracts them by itself. Each validator can run by itself and doesn't require any other validator to be run before.
I don't see any added complexity.

> Validators should only perform validation.

Yes, but they can produce some info as result of the validation itself. The real constraints is that they don't have to rely on data from other validators to be present.

machi1990:
> Yes, but they can produce some info as result of the validation itself. The real constraints is that they don't have to rely on data from other validators to be present.

I think the info the validators produce is: `This validation run successful` or `This validation failed because of this and that`.
And not `This is how you initialize the object, and that object and that other object`. That does not seem to be the responsibility of the validator.

ziccardi:
> I think the info the validators produce is: `This validation run successful` or `This validation failed because of this and that`. 

That's the output, but what is produced.

> And not `This is how you initialize the ozject, and that object and that other object`. That does not seem to be the responsibility of the validator.

I think there is some misunderstanding about what the current approach does. It doesn't show how to initialise anything at all.
The validator does some work to perform the validation, and that work produces something.
I think it make no sense to repeat again that work only because we don't want to save that into a context.

I'd agree with you if the validators would rely/require those data to be in the context, but they don't. If they are there they use them, if they are not, they produce them. Each validator can be used independently of the others and no particular order is required.

internal/kafka/internal/handlers/validation.go
```diff
@@ -51,45 +85,113 @@ func ValidateKafkaClusterNameIsUnique(name *string, kafkaService services.KafkaS
...
+		} else {
+			region, _ = provider.Regions.GetByName(kafkaRequest.Region)
+		}
+
 		if !region.IsInstanceTypeSupported(config.InstanceType(instanceType)) {
 			return errors.InstanceTypeNotSupported("instance type '%s' not supported for region '%s'", instanceType.String(), region.Name)
 		}
+
+		validatorContext[CLOUD_PROVIDER] = provider.Name
+		validatorContext[REGION] = region.Name
```
machi1990, line 144: This is side effect..

machi1990: There are a lot of these in the validators..

ziccardi: Sure it is. But, IMO, it makes the code simpler, while the proposed approach makes it unreadable.
Anyway, looks like I'm alone in this battle, so I will follow your suggestions :)

machi1990: > Sure it is

it is not different from what we've at the moment which is what the JIRA is proposing to get rid of.

> But, IMO, it makes the code simpler

It is hard to reason about the code and violates the principle of least surprises ;-) i.e I'd not expect the validator to have side effects and its the last place someone will look to see where the initialization of a `foo` or `bar` field is being made.

> so I will follow your suggestions

Thanks.

ziccardi: > it is not different from what we've at the moment which is what the JIRA is proposing to get rid of.
It is very different. Currently the validator expects some data to be ready and the order of validators is important.
With this change the validator does not expect anything and doesn't care about the order. You can even remove validators and it will work anyway.

> someone will look to see where the initialization of a foo or bar field is being made.

Nope . As I stated before, validator doesn't have to rely to the content of the context. They can check if a value is there, but if it is not there, they have to produce that themselves.

ziccardi:
Superseded by #998

PR #989, opened 2022-04-29
fix: IsJWTTokenExpired returning true for every input

VanillaSpoon:
## Description
`IsJWTTokenExpired()` moved from pkg/shared/is_jwt_token_expired.go --> pkg/client/keycloak/client.go 

IsJWTTokenExpired() has been altered as this function was returning true for every input due to jwt.Parse() returning `no Keyfunc was provided` error. This has been resolved.

Unit tests have been added

[MGDSTRM-8381](https://issues.redhat.com/browse/MGDSTRM-8381)

- [ ] Code Review completed


pkg/auth/is_jwt_token_expired.go
```diff
@@ -0,0 +1,17 @@
+package auth
+
+import (
+	"time"
+
+	"github.com/golang-jwt/jwt/v4"
+)
+
+// This method returns true if JWT token is expired, otherwise returns false
```
machi1990, line 9:
```suggestion
// IsJWTTokenExpired returns true if JWT token is expired, otherwise returns false
```

pkg/auth/is_jwt_token_expired.go
```diff
@@ -0,0 +1,17 @@
+package auth
```
Rajagopalan-Ranganathan, line 1: Cannot find a better way to put this comment: the file name doesnt follow Go conventions, "_" must not be present and the filename is specific to a method for now, would it be better to rename it as jwtUtils or something? anticpating future JWT related common utils will go in here?

pkg/auth/is_jwt_token_expired_test.go
```diff
@@ -0,0 +1,71 @@
...
+		},
+		{
+			name: "return false for unExpired accessToken",
+			args: args{
+				accessToken: unExpiredToken,
+			},
+			want: false,
+		},
+		{
+			name: "return false for invalid accessToken",
```
Rajagopalan-Ranganathan, line 58: we are expecting a a return value of "true" the text there needs to be updated to reflect the same.

pkg/auth/is_jwt_token_expired.go
VanillaSpoon: Thanks for the input @Rajagopalan-Ranganathan 
I agree, jwtUtils would be a better alternative. Also, I wonder if moving this to auth/helper.go would be out of the question?

pkg/auth/is_jwt_token_expired.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/golang-jwt/jwt/v4"
+)
+
+// This method returns true if JWT token is expired, otherwise returns false
+func IsJWTTokenExpired(accessToken string) bool {
+	if token, _ := jwt.Parse(accessToken, nil); token != nil {
+		tokenClaims := token.Claims.(jwt.MapClaims)
+		exp := tokenClaims["exp"].(float64)
+		return exp-float64(time.Now().Unix()) <= 0
```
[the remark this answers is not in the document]
VanillaSpoon: Thanks @ziccardi I'll check this out

pkg/auth/is_jwt_token_expired.go
VanillaSpoon: Thanks @machi1990

pkg/auth/is_jwt_token_expired.go
Rajagopalan-Ranganathan: The debate is :) is this a helper or util method, i think this falls under util category, but please let me know yours (and others thoughts) as well

pkg/auth/is_jwt_token_expired.go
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: https://pkg.go.dev/time#example-Time.Before - Arithmetic operations on "time" should always be avoided if there are inbuilt methods.

pkg/auth/is_jwt_token_expired.go
miguelsorianod: >  Cannot find a better way to put this comment: the file name doesnt follow Go conventions, "_" must not be present and the filename is specific to a method for now, would it be better to rename it as jwtUtils or something? anticpating future JWT related common utils will go in here?

Where is the ` "_" must not be present ` in a filename convention specified? I can observe underscore being used in several well-known projects like OpenShift origin and K8s. I can also see file names with compound names containing underscore. I can't also find any mention of it in https://go.dev/doc/code#Overview.

I agree however that the file `is_jwt_token_expired` seems to specific for a file. `jwtutils.go` or `jwt_utils.go` might be a better name.

pkg/auth/is_jwt_token_expired.go
```diff
@@ -0,0 +1,17 @@
+package auth
+
+import (
+	"time"
+
+	"github.com/golang-jwt/jwt/v4"
+)
+
+// This method returns true if JWT token is expired, otherwise returns false
+func IsJWTTokenExpired(accessToken string) bool {
```
miguelsorianod, line 10: This method returns false if the provided accessToken is invalid. If we are going this route I think this should be documented in the method's doc.

I am also wondering whether this is the way we want to approach it. We could decide to return a `bool, error` instead.
Regarding the `no Keyfunc was provided` error being returned by Parse, that's because the `Parse` method requires a `keyFunc`.
Then the next question is whether we want to ensure signature verification when parsing the token. Depending on the decision taken this should be at least documented if it is not evident. If Token verification is desired then maybe the `GetJWTFromSignedToken` from the AuthHelper data type could be used (although I haven't looked into details what it implies). If no token verification is desired then https://pkg.go.dev/github.com/golang-jwt/jwt/v4@v4.2.0#Parser.ParseUnverified can be used, although with extreme caution as verification is not being done.

An insight on `IsJWTTokenExpired` is that it performs two things:
1. Parsing a raw token
2. Check the expiration

An alternative would be accepting a jwt v4 token (which is already parsed) and just perform the expiration check.

pkg/auth/is_jwt_token_expired.go
miguelsorianod, line 14: I think 	"github.com/golang-jwt/jwt/v4" itself provides you with that information.

See https://pkg.go.dev/github.com/golang-jwt/jwt/v4@v4.2.0#pkg-constants: `The errors that might occur when parsing and validating a token `.
In that case you can check for the `jwt.ValidationErrorExpired` error and maybe `|jwt.ValidationErrorNotValidYet` (depending on what we understand by "Is JWT token expired" method and where it is used. You can see an example of the error handling in the examples in https://pkg.go.dev/github.com/golang-jwt/jwt/v4@v4.2.0#Parse

pkg/auth/is_jwt_token_expired.go
Rajagopalan-Ranganathan: @miguelsorianod https://go.dev/doc/effective_go#mixed-caps - it talks about names in general :). File names i assume are a part of that. But, like you mentioned i see that "_" is used in lot of big projects, so it is something we can agree on KFM level i assume.

miguelsorianod: :+1: . mixedCaps in file names nor package names is definitely not something that I've seen (or at least seen commonly)

pkg/auth/is_jwt_token_expired.go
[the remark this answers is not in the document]
VanillaSpoon: Thanks @Rajagopalan-Ranganathan 
The arithmetic operations are in place as there are some difficulties with the value types. I can't seem to find an inbuilt method which will enable the comparison of the `exp` (float) with a `time.Time` value. Would either of you know how to approach this?

pkg/auth/is_jwt_token_expired.go
VanillaSpoon: Thanks for chiming in too @miguelsorianod 
After thinking over it and going through both of your points, I think jwt_utils.go would be the best naming for here. This removes the focus from a single method, allows us to add other util methods in the future, and satisfies the naming convention of what is already in place. Would you both agree?

pkg/auth/is_jwt_token_expired.go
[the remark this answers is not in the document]
miguelsorianod: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/989#discussion_r862781213 for a possible approach to take into consideration

pkg/auth/is_jwt_token_expired.go
VanillaSpoon: Yea, considering the specified use, this could lead to some future issues if misunderstood. Thanks for the clarification @akoserwal 

As this is called in:

VanillaSpoon: Yea, considering the specified use, this could lead to some future issues if misunderstood. Thanks for the clarification @akoserwal

As this is called in:
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/client/keycloak/client.go#L166
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/client/redhatsso/client.go#L97

Would I keep it separate in the client folder, or in keycloak/client.go as previously suggested?

VanillaSpoon:
Hey @machi1990 @akoserwal @miguelsorianod @ziccardi @Rajagopalan-Ranganathan 
I have pushed an alternative approach to this implementation. 

IsJWTTokenExpired()  has been moved to pkg/client/keycloak/client.go, and warnings about the lack of validation have been added as a comment above. 

Unfortunately the arithmetic operations on time.Now() are still standing.

 Thanks for the feedback so far, I would love further feedback if you feel there are further adjustments which could be made.

machi1990:
> Hey @machi1990 @akoserwal @miguelsorianod @ziccardi @Rajagopalan-Ranganathan I have pushed an alternative approach to this implementation.
> 
> IsJWTTokenExpired() has been moved to pkg/client/keycloak/client.go, and warnings about the lack of validation have been added as a comment above.
> 
> Unfortunately the arithmetic operations on time.Now() are still standing.
> 
> Thanks for the feedback so far, I would love further feedback if you feel there are further adjustments which could be made.

I'll let others have a look at this to see if their feedback have been addressed.

pkg/client/keycloak/client.go
```diff
@@ -175,6 +175,17 @@ func (kc *kcClient) GetToken() (string, error) {
 	return tokenResp.AccessToken, nil
 }
 
+// IsJWTTokenExpired returns false if JWT token is not expired, otherwise returns false
+// This method does NOT validate the token, and should not be used when validation is necessary
+func IsJWTTokenExpired(accessToken string) bool {
+	if token, _ := jwt.Parse(accessToken, nil); token != nil {
+		tokenClaims := token.Claims.(jwt.MapClaims)
+		exp := tokenClaims["exp"].(float64)
```
Rajagopalan-Ranganathan, line 183: You can follow this to get expiry in time format and using time before method to return true or false: https://stackoverflow.com/questions/58441574/how-to-parse-the-expiration-date-of-a-jwt-to-a-time-time-in-go

VanillaSpoon: Brilliant! Thanks @Rajagopalan-Ranganathan 
I've added that adjustment. Do you think I should separate `exp` into two steps or is it ok?

pkg/client/keycloak/client.go
```diff
@@ -175,6 +175,17 @@ func (kc *kcClient) GetToken() (string, error) {
 	return tokenResp.AccessToken, nil
 }
 
+// IsJWTTokenExpired returns false if JWT token is not expired, otherwise returns false
+// This method does NOT validate the token, and should not be used when validation is necessary
+func IsJWTTokenExpired(accessToken string) bool {
+	if token, _ := jwt.Parse(accessToken, nil); token != nil {
+		tokenClaims := token.Claims.(jwt.MapClaims)
+		exp := time.Unix(int64(tokenClaims["exp"].(float64)), 0)
```
Rajagopalan-Ranganathan, line 183: expTime would be better a variable name. WDYT?

Rajagopalan-Ranganathan: better to separate this as 2  steps and validate that expiry time is present, before performing any operations on it. According to the RFC: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4 use of this claim is optional, so better to check if exp is there, and valid and get it and then convert/typecast it to time.

VanillaSpoon: Good point @Rajagopalan-Ranganathan, thanks
@akoserwal  what are your thoughts on this? Would passing an empty exp return true?

pkg/client/keycloak/client.go
```diff
@@ -175,6 +175,17 @@ func (kc *kcClient) GetToken() (string, error) {
 	return tokenResp.AccessToken, nil
 }
 
+// IsJWTTokenExpired returns false if JWT token is not expired, otherwise returns false
```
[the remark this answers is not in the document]
VanillaSpoon: Thanks @miguelsorianod

pkg/client/keycloak/client.go
akoserwal: exp would be present in every jwt token. if it is not present than return true

machi1990:
@VanillaSpoon This PR needs a rebase.

Also @miguelsorianod @Rajagopalan-Ranganathan @akoserwal @ziccardi can you have a look if your comments have been addressed?

pkg/client/keycloak/client.go
VanillaSpoon: Thanks for the clarification @akoserwal. How is this?

```golang
func IsJWTTokenExpired(accessToken string) bool {
	if token, _ := jwt.Parse(accessToken, nil); token != nil {
		tokenClaims := token.Claims.(jwt.MapClaims)
		if _, ok := tokenClaims["exp"]; ok {
			expTime := time.Unix(int64(tokenClaims["exp"].(float64)), 0)
			return time.Now().After(expTime)
		}
	}
	return true
}
```

akoserwal (approved):
lgtm

PR #991, opened 2022-05-03
feat(connectors): fixes APPSRE-4651, added support for vault-secret-prefix flag to add a prefix to all aws vault secret keys, defaults to managed-connectors/

dhirajsb:
## Description
Added support for vault-secret-prefix flag to add a prefix to al aws vault secret keys, defaults to managed-connectors/

## Verification Steps
Included unit test can be run with valid aws credentials file. 

- [x] Code Review completed

### DIFF
--- a/internal/connector/internal/services/vault/vault_service_aws.go
+++ b/internal/connector/internal/services/vault/vault_service_aws.go
@@ -15,9 +15,10 @@ var OwnerResourceTagKey = "owner-resource"
 var _ VaultService = &awsVaultService{}
 
 type awsVaultService struct {
-	secretCache  *secretcache.Cache
-	secretClient *secretsmanager.SecretsManager
-	secretPrefix string
+	secretCache        *secretcache.Cache
+	secretClient       *secretsmanager.SecretsManager
+	secretPrefixEnable bool
+	secretPrefix       string
 }
 
 func NewAwsVaultService(vaultConfig *Config) (*awsVaultService, error) {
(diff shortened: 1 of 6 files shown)


heiko-braun:
@dhirajsb Can you hold this change for now until we had a chance to discuss the ramifications?

BumbleFeng:
I suggest making SecretPrefix optional, so we can separate promoting code with migrating secrets in two steps.

dhirajsb:
@BumbleFeng I added a flag to enable/disable secret prefix. I'll also add that to the cos-fleet-manager templates once this PR is merged. 
@lburgazzoli we need to get this merged for sre acceptance.

internal/connector/internal/services/vault/vault_service_aws.go
```diff
@@ -15,8 +15,10 @@ var OwnerResourceTagKey = "owner-resource"
 var _ VaultService = &awsVaultService{}
 
 type awsVaultService struct {
-	secretCache  *secretcache.Cache
-	secretClient *secretsmanager.SecretsManager
+	secretCache        *secretcache.Cache
+	secretClient       *secretsmanager.SecretsManager
+	secretPrefixEnable bool
```
valdar, line 20: Is an enabling boolean necessary?
After all if you specify a prefix you implicitly enable the feature right? What is the point of specifying a prefix and disabling the feature?

valdar (changes requested):

internal/connector/internal/services/vault/vault_service_aws.go
dhirajsb: Yes, it's necessary if one also wants to support a default value for a prefix. Otherwise it's a nightmare to figure out whether it's enabled or not by having to look at every point of configuration i.e, code, template, and app-services config. With a switch in the app-services config, it will be self documenting.

valdar (approved):

PR #998, opened 2022-05-04
fix: removed validators side effects

ziccardi:
## Description

Reimplementation of #988 without the use of a shared context.

We decided that we don't want to have a context, even if it didn't tie any of the validators to using it or expecting anything to be there. This has some drawback: some code needs to be executed more times to retrieve the same values: one time to validate it and one time to put it into the resulting object.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/handlers/kafka.go
+++ b/internal/kafka/internal/handlers/kafka.go
@@ -1,10 +1,8 @@
 package handlers
 
 import (
-	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/auth"
 	"net/http"
-
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/public"
 	config "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
@@ -40,52 +38,32 @@ func NewKafkaHandler(service services.KafkaService, providerConfig *config.Provi
 }
 
 func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
-	var kafkaRequest public.KafkaRequestPayload
+	var kafkaRequestPayload public.KafkaRequestPayload
 	ctx := r.Context()
-	convKafka := &dbapi.KafkaRequest{}
 
 	cfg := &handlers.HandlerConfig{
-		MarshalInto: &kafkaRequest,
+		MarshalInto: &kafkaRequestPayload,
 		Validate: []handlers.Validate{
 			handlers.ValidateAsyncEnabled(r, "creating kafka requests"),
-			handlers.ValidateLength(&kafkaRequest.Name, "name", handlers.MinRequiredFieldLength, &MaxKafkaNameLength),
-			ValidKafkaClusterName(&kafkaRequest.Name, "name"),
-			ValidateKafkaClusterNameIsUnique(&kafkaRequest.Name, h.service, r.Context()),
-			ValidateKafkaClaims(ctx, &kafkaRequest, convKafka),
-			ValidateCloudProvider(&h.service, convKafka, h.providerConfig, "creating kafka requests"),
-			func() *errors.ServiceError { // Validate plan
-				instanceType, err := h.service.AssignInstanceType(convKafka)
-				if err != nil {
-					return err
-				}
-				if stringSet(&kafkaRequest.Plan) {
-					plan := config.Plan(kafkaRequest.Plan)
-					instTypeFromPlan, e := plan.GetInstanceType()
-					if e != nil || instTypeFromPlan != string(instanceType) {
-						return errors.New(errors.ErrorBadRequest, fmt.Sprintf("Unable to detect instance type in plan provided: '%s'", kafkaRequest.Plan))
-					}
-					size, e1 := plan.GetSizeID()
-					if e1 != nil {
-						return errors.New(errors.ErrorBadRequest, fmt.Sprintf("Unable to detect instance size in plan provided: '%s'", kafkaRequest.Plan))
-					}
-					_, e2 := h.kafkaConfig.GetKafkaInstanceSize(instTypeFromPlan, size)
-
-					if e2 != nil {
-						return errors.InstancePlanNotSupported("Unsupported plan provided: '%s'", kafkaRequest.Plan)
-					}
-					convKafka.SizeId = size
-				} else {
-					rSize, err := h.kafkaConfig.GetFirstAvailableSize(instanceType.String())
-					if err != nil {
-						return errors.InstanceTypeNotSupported("Unsupported kafka instance type: '%s' provided", instanceType.String())
-					}
-					convKafka.SizeId = rSize.Id
-				}
-				convKafka.InstanceType = instanceType.String()
-				return nil
-			},
+			handlers.ValidateLength(&kafkaRequestPayload.Name, "name", handlers.MinRequiredFieldLength, &MaxKafkaNameLength),
+			ValidKafkaClusterName(&kafkaRequestPayload.Name, "name"),
+			ValidateKafkaClusterNameIsUnique(&kafkaRequestPayload.Name, h.service, r.Context()),
+			ValidateKafkaClaims(ctx, ValidateUsername(), ValidateOrganisationId()),
+			ValidateCloudProvider(ctx, &h.service, &kafkaRequestPayload, h.providerConfig, "creating kafka requests"),
+			ValidateKafkaPlan(ctx, &h.service, h.kafkaConfig, &kafkaRequestPayload),
 		},
 		Action: func() (interface{}, *errors.ServiceError) {
+			convKafka := presenters.ConvertKafkaRequest(kafkaRequestPayload)
+
+			claims, _ := getClaims(ctx)
+			convKafka.Owner = auth.GetUsernameFromClaims(claims)
+			convKafka.OrganisationId = auth.GetOrgIdFromClaims(claims)
+			convKafka.OwnerAccountId = auth.GetAccountIdFromClaims(claims)
+
+			convKafka.InstanceType, convKafka.SizeId, _ = getInstanceTypeAndSize(ctx, &h.service, h.kafkaConfig, &kafkaRequestPayload)
+
+			convKafka.CloudProvider, convKafka.Region, _ = getCloudProviderAndRegion(ctx, &h.service, &kafkaRequestPayload, h.providerConfig)
+
 			svcErr := h.service.RegisterKafkaJob(convKafka)
 			if svcErr != nil {
 				return nil, svcErr
--- a/internal/kafka/internal/handlers/kafka.go
+++ b/internal/kafka/internal/handlers/kafka.go
@@ -1,7 +1,6 @@
 package handlers
 
 import (
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/auth"
 	"net/http"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/public"
@@ -56,9 +55,9 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
 			convKafka := presenters.ConvertKafkaRequest(kafkaRequestPayload)
 
 			claims, _ := getClaims(ctx)
-			convKafka.Owner = auth.GetUsernameFromClaims(claims)
-			convKafka.OrganisationId = auth.GetOrgIdFromClaims(claims)
-			convKafka.OwnerAccountId = auth.GetAccountIdFromClaims(claims)
+			convKafka.Owner, _ = claims.GetUsername()
+			convKafka.OrganisationId, _ = claims.GetOrgId()
+			convKafka.OwnerAccountId, _ = claims.GetAccountId()
 
 			convKafka.InstanceType, convKafka.SizeId, _ = getInstanceTypeAndSize(ctx, &h.service, h.kafkaConfig, &kafkaRequestPayload)
 
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -222,7 +222,7 @@ func ValidateUsername() ValidateKafkaClaimsOptions {
 	return func(claims *jwt.MapClaims) *errors.ServiceError {
 		username := auth.GetUsernameFromClaims(*claims)
 		if username == "" {
-			return errors.New(errors.ErrorBadRequest, "can't find 'user' attribute in claims")
+			return errors.New(errors.ErrorForbidden, "can't find neither 'user' or 'preferred_username' attribute in claims")
 		}
 		return nil
 	}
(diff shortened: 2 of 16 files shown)


internal/kafka/internal/handlers/validation.go
```diff
@@ -145,18 +208,19 @@ func ValidateKafkaUserFacingUpdateFields(ctx context.Context, authService author
...
+	return claims, nil
+}
 
-		return nil
+// ValidateKafkaClaims - Verifies that the context contains the required claims
+func ValidateKafkaClaims(ctx context.Context) handlers.Validate {
+	return func() *errors.ServiceError {
+		_, err := getClaims(ctx)
+		return err
 	}
```
machi1990, line 224: What claims are being validated? I think this can be removed all together if we do not validate specific claims

machi1990 (commented):
overall, looks good to me.

There is a room to review how certain values are set from the defaults if they are not provided in the request body.

internal/kafka/internal/handlers/kafka.go
```diff
@@ -36,52 +34,32 @@ func NewKafkaHandler(service services.KafkaService, providerConfig *config.Provi
...
+			convKafka := presenters.ConvertKafkaRequest(kafkaRequestPayload)
+
+			claims, _ := getClaims(ctx)
+			convKafka.Owner = auth.GetUsernameFromClaims(claims)
+			convKafka.OrganisationId = auth.GetOrgIdFromClaims(claims)
+			convKafka.OwnerAccountId = auth.GetAccountIdFromClaims(claims)
+
+			convKafka.InstanceType, convKafka.SizeId, _ = getInstanceTypeAndSize(ctx, &h.service, h.kafkaConfig, &kafkaRequestPayload)
+
+			convKafka.CloudProvider, convKafka.Region, _ = getCloudProviderAndRegion(ctx, &h.service, &kafkaRequestPayload, h.providerConfig)
```
machi1990, line 61: This is not related to the PR, but I think cloud provider and cloud region should always be given as part of the request payload. I'll start a separate discussion

internal/kafka/internal/handlers/validation.go
ziccardi: It ensure that the request contains the claims.

machi1990: > It ensure that the request contains the claims.

...but this is done in other validators already that rely on claims.
 
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/f26d8f9fce7075e3b309928fdfe8bf24eca8bdc1/internal/kafka/internal/handlers/validation.go#L79-L82
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/f26d8f9fce7075e3b309928fdfe8bf24eca8bdc1/internal/kafka/internal/handlers/validation.go#L116-L119

ziccardi: true, but that is only by incident, because those validators needs authentication.
If we want to be purist in the rule design pattern, then each rule should make one check. That means that one check will be used to check for claims to be present, while the other checks will be relied upon only for their respective responsibilities.

The other two checks you linked are responsible of checking the provider and region values and the instance type and size values. Relying on those for checking the claims presence would break the rule design pattern contract.

machi1990: > true, but that is only by incident, because those validators needs authentication.

You nailed it. This is where I wanted to arrive.. to me the validation as it is at the moment, seems to be performing authentication related validation and I think we have a middleware for that. Is this correct?

ziccardi: yup. And I don't think the validators should rely upon the presence or not of such middleware.

machi1990: > I think the validators should check that everything that is needed by the handler is present and is valid.

+1

> So, if the handler needs the claims to be present, then the validators should check for the claims to be present

+1 on this if we performed specific validation on specific claims, which is not the case for the moment. And the validator is reduced to performing a role which is already performed at a layer above (the middleware).

ziccardi: Added options to the validator so that the claims validator can be configured enabling or disabling validations.
Mind giving another look?

machi1990: Thanks. Now the validation makes sense.

internal/kafka/internal/services/kafka.go
```diff
@@ -83,7 +83,7 @@ type KafkaService interface {
 	Updates(kafkaRequest *dbapi.KafkaRequest, values map[string]interface{}) *errors.ServiceError
 	ChangeKafkaCNAMErecords(kafkaRequest *dbapi.KafkaRequest, action KafkaRoutesAction) (*route53.ChangeResourceRecordSetsOutput, *errors.ServiceError)
 	GetCNAMERecordStatus(kafkaRequest *dbapi.KafkaRequest) (*CNameRecordStatus, error)
-	AssignInstanceType(kafkaRequest *dbapi.KafkaRequest) (types.KafkaInstanceType, *errors.ServiceError)
+	AssignInstanceType(owner string, organisationID string) (types.KafkaInstanceType, *errors.ServiceError)
```
miguelsorianod, line 86: I like this change :+1: , the smaller the number of dependencies the better

internal/kafka/internal/handlers/validation.go
```diff
@@ -51,46 +51,109 @@ func ValidateKafkaClusterNameIsUnique(name *string, kafkaService services.KafkaS
...
-// provider and region
-func ValidateCloudProvider(kafkaService *services.KafkaService, kafkaRequest *dbapi.KafkaRequest, providerConfig *config.ProviderConfig, action string) handlers.Validate {
-	return func() *errors.ServiceError {
-		// Set Cloud Provider default if not received in the request
-		supportedProviders := providerConfig.ProvidersConfig.SupportedProviders
-		if kafkaRequest.CloudProvider == "" {
-			defaultProvider, _ := supportedProviders.GetDefault()
-			kafkaRequest.CloudProvider = defaultProvider.Name
-		}
+func getCloudProviderAndRegion(
```
miguelsorianod, line 54: Just as a thought, do you think this could be further split?

Quickly looking at it, it seems it validates the value if it is provided. If it is not provided then a default is assigned.

We could have a method that only validates, and then on the action handler it could be checked whether the attribute is defined. If it is defined we know it is valid so we can just assign it. If it is not defined we know we should set the default value.
In this way we would avoid repeating the validation. It might not be worth it though

internal/kafka/internal/handlers/validation.go
```diff
@@ -51,46 +51,109 @@ func ValidateKafkaClusterNameIsUnique(name *string, kafkaService services.KafkaS
...
+
+// ValidateCloudProvider returns a validator that validates provided provider and region
+func ValidateCloudProvider(ctx context.Context, kafkaService *services.KafkaService, kafkaRequest *public.KafkaRequestPayload, providerConfig *config.ProviderConfig, action string) handlers.Validate {
+	return func() *errors.ServiceError {
+		_, _, err := getCloudProviderAndRegion(ctx, kafkaService, kafkaRequest, providerConfig)
+		return err
+	}
+}
+
+func getInstanceTypeAndSize(ctx context.Context, kafkaService *services.KafkaService, kafkaConfig *config.KafkaConfig, kafkaRequestPayload *public.KafkaRequestPayload) (string, string, *errors.ServiceError) {
```
miguelsorianod, line 115: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/998/files#r865011118

internal/kafka/internal/handlers/validation.go
```diff
@@ -145,21 +208,50 @@ func ValidateKafkaUserFacingUpdateFields(ctx context.Context, authService author
...
-			return errors.Unauthenticated("user not authenticated")
+func getClaims(ctx context.Context) (jwt.MapClaims, *errors.ServiceError) {
+	claims, err := auth.GetClaimsFromContext(ctx)
+	if err != nil {
+		return nil, errors.Unauthenticated("user not authenticated")
+	}
+	return claims, nil
+}
+
+type ValidateKafkaClaimsOptions func(claims *jwt.MapClaims) *errors.ServiceError
```
miguelsorianod, line 219: Why the "options" name?  might `KafkaClaimValidation` be a better name?

internal/kafka/internal/handlers/kafka.go
```diff
@@ -36,52 +34,32 @@ func NewKafkaHandler(service services.KafkaService, providerConfig *config.Provi
...
 		},
 		Action: func() (interface{}, *errors.ServiceError) {
+			convKafka := presenters.ConvertKafkaRequest(kafkaRequestPayload)
+
+			claims, _ := getClaims(ctx)
+			convKafka.Owner = auth.GetUsernameFromClaims(claims)
+			convKafka.OrganisationId = auth.GetOrgIdFromClaims(claims)
+			convKafka.OwnerAccountId = auth.GetAccountIdFromClaims(claims)
+
+			convKafka.InstanceType, convKafka.SizeId, _ = getInstanceTypeAndSize(ctx, &h.service, h.kafkaConfig, &kafkaRequestPayload)
```
miguelsorianod, line 59: Is there some reason we are ignoring the error? I understand it is not being done because at this point "it should not fail" but still :thinking:

internal/kafka/internal/handlers/validation.go
ziccardi: same as before, the instance type is used to get the size. 
Without a validation context, they can't be splitted because to validate the size you need the instance type.
We could split and have a `getInstanceType` and a `getSize(instancetype)` but since we use them always together we wouldn't have any benefit.

internal/kafka/internal/handlers/kafka.go
ziccardi: We are ignoring the error because those values has been already validate by the validators so we are sure they won't fail.

internal/kafka/internal/handlers/validation.go
miguelsorianod: I didn't mean splitting the region and provider logic, but splitting the validation from the retrieval

internal/kafka/internal/handlers/validation.go
miguelsorianod: :+1: , as you prefer then. I thought `KafkaClaimValidation` fit there as it is a validation element that is passed to the ValidateClaims method

miguelsorianod:
> I don't really agree. We are duplicating database and AMS calls just to avoid having a context (which is not an anti-pattern and is not uncommon). I believe the previous approach (use it from validation context or get it yourself) didn't add any complexities at all since it didn't create any relation/dependency between the validators. On the other side, it allowed to avoid any duplication. But, as I said, I'm democratic and will follow the majority :D

Let's go ahead with this. It is a tradeoff matter. If we merge this and we notice an impact on performance then we can reevaluate the solution to take

internal/kafka/internal/handlers/validation.go
```diff
@@ -145,21 +208,50 @@ func ValidateKafkaUserFacingUpdateFields(ctx context.Context, authService author
...
+
+func ValidateUsername() ValidateKafkaClaimsOptions {
+	return func(claims *jwt.MapClaims) *errors.ServiceError {
+		username := auth.GetUsernameFromClaims(*claims)
+		if username == "" {
+			return errors.New(errors.ErrorBadRequest, "can't find 'user' attribute in claims")
```
machi1990, line 225: Should this be a forbidden code instead of bad request?

internal/kafka/internal/handlers/validation.go
```diff
@@ -145,21 +208,50 @@ func ValidateKafkaUserFacingUpdateFields(ctx context.Context, authService author
...
-		(*kafkaRequest).OwnerAccountId = auth.GetAccountIdFromClaims(claims)
+		return nil
+	}
+}
 
+func ValidateOrganisationId() ValidateKafkaClaimsOptions {
+	return func(claims *jwt.MapClaims) *errors.ServiceError {
+		orgId := auth.GetOrgIdFromClaims(*claims)
+		if orgId == "" {
+			return errors.New(errors.ErrorBadRequest, "can't find 'organisation id' attribute in claims")
```
machi1990, line 235: Should this be a forbidden code instead of bad request?

internal/kafka/internal/handlers/validation.go
machi1990, line 225: `user` in the error reason message can be misleading. The claims we are interested are configurable at https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/4e3302d57225475aad5e54e286041946fe8a4c8b/pkg/auth/context.go#L25 or the alternative one at https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/4e3302d57225475aad5e54e286041946fe8a4c8b/pkg/auth/context.go#L30 so I think we either:
- include the right claim key in the message 
- or reword the message a bit to make it generic and avoid sending the claim key which can be confusing to the consumer of the API

internal/kafka/internal/handlers/validation.go
machi1990, line 235: the `organisation id` attribute can be confusing:
The claim key in use are defined in:
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/4e3302d57225475aad5e54e286041946fe8a4c8b/pkg/auth/context.go#L26
- or the alternative one in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/4e3302d57225475aad5e54e286041946fe8a4c8b/pkg/auth/context.go#L35

and these can be configured as well i.e they are dynamic so I think we should use the right claims key or reword the message a bit to avoid mentioning wrong claim key

machi1990 (changes requested):
lgtm as well.

left a few minor comments. Let me know what you think

internal/kafka/internal/handlers/validation.go
ziccardi: The message is already generic and is not mentioning the claim key. However, I believe it should be coherent with the same error about the `username`, so I'm going to change it here to include correct key names

machi1990: Thanks @ziccardi

machi1990 (approved):
lgtm

I'd have made sure that we are using KFMClaims object everywhere in the same PR.
And also add a documentation if possible.
I'll let you decide if you want to add it on this PR or the next one

Also /cc @miguelsorianod can you have another look? There has been some new additions since your last review.

internal/kafka/internal/handlers/validation.go
```diff
@@ -51,46 +51,109 @@ func ValidateKafkaClusterNameIsUnique(name *string, kafkaService services.KafkaS
...
-		}
+	// Set Cloud Provider default if not received in the request
+	supportedProviders := providerConfig.ProvidersConfig.SupportedProviders
 
-		// Set Cloud Region default if not received in the request
-		if kafkaRequest.Region == "" {
-			defaultRegion, _ := provider.GetDefaultRegion()
-			kafkaRequest.Region = defaultRegion.Name
-		}
+	defaultProvider, _ := supportedProviders.GetDefault()
```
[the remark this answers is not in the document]
ziccardi: I guess the reaso is that if a default provider is not found, an empty provider instance is returned, together with the error.
Since the empty provider will have an empty name, the next `supportedProvider.GetByName` will fail.

miguelsorianod (approved):
It looks good to me. I left a minor comment with a possible change

pkg/auth/context_config.go
```diff
@@ -4,6 +4,21 @@ import (
 	"github.com/spf13/pflag"
 )
 
+var (
```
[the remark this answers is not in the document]
miguelsorianod: oh, I see. :+1:  then

PR #999, opened 2022-05-04
 feat: adding option to configure kafka owners

akoserwal:
Creating a draft pr to get input on the approach.

Only for development/testing allow to setup kafka owners using a configuration file.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/config/kafka.go
+++ b/internal/kafka/internal/config/kafka.go
@@ -30,6 +30,9 @@ type KafkaConfig struct {
 	KafkaLifespan          *KafkaLifespanConfig               `json:"kafka_lifespan"`
 	Quota                  *KafkaQuotaConfig                  `json:"kafka_quota"`
 	SupportedInstanceTypes *KafkaSupportedInstanceTypesConfig `json:"kafka_supported_sizes"`
+	EnableKafkaOwnerConfig bool                               `json:"enable_kafka_owner_config"`
+	KafkaOwnerList         []string                           `json:"kafka_owner_list"`
+	KafkaOwnerListFile     string                             `json:"kafka_owner_list_file"`
 }
 
 func NewKafkaConfig() *KafkaConfig {
@@ -37,10 +40,12 @@ func NewKafkaConfig() *KafkaConfig {
 		KafkaTLSCertFile:               "secrets/kafka-tls.crt",
 		KafkaTLSKeyFile:                "secrets/kafka-tls.key",
 		EnableKafkaExternalCertificate: false,
+		EnableKafkaOwnerConfig:         false,
 		KafkaDomainName:                "kafka.bf2.dev",
 		KafkaLifespan:                  NewKafkaLifespanConfig(),
 		Quota:                          NewKafkaQuotaConfig(),
 		SupportedInstanceTypes:         NewKafkaSupportedInstanceTypesConfig(),
+		KafkaOwnerListFile:             "config/kafka-owner-list.yaml",
 		BrowserUrl:                     "http://localhost:8080/",
 	}
 }
@@ -56,6 +61,8 @@ func (c *KafkaConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&c.Quota.AllowDeveloperInstance, "allow-developer-instance", c.Quota.AllowDeveloperInstance, "Allow the creation of kafka developer instances")
 	fs.StringVar(&c.SupportedInstanceTypes.ConfigurationFile, "supported-kafka-instance-types-config-file", c.SupportedInstanceTypes.ConfigurationFile, "File containing the supported instance types configuration")
 	fs.StringVar(&c.BrowserUrl, "browser-url", c.BrowserUrl, "Browser url to kafka admin UI")
+	fs.BoolVar(&c.EnableKafkaOwnerConfig, "enable-kafka-owner-config", c.EnableKafkaOwnerConfig, "Enable configuration for setting kafka owners")
+	fs.StringVar(&c.KafkaOwnerListFile, "kafka-owner-list-file", c.KafkaOwnerListFile, "File containing list of kafka owners")
 }
 
 func (c *KafkaConfig) ReadFiles() error {
@@ -66,6 +73,12 @@ func (c *KafkaConfig) ReadFiles() error {
 	err = shared.ReadFileValueString(c.KafkaTLSKeyFile, &c.KafkaTLSKey)
 	if err != nil {
 		return err
+	}
+	if c.EnableKafkaOwnerConfig {
+		err = shared.ReadYamlFile(c.KafkaOwnerListFile, &c.KafkaOwnerList)
+		if err != nil {
+			return err
+		}
 	}
 
 	err = shared.ReadYamlFile(c.SupportedInstanceTypes.ConfigurationFile, &c.SupportedInstanceTypes.Configuration)
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -1077,9 +1077,7 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 				KafkaIBP: kafkaRequest.DesiredKafkaIBPVersion,
 			},
 			Deleted: kafkaRequest.Status == constants2.KafkaRequestStatusDeprovision.String(),
-			Owners: []string{
-				kafkaRequest.Owner,
-			},
+			Owners:  buildKafkaOwner(kafkaRequest, kafkaConfig),
 		},
 		Status: managedkafka.ManagedKafkaStatus{},
 	}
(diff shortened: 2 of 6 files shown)


internal/kafka/internal/config/kafka.go
```diff
@@ -56,6 +61,8 @@ func (c *KafkaConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&c.Quota.AllowDeveloperInstance, "allow-developer-instance", c.Quota.AllowDeveloperInstance, "Allow the creation of kafka developer instances")
 	fs.StringVar(&c.SupportedInstanceTypes.ConfigurationFile, "supported-kafka-instance-types-config-file", c.SupportedInstanceTypes.ConfigurationFile, "File containing the supported instance types configuration")
 	fs.StringVar(&c.BrowserUrl, "browser-url", c.BrowserUrl, "Browser url to kafka admin UI")
+	fs.BoolVar(&c.EnableKafkaOwnerConfig, "enable-kafka-owner-config", c.EnableKafkaOwnerConfig, "Enable configuration for setting kafka owners")
+	fs.StringVar(&c.KafkaOwnerListFile, "kafka-owner-list-file", c.KafkaOwnerListFile, "File containing list of kafka owners")
```
machi1990, line 65: great.
We may also need to:
- add the corresponding parameters and configmap of the file in the `service-template.yaml` file: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3debe34479689bf45ee61292fe627537742fee82/templates/service-template.yml. 
- add the way  to override the parameters in the `make deploy/service` target in the `Makefile`. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3debe34479689bf45ee61292fe627537742fee82/Makefile#L714-L823
- add a documentation in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3debe34479689bf45ee61292fe627537742fee82/docs/deploying-kas-fleet-manager-to-openshift.md#deploy-kas-fleet-manager

internal/kafka/internal/services/kafka.go
```diff
@@ -1127,6 +1125,15 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 	return managedKafkaCR, nil
 }
 
+func buildKafkaOwner(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.KafkaConfig) []string {
+	if kafkaConfig.EnableKafkaOwnerConfig {
+		return append(kafkaConfig.KafkaOwnerList, kafkaRequest.Owner)
+	}
```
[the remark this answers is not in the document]
akoserwal: yes, make sense. updated

templates/service-template.yml
```diff
@@ -832,6 +836,9 @@ objects:
             - name: service-account-limits-check-skip-org-id-list
               mountPath: /config/service-account-limits-check-skip-org-id-list.yaml
               subPath: service-account-limits-check-skip-org-id-list.yaml
+            - name: kas-fleet-manager-kafka-owner-list
+              mountPath: /config/kafka-owner-list.yaml
+              subPath: kafka-owner-list.yaml
```
machi1990, line 841: We also need to expose a parameter that takes a list of owners and use it to construct the ConfigMap
For example, see how it was done for the supported instance type:
 
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/250a94108e3741e55cc9b0f9511667d6cfdc8291/templates/service-template.yml#L160
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/250a94108e3741e55cc9b0f9511667d6cfdc8291/templates/service-template.yml#L549-L557
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/250a94108e3741e55cc9b0f9511667d6cfdc8291/templates/service-template.yml#L733-L735

Rajagopalan-Ranganathan:
Trying to understand , why we have a "list" of owners for a Kafka, rather than a single owner?

templates/service-template.yml
```diff
@@ -732,6 +749,9 @@ objects:
           - name: kas-fleet-manager-allowed-users-config
             configMap:
               name: kas-fleet-manager-allowed-users-config
+          - name: kas-fleet-manager-kafka-owner-list
+              configMap:
+                name: kas-fleet-manager-kafka-owner-list
```
machi1990, line 754: This fails with
```
kas-fleet-manager/templates/service-template.yml
 753:24  error  parser  Bad indentation of a mapping entry
 753:25  error  parser  Incomplete explicit mapping pair; a key node is missed
```
looks like the indentation is not okay

akoserwal:
> Trying to understand , why we have a "list" of owners for a Kafka, rather than a single owner?

This is an additional config for development/automated script. Kafka instance team automated tests are blocked. So, using this config additional owners can be configured.

akoserwal:
@MikeEdgar

MikeEdgar:
This would allow developers to create an instance where their user account is listed as an owner _and_ a service account (id known at time of fleet-manager deployment) is also an owner. This would allow the service account to manage the new instance immediately without additional ACL setup.

machi1990:
> @MikeEdgar

Tagging @daveconde @franvila as they'll be interested on this as well

MikeEdgar:
@akoserwal - confirmed with kas-installer :
1. Create service account in mas-sso
2. Deploy fleet-manager with `ENABLE_KAFKA_OWNER=true` and `KAFKA_OWNERS` containing the service account
3. Create Kafka instance
4. ManagedKafka CR contains two owners (including the service account)
5. Able to create a topic using the service account without any ACL configuration

akoserwal:
> @akoserwal - confirmed with kas-installer :
> 
> 1. Create service account in mas-sso
> 2. Deploy fleet-manager with `ENABLE_KAFKA_OWNER=true` and `KAFKA_OWNERS` containing the service account
> 3. Create Kafka instance
> 4. ManagedKafka CR contains two owners (including the service account)
> 5. Able to create a topic using the service account without any ACL configuration

Correct

Rajagopalan-Ranganathan:
@MikeEdgar - this a tangent with respect to our current authz policies, where there is only one owner. What is the need for this? To skip ACL configuration?

Rajagopalan-Ranganathan:
> > What is the need for this? To skip ACL configuration?
> 
> Yes, for use within testing pipelines. I believe this is a temporary solution until we are using a single SSO and service accounts can be used to create Kafka instances.

Thanks for the clarification @MikeEdgar , @akoserwal if this to be used for developer/testing/standalone purpose only, should we wrap owner list config with that additional check as well? i couldnt find that

akoserwal:
> > > What is the need for this? To skip ACL configuration?
> > 
> > 
> > Yes, for use within testing pipelines. I believe this is a temporary solution until we are using a single SSO and service accounts can be used to create Kafka instances.
> 
> Thanks for the clarification @MikeEdgar , @akoserwal if this to be used for developer/testing/standalone purpose only, should we wrap owner list config with that additional check as well? i couldnt find that

ENABLE_KAFKA_OWNER flag is kept as default false. It can be enabled when require for dev/testing purpose

Rajagopalan-Ranganathan:
> > > > What is the need for this? To skip ACL configuration?
> > > 
> > > 
> > > Yes, for use within testing pipelines. I believe this is a temporary solution until we are using a single SSO and service accounts can be used to create Kafka instances.
> > 
> > 
> > Thanks for the clarification @MikeEdgar , @akoserwal if this to be used for developer/testing/standalone purpose only, should we wrap owner list config with that additional check as well? i couldnt find that
> 
> ENABLE_KAFKA_OWNER flag is kept as default false. It can be enabled when require for dev/testing purpose

That is not enough, IMO, its a flag that should be used only for dev/testing purpose, if there is way to secure it further with a deployment model that should be fine. there should be no chance of misusing this.

MikeEdgar:
> Does Data plane authroize the service account present in the owners list? In a sense, User A is owner and only user A service account can be used to manage that Kafka? And if there is a user B service account added to that list , where owner is user A what happens?

@Rajagopalan-Ranganathan the ManagedKafka CR supports a list of owners, They are all treated the same regardless of the number of entries in the list.

Rajagopalan-Ranganathan:
> > Does Data plane authroize the service account present in the owners list? In a sense, User A is owner and only user A service account can be used to manage that Kafka? And if there is a user B service account added to that list , where owner is user A what happens?
> 
> @Rajagopalan-Ranganathan the ManagedKafka CR supports a list of owners, They are all treated the same regardless of the number of entries in the list.

My question is not about the number of entries in the list, rather content of the list. So in theory, User B service account can be treated as owner for a user A Kafka, that is correct ? via this change

MikeEdgar:
> User B service account can be treated as owner for a user A Kafka, that is correct ? via this change

Yes, that's right.

Rajagopalan-Ranganathan:
> > User B service account can be treated as owner for a user A Kafka, that is correct ? via this change
> 
> Yes, that's right.

that being the case, i am highly skeptical about this temporary solution. This requires a wider discussion , probably can be discussed in security stream call tomorrow

Rajagopalan-Ranganathan:
No objections were raised in the Security stream call regarding this temporary solution.

machi1990 (approved):
lgtm

I'll let @Rajagopalan-Ranganathan and @MikeEdgar have another look as well

MikeEdgar (approved):
No concerns from me. Worked as expected.

wtrocki:
> This would allow developers to create an instance where their user account is listed as an owner and a service account (id known at time of fleet-manager deployment) is also an owner. This would allow the service account to manage the new instance immediately without additional ACL setup.

Wow! This is amazing! Could that be something that we can reuse in user facing artifacts?

akoserwal:
@wtrocki Please can elaborate on the "user facing artifacts?"

wtrocki:
UI/CLI etc.

akoserwal:
> UI/CLI etc.
This can't be enabled for stage/production development. Only used for local dev

wtrocki:
Ah.. ok. Thank you

PR #1000, opened 2022-05-04
feat: orderBy parameter validation.

valdar:
## Description
Added orderBy parameter validation.

ListArguments.Validate() has been enhanced with a version that takes allowed parameters names as an argument.

## Verification Steps

- [ ] Code Review completed


lburgazzoli (approved):

pkg/services/types.go
```diff
@@ -75,7 +80,7 @@ func (la *ListArguments) Validate() error {
 				return errors.Errorf("invalid order by clause '%s'", orderByClause)
 			}
 
-			if !arrays.Contains(GetAcceptedOrderByParams(), keywords[0]) {
+			if !arrays.Contains(acceptedOrderByParams, keywords[0]) {
```
machi1990, line 83: I wonder whether split the accepted order by params per service?
in the array `acceptedOrderByParams` there are mixed columns that applies to only connectors or kafkas.

/cc @ziccardi

ziccardi: @machi1990 Not sure I get what you mean. The `acceptedOrderByParams` is a parameter containing only the columns valid for `connectors` order by IMU

machi1990: You are right. 

I was looking https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/a3566fed2168e91b10299b213968d9d7057e0050/pkg/services/types.go#L15 and I thought this is to be used by all the fleet-managers.

But it seems like that list of columns seems to be specific for Kafka, and I wonder if we should just move it to `internal/kafka` folder.

ziccardi: Agree

valdar: Well yes probably moving it would make sense, I have paved the way for that move to be smother :)

machi1990: Thanks @valdar, is this something that you'd like to address as part of the PR while we are at it?

dhirajsb: We did something similar for query search columns.

dhirajsb: We should probably move the kafka query search columns as well at pkg/services/queryparser/query_parser.go:12

machi1990: I think that's fine. What do you think @ziccardi ?

Do we need a method or a variable could do the job as well? If we still need a method, then not making it public i.e not starting it with caps, will be my only suggestion.

ziccardi (approved):
lgtm

pkg/services/types.go
ziccardi: I believe the method is not needed. A variable could do the job as well.

valdar: The rationale behind the method is that you get each time you use it a clean slice, there is less chance of accidental modification in a part of the code that affects in unpredictable ways other part of the code, it is the closest thing to a "constant slice" that can not be declared.
That said, it is used just once and never modified...

PR #1001, opened 2022-05-04
feat: Set minimum TLS version as TLS 1.2 for the HTTPS servers

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-2123.

TLS v1.2 should be the minimum accepted version when offering a HTTPS server.

KAS Fleet Manager offers the following HTTP/S servers, where HTTPS can be enabled individually for each of them:
* KAS Fleet Manager API server
* KAS Fleet Manager metrics server
* KAS Fleet Manager healthcheck server

This change should be transparent for the kas fleet manager stage and production environments. See the referenced Jira issue for details and an explanation on why this should be the case.

This also has an impact on connectors part of the code.

## Verification Steps
To verify it, compile the binary and then:

1. Generate a self-signed certificate. For example with: 
```
mkdir samplecerts
cd samplecerts
openssl req -newkey rsa:2048 -nodes -keyout key.pem -x509 -days 365 -out certificate.pem # When asked about input, any values will work for this purpose
```
1. Run kas fleet manager enabling https for all the servers and providing the certificates. For example:
```
./kas-fleet-manager serve --dataplane-cluster-scaling-type=none --enable-https --enable-metrics-https --enable-health-check-https --https-cert-file=samplecerts/certificate.pem --https-key-file=samplecerts/key.pem
```

Then perform requests to the https servers forcing the tls version to be a smaller tls version than 1.2. For example:
```
curl --tls-max 1.0 -k  -H "Authorization: Bearer $(ocm token)" https://127.0.0.1:8000/api/kafkas_mgmt/v1/ 
curl --tls-max 1.1 -k  -H "Authorization: Bearer $(ocm token)" https://127.0.0.1:8000/api/kafkas_mgmt/v1/ 
curl --tls-max 1.0 -k  -H "Authorization: Bearer $(ocm token)" https://127.0.0.1:8080/metrics
curl --tls-max 1.1 -k  -H "Authorization: Bearer $(ocm token)" https://127.0.0.1:8080/metrics
curl --tls-max 1.0 -k  -H "Authorization: Bearer $(ocm token)" https://127.0.0.1:8083/healthcheck 
curl --tls-max 1.1 -k  -H "Authorization: Bearer $(ocm token)" https://127.0.0.1:8083/healthcheck 
```

For all of those errors should be returned.

Then verify that requests without forcing any TLS version work for all the servers. With curl you should also be able to see what is the TLS version that is automatically negotiated. It should be at minimum TLS 1.2.
Then verify that forcing a request using tls 1.2 and using tls 1.3 works too.

- [ ] Code Review completed

### DIFF
--- a/pkg/server/api_server.go
+++ b/pkg/server/api_server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net"
 	"net/http"
@@ -115,6 +116,9 @@ func NewAPIServer(options ServerOptions) *ApiServer {
 	s.httpServer = &http.Server{
 		Addr:    options.ServerConfig.BindAddress,
 		Handler: mainHandler,
+		TLSConfig: &tls.Config{
+			MinVersion: options.ServerConfig.MinTLSVersion,
+		},
 	}
 
 	return s
--- a/pkg/server/healthcheck_config.go
+++ b/pkg/server/healthcheck_config.go
@@ -1,18 +1,26 @@
 package server
 
 import (
+	"crypto/tls"
+
 	"github.com/spf13/pflag"
 )
 
 type HealthCheckConfig struct {
 	BindAddress string `json:"bind_address"`
 	EnableHTTPS bool   `json:"enable_https"`
+	// Minimum TLS version accepted by the healthcheck server. Only takes effect when
+	// EnableHTTPS is true. The data type is uint16 due to golang's
+	// tls package accepts the versions in uint16 format, whose values
+	// are available as constants in that same package
+	MinTLSVersion uint16
 }
 
 func NewHealthCheckConfig() *HealthCheckConfig {
 	return &HealthCheckConfig{
-		BindAddress: "localhost:8083",
-		EnableHTTPS: false,
+		BindAddress:   "localhost:8083",
+		EnableHTTPS:   false,
+		MinTLSVersion: tls.VersionTLS12,
 	}
 }
 
--- a/pkg/server/healthcheck_server.go
+++ b/pkg/server/healthcheck_server.go
@@ -2,11 +2,13 @@ package server
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services/sentry"
 	"net"
 	"net/http"
 	"time"
+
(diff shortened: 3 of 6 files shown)


pkg/server/api_server.go
```diff
@@ -115,6 +116,9 @@ func NewAPIServer(options ServerOptions) *ApiServer {
 	s.httpServer = &http.Server{
 		Addr:    options.ServerConfig.BindAddress,
 		Handler: mainHandler,
+		TLSConfig: &tls.Config{
+			MinVersion: options.ServerConfig.MinTLSVersion,
+		},
```
machi1990, line 121: Do we need to set this conditionally based on whether https is enabled?

miguelsorianod:
cc @dhirajsb @lburgazzoli as I understand this affects connectors too.

machi1990 (commented):
This is worth addition also in the ffm project

pkg/server/api_server.go
miguelsorianod: I don't think so.
See the description of the TLSConfig field in http#Server: https://pkg.go.dev/net/http#Server :

```
// TLSConfig optionally provides a TLS configuration for use
	// by ServeTLS and ListenAndServeTLS. Note that this value is
	// cloned by ServeTLS and ListenAndServeTLS, so it's not
	// possible to modify the configuration with methods like
	// tls.Config.SetSessionTicketKeys. To use
	// SetSessionTicketKeys, use Server.Serve with a TLS Listener
	// instead.
	TLSConfig *[tls](https://pkg.go.dev/crypto/tls).[Config](https://pkg.go.dev/crypto/tls#Config)
```

So my understanding is that is only used when ListenAndServeTLS or ServeTLS are called, and not when ListenAndServe or Serve are called.

machi1990: That make sense. Thanks for checking.

miguelsorianod:
ping @dhirajsb @lburgazzoli could you review this and tell us if we can merge this change? this will potentially impact connectors.

Thanks

lburgazzoli:
@dhirajsb @valdar can you please have a look ?

dhirajsb:
I used this kfm PR to build a cfm binary. It throws errors for protocol version less that 1.2 as expected. :+1:

dhirajsb (approved):

dhirajsb:
I do have a follow up question. Does the TLS version need to be exposed as a config option in the future?

miguelsorianod:
If a need arises for it we could expose it, yes

machi1990 (approved):

PR #1006, opened 2022-05-05
Add new kafkafields largerkafka functionality

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-8385.

Adds the following fields in the response returned by the getInstanceTypesByCloudProviderAndRegion endpoint of the instance types api. Specifically, on the instance size part of the response.
* `max_message_size`. an object that contains a `bytes` attribute. The format is in bytes. This follows the same format as other fields available in the response like `ingress_throughput_per_sec`. The data type for the `bytes` attribute is an int64
* `supported_az_modes`. a list that can contain `single` or `multi` (or both) as its values returned
* `display_name`. string.
* `min_in_sync_replicas`. int32.
* `replication_factor`. int32.
* `lifespan_seconds`. int32. nullable. It is not returned if null or not set

Adds the following fields in the `config/kafka-instance-types-configuration.yaml` file:
* `maxMessageSize`. string. The format is a K8s Quantity
* `supportedAZModes`. list of string. The values of the list can be `single`, `multi`(or both).
* `displayName`. string.
* `minInSyncReplicas`. int.
* `replicationFactor`. int.
* `lifespanSeconds`. *int. If not set or set to nil the instance type is considered to not have an expiration time

Additionally, to fix an accuracy issue the data type of the following already existing fields has been changed from float32 to int64:
* `ingress_throughput_per_sec`
* `egress_throughput_per_sec`
* `max_data_retention_size`

## Verification Steps

* Verify unit tests and integration tests pass

* Run kas fleet manager and perform a request to the /api/kafkas_mgmt/v1/instance_types/aws/us-east-1 endpoint. Verify that the following new fields appear and they appear with the values configured in the config file:
```
min_in_sync_replicas
replication_factor
supported_az_nodes
max_message_size (an object with a bytes attribute)
display_name (on the kafka size object)
lifespan_seconds. This attribute should only appear for developer instance type.
```

* Verify that the values in bytes reported for the following fields are accurate and there are no rounding errors compared to what's been specified in the config file:
```
max_message_size
ingress_throughput_per_sec
egress_throughput_per_sec
max_data_retention_size
```
* Verify that the newly configured attributes in config/kafka-instance-types-configuration.yaml match the kafka instance limits defined in the spreadsheet

- [ ] Code Review completed


miguelsorianod:
cc @wtrocki , @JameelB  too.

Let me know if the current PR is ok to you, including the formats units etc....
If we deem it correct then I will do verifications and see if some test would be useful for this changes.

Thanks

miguelsorianod:
cc @wtrocki , @JameelB I added the new requested fields to the original PR and I also included the fix regarding accuracy issues that we detected for some fields returned in the instance type endpoints.

Let me know if the current PR is ok to you, including the formats units etc....
If we deem it correct then I will do verifications and see if some test would be useful for this changes.

Thanks

openapi/kas-fleet-manager.yaml
```diff
@@ -1016,6 +1016,15 @@ components:
               type: string
             max_connection_attempts_per_sec:
               type: integer
+            max_message_size:
+              description: "Maximum message size"
+              type: string
+            min_in_sync_replicas:
+              description: "Minimum number of in-sync replicas"
+              type: integer
+            replication_factor:
```
JameelB, line 1025: can we add these new fields to the [KafkaRequestExample](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/dd5c689c8372d5e155cefcfeb47b53e7477d8028/openapi/kas-fleet-manager.yaml#L1656)?

openapi/kas-fleet-manager.yaml
```diff
@@ -1114,6 +1123,9 @@ components:
         id:
           description: 'Unique identifier of this Kafka instance size.'
           type: string
+        display_name:
```
JameelB, line 1126: Can we add these new fields to the [SupportedKafkaInstanceTypeListExample](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/dd5c689c8372d5e155cefcfeb47b53e7477d8028/openapi/kas-fleet-manager.yaml#L1698)?

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -49,16 +49,21 @@ func (kp *KafkaInstanceType) validate() error {
...
+	MaxMessageSize              Quantity `yaml:"maxMessageSize"`
 	QuotaConsumed               int      `yaml:"quotaConsumed"`
 	QuotaType                   string   `yaml:"quotaType"`
 	CapacityConsumed            int      `yaml:"capacityConsumed"`
+	SupportedAZModes            []string `yaml:"supportedAZModes"`
+	MinInSyncReplicas           int      `yaml:"minInSyncReplicas"` // also abbreviated as ISR in Kafka terminology
+	ReplicationFactor           int      `yaml:"replicationFactor"` // also abbreviated as RF in Kafka terminology
 }
 
 // validates Kafka instance size configuration to ensure the following:
```
JameelB, line 69: Can we add some validation for the new fields? 

- `DisplayName` shouldn't be an empty string.
- `MaxMessageSize` should be a valid K8s quantity. 
- `SupportedAZModes` should have at least 1 length, optionally we could even validate that the list only includes valid input ("multi" or "single"). 
- `MinInSyncReplicas` and `ReplicationFactor` should be greater than 0.

JameelB: ah just saw your comment there about verification :) Can ignore this!

The new field names/format looks good to me. Will wait on @wtrocki to confirm.

JameelB:
The new field names/format looks good to me @miguelsorianod. Will wait on @wtrocki on his end.

wtrocki:
All good! Amazing work

wtrocki:
FYI @riccardo-forina - It would require SDK update once merged.

config/kafka-instance-types-configuration.yaml
```diff
@@ -8,13 +8,18 @@
...
 #   - egressThroughputPerSec: Maximum limit for egress throughput per second (MB/s).
 #   - totalMaxConnections: Maximum client connections allowed.
 #   - maxConnectionAttemptsPerSec: Maximum client connection attempts per second.
 #   - maxDataRetentionSize: Maximum storage (GB).
 #   - maxDataRetentionPeriod: Maximum duration(ISO8601) for retaining data.
 #   - maxPartitions: Maximum partitions for all topics combined.
+#   - maxMessageSize: Maximum message size (MB)
+#   - minInSyncReplicas: Minimum number of in sync replicas
+#   - replicationFactor: Replication factor
+#   - supportedAZModes: a list of the supported AZ modes. The possible values are "single", "multi"
```
machi1990, line 22: Let's also add these parameters in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/212e5710729dd534b87f7fc98da0aab85851a03b/templates/service-template.yml#L163

and if possible and applicable create an issue/PR in kas-installer or just let people know of the change in parameters or addition of new parameters. 
/cc @MikeEdgar @agullon @daveconde FYI, once these PR is in, there will be extra parameters that will need to be configured for the `SUPPORTED_INSTANCE_TYPES` parameter

machi1990 (changes requested):
let's also update the default values for `SUPPORTED_INSTANCE_TYPES` in service-template.yml and possibly keep kas-installer or its users in Sync with this change

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -2035,6 +2053,15 @@ components:
           type: string
         max_connection_attempts_per_sec:
           type: integer
+        max_message_size:
+          description: Maximum message size
+          type: string
```
riccardo-forina, line 2058: I find it a bit confusing to have a K8s size string here and the bytes object in the other API returning this. Can't we have the same bytes object here?

miguelsorianod:
> I left a comment about max_message_size being exposed in two different formats, where I think we should select one and stick with it everywhere. For the rest, it LGTM

We are aware of this. This is intentional for now. Other fields also have the same inconsistency so we decided to be consistent with the inconsistency in here. We opened a jira issue to evaluate changes for all those fields including this one in https://issues.redhat.com/browse/MGDSTRM-8442. That will be dealt with in the future as this will imply deprecations/removals at API level

internal/kafka/internal/api/public/api/openapi.yaml
miguelsorianod: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1006#issuecomment-1121186214 comment

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -92,10 +98,26 @@ func (k *KafkaInstanceSize) validate(instanceTypeId string) error {
...
+			return fmt.Errorf("value '%s' in supportedAZModes for Kafka instance type '%s', size '%s' is invalid", supportedAZMode, k.Id, instanceTypeId)
+		}
+	}
+
 	if maxDataRetentionPeriod.IsZero() || egressThroughputQuantity.CmpInt64(1) < 0 ||
 		ingressThroughputQuantity.CmpInt64(1) < 0 || maxDataRetentionSize.CmpInt64(1) < 0 ||
 		k.TotalMaxConnections <= 0 || k.MaxPartitions <= 0 || k.MaxConnectionAttemptsPerSec <= 0 ||
-		k.QuotaConsumed < 1 || k.CapacityConsumed < 1 {
+		k.QuotaConsumed < 1 || k.CapacityConsumed < 1 || k.MinInSyncReplicas < 1 ||
+		k.ReplicationFactor < 1 || maxMessageSize.CmpInt64(0) < 0 || len(k.SupportedAZModes) == 0 {
```
JameelB, line 120: I think maxMessageSize should be larger than 0. wdyt of changing this check to the following?

```suggestion
		k.ReplicationFactor < 1 || maxMessageSize.CmpInt64(1) < 0 || len(k.SupportedAZModes) == 0 {
```

miguelsorianod: I set it at 0 due to theoretically the field accepts 0 as a value: https://kafka.apache.org/documentation/#brokerconfigs_message.max.bytes

JameelB: ah didn't see that, thanks for confirming!

config/kafka-instance-types-configuration.yaml
miguelsorianod: Parameters updated

openapi/kas-fleet-manager.yaml
miguelsorianod: Added

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: Added validations

openapi/kas-fleet-manager.yaml
```diff
@@ -1675,6 +1706,7 @@ components:
           [
             {
               id: x1,
+              display_name: " 1",
```
machi1990, line 1709:
```suggestion
              display_name: "1",
```
trailing space?

openapi/kas-fleet-manager.yaml
```diff
@@ -1138,6 +1150,21 @@ components:
...
+        min_in_sync_replicas:
+          description: "Minimum number of in-sync replicas."
+          type: integer
+        replication_factor:
+          description: "Replication factor available to this Kafka instance size."
+          type: integer
+        supported_az_modes:
+          description: 'List of Availability Zone modes that this Kafka instance size supports. The possible values are "single", "multi".'
+          type: array
+          items:
```
machi1990, line 1166: PS: The question below might not be applicable as by looking at their description and the validation (corrent??) these values are never going to have a value asked in a question. But thought I could still ask in case of future changes. 

What happens if some of these are empty i.e if an `integer` is `0` and the array is an empty array? 
Is the API expected to send the equivalent values or the desired behaviour is that they'll be omitted from the returned response?

config/kafka-instance-types-configuration.yaml
```diff
@@ -24,52 +29,76 @@ supported_instance_types:
...
       maxDataRetentionPeriod: "P14D"
       maxPartitions: 1000
+      maxMessageSize: "1Mi"
+      minInSyncReplicas: 2
+      replicationFactor: 3
       quotaConsumed: 1
       quotaType: "RHOSAKTrial"
       capacityConsumed: 1
+      supportedAZModes:
+      - multi
```
machi1990, line 104:
```suggestion
      - multi

```

config/kafka-instance-types-configuration.yaml
machi1990: Thanks. There are some people / pipeline that have this deployment parameter overridden. 
When this PR is merged, let's give a heads up in kas-installer chat with the fields to add to avoid surprise failures in their setup.

config/kafka-instance-types-configuration.yaml
miguelsorianod: fixed :+1:

openapi/kas-fleet-manager.yaml
miguelsorianod: This fields are always returned by a GET/LIST endpoint so the user will never set them. The described scenario should never happen. As you mentioned we have validation to ensure that doesn't happen when reading from the configuration file.
In any case, if for some reason they end up with the zero-value whether they will be omitted from the returned response will depend on the marshaling behavior specified in OpenAPI. For example for supported_az_modes if the value of it is `nil` then it would be ommited from the response as the generated data type from OpenAPI is:
```
	SupportedAzModes []string `json:"supported_az_modes,omitempty"`
```
(notice the omitempty).
But as commented, this should never happen in this specific fields (if it does is a bug)

machi1990: Perfect, thanks.

internal/kafka/internal/api/public/model_supported_kafka_size.go
```diff
@@ -23,7 +25,14 @@ type SupportedKafkaSize struct {
...
 	// Maximium connection attempts per second available to this Kafka instance size.
-	MaxConnectionAttemptsPerSec int32 `json:"max_connection_attempts_per_sec,omitempty"`
+	MaxConnectionAttemptsPerSec int32                            `json:"max_connection_attempts_per_sec,omitempty"`
+	MaxMessageSize              SupportedKafkaSizeBytesValueItem `json:"max_message_size,omitempty"`
+	// Minimum number of in-sync replicas.
+	MinInSyncReplicas int32 `json:"min_in_sync_replicas,omitempty"`
+	// Replication factor available to this Kafka instance size.
+	ReplicationFactor int32 `json:"replication_factor,omitempty"`
+	// List of Availability Zone modes that this Kafka instance size supports. The possible values are \"single\", \"multi\".
+	SupportedAzModes []string `json:"supported_az_modes,omitempty"`
```
ziccardi, line 35: run `make code fix` so that the code gets formatted

ziccardi: Hmmm... The field types should be aligned (see DisplayName, IngressThroughputPerSec, etc).
I believe there is a bug/limitation of the formatter when there is a comment between two fields.

openapi/kas-fleet-manager.yaml
JameelB: thanks!

openapi/kas-fleet-manager.yaml
JameelB: thanks!

JameelB (approved):
Looks good to me! Verified the following:
* Limits in the config file matches the values in the service limits spreadsheet
* Limits returned in 'Bytes' format now has the correct values
* `max_message_size`, `min_in_sync_replicas` and `replication_factor` are returned in the response of the GET, LIST, PATCH and POST /kafkas endpoints. (see example response [here](https://gist.github.com/JameelB/c4c5017324502269c846b625f186127f#file-kafka_list-json))
* `display_name`, `max_message_size`, `min_in_sync_replicas`, `replication_factor` and `supported_az_modes` are now returned in the GET /instance_types endpoint. (see example response [here](https://gist.github.com/JameelB/c4c5017324502269c846b625f186127f#file-supported_instance_types-json))

Will wait on @wtrocki or @riccardo-forina for final approval

miguelsorianod:
A new field `lifespan_seconds` has been added:
* Added it as lifespanSeconds in the `config/kafka-instance-types-configuration.yaml` configuration file. Its data type is *int. If not set or set as nil it will be interpreted as no expiration for that type
* Added it as `lifespan_seconds in the instance types endpoint. It is a nullable int32. If not specified or null then it won't be returned in the response at all
* It has not been added in the kafka get/list endpoints due to there's a pending task where a potential redesign of returned kafka schema for get/list endpoints will be rethought

JameelB (commented):
New changes for `lifespan_seconds` looks good to me. 

Verified that it returns in the response of the /instance_types/{cloud_provider}/{cloud_region} endpoint (see updated response [here](https://gist.github.com/JameelB/c4c5017324502269c846b625f186127f#file-supported_instance_types-json)).

pb82:
@riccardo-forina @rareddy can you please check Jameels link on how the responses now look: https://gist.github.com/JameelB/c4c5017324502269c846b625f186127f#file-supported_instance_types-json

Can you confirn that this is what you expect?

riccardo-forina:
> @riccardo-forina @rareddy can you please check Jameels link on how the responses now look: https://gist.github.com/JameelB/c4c5017324502269c846b625f186127f#file-supported_instance_types-json
> 
> Can you confirn that this is what you expect?

For the `KafkaRequestList` response:
* I don't see the `display_name` value that matches the `size_id` that has been added.
* the example is about a `standard` instance. Will a `developer` instance have the `lifespan_seconds` and/or `expire_at` fields we talked about?

The supported instance types example looks ok to me.

miguelsorianod:
> I don't see the display_name value that matches the size_id that has been added.
On Kafka Get/List the display_name of the size is not added. See original comment. It has only been added into the instance types endpoint.

> the example is about a standard instance. Will a developer instance have the lifespan_seconds and/or expire_at fields we talked about?

For the instance types endpoint:

developer instances do not have time limit so the value is not set for them. See original comment:
```
lifespanSeconds. *int. If not set or set to nil the instance type is considered to not have an expiration time
...
lifespan_seconds. int32. nullable. It is not returned if null or not set
```

For the get/list endpoint:

An expires_at field will be added. This will be dealt with in another PR. The behavior will be the same: If not set or set to nil (the case of developer instances) the instance is considered to not have an expiration time

pb82:
> I don't see the display_name value that matches the size_id that has been added.

@riccardo-forina Is this needed? I would argue that information about instance types should come from the instance type endpoint. If you want to display that information for a given instance, you can also match it against the response from that endpoint.

I'm hesitant about including those fields in the Kafka crudl endpoints and in general about refactoring the response from that endpoints. This is quite complex in terms of deprecation and backwards compatibility for what seems to be only a small benefit (one additional request saved) for the UI?

Can we agree on limiting this change to only the instance type endpoint?

riccardo-forina:
We need to show the size of the instance in the drawer panel, together with all the limits and the eventual expiry date. You can see a [demo for standard instances here](https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-kafka-instance-drawer-kafka-details-tab--standard-instance-created) and one for [trial/developer instances here](https://redhat-developer.github.io/app-services-ui-components/?path=/story/kafka-kafka-instance-drawer-kafka-details-tab--trial-instance-just-created).

This drawer is fed with the data coming from the KafkaRequest object. Having to call another API endpoint in a waterfall fashion (since we need to know the provider and region and size id to call the sizes API) just to get the display name for the size feels like a big loss in terms of UX, since we need to convert something that's synchronous to asynchronous, with the handling of the loading states and error states.

The same is valid for the expiry date.

If we can't get the display name in the KafkaRequest object then sure, we'll change the UI to handle this asynchronicity, but then I suppose we could use the limits coming from the size API as well, making the changes done to enrich the KafkaRequest a bit of a wasted effort.

miguelsorianod:
I updated the PR to undo the changes to the kafka request data structure and its get/list endpoints.

cc @riccardo-forina @JameelB please review and approve if it looks ok so we can merge it.

Thank you.

JameelB (approved):

riccardo-forina (approved):

PR #1007, opened 2022-05-06
feat: Add SelectSSOProvider flag/Add rh clientSecret/id

JasonPowr:
## Description
Adding a flag in order to be able to select which so provider to choose from i.e `redhat_sso` or `mas_sso`. Also adding the env vars for ClientIDFile and ClientSecretFile to the service template as well.

Jira: [MGDSTRM-8210](https://issues.redhat.com/browse/MGDSTRM-8210)

## Verification Steps
1. Create cluster.
2. Deploy kas-fleet-manager to cluster. 
3. Run the command `make deploy/service IMAGE_TAG=<your-image-tag-here> SSO_PROVIDER_TYPE=<provider>` when deploying kas-fleet-manager to the cluster
4. Use this endpoint  [here](https://api.openshift.com/?urls.primaryName=kafka%20service%20fleet%20manager%20service#/security/getSsoProviders) to test and ensure the provider is being changed.

Supporting docs : [Here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/docs/deploying-kas-fleet-manager-to-openshift.md)

- [ ] Code Review completed


JasonPowr:
Few questions about this.

1. Is it better to set mas_sso as provider or error out when the a incorrect provider is given?
2. If error out is there a better way than using panic()?
3. Is it ok to change SelectSSOProvider to a string as it was causing issues when adding the flag otherwise?

Makefile
```diff
@@ -599,6 +599,8 @@ keycloak/setup:
 	@echo -n "$(MAS_SSO_CLIENT_SECRET)" > secrets/keycloak-service.clientSecret
 	@echo -n "$(OSD_IDP_MAS_SSO_CLIENT_ID)" > secrets/osd-idp-keycloak-service.clientId
 	@echo -n "$(OSD_IDP_MAS_SSO_CLIENT_SECRET)" > secrets/osd-idp-keycloak-service.clientSecret
+	@echo -n "$(REDHAT_SSO_CLIENT_ID)" > secrets/redhatsso-service.clientId
+	@echo -n "$(REDHAT_SSO_CLIENT_SECRET)" > secrets/redhatsso-service.clientSecret
```
machi1990, line 603: How does this change relates to the `redhatsso/setup` target down below?
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8df317f8d17c5618e651300e444dc042b0e6d3c6/Makefile#L606-L609

machi1990: Do we want to keep one or the other?

internal/kafka/internal/services/util.go
```diff
@@ -70,8 +69,8 @@ func BuildKeycloakClientNameIdentifier(kafkaRequestID string) string {
 	return fmt.Sprintf("%s-%s", "kafka", strings.ToLower(kafkaRequestID))
 }
 
-func BuildCustomClaimCheck(kafkaRequest *dbapi.KafkaRequest, ssoconfigProvider keycloak.SSOProvider) string {
-	if ssoconfigProvider == keycloak.REDHAT_SSO {
+func BuildCustomClaimCheck(kafkaRequest *dbapi.KafkaRequest, ssoconfigProvider string) string {
+	if ssoconfigProvider == "redhat_sso" {
```
machi1990, line 73: Why this change?

pkg/client/keycloak/config.go
```diff
@@ -9,12 +9,8 @@ import (
 	"github.com/spf13/pflag"
 )
 
-type SSOProvider string
-
 const (
-	MAS_SSO                       SSOProvider = "mas_sso"
-	REDHAT_SSO                    SSOProvider = "redhat_sso"
```
machi1990, line 16: Why this change?

pkg/client/keycloak/config.go
```diff
@@ -130,6 +132,7 @@ func (kc *KeycloakConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&kc.SsoBaseUrl, "redhat-sso-base-url", kc.SsoBaseUrl, "The base URL of the mas-sso, integration by default")
 	fs.StringVar(&kc.SSOSpecialManagementOrgID, "sso-special-management-org-id", SSO_SPEICAL_MGMT_ORG_ID_STAGE, "The Special Management Organization ID used for creating internal Service accounts")
 	fs.StringVar(&kc.ServiceAccounttLimitCheckSkipOrgIdListFile, "service-account-limits-check-skip-org-id-list-file", kc.ServiceAccounttLimitCheckSkipOrgIdListFile, "File containing a list of Org IDs for which service account limits check will be skipped")
+	fs.StringVar(&kc.SelectSSOProvider, "select-sso-provider", kc.SelectSSOProvider, "Option to choose between sso providers i.e, mas-sso or redhat_sso, mas-sso by default")
```
machi1990, line 135:
```suggestion
	fs.StringVar(&kc.SelectSSOProvider, "sso-provider-type", kc.SelectSSOProvider, "Option to choose between sso providers i.e, mas-sso or redhat_sso, mas-sso by default")
```
`sso-provider-type` maybe?

internal/kafka/internal/services/util.go
JasonPowr: Hey @machi1990 this change is in relation to question 3 here. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1007#issuecomment-1119402717

pkg/client/keycloak/config.go
```diff
@@ -184,6 +187,7 @@ func (kc *KeycloakConfig) ReadFiles() error {
 		}
 	}
 
+	validateSSOProviderRealm(kc.SelectSSOProvider)
```
machi1990, line 190: I think we can have a separate `.Validate` method on the ConfigModule.
This way, it'll be validated automatically by the config module setup.

Something like this: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8df317f8d17c5618e651300e444dc042b0e6d3c6/internal/kafka/internal/config/kafka.go#L79-L81

This way, the `.ReadFiles()` does not have two responsibilities
1. to read a file
2. perform validation

pkg/client/keycloak/config.go
JasonPowr: Same as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1007#discussion_r866635039

templates/service-template.yml
```diff
@@ -372,7 +372,11 @@ parameters:
   displayName: MAS SSO REALM for OSD Cluster Identity Provider Clients
   description: MAS SSO realm for OSD Cluster Identity Provider Clients
   value: "interim-test-realm"
-
+  
+- name: SELECT_SSO_PROVIDER
```
machi1990, line 376:
```suggestion
- name: SSO_PROVIDER_TYPE
```

`SSO_PROVIDER_TYPE` maybe?

JasonPowr:
CC: @VanillaSpoon

machi1990:
> Few questions about this.
> 
> 1. Is it better to set mas_sso as provider or error out when the a incorrect provider is given?

I'd say an error is okay. If someone selected the wrong provider type its quite possible that even the credentials they have put is wrong. So let's send an error by performing a proper validation


> 2. If error out is there a better way than using panic()?

We should never return panic but a proper error. See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1007#discussion_r866635329 for a suggested validation path

> 3. Is it ok to change SelectSSOProvider to a string as it was causing issues when adding the flag otherwise?

Yes it is okay but let's keep the constants to avoid cascading changes everywhere.
It's okay to change it to a string to be able to add a flag and have it properly initialized. 
But let's use the predefined constants in places where we are referencing the values instead of hardcoding them each time.

JasonPowr:
> 

Thanks for the answering my questions @machi1990 Ill go ahead and make the changes you suggested aswell :)

pkg/client/keycloak/config.go
```diff
@@ -1,16 +1,22 @@
...
 
 	"github.com/golang/glog"
 	"github.com/spf13/pflag"
 )
 
+type SSOProvider string
+
 const (
-	SSO_SPEICAL_MGMT_ORG_ID_STAGE string = "13640203"
+	MAS_SSO                       SSOProvider = "mas_sso"
```
machi1990, line 17: you can make this a string too.

That may avoid the need to `string(keycloak.MAS_SSO) `
```suggestion
	MAS_SSO                       string = "mas_sso"
```

pkg/client/keycloak/config.go
```diff
@@ -153,7 +160,7 @@ func (kc *KeycloakConfig) ReadFiles() error {
 	if err != nil {
 		return err
 	}
-	if kc.SelectSSOProvider == REDHAT_SSO {
+	if kc.SelectSSOProvider == "redhat_sso" {
```
machi1990, line 163:
```suggestion
	if kc.SelectSSOProvider == REDHAT_SSO {
```

akoserwal:
A followup PR for app-interface would also be required to add sso provider flag for stage/prod

internal/kafka/internal/services/util_test.go
```diff
@@ -425,7 +426,7 @@ func Test_BuildCustomClaimCheck(t *testing.T) {
 	RegisterTestingT(t)
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			receivedBuiltCustomClaimCheck := BuildCustomClaimCheck(tt.args.kafkaRequest, tt.args.ssoconfigProvider)
+			receivedBuiltCustomClaimCheck := BuildCustomClaimCheck(tt.args.kafkaRequest, string(tt.args.ssoconfigProvider))
```
dhirajsb, line 429: Is the type cast needed if the type is also being changed to string on line 401?

dhirajsb (changes requested):

internal/kafka/internal/services/util.go
```diff
@@ -70,7 +70,7 @@ func BuildKeycloakClientNameIdentifier(kafkaRequestID string) string {
 	return fmt.Sprintf("%s-%s", "kafka", strings.ToLower(kafkaRequestID))
 }
 
-func BuildCustomClaimCheck(kafkaRequest *dbapi.KafkaRequest, ssoconfigProvider keycloak.SSOProvider) string {
+func BuildCustomClaimCheck(kafkaRequest *dbapi.KafkaRequest, ssoconfigProvider string) string {
 	if ssoconfigProvider == keycloak.REDHAT_SSO {
 		return fmt.Sprintf("@.rh-org-id == '%s'|| @.org_id == '%s' || @.clientId == '%s'", kafkaRequest.OrganisationId, kafkaRequest.OrganisationId, kafkaRequest.CanaryServiceAccountClientID)
 	} else {
```
ziccardi, line 76: If we keep `ssoconfigProvider` as object, this function could become a method of `SSOProvider`.

This way, code that needs the customClaim will be able to do something like this:
```go
customClaim := ssoconfigProvider.BuildCustomClaimCheck(kafkaRequest)
```

I'd prefer that approach, because that way we avoid the spreading of `if provider == xxx`

pkg/client/keycloak/config.go
```diff
@@ -130,6 +129,14 @@ func (kc *KeycloakConfig) AddFlags(fs *pflag.FlagSet) {
...
 	fs.StringVar(&kc.SSOSpecialManagementOrgID, "sso-special-management-org-id", SSO_SPEICAL_MGMT_ORG_ID_STAGE, "The Special Management Organization ID used for creating internal Service accounts")
 	fs.StringVar(&kc.ServiceAccounttLimitCheckSkipOrgIdListFile, "service-account-limits-check-skip-org-id-list-file", kc.ServiceAccounttLimitCheckSkipOrgIdListFile, "File containing a list of Org IDs for which service account limits check will be skipped")
+	fs.StringVar(&kc.SelectSSOProvider, "sso-provider-type", kc.SelectSSOProvider, "Option to choose between sso providers i.e, mas-sso or redhat_sso, mas-sso by default")
+}
+
+func (kc *KeycloakConfig) Validate(env *environments.Env) error {
+	if kc.SelectSSOProvider != REDHAT_SSO && kc.SelectSSOProvider != MAS_SSO {
+		return fmt.Errorf("Invalid sso provider selcted must be `mas-sso` or `redhat_sso`")
+	}
+	return nil
```
ziccardi, line 139: As in my previous comment: if we keep `selectSSOProvider` as object, we can add a `Validate` method to that object. 
That way if in the future the list of supported provider should change, the client code won't need any update.

docs/deploying-kas-fleet-manager-to-openshift.md
```diff
@@ -179,6 +179,7 @@ make deploy/service IMAGE_TAG=<your-image-tag-here> <OPTIONAL_PARAMETERS>
...
 - `KAS_FLEETSHARD_OPERATOR_SUBSCRIPTION_CONFIG`: Kas-fleetshard operator subscription config. This is applied for standalone clusters only. The configuration must be of type [SubscriptionConfig](https://pkg.go.dev/github.com/operator-framework/api@v0.3.25/pkg/operators/v1alpha1?utm_source=gopls#SubscriptionConfig). Defaults to an empty object i.e `{}`. See the [config/kas-fleetshard-operator-subscription-spec-config.yaml](../config/kas-fleetshard-operator-subscription-spec-config.yaml) file for example values.
 - `STRIMZI_OPERATOR_SUBSCRIPTION_CONFIG`: Strimzi operator subscription config. This is applied for standalone clusters only. The configuration must be of type [SubscriptionConfig](https://pkg.go.dev/github.com/operator-framework/api@v0.3.25/pkg/operators/v1alpha1?utm_source=gopls#SubscriptionConfig). Defaults to an empty object i.e `{}`. See the [config/strimzi-operator-subscription-spec-config.yaml](../config/strimzi-operator-subscription-spec-config.yaml) file for example values. 
+- `SSO_PROVIDER_TYPE`: Option to choose between sso providers i.e, mas-sso or redhat_sso, mas-sso by default.
```
akoserwal, line 182: `mas_sso` instead of `mas-sso`

internal/kafka/internal/services/util.go
JasonPowr: I think this also applies to my comment above. If I am wrong could you clarify please? https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1007#discussion_r873779876

pkg/client/keycloak/config.go
JasonPowr: I think this also applies to my comment above. If I am wrong could you clarify please? https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1007#discussion_r873779876

MikeEdgar:
I don't see where the RH SSO clientId/secret are used. Are they meant to be referenced in a future PR?

internal/kafka/internal/services/util_test.go
JasonPowr: @dhirajsb Could you give this another review please when you have the time.

JameelB (commented):
Look

pkg/client/keycloak/config.go
```diff
@@ -128,6 +129,14 @@ func (kc *KeycloakConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&kc.SsoBaseUrl, "redhat-sso-base-url", kc.SsoBaseUrl, "The base URL of the mas-sso, integration by default")
 	fs.StringVar(&kc.SSOSpecialManagementOrgID, "sso-special-management-org-id", SSO_SPEICAL_MGMT_ORG_ID_STAGE, "The Special Management Organization ID used for creating internal Service accounts")
 	fs.StringVar(&kc.ServiceAccounttLimitCheckSkipOrgIdListFile, "service-account-limits-check-skip-org-id-list-file", kc.ServiceAccounttLimitCheckSkipOrgIdListFile, "File containing a list of Org IDs for which service account limits check will be skipped")
+	fs.StringVar(&kc.SelectSSOProvider, "sso-provider-type", kc.SelectSSOProvider, "Option to choose between sso providers i.e, mas-sso or redhat_sso, mas-sso by default")
+}
+
+func (kc *KeycloakConfig) Validate(env *environments.Env) error {
+	if kc.SelectSSOProvider != REDHAT_SSO && kc.SelectSSOProvider != MAS_SSO {
+		return fmt.Errorf("Invalid sso provider selcted must be `mas-sso` or `redhat_sso`")
```
JameelB, line 137:
```suggestion
		return fmt.Errorf("Invalid sso provider selected must be `mas_sso` or `redhat_sso`")
```

JameelB (approved):
Looks good! Also verified locally, working as expected :+1:

internal/kafka/internal/services/util_test.go
machi1990: Yes in my opinion as the scope of this PR has been reduced significantly following other earlier changes around this.

JameelB:
yep go ahead, i've verified and approved it already :slightly_smiling_face:

PR #1009, opened 2022-05-06
refactor: reposition Expect(<values>) to below errors, in providers_test.go

VanillaSpoon:
If an unexpected error occurs, got will be nil and the Expect(got).To... will fail, halting the test execution.
That means that the Expect(err != nil) won't be executed, resulting in confusing errors (error will fail for unmatched got instead than unexpected error).


VanillaSpoon:
Thanks @ziccardi for the explanation

ziccardi (approved):

PR #1011, opened 2022-05-09
test: added unit tests for pkg/logger

pawelpaszki:
## Description
test: added unit tests for pkg/logger

- [ ] Code Review completed


pkg/logger/operationid_middleware_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}{
+		{
+			name: "should return OpID value when its set in the context",
+			args: args{
+				ctx: context.Background(),
+			},
+			want: "",
+		},
+		{
+			name: "should return an empty string if OpID value is not set in the context",
```
vmanley, line 71: Should name of this case be swapped with name of test case in L64 above?

pawelpaszki: yes. I have swapped them. good catch

pkg/logger/log_events.go
```diff
@@ -0,0 +1,52 @@
+package logger
```
machi1990, line 1: Why is this a separate file?

pawelpaszki: because it doesn't make sense to write tests for the content of the file (it is ignored in the makefile)

machi1990: Why not just ignore the original file `logger.go`?

pawelpaszki: because some of the logger context needs to be tested (see the logger_test.go). The rest doesn't return any value and doesn't make sense to test it

machi1990: Yes, but we can still have the test and ignore the file for code coverage.  

I saying that because I am not convinced by the split, as the design/code separation is then driven by if something is testable or not to be added to the code coverage.

pawelpaszki: ok. I can change it

pawelpaszki: @machi1990 - done

machi1990: Thanks @pawelpaszki

PR #1014, opened 2022-05-10
MGDCTRS-927: add an endpoint to list connector type and related shard metadata

lburgazzoli:

- [ ] Code Review completed


internal/connector/internal/services/connector_types.go
```diff
@@ -350,3 +352,30 @@ func (cts *connectorTypesService) DeleteUnusedAndNotInCatalog() *errors.ServiceE
...
+		end = pagingMeta.Total
+	}
+
+	if pagingMeta.Size > pagingMeta.Total {
+		pagingMeta.Size = pagingMeta.Total
+	}
+
+	// TODO it may be better to copy the entries to avoid potential changes outside the control of the service
+	// TODO it may be better to load it from the db ?
+	return cts.connectorsConfig.CatalogEntries[begin:end], &pagingMeta, nil
```
lburgazzoli, line 380: @valdar @dhirajsb this is likely not to be the perfect solution but the it is mainly aimed to get information about the deployed catalogs hence it does not rely on the information in the database. 

Eventually we could add an additional endpoint to check for inconsistencies and/or include some stats in the metrics.

test/cucumber/http_response.go
```diff
@@ -53,7 +53,7 @@ func init() {
 		ctx.Step(`^the response should match "([^"]*)"$`, s.theResponseShouldMatchText)
 		ctx.Step(`^I store the "([^"]*)" selection from the response as \${([^"]*)}$`, s.iStoreTheSelectionFromTheResponseAs)
 		ctx.Step(`^I store json as \${([^"]*)}:$`, s.iStoreJsonAsInput)
-		ctx.Step(`^the "([^"]*)" selection from the response should match "([^"]*)"$`, s.theSelectionFromTheResponseShouldMatch)
+		ctx.Step(`^the "(.*)" selection from the response should match "([^"]*)"$`, s.theSelectionFromTheResponseShouldMatch)
```
lburgazzoli, line 56: This change is required to allow more complex JQ expression such as `.items[] | select(.id == "aws-sqs-source-v1alpha1") | has("schema")`  where double quotes are required by the JQ syntax but were excluded by the original regexp.

internal/connector/test/integration/features/connector-admin-api.feature
```diff
@@ -0,0 +1,50 @@
...
+    Then the response code should be 200
+    And the ".kind" selection from the response should match "ConnectorTypeAdminViewList"
+    And the ".page" selection from the response should match "1"
+    And the ".size" selection from the response should match "2"
+    And the ".total" selection from the response should match "2"
+    And the ".items | length" selection from the response should match "2"
+    And the ".items[] | select(.id == "log_sink_0.1") | has("schema")" selection from the response should match "true"
+    And the ".items[] | select(.id == "log_sink_0.1") | .channels.stable.shard_metadata.connector_image" selection from the response should match "quay.io/mcs_dev/log-sink:0.0.1"
+    And the ".items[] | select(.id == "aws-sqs-source-v1alpha1") | has("schema")" selection from the response should match "true"
+    And the ".items[] | select(.id == "aws-sqs-source-v1alpha1") | .channels.stable.shard_metadata.connector_image" selection from the response should match "quay.io/mock-image:77c0b8763729a9167ddfa19266d83a3512b7aa8124ca53e381d5d05f7d197a24"
```
lburgazzoli, line 28: The checks here do not test the full payload but validates that some of the constituents are part of the response.

valdar (approved):

internal/connector/internal/services/connector_types.go
dhirajsb: We should load it from the DB, as it's theoretically possible that there are used older types no longer in the catalog.

internal/connector/internal/api/admin/private/api/openapi.yaml
```diff
@@ -1621,6 +1674,22 @@ components:
       allOf:
       - $ref: '#/components/schemas/List'
       - $ref: '#/components/schemas/ConnectorDeploymentAdminViewList_allOf'
+    ConnectorTypeAdminViewList:
```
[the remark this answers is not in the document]
lburgazzoli: It is needed as the admin view also includes the channels and related shared meta that is not part of the public APIs AFAIK

internal/connector/internal/services/connector_types.go
lburgazzoli: I think we need to reason a little bit about that as if a connector type is not more in the catalog, then it should not be possible to create a connector from that connector_type hence, it should not be listed among the connector_types 

I don't think that as today we handle such case so theoretically one would be able to create connectors on something we have removed as long an instance of such connector exists which does not seems right. 

I think it could make sense to have an endpoint to get only the shard meta which is where it still make sense to keep older version around.

What do you think ?

dhirajsb:
@lburgazzoli since we do have a public API to list types, what's the motivation for duplicating that in the admin API? Shouldn't we restrict that to either bypass authz checks for visibility into all clusters and namespaces and to perform admin privileged actions?

internal/connector/internal/services/connector_types.go
dhirajsb: I agree about the edge case with obsolete connector types. That's the way we are handling the import right now, but perhaps we can delete it and mark older connectors as not having a spec anymore, so one could only delete such connectors. Can you create another issue for that?

lburgazzoli:
it is to get a view on the catalog including the shard meta that is not visible with the public api. 

Also the connector types/catalogs visibility may be restricted at some point for some users/orgs.

internal/connector/test/integration/features/connector-admin-api.feature
dhirajsb: ltgm

test/cucumber/http_response.go
dhirajsb: lgtm

internal/connector/internal/api/admin/private/api/openapi.yaml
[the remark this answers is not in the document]
dhirajsb: ok, I missed that part in the first read. lgtm

openapi/connector_mgmt-private-admin.yaml
```diff
@@ -879,6 +879,86 @@ paths:
...
+      tags:
+        - Connector Types
+      security:
+        - Bearer: [ ]
+      operationId: getConnectorTypes
+      summary: Returns a list of connector types
+      description: Returns a list of connector types
+      parameters:
+        - $ref: "connector_mgmt.yaml#/components/parameters/page"
+        - $ref: "connector_mgmt.yaml#/components/parameters/size"
```
dhirajsb, line 893: Can you also add `search` and `orderBy` parameters to make it uniform with all the other list operations? Maybe we can then also use it in `rhoc cli` (if it doesn't do it already) to help filter and order results.

dhirajsb (changes requested):

dhirajsb:
> it is to get a view on the catalog including the shard meta that is not visible with the public api.
> 
> Also the connector types/catalogs visibility may be restricted at some point for some users/orgs.
> 
> We can use gebi to get into the db but having access to the api is much more practical for scripting.
> 
> If you think it is not useful I have no problem to close the PR.

I think it's ok for the reasons you stated. Perhaps we could add the missing data back to the public API in the future once the model solidifies, and then we won't need this in the admin API.

lburgazzoli:
> > it is to get a view on the catalog including the shard meta that is not visible with the public api.
> > Also the connector types/catalogs visibility may be restricted at some point for some users/orgs.
> > We can use gebi to get into the db but having access to the api is much more practical for scripting.
> > If you think it is not useful I have no problem to close the PR.
> 
> I think it's ok for the reasons you stated. Perhaps we could add the missing data back to the public API in the future once the model solidifies, and then we won't need this in the admin API.

Since the shard_meta is an implementation detail, I don't think it should be leaked out in the public APIs

openapi/connector_mgmt-private-admin.yaml
lburgazzoli: fixed

test/cucumber/http_response.go
lburgazzoli: @pb82 @machi1990 @ziccardi is that change ok with you ?

machi1990: Yes. It is okay with us. The kafka side does not use cucumber at the moment...

machi1990 (approved):
reviewed the change in the common folders and it looks good to me

PR #1021, opened 2022-05-13
test: added unit tests for pkg/keycloak

pawelpaszki:
## Description
test: added unit tests for pkg/keycloak

- [ ] Code Review completed

### DIFF
--- a/pkg/client/keycloak/client.go
+++ b/pkg/client/keycloak/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/Nerzal/gocloak/v11"
@@ -16,6 +17,13 @@ const (
 	// gocloak access token duration before expiration
 	tokenLifeDuration    = 5 * time.Minute
 	cacheCleanupInterval = 5 * time.Minute
+	OrgKey               = "rh-org-id"
+	UserKey              = "rh-user-id"
+)
+
+var (
+	protocol = "openid-connect"
+	mapper   = "oidc-usermodel-attribute-mapper"
 )
 
 //go:generate moq -out client_moq.go . KcClient
@@ -84,7 +92,6 @@ func NewClient(config *KeycloakConfig, realmConfig *KeycloakRealmConfig) *kcClie
 }
 
 func (kc *kcClient) ClientConfig(client ClientRepresentation) gocloak.Client {
-	protocol := "openid-connect"
 	publicClient := false
 	directAccess := false
 	return gocloak.Client{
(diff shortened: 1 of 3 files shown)


pkg/client/keycloak/client.go
```diff
@@ -192,15 +197,11 @@ func (kc *kcClient) GetClientSecret(internalClientId string, accessToken string)
...
-	return value, err
+	return *resp.Value, err
 }
 
 func (kc *kcClient) DeleteClient(internalClientID string, accessToken string) error {
 	err := kc.kcClient.DeleteClient(kc.ctx, accessToken, kc.realmConfig.Realm, internalClientID)
-	if err != nil {
-		return err
-	}
 	return err
```
ziccardi, line 205: return kc.kcClient.DeleteClient(kc.ctx, accessToken, kc.realmConfig.Realm, internalClientID)

pkg/client/keycloak/client.go
```diff
@@ -291,20 +292,18 @@ func (kc *kcClient) IsSameOrg(client *gocloak.Client, orgId string) bool {
...
 func (kc *kcClient) IsOwner(client *gocloak.Client, userId string) bool {
 	if userId == "" {
 		return false
 	}
 	attributes := *client.Attributes
-	if rhUserId, found := attributes["rh-user-id"]; found {
+	if rhUserId, found := attributes[UserKey]; found {
 		if rhUserId == userId {
 			return true
 		}
```
ziccardi, line 306:
```suggestion
		return rhUserId == userId
```

pkg/client/keycloak/client.go
```diff
@@ -368,7 +367,7 @@ func (kc *kcClient) AddRealmRoleToUser(accessToken string, userId string, role g
 
 func isNotFoundError(err error) bool {
 	if e, ok := err.(*gocloak.APIError); ok {
-		if e.Code == 404 {
+		if e.Code == http.StatusNotFound {
 			return true
 		}
```
ziccardi, line 372:
```suggestion
		return e.Code == http.StatusNotFound
```

ziccardi (commented):
Added a couple of comments.
Great job!

pawelpaszki:
@ziccardi - thanks for your review. I have addressed your comments. would you mind taking another look please?

pkg/client/keycloak/client.go
```diff
@@ -291,20 +291,16 @@ func (kc *kcClient) IsSameOrg(client *gocloak.Client, orgId string) bool {
...
 	attributes := *client.Attributes
-	if rhUserId, found := attributes["rh-user-id"]; found {
-		if rhUserId == userId {
-			return true
-		}
-	} else {
-		return true
+	if rhUserId, found := attributes[UserKey]; found {
+		return rhUserId == userId
 	}
```
ziccardi, line 304: I believe we removed too much. According to previous code, if the key is not found, it should return `true`.
If removing this else branch we don't have any test failing, I believe we are missing one test case :)

ziccardi (changes requested):
Just a small fix required. Other than that, looks good!

pkg/client/keycloak/client.go
pawelpaszki: should it not return false?

ziccardi: That's a good question. Before it was doing 
```go
if rhUserId, found := attributes["rh-user-id"]; found {
		if rhUserId == userId {
			return true
		}
	} else {
		return true
```

so when `found` was false, was returning `true`. Now when `found` is `false` it returns `false`.
@akoserwal What is the correct behaviour here? The old one or the one introduced with this PR?

akoserwal: `else { return true } ` was added because some of the initial service accounts didn't have the rh-user-id claims. Now it can be removed. What @pawelpaszki added in correct going forward

ziccardi (approved):
lgtm

PR #1027, opened 2022-05-16
Add expires_at field to kafka request

miguelsorianod:
## Description

Add a new read-only field called `expires_at` to the KafkaRequest schema which contains the calculation of the expiration time for instances that have a limit lifespan defined (in config/kafka-instance-types-configuration.yaml).
If a given instance size does not have a limit lifespan decided then expires_at is not set and thus not returned in the response.

Implications of this implementation:
* expires_at is derived from the creation time of the kafka instance and the lifespanSeconds from the size configuration. It is not stored in the DB
* If a kafka size has a lifespan set we ALWAYS set the expires_at
* If a kafka size does not have a lifespan we never set the expires_at
* expires_at can't be set by a user. If that is needed then we will change the behavior. The risk here is that the semantics of the expires_at field might change in that case which might be potentially breaking behavior
* If the lifespan of a kafka instance size is changed, the expiration time of already pre-existing instances will change too and thus the instances will be affected by the change. This happens in both scenarios of increasing the lifespan or decreasing it. This behavior was already happening before the implementation of this PR.

## Verification Steps

* Create a standard instance and verify that when getting that kafka instance through the API the expires_at field is not returned.
* Create a developer instance and verify that when getting that kafka instance through the API the expires_at is set at a time equal to created_at + lifespanSeconds. lifespanSeconds value should be the one set in the config/kafka-instance-types-configuration.yaml for the corresponding instance type-size.

- [ ] Code Review completed


miguelsorianod:
This will need changes in app-interface. I think changes in app-interface should be done before merging this PR and they can be done at any moment

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -2,13 +2,15 @@ package dbapi
...
 	"encoding/json"
+	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	"gorm.io/gorm"
 )
 
 type KafkaRequest struct {
 	api.Meta
+	ExpiresAt                        *time.Time
```
machi1990, line 13: Do we need to persist the information in the database?

the question is more or less: do we want to record the expiration time of an instance and make sure that any future change of the configuration value does not impact the value set for already existing instances?

miguelsorianod: I'd say so. It is also useful information to consumers of the get/list endpoint

internal/kafka/internal/migrations/20220511164100_add_expires_at_column.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/environments"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// addExpiresAtColumn adds the expires_at column in the kafka_requests table
+// and sets its value for potentially already existing kafka instances of
+// 'developer' instance_type. The lifespan set for the kafka instance depends
+// on the OCM_ENV value that the migration is run with.
+func addExpiresAtColumn() *gormigrate.Migration {
```
machi1990, line 15: Assuming that we still need to persist the expired_at time (I've asked a question in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1027/files#r873909065) 

The way we update existing values does not take into account the fact that a new developer instance might be created during a rollout of this change. Specifically if a developer kafka is created on a existing old pod during a time when the execution of this migration is between line 44-48 with.

We've two options:
- do a phased rollout
- or a use a reconciler to reconciler expires_at time for developer instance.

I think the second option might be better since it also reads the correct values from the configuration and not assumption taken in line 34, 36 and 38 (hardcoding the values). Chances of them changing a low but they could still change.

internal/kafka/internal/api/dbapi/kafka_request_types.go
miguelsorianod: > Can you expand on this? How easier?

You only need to compare the field value and the current time

internal/kafka/internal/migrations/20220511164100_add_expires_at_column.go
miguelsorianod: > The way we update existing values does not take into account the fact that a new developer instance might be created during a rollout of this change. Specifically if a developer kafka is created on a existing old pod during a time when the execution of this migration is between line 44-48 with.

That's not accurate at the moment. The calculation for kafka instances expiration is not done from the value of the config file but from the KAFKA_LIFESPAN environment variable (see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1027#issuecomment-1127830297).
This means we could:
* merge this functionality
* wait until changes are deployed for 2 hours in stage and for 48 hours in production
* merge the changes that update the internal logic to use expire_at to expire the kafka instances

machi1990: > > The way we update existing values does not take into account the fact that a new developer instance might be created during a rollout of this change. Specifically if a developer kafka is created on a existing old pod during a time when the execution of this migration is between line 44-48 with.
> 
> That's not accurate at the moment. The calculation for kafka instances expiration is not done from the value of the config file but from the KAFKA_LIFE_SPAN environment variable at the moment (see [#1027 (comment)](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1027#issuecomment-1127830297)). 

The reconcilers have access to the CLI configurations you mentioned and all the other configurations coming from files (configmap). Generally, if need to access values dynamically that are not reliable accessible at a migration time (e.g data from external services, data that may be modified by a running old pod etc), then a reconciler (temporarily one) is a a fallback option. 

Another option is of course doing a phased rollout which I think is what you’ll go with from the last comment

internal/kafka/internal/api/dbapi/kafka_request_types.go
machi1990: > > Can you expand on this? How easier?
> 
> You only need to compare the field value and the current time stored in the kafka request

This is more or less what is there current but doing the comparison on the fly as well. The lifespan is passed in as an argument and it is used to perform the db update. If no expires_at is stored then the logic requires no change other than continuing to pass in the lifespan duration. If the expires_at is stored then the deprovision logic will need to be revisited. But as far as the comparison goes, I do find the two ways of doing things at the same level of understandability.

machi1990: > > The question was specifically about persisting the information in the database. The API can still return the expires_at information without persisting it in the database. It'll return it on the fly on the presentation layer.
> 
> It is a possible option, although as you said the implication is that then if we change the values in the supported kafka instance types it would imply a change of the expiration time in both the returned responses and in the expiration logic (when implemented). 

This is what I’d like to get clarity on: what’s the expected behaviour? At the moment, the behavious is - if the lifespan flag is changed, the developer instance (old and existing ones) will expires at a time equals to “created_at + new_lifespan_duration”. 
if we store the value and updated the deprovision logic based on to use the expires_at, we are changing the current behaviour, is this desirable or a side effect of work in the PR?

> Aside from that, if for some reason in the future we let POST/PATCH specify an expire time that would not work

In the future, if it is needed, it can be added.

miguelsorianod: > This is what I’d like to get clarity on: what’s the expected behaviour? At the moment, the behavious is - if the lifespan flag is changed, the developer instance (old and existing ones) will expires at a time equals to “created_at + new_lifespan_duration”.
if we store the value and updated the deprovision logic based on to use the expires_at, we are changing the current behaviour, is this desirable or a side effect of work in the PR?

I guess that is not really defined/decided. Both options seem acceptable to me.
I would like other thoughts about this from @JameelB.

> In the future, if it is needed, it can be added.

That is correct.

internal/kafka/internal/migrations/20220511164100_add_expires_at_column.go
miguelsorianod: > The reconcilers have access to the CLI configurations you mentioned and all the other configurations coming from files (configmap). Generally, if need to access values dynamically that are not reliable accessible at a migration time (e.g data from external services, data that may be modified by a running old pod etc), then a reconciler (temporarily one) is a a fallback option.

I understand this but I am not sure I am getting what's the point you are trying to explain. The risk of that happening is non-existent due to the logic of kafka expiration calculation does not rely on the new changes that would be applied at the moment.

machi1990: Since the `expires_at` column is used in the presenter, any new kafka created on an old Pod in between the lines I mentioned above won't have the `expires_at`, the information returned for these developers kafkas will be empty.

riccardo-forina:
Just chiming in to ask, if `expires_at` is a calculated field as stated in the PR's description, why does it need to be stored in the DB? Isn't it enough to derive the value from the `created_at` value by subtracting the recently added field that holds the instance duration length?

machi1990:
> Just chiming in to ask, if `expires_at` is a calculated field as stated in the PR's description, why does it need to be stored in the DB? Isn't it enough to derive the value from the `created_at` value by subtracting the recently added field that holds the instance duration length?

This is what I am asking in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1027#discussion_r873909065 as well.

internal/kafka/internal/api/dbapi/kafka_request_types.go
miguelsorianod: If we want to keep the current behavior, where changing KAFKAS_LIFE_SPAN value impacts all already pre-existing kafka instances lifetime then the approach would be not to persist the value in the DB

JameelB: Good point @machi1990. For me both options are reasonable as well. +1 with what @miguelsorianod said, we haven't really discussed it. Is the current behaviour the actual desired behaviour? 

I think adding the expires_at in the db would definitely be more future proof. It will enable us to allow users to request an extension for short-lived Kafkas. 

We don't really change the lifespan duration that much. With the current behaviour though, if i was the user of a Kafka and was told I had an x amount of hours left at the start then it was changed to a shorter time limit and the Kafka expired without me knowing, i'd find it confusing as to why that Kafka disappeared all of the sudden. 

I guess question is, if we told the user at the start they have x amount of hours on their Kafka, do we need to ensure that their Kafka stays available for that period of time?

riccardo-forina: Can’t `expire_at` be used only for the eventual trial extend case, while deriving it for all other instances?

pb82: @riccardo-forina `expires_at` is only relevant for trial/developer instances. standard instances don't expire.

miguelsorianod: Decision. Keep the behavior we have for now, and we return expires_at always if the instance has a lifespan:
1. Do not store it in the DB
2. If a kafka size has a lifespan set we ALWAYS set the expires_at
3. If a kafka size does not have a lifespan we never set the expires_at
4. expires_at can't be set by a user. If that is needed then we will change the behavior. The risk here is that the semantics of the expires_at field might change in that case which might be potentially breaking behavior
5. This means if we change the lifespan of an instance size the already existing instances are impacted. This is what is already happening

machi1990 (approved):
the change looks good to me

I'll let @JameelB @pb82 have a look as well

ziccardi (approved):
lgtm

pb82 (approved):

PR #1016, opened 2022-05-11
MGDCTRS-933: add an admin endpoint to get cluster details

lburgazzoli:
- MGDCTRS-927: add an endpoint to list connector type and related shard metadata
- MGDCTRS-933: add an admin endpoint to get cluster details

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has been created for changes required on the client side

PR #1029, opened 2022-05-17
Mgdstrm 8476

pb82:
## Description

Returns a new field `maturity_status` from the supported instance types endpoint.

## Verification Steps

1. Compile and run the fleet manager locally: `make binary && ./kas-fleet-manager serve`
2. Run the OpenAPI server: `make run/docs`
3. Query the `/api/kafkas_mgmt/v1/instance_types/{cloud_provider}/{cloud_region}` endpoint using `aws` and `us-east-1`.
4. Make sure that all sizes have a field `maturity_status`, but only `standard.x1` is `stable`. The others should be `preview`.

- [ ] Code Review completed
- [x] JIRA has been created for changes required on the client side


internal/connector/internal/api/admin/private/api/openapi.yaml
```diff
@@ -133,6 +133,48 @@ paths:
       summary: Returns a list of connector clusters
       tags:
       - Connector Clusters Admin
+  /api/connector_mgmt/v1/admin/kafka_connector_clusters/{connector_cluster_id}:
```
pb82, line 136: not quite sure where this came from. Connectors related change that appeared after `make generate`.

internal/connector/internal/api/admin/private/api_connector_clusters_admin.go
```diff
@@ -743,6 +743,102 @@ func (a *ConnectorClustersAdminApiService) GetConnector(ctx _context.Context, co
 	return localVarReturnValue, localVarHTTPResponse, nil
 }
 
+/*
+GetConnectorCluster Get a connector cluster
+ * @param ctx _context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+ * @param connectorClusterId The id of the connector cluster
+@return ConnectorCluster
+*/
+func (a *ConnectorClustersAdminApiService) GetConnectorCluster(ctx _context.Context, connectorClusterId string) (ConnectorCluster, *_nethttp.Response, error) {
```
pb82, line 752: not quite sure where this came from. Connectors related change that appeared after `make generate`.
ping @lburgazzoli

internal/connector/internal/api/admin/private/api/openapi.yaml
miguelsorianod: This is an uncommited change from someone else's PR. Please undo changes not related to this PR.
This is a recurring issue. I created a Jira issue to deal with this https://issues.redhat.com/browse/MGDSTRM-8508 to try to prevent it from happening.

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -12,12 +12,23 @@ import (
...
+	MaturityStatusStable      MaturityStatus = "stable"
+)
+
 type KafkaInstanceType struct {
 	Id          string              `yaml:"id"`
 	DisplayName string              `yaml:"display_name"`
 	Sizes       []KafkaInstanceSize `yaml:"sizes"`
 }
 
+func getValidMaturityStates() []MaturityStatus {
```
miguelsorianod, line 28: In https://issues.redhat.com/browse/MGDSTRM-8476 more states are listed. Is having this two values only for now intentional?

miguelsorianod:
This will need changes in appInterface

internal/kafka/internal/config/kafka_supported_instance_types_test.go
```diff
@@ -772,6 +772,26 @@ func TestKafkaSupportedSizesConfig_Validate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "Should return an error if maturity status is invalid",
```
miguelsorianod, line 776: Another test to check that it can't be empty would be good too

internal/kafka/internal/config/kafka_supported_instance_types.go
pb82: As per Jira: `For now, preview and stable will suffice.`. No harm in adding the others already.

miguelsorianod: :+1: we can leave it as it is for now then

internal/connector/internal/api/admin/private/api/openapi.yaml
lburgazzoli: seems to be something missing from here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1016/

miguelsorianod (changes requested):
I left some comments. Changes look good to me.
There are some tests failing.

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -118,6 +130,17 @@ func (k *KafkaInstanceSize) validate(instanceTypeId string) error {
 		return fmt.Errorf("Kafka instance size '%s' for instance type '%s' specifies a lifespanSeconds seconds value less than or equals to Zero.", k.Id, instanceTypeId)
 	}
 
+	maturityStatusKnown := false
+	for _, status := range getValidMaturityStates() {
+		if k.MaturityStatus == status {
+			maturityStatusKnown = true
```
ziccardi, line 136: for can be interrupted here:
```suggestion
			maturityStatusKnown = true
			break
```

miguelsorianod:
@pb82 I created a `needs-app-interface-changes` label and added it to this PR. Do you think this might be useful?

config/kafka-instance-types-configuration.yaml
```diff
@@ -46,6 +46,7 @@ supported_instance_types:
       capacityConsumed: 1
       supportedAZModes:
       - multi
+      maturityStatus: stable
```
machi1990, line 49: let's also make sure that https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e6604e318d78ea293ca043bfc1ba9e16c69a1417/templates/service-template.yml#L160-L163 is updated.

machi1990 (changes requested):
I added a comment about updating service template. Let's also notify kas-installer users or users who are directly deploying KFM on their openshift instance and who are overriding the supported_instances_types parameter to update their config to include the change of maturityStatus info.

/cc @daveconde @MikeEdgar

config/kafka-instance-types-configuration.yaml
pb82: :+1:

daveconde:
Is this backwards compatible, will something break if this isn't set?

pb82:
@daveconde it will break, but we will make sure that is in app-interface before merging this.

machi1990:
> Maybe we could push some default values likes these into KFM Makefile or an env script so the updates flow all the way down.

We try to best make sure that the default values are updated i.e the one maintained in the service template as commented in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1029#discussion_r874789255 or the one we have in Makefile.

The issue is more if the user has overridden these values to something else. When the underlying struct is modified, it may come with an update penalty to the user since they'll have to update the value as well.

in my opinion, KFM can still strive to make the change less breaking change by either ignoring validation when the value is not set etc and default to some values in the code but that also comes with a pain in maintaining the temporary code for the time until all users have migrated to using the new OpenShift parameter  - which we do not know yet . So at the moment, we rely on giving users a heads up, until we've a clear policy around these parameters changes.

pkg/client/aws/client_moq.go
```diff
@@ -14,25 +14,25 @@ var _ AWSClient = &AWSClientMock{}
```
miguelsorianod, line 14: Can you undo this change? It does not belong to this PR either

miguelsorianod (approved):

machi1990 (approved):
lgtm, thanks @pb82 

can we run the make generate command again to fix the conflict?

pb82:
@machi1990 won't that re-introduce the conflicts? I've specifically removed those changes now.

PR #1031, opened 2022-05-17
ci: add a check to verify that make generate has been run

machi1990:
## Description

Add a check to verify that make generate has been run
This is only added for Github CI for now. No need to add it to midstream pipelines

PS: I'll add another commit that introduce the required changes with `make generate code/fix` output.
Opening it to get the feedback from CI if it is working. 
 
/cc @miguelsorianod what do you think?
## Verification Steps

1. See the failure reason in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/runs/6472293220?check_suite_focus=true which is an output we get when `make generate code/fix` has not been run.
2. If the `make generate code/fix` command has been run, the CI should pass as normal

- [ ] Code Review completed


check_openapi_generator_has_been_run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
```
miguelsorianod, line 1: Any reason this is in the project's main directory instead of in scripts or .github?

.github/workflows/ci.yaml
```diff
@@ -114,6 +114,10 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
+      - name: Make Sure That OpenAPI code has been generated
```
miguelsorianod, line 117: The name is not fully accurate. It should be something like `Make sure there is no uncommited autogenerated code` or something like that

check_openapi_generator_has_been_run.sh
machi1990: No reason at all. It can be moved there

.github/workflows/ci.yaml
machi1990: I like the way you worded it.

check_openapi_generator_has_been_run.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# First attempt to generate new code
+make generate code/fix
+
+# Then make sure that no new code was generated 
+changed_file_count=$(git status -s | wc -l)
```
miguelsorianod, line 7: I am being very nitpicky already but a possible alternative is using `git status --porcelain`.
The description of the `porcelain` flag is:
```
       --porcelain[=<version>]
           Give the output in an easy-to-parse format for scripts. This is similar to the short output, but will remain stable across Git versions and regardless of user configuration. See below for details.

           The version parameter is used to specify the format version. This is optional and defaults to the original version v1 format.
```

Up to you how to implement it. Both approaches are good

check_openapi_generator_has_been_run.sh
machi1990: Actually, I spoke too early. 
If it is moved to .gihub 
`make` doesn't seem quite happy about it 
```
-t openapi/templates --ignore-file-override ./.openapi-generator-ignore
^Cmake: *** [Makefile:421: openapi/generate/kas-public] Interrupt
```
would you know why?

miguelsorianod: I can think of a possible reason without having looked at it in detail:

Some of the tasks being run are run relative to the current directory so I guess it can't find the files then?

The appropriate approach in that case is to reference paths using `PROJECT_PATH` variable in the `Makefile`

check_openapi_generator_has_been_run.sh
machi1990: Good to know, thanks.

.github/workflows/ci.yaml
machi1990: Yes. That's true. I've changed the script name as well.

check_openapi_generator_has_been_run.sh
machi1990: I did indeed run the script from a different directory. 
This has been fixed now.

machi1990:
@miguelsorianod mind having another look?

wtrocki:
Can we add validation of API as well?

machi1990:
> Can we add validation of API as well? If not there

It is there already and being run as part of the `make lint` target. 
We use spectral rules for that.

JameelB (approved):
Thanks for this @machi1990! :)

miguelsorianod (commented):
:+1:

PR #1033, opened 2022-05-17
test: fix and refactor kafka integration tests to work with rhsso

JameelB:
## Description
Refactor and fix any issues found when running integration tests with both MAS SSO and RHSSO clients. 

Please note that when `redhat_sso` is set as the selected SSO provider, this will call to the actual sso instance whether the environment selected is integration or development. Because of this, we will need to get an actual token from rhsso within our tests. If mas_sso is selected, we can continue on using the mock token. 

There are also a few differences in the responses between these two clients. Any assertions that has different results between the two have been commented out for now.

Note that this PR only fixes kafka related tests

## Verification Steps
1. Ensure CI is passing (Runs the tests with `OCM_ENV=integration` AND `SelectSSOProvider=MAS_SSO ` configuration)
2. Run the kafka test with the following configurations:
    * `OCM_ENV=integration` and `SelectSSOProvider: REDHAT_SSO`
    * `OCM_ENV=development` and `SelectSSOProvider: MAS_SSO`
       * See results here: [mas_sso_dev_env_test_results](https://gist.github.com/JameelB/afd2aa3d439765d36b3f634148cf804b#file-mas_sso_dev_env_test_results)
    * `OCM_ENV=development` and `SelectSSOProvider: REDHAT_SSO`
      * See results here: [rhsso_dev_env_test_results](https://gist.github.com/JameelB/afd2aa3d439765d36b3f634148cf804b#file-rhsso_dev_env_test_results)
      * Note that connector tests are still failing when using rhsso provider. This will be fixed in a separate PR

When running with 'redhat_sso' + `OCM_ENV=integration`, please update any references of `sso.redhat.com` to `sso.qa.redhat.com` in the [keycloakConfig](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/client/keycloak/config.go#L89) and [serverConfig](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/server/server_config.go). Do not modify the mas-sso-base-url as this is still use for configuration of osd idp. A valid rhsso service account client id and secret is required to run this, please message me for that information when verifying these changes.

- [ ] Code Review completed


internal/kafka/internal/handlers/service_accounts.go
```diff
@@ -124,11 +130,17 @@ func (s serviceAccountsHandler) ResetServiceAccountCredential(w http.ResponseWri
...
 	clientId := r.FormValue("client_id")
+	validators := []handlers.Validate{
+		handlers.ValidateLength(&clientId, "client_id", handlers.MinRequiredFieldLength, &handlers.MaxServiceAccountClientId),
+	}
+
+	if s.service.GetConfig().SelectSSOProvider == keycloak.MAS_SSO {
+		// only service accounts from mas sso are prefixed with "srvc-acc-"
+		validators = append(validators, handlers.ValidateServiceAccountClientId(&clientId, "client_id"))
+	}
+
```
ziccardi, line 141: I think it would be more readable if, instead of this change, we would add the `SelectSSOProvider` as parameter to `ValidateServiceAccountClientId` which, in turn, will just return `true` if the provider is not `MAS_SSO`.
WDYT?

internal/kafka/test/mocks/kafkas/kafka_instance_types.go
```diff
@@ -0,0 +1,35 @@
...
+	DefaultTotalMaxConnections         = 100
+	DefaultMaxDataRetentionSize        = "1Gi"
+	DefaultMaxDataRetentionPeriod      = "P14D"
+	DefaultMaxConnectionAttemptsPerSec = 100
+	DefaultQuotaconsumed               = 1
+	DefaultQuotaType                   = "RHOSAK"
+	DefaultCapacityConsumed            = 1
+)
+
+func BuildKafkaInstanceSize(modifyFn func(kafkaInstanceSize *config.KafkaInstanceSize)) *config.KafkaInstanceSize {
```
ziccardi, line 18: this is a good candidate for the functional options pattern

internal/kafka/internal/handlers/service_accounts.go
JameelB: makes sense, i'll change that

JameelB: updated, had to pass in the `SelectSSOProvider` as param to the validator

internal/kafka/test/mocks/kafkas/kafka_instance_types.go
JameelB: updated it there, some function and var names were clashing with the kafka mock so i put it in its own package. Please take a look when you get a chance :)

internal/kafka/internal/presenters/kafka_test.go
```diff
@@ -37,12 +39,12 @@ func TestConvertKafkaRequest(t *testing.T) {
 				kafkaRequestPayload: *mock.BuildKafkaRequestPayload(nil),
 				dbKafkaRequests:     []*dbapi.KafkaRequest{},
 			},
-			want: mock.BuildKafkaRequest(
-				mock.With(mock.STATUS, ""),
-				mock.With(mock.OWNER, ""),
-				mock.With(mock.CLUSTER_ID, ""),
```
ziccardi, line 43: Isn't 'empty string' the default? I believe you can remove the `mock.With` calls here.

internal/kafka/test/integration/data_plane_cluster_test.go
```diff
@@ -110,15 +112,16 @@ func TestDataPlaneCluster_BadRequestWhenNonexistingCluster(t *testing.T) {
...
-	ctx := kasfleetshardsync.NewAuthenticatedContextForDataPlaneCluster(h, testDataPlaneclusterID)
+	ctx, err := kasfleetshardsync.NewAuthenticatedContextForDataPlaneCluster(h, testDataPlaneclusterID)
+	Expect(err).ToNot(HaveOccurred())
 	privateAPIClient := test.NewPrivateAPIClient(h)
 
-	resp, err := privateAPIClient.AgentClustersApi.UpdateAgentClusterStatus(ctx, testDataPlaneclusterID, private.DataPlaneClusterUpdateStatusRequest{})
-	Expect(resp.StatusCode).To(Equal(http.StatusNotFound)) // We expect 404 error in this test because the cluster ID does not exist
+	resp, err := privateAPIClient.AgentClustersApi.UpdateAgentClusterStatus(ctx, testDataPlaneclusterID, *dataplanemocks.BuildValidDataPlaneClusterUpdateStatusRequest(nil))
+	Expect(resp.StatusCode).To(Equal(http.StatusBadRequest)) // We expect 400 error in this test because the cluster ID does not exist
 	Expect(err).To(HaveOccurred())
```
ziccardi, line 121: err should be checked first to avoid `panic` errors
```suggestion
	Expect(err).To(HaveOccurred())
	Expect(resp.StatusCode).To(Equal(http.StatusBadRequest)) // We expect 400 error in this test because the cluster ID does not exist
```

internal/kafka/test/integration/data_plane_cluster_test.go
```diff
@@ -110,15 +112,16 @@ func TestDataPlaneCluster_BadRequestWhenNonexistingCluster(t *testing.T) {
...
 
 	_, resp, err = privateAPIClient.AgentClustersApi.GetKafkaAgent(ctx, testDataPlaneclusterID)
-	Expect(resp.StatusCode).To(Equal(http.StatusNotFound)) // We expect 404 error in this test because the cluster ID does not exist
+	Expect(resp.StatusCode).To(Equal(http.StatusBadRequest)) // We expect 400 error in this test because the cluster ID does not exist
 	Expect(err).To(HaveOccurred())
```
ziccardi, line 125: See previous comment
```suggestion
	Expect(err).To(HaveOccurred())
	Expect(resp.StatusCode).To(Equal(http.StatusBadRequest)) // We expect 400 error in this test because the cluster ID does not exist
```

internal/kafka/test/integration/data_plane_cluster_test.go
```diff
@@ -197,7 +200,9 @@ func TestDataPlaneCluster_GetManagedKafkaAgentCRSuccess(t *testing.T) {
 		t.Fatalf("Cluster not found")
 	}
 
-	ctx := kasfleetshardsync.NewAuthenticatedContextForDataPlaneCluster(h, testDataPlaneclusterID)
+	ctx, err := kasfleetshardsync.NewAuthenticatedContextForDataPlaneCluster(h, testDataPlaneclusterID)
+	Expect(err).ToNot(HaveOccurred())
+
 	privateAPIClient := test.NewPrivateAPIClient(h)
 	config, resp, err := privateAPIClient.AgentClustersApi.GetKafkaAgent(ctx, testDataPlaneclusterID)
 	Expect(resp.StatusCode).To(Equal(http.StatusOK))
```
ziccardi, line 208: Move this after the error check

internal/kafka/test/integration/data_plane_cluster_test.go
```diff
@@ -613,7 +625,9 @@ func TestDataPlaneCluster_WhenReportedStrimziVersionsIsEmptyAndClusterStrimziVer
 		t.Fatalf("Cluster not found")
```
ziccardi, line 625: Not part of this PR, but I believe this should be changed to 
```go
Expect(testDataPlaneclusterID).ToNot(BeEqual(""));
```

Same should be done for the other check (`if err != nil` ==> `Expect(err).ToNot(HaveOccurred())`

internal/kafka/test/integration/kafkas_test.go
```diff
@@ -1679,7 +1679,11 @@ func TestKafkaDelete_DeleteDuringCreation(t *testing.T) {
 
 		// only update kafka status if a data plane cluster is available
 		if dataplaneCluster != nil {
-			ctx := kasfleetshardsync.NewAuthenticatedContextForDataPlaneCluster(helper, dataplaneCluster.ClusterID)
+			ctx, err := kasfleetshardsync.NewAuthenticatedContextForDataPlaneCluster(helper, dataplaneCluster.ClusterID)
+			if err != nil {
+				return err
+			}
```
ziccardi, line 1685:
```suggestion
Expect(err).ToNot(BeOccurred())
```

internal/kafka/test/integration/serviceaccounts_test.go
```diff
@@ -88,38 +105,43 @@ func TestServiceAccounts_Success(t *testing.T) {
...
 	Expect(sa.Id).NotTo(BeEmpty())
 	Expect(sa.CreatedAt).Should(BeTemporally(">=", createdAt))
+	// skipping for now as createdBy is empty when using redhat_sso client, will need to generate new version of sdk
+	// Expect(sa.CreatedBy).Should(Equal(account.Username()))
 
 	// verify get by id
 	id := sa.Id
 	sa, resp, err = client.SecurityApi.GetServiceAccountById(ctx, id)
 	Expect(resp.StatusCode).To(Equal(http.StatusOK))
 	Expect(err).ShouldNot(HaveOccurred())
```
ziccardi, line 117: Move this as first check

internal/kafka/test/integration/serviceaccounts_test.go
```diff
@@ -88,38 +105,43 @@ func TestServiceAccounts_Success(t *testing.T) {
...
+	// Expect(sa.CreatedBy).Should(Equal(account.Username())) // createdBy is empty
+	// Expect(sa.CreatedBy).NotTo(BeEmpty())
 
 	//verify delete
 	_, _, err = client.SecurityApi.DeleteServiceAccountById(ctx, id)
 	Expect(err).ShouldNot(HaveOccurred())
 
 	// verify deletion of non-existent service account throws http status code 404
-	_, resp, _ = client.SecurityApi.DeleteServiceAccountById(ctx, id)
-	Expect(resp.StatusCode).To(Equal(http.StatusNotFound))
```
ziccardi, line 122: I believe a test on the error should be added here too.

ziccardi (changes requested):

internal/kafka/internal/presenters/kafka_test.go
JameelB: Not with `BuildKafkaRequest`, these fields are set now.

I did replace this one to use `BuildEmptyKafkaRequest` though which returns a kafka request object with no fields set so i've removed the mock.With calls here that sets these fields to an empty string. Reason for that is, the majority of the kafka requests expected by other tests requires most of the fields to be set. This particular test requires the majority of them to be empty. Rather than calling each mock.With to set these fields to an empty string, I just build an empty Kafka request and modify the fields to be set as what is expected for this test. :thinking: we don't really have a `BuildEmpty..` object for the rest of our mock objects yet. wdyt of doing it this way? :thinking:

internal/kafka/test/integration/serviceaccounts_test.go
JameelB: The error is tested here, skipping the response code test for now since mas sso and rhsso returns different error codes.

ziccardi: I mean the value of `err`: if `err` is `not nil`, `resp` will be `nil` and the `Expect` will panic

ziccardi: forget it. Commented the wrong row!

internal/kafka/internal/presenters/kafka_test.go
ziccardi: I commented the removed lines instead of the added ones... I guess I need some sleep... 

I would do it like this:

- `BuildKafkaRequest(Options options...)`: create an empty KafkaRequest that can be customised using the options

The I would add a `WithPredefinedTestValues` option that populates the request with the predefined values used by most tests.

This way, if you need the `pre populated` KafkaRequest you will do:
```go
kafkaRequest := BuildKafkaRequest(WithPredefinedTestValues())
````

WDYT?

File: internal/kafka/test/integration/data_plane_cluster_test.go (hunk 4 above)
Comment (JameelB) [reply internal/kafka/test/integration/data_plane_cluster_test.go]: There's quite a lot of this actually throughout our tests, I just changed it for all, at least the ones I can find. I've isolated that change in its own commit.

File: internal/kafka/test/integration/data_plane_cluster_test.go (hunk 7 above)
Comment (JameelB) [reply internal/kafka/test/integration/data_plane_cluster_test.go]: I changed occurrences of checking errors like this `if err != nil` and failing with `t.Fatal` and `t.Error` to use gomega expect assertion. I only changed it for the data plane cluster test though but saw several instances where we do this in other test files. 

I think I can create a follow up PR to refactor that to use gomega instead, wdyt?

File: internal/kafka/internal/presenters/kafka_test.go (hunk 3 above)
Comment (JameelB) [reply internal/kafka/internal/presenters/kafka_test.go:43]: that sounds good :+1: i'll update that now.

File: internal/kafka/test/integration/data_plane_cluster_test.go (hunk 7 above)
Comment (ziccardi) [reply internal/kafka/test/integration/data_plane_cluster_test.go]: Agree

**Comment by JameelB** [comment]
@ziccardi i've addressed your comments, please have a look again when you get the chance.

File: internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr_test.go [hunk 11]
```diff
@@ -161,6 +161,7 @@ func TestAcceptedKafkaManager_reconcileAcceptedKafka(t *testing.T) {
 			},
 			args: args{
 				kafka: mockKafkas.BuildKafkaRequest(
+					mockKafkas.With(mockKafkas.CLUSTER_ID, mockKafkas.DefaultClusterID),
 					mockKafkas.With(mockKafkas.STATUS, ""),
```
ziccardi, line 165: I think this can be removed. Default is empty string.

internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr_test.go
```diff
@@ -189,6 +190,7 @@ func TestAcceptedKafkaManager_reconcileAcceptedKafka(t *testing.T) {
...
```
ziccardi, line 194: can be removed

internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr_test.go
```diff
@@ -222,6 +224,7 @@ func TestAcceptedKafkaManager_reconcileAcceptedKafka(t *testing.T) {
...
```
ziccardi, line 228: can be removed

internal/kafka/test/integration/kafkas_test.go
ziccardi: This one is still here :)

ziccardi: My bad. It is correct. Sorry.

ziccardi (approved):
Amazing work! Just left a couple of minor comments

LGTM

JameelB:
@ziccardi got latest changes from master and addressed your comments there :) 

I also re-added the check for verifying the service account created by field since the service account sdk version has been bumped. cc @akoserwal

ziccardi (approved):
Amazong job! Thanks Jameel!

PR #1037, opened 2022-05-19
chore: remove image pull secret

JameelB:
## Description
Remove image pull secret so that any changes to this file can be ignored by git. This is already included in the [.gitignore](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/.gitignore#L60)

## Verification Steps
1. Ensure any changes to image pull secret file is not included in git changes

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/config/dataplane_cluster_config.go
+++ b/internal/kafka/internal/config/dataplane_cluster_config.go
@@ -292,8 +292,12 @@ func (c *DataplaneClusterConfig) AddFlags(fs *pflag.FlagSet) {
 func (c *DataplaneClusterConfig) ReadFiles() error {
 	if c.ImagePullDockerConfigContent == "" && c.ImagePullDockerConfigFile != "" {
 		err := shared.ReadFileValueString(c.ImagePullDockerConfigFile, &c.ImagePullDockerConfigContent)
-		if err != nil {
+		if err != nil && !os.IsNotExist(err) {
 			return err
+		}
+
+		if (err != nil && os.IsNotExist(err)) || c.ImagePullDockerConfigContent == "" {
+			logger.Logger.Warningf("Specified image pull secret file does not exist or has no content. Data plane cluster terraform may fail if operators to be installed use images from a private repository")
 		}
 	}
 
--- a/internal/kafka/internal/config/dataplane_cluster_config_test.go
+++ b/internal/kafka/internal/config/dataplane_cluster_config_test.go
@@ -846,7 +846,7 @@ func Test_ReadFiles(t *testing.T) {
 		wantErr  bool
 	}{
 		{
-			name: "should return an error with misconfigured ImagePullDockerConfig",
+			name: "should not return an error when specified ImagePullDockerConfig file does not exist",
 			fields: fields{
 				config: NewDataplaneClusterConfig(),
 			},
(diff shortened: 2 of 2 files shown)


pawelpaszki (approved):

Rajagopalan-Ranganathan (approved):

JameelB:
@machi1990 you're right, i think it might be. Reading the file is going to fail. Let me check the ci config

machi1990:
> @machi1990 you're right, i think it might be. Reading the file is going to fail. Let me check the ci config

Maybe, we could ignore the the error if the image pull secret file does not exists.

JameelB:
@machi1990 maybe it should be required since installation of the strimzi operator is going to fail without it? I'm not sure if that's the same case for standalone clusters though.

machi1990:
> @machi1990 maybe it should be required since installation of the strimzi operator is going to fail without it? I'm not sure if that's the same case for standalone clusters though.

Very good point. The standalone case depends on the image being public or not. "If" there are public operators images then the pull secret configuration is not required at all. Its quite possible that someone can have some public operators images that they want to use without configuring this secret. 
I guess ignoring not exists error in those scenario makes sense. Possibly, accompanied with a warning message of what might go wrong.

JameelB:
> > @machi1990 maybe it should be required since installation of the strimzi operator is going to fail without it? I'm not sure if that's the same case for standalone clusters though.
> 
> Very good point. The standalone case depends on the image being public or not. "If" there are public operators images then the pull secret configuration is not required at all. Its quite possible that someone can have some public operators images that they want to use without configuring this secret. I guess ignoring not exists error in those scenario makes sense. Possibly, accompanied with a warning message of what might go wrong.

Thanks for confirmation and +1 to that idea. I think it'll be difficult to determine the cluster provider at this stage so we could do what you suggested: ignore readfile error if it doesn't exist and just accompany that with the warning message on kas fleet manager start up. wdyt? 

As for the CI checks, since that just runs with a mock server, I don't think we need to set the image pull secret in the ci config?

machi1990:
> > > @machi1990 maybe it should be required since installation of the strimzi operator is going to fail without it? I'm not sure if that's the same case for standalone clusters though.
> > 
> > 
> > Very good point. The standalone case depends on the image being public or not. "If" there are public operators images then the pull secret configuration is not required at all. Its quite possible that someone can have some public operators images that they want to use without configuring this secret. I guess ignoring not exists error in those scenario makes sense. Possibly, accompanied with a warning message of what might go wrong.
> 
> Thanks for confirmation and +1 to that idea. I think it'll be difficult to determine the cluster provider at this stage so we could do what you suggested: ignore readfile error if it doesn't exist and just accompany that with the warning message on kas fleet manager start up. wdyt?
> 
> As for the CI checks, since that just runs with a mock server, I don't think we need to set the image pull secret in the ci config?

No, we do not need to get the image pull secret in the Github config and for the midstream PR check it is not needed as well. But it is needed for the nightly run that @pawelpaszki is doing its configuration in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1038

machi1990 (approved):
lgtm

love the clarity of the log message. Thanks for this @JameelB

PR #1044, opened 2022-05-23
fix: developer instances should be marked as stable

pb82:
update the maturity status of developer instances to `stable` as per BU requirement.


machi1990 (approved):
Thanks @pb82

wtrocki (approved):
Thank you so much for rapid resolution

PR #1059, opened 2022-05-27
Mgdstrm 8579

pb82:
## Description

https://issues.redhat.com/browse/MGDSTRM-8579

## Verification Steps

To test this we need an organization that has linked cloud accounts. Not quite sure how to obtain this.

### DIFF
--- a/internal/kafka/internal/api/dbapi/kafka_request_types.go
+++ b/internal/kafka/internal/api/dbapi/kafka_request_types.go
@@ -52,6 +52,8 @@ type KafkaRequest struct {
 	ReauthenticationEnabled bool   `json:"reauthentication_enabled"`
 	RoutesCreationId        string `json:"routes_creation_id"`
 	SizeId                  string `json:"size_id"`
+	BillingCloudAccountId   string `json:"billing_cloud_account_id"`
+	Marketplace             string `json:"marketplace"`
 }
 
 type KafkaList []*KafkaRequest
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -31,12 +31,12 @@ func ValidateBillingCloudAccountIdAndMarketplace(ctx context.Context, kafkaServi
 
 	return func() *errors.ServiceError {
 		// both fields are optional
-		if blank(kafkaRequestPayload.CloudAccountId) && blank(kafkaRequestPayload.Marketplace) {
+		if blank(kafkaRequestPayload.BillingCloudAccountId) && blank(kafkaRequestPayload.Marketplace) {
 			return nil
 		}
 
 		// marketplace without a billing account provided
-		if blank(kafkaRequestPayload.CloudAccountId) && !blank(kafkaRequestPayload.Marketplace) {
+		if blank(kafkaRequestPayload.BillingCloudAccountId) && !blank(kafkaRequestPayload.Marketplace) {
 			return errors.InvalidBillingAccount("no billing account provided for marketplace: %s", kafkaRequestPayload.Marketplace)
 		}
 
@@ -53,7 +53,7 @@ func ValidateBillingCloudAccountIdAndMarketplace(ctx context.Context, kafkaServi
 			return errors.NewWithCause(errors.ErrorGeneral, err, "error assigning instance type: %s", err.Error())
 		}
 
-		return (*kafkaService).ValidateBillingAccount(organisationId, instanceType, *kafkaRequestPayload.CloudAccountId, kafkaRequestPayload.Marketplace)
+		return (*kafkaService).ValidateBillingAccount(organisationId, instanceType, *kafkaRequestPayload.BillingCloudAccountId, kafkaRequestPayload.Marketplace)
 	}
 }
 
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"context"
 	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	"regexp"
 	"strings"
@@ -25,18 +26,14 @@ var ValidKafkaClusterNameRegexp = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])
 var MaxKafkaNameLength = 32
 
 func ValidateBillingCloudAccountIdAndMarketplace(ctx context.Context, kafkaService *services.KafkaService, kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
-	blank := func(s *string) bool {
-		return s == nil || *s == ""
-	}
-
 	return func() *errors.ServiceError {
 		// both fields are optional
-		if blank(kafkaRequestPayload.BillingCloudAccountId) && blank(kafkaRequestPayload.Marketplace) {
+		if shared.SafeString(kafkaRequestPayload.BillingCloudAccountId) == "" && shared.SafeString(kafkaRequestPayload.Marketplace) == "" {
 			return nil
 		}
 
 		// marketplace without a billing account provided
-		if blank(kafkaRequestPayload.BillingCloudAccountId) && !blank(kafkaRequestPayload.Marketplace) {
+		if shared.SafeString(kafkaRequestPayload.BillingCloudAccountId) == "" && shared.SafeString(kafkaRequestPayload.Marketplace) != "" {
 			return errors.InvalidBillingAccount("no billing account provided for marketplace: %s", kafkaRequestPayload.Marketplace)
 		}
 
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -24,6 +24,39 @@ var ValidKafkaClusterNameRegexp = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])
 
 var MaxKafkaNameLength = 32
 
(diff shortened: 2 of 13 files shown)


internal/kafka/internal/handlers/validation.go
```diff
@@ -24,6 +24,39 @@ var ValidKafkaClusterNameRegexp = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])
...
 
+func ValidateBillingCloudAccountIdAndMarketplace(ctx context.Context, kafkaService *services.KafkaService, kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
+	blank := func(s *string) bool {
+		return s == nil || *s == ""
+	}
+
+	return func() *errors.ServiceError {
+		// both fields are optional
+		if blank(kafkaRequestPayload.BillingCloudAccountId) && blank(kafkaRequestPayload.Marketplace) {
+			return nil
```
pawelpaszki, line 35: should it not continue to `AssignInstanceType` if two fields are empty?

pb82: No, i don't think so. `AssignInstanceType` is not a good name for what the method does: it only returns if the given org has standard or developer quota. It doesn't really assign anything.

pawelpaszki: yes- you are right, but looking further into the code - you can see in the handler, that AssignInstanceType is called twice. Once here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1059/files#diff-223cf9d875575c2b0cb237848a3d6055a32acf843ff8fce340d22bfbef6ac5faR63 and once in this block you added. do you think this could be somehow unified?

internal/kafka/internal/handlers/validation.go
```diff
@@ -24,6 +24,39 @@ var ValidKafkaClusterNameRegexp = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])
 
 var MaxKafkaNameLength = 32
 
+func ValidateBillingCloudAccountIdAndMarketplace(ctx context.Context, kafkaService *services.KafkaService, kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
+	blank := func(s *string) bool {
```
ziccardi, line 28: I think it would be more readable to call this `isBlank`

internal/kafka/internal/handlers/validation.go
```diff
@@ -24,6 +24,39 @@ var ValidKafkaClusterNameRegexp = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])
...
+		return s == nil || *s == ""
```
ziccardi, line 29:
```suggestion
		return SafeString(s) == ""
```

internal/kafka/internal/presenters/kafka.go
```diff
@@ -26,6 +26,14 @@ func ConvertKafkaRequest(kafkaRequestPayload public.KafkaRequestPayload, dbKafka
 	kafka.CloudProvider = kafkaRequestPayload.CloudProvider
 	kafka.MultiAZ = kafkaRequestPayload.DeprecatedMultiAz
 
+	if kafkaRequestPayload.BillingCloudAccountId != nil {
+		kafka.BillingCloudAccountId = *kafkaRequestPayload.BillingCloudAccountId
+	}
+
+	if kafkaRequestPayload.Marketplace != nil {
+		kafka.Marketplace = *kafkaRequestPayload.Marketplace
+	}
```
ziccardi, line 35:
```suggestion
		kafka.Marketplace = shared.SafeString(kafkaRequestPayload.Marketplace)
```

internal/kafka/internal/services/kafka.go
```diff
@@ -131,6 +132,15 @@ func NewKafkaService(connectionFactory *db.ConnectionFactory, clusterService Clu
 	}
 }
 
+func (k *kafkaService) ValidateBillingAccount(externalId string, instanceType types.KafkaInstanceType, billingCloudAccountId string, marketplace *string) *errors.ServiceError {
+	quotaService, factoryErr := k.quotaServiceFactory.GetQuotaService(api.QuotaType(k.kafkaConfig.Quota.Type))
```
pawelpaszki, line 136: maybe quotaErr would make more sense here (instead of `factoryErr`)?

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -32,6 +32,50 @@ var supportedAMSBillingModels map[string]struct{} = map[string]struct{}{
...
+		return nil
+	} else if matchingBillingAccounts > 1 {
+		return errors.InvalidBillingAccount("Multiple matching billing accounts found, only one expected. Available billing accounts: %v", billingAccounts)
+	}
+
+	if len(billingAccounts) == 0 {
+		return errors.InvalidBillingAccount("No billing accounts available in quota")
+	}
+
+	return errors.InvalidBillingAccount("No matching billing account found. Provided: %s, Available: %v", orgId, billingCloudAccountId, billingAccounts)
```
pawelpaszki, line 76: isn't this line and `if len(billingAccounts) == 0 {` doing the same thing?

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -15,6 +15,133 @@ import (
...
+				ocmClient:   ocmClientMockWithCloudAccounts,
+				kafkaConfig: &defaultKafkaConf,
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gomega.RegisterTestingT(t)
```
pawelpaszki, line 136: this could be moved before the for loop and `gomega.` could be removed, if the import line looks like this: `. "github.com/onsi/gomega"`

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -15,6 +15,133 @@ import (
...
+			factory := NewDefaultQuotaServiceFactory(tt.fields.ocmClient, nil, nil, tt.fields.kafkaConfig)
+			quotaService, _ := factory.GetQuotaService(api.AMSQuotaType)
+			err := quotaService.ValidateBillingAccount(tt.args.orgId, types.STANDARD, tt.args.billingAccountId, tt.args.marketplace)
+			gomega.Expect(err != nil).To(gomega.Equal(tt.wantErr))
```
pawelpaszki, line 140: see comment above about `gomega`

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -52,6 +52,8 @@ type KafkaRequest struct {
 	ReauthenticationEnabled bool   `json:"reauthentication_enabled"`
 	RoutesCreationId        string `json:"routes_creation_id"`
 	SizeId                  string `json:"size_id"`
+	BillingCloudAccountId   string `json:"billing_cloud_account_id"`
+	Marketplace             string `json:"Marketplace"`
```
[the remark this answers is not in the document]
pb82: Good catch!

internal/kafka/internal/services/kafka.go
pb82: this is calling the quota service factory, not the quota service itself. So I think the name is ok here.

internal/kafka/internal/services/quota/ams_quota_service.go
pb82: but i'll add an extra test for this scenario :+1:

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -509,17 +676,17 @@ func Test_AMSReserveQuota(t *testing.T) {
...
-			gomega.Expect(subId).To(gomega.Equal(tt.want))
-			gomega.Expect(err != nil).To(gomega.Equal(tt.wantErr))
+			Expect(subId).To(Equal(tt.want))
+			Expect(err != nil).To(Equal(tt.wantErr))
 
 			if tt.wantBillingModel != "" {
 				ocmClientMock := tt.fields.ocmClient.(*ocm.ClientMock)
 				clusterAuthorizationCalls := ocmClientMock.ClusterAuthorizationCalls()
-				gomega.Expect(len(clusterAuthorizationCalls)).To(gomega.Equal(1))
+				Expect(len(clusterAuthorizationCalls)).To(Equal(1))
```
ziccardi, line 685:
```suggestion
				Expect(clusterAuthorizationCalls).To(HaveLen(1))
```

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -509,17 +676,17 @@ func Test_AMSReserveQuota(t *testing.T) {
...
 				clusterAuthorizationResources := clusterAuthorizationCalls[0].Cb.Resources()
-				gomega.Expect(len(clusterAuthorizationResources)).To(gomega.Equal(1))
+				Expect(len(clusterAuthorizationResources)).To(Equal(1))
```
ziccardi, line 687:
```suggestion
				Expect(clusterAuthorizationResources).To(HaveLen(1))
```

internal/kafka/internal/handlers/validation.go
```diff
@@ -24,6 +25,35 @@ var ValidKafkaClusterNameRegexp = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])
...
 var MaxKafkaNameLength = 32
 
+func ValidateBillingCloudAccountIdAndMarketplace(ctx context.Context, kafkaService *services.KafkaService, kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
+	return func() *errors.ServiceError {
+		// both fields are optional
+		if shared.SafeString(kafkaRequestPayload.BillingCloudAccountId) == "" && shared.SafeString(kafkaRequestPayload.Marketplace) == "" {
+			return nil
+		}
+
+		// marketplace without a billing account provided
```
miguelsorianod, line 35: Is the use case of a billing account without a marketplace supported?

PR #1065, opened 2022-05-30
bump quota for kafka instances team to allow easy testing of 6 broker…

SamBarker:
… clusters.

## Description
The kafka integrations team are doing a lot of work with 6+ node broker clusters and have various different customisations in their kas-installer settings. This PR increases the default quota for the team to simplify testing.

- [ ] Code Review completed

### DIFF
--- a/config/quota-management-list-configuration.yaml
+++ b/config/quota-management-list-configuration.yaml
@@ -40,7 +40,7 @@ registered_users_per_organisation:
     # The Kafka Instances team
   - id: 13639843
     any_user: true
-    max_allowed_instances: 1
+    max_allowed_instances: 5
     registered_users: []
     # The MK Security team
   - id: 13785172


JameelB:
@SamBarker This change will only configure quota for a local kas fleet manager deployment, this file is not used when deploying kas fleet manager to OpenShift. Does kas installer use this file directly? 

I assumed the kas-installer uses the kas fleet manager service-template in order to deploy KAS Fleet Manager to OpenShift? If so, you'll need to update the default value for the `REGISTERED_USERS_PER_ORGANISATION` parameter: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/templates/service-template.yml#L291

We have also recently added changes to allow people to configure this parameter via the `deploy/service` make target (see [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/Makefile#L773)). If kas-installer uses that, it would also be possible to configure its value by specifying that param when calling that make target.

machi1990 (changes requested):
See the above comment:

Also, we'll need to bump the values in:
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/81c2bea272b1075ef39e6d0230c63b936e5cb78c/Makefile#L773
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/81c2bea272b1075ef39e6d0230c63b936e5cb78c/templates/service-template.yml#L291

JameelB:
Just a thought, should this change really be here or should it be done in kas-installer? If kas-installer needs to configure the number of max allowed instances for that org, I think we should do it there by configuring this parameter rather than changing the default value in the template. It just happens that we had multiple organisations as the default value here. 

Not part of this PR but for me the default value for the template should really be minimal. Any modification should be done by those that uses it by configuring the parameter when using the template.

machi1990:
> Just a thought, should this change really be here or should it be done in kas-installer? If kas-installer needs to configure the number of max allowed instances for that org, I think we should do it there by configuring this parameter rather than changing the default value in the template. It just happens that we had multiple organisations as the default value here.

kas-installer does allow for various parameters to be overridden: since https://github.com/bf2fc6cc711aee1a0c2a/kas-installer/pull/163. So this could easily be done there, I think the integration pipeline are not using kas-installer but they are deploying KFM and other components directly using the make targets and they could override the quota settings at `REGISTERED_USERS_PER_ORGANISATION` which was exposed in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1062

> Not part of this PR but for me the default value for the template should really be minimal. Any modification should be done by those that uses it by configuring the parameter when using the template.

+1. 
I think the default values for other orgs were added for convenience. As long as the value are still there, I think they should be updated in a way we'll enable them to create various supported kafka sizes. 
Or we could think of removing them and ask the impacted users to start overriding the parameter.

JameelB:
@SamBarker what's the state of this PR? is it still needed?

SamBarker:
My original intent of trying to fix the default for the whole team is still needed but its clear this PR isn't the right way to do that.

I haven't found time to look at the suggested alternatives yet so I'll close this for now.

PR #1071, opened 2022-06-02
[WIP] Update billing model association logic to support new marketplaces

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-8599.

This PR updates the billing model assignment logic when a kafka instance is created to support additional billing models for new marketplaces.

It also contains a refactoring of some of some of the logic that was used to validate the billing information provided with the kafka request (billing_cloud_account_id and marketplace) against AMS.

The logic has become quite complex due to several assumptions and priority orders that are currently stablished.

The summary of the changes made and insights at the moment of creating the PR are:
* OCM Go sdk library has been updated to be able to use the new accepted billing models
* The `ValidateBillingInformation` method in the kafka service type and in the AMS quota service type has been renamed to `GetMarketplaceFromBillingAccountInformation`. Now, aside from doing the validation the calculated marketplace value from the Cloud Provider Account ID in AMS side is returned.
* In the kafka create handler, we now perform a call to `GetMarketplaceFromBillingAccountInformation` to set the calculated `marketplace` value (if it applies) in the kafka request when only the billing_cloud_account_id has been  provided so the marketplace value is stored in the database to avoid having to go to AMS multiple times to gather it. We are aware that calling `GetMarketplaceFromBillingAccountInformation` implies calling again AMS after validation. If performance becomes a problem we can think about optimizations around that (same case as with `ValidateKafkaPlan`)
* The `getAvailableBillingModelFromKafkaInstanceType` method has been updated to support the detection of AWS marketplace billing model. The billing model detection logic is now the following one:
  * If the marketplace attribute for the kafka instance is not empty, we check if there is quota for the AWS billing model for the provided billing_cloud_account_id. If there is then we return that billing model
  * If the marketplace attribute for the kafka instance is empty, the following logic applies:
    * If the user has standard quota defined and it has quota available then standard is used and returned
    * If the user does not have standard quota or has standard quota but without quota available then we check if there is quota available for the red hat marketplace billing model or for the aws marketplace billing model. If both billing models have quota available we treat that as an error as it shouldn't happen.
* Due to until now there was not an explicit reference to the Red Hat Marketplace type (it was just 'marketplace' value in AMS side) no value is stored for the `marketplace` kafka attribute in that case. We might need to change this in the future and it would imply (at least) performing migrations to keep compatibility

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -147,32 +147,26 @@ func (q amsQuotaService) getAvailableBillingModelFromKafkaInstanceType(externalI
 
 	billingModel := ""
 	matchedBillingModels := 0
+	foundBillingAccount := false
 	for _, qc := range quotaCosts {
-		foundBillingCloudAccount := false
-		if marketplace != "" {
+		if billingCloudAccountId != "" && marketplace != "" {
 			for _, ca := range qc.CloudAccounts() {
-				if billingCloudAccountId == ca.CloudAccountID() {
-					foundBillingCloudAccount = true
+				if billingCloudAccountId == ca.CloudAccountID() && ca.CloudProviderID() == marketplace {
+					foundBillingAccount = true
 					break
 				}
 			}
 		}
-		if foundBillingCloudAccount {
-			continue
-		}
 		for _, rr := range qc.RelatedResources() {
 			if qc.Consumed()+sizeRequired <= qc.Allowed() || rr.Cost() == 0 {
-				if marketplace != "" {
-					if marketplace == "aws" && rr.BillingModel() == string(amsv1.BillingModelMarketplaceAWS) {
-						return rr.BillingModel(), nil
-					}
-				} else {
-					if rr.BillingModel() == string(amsv1.BillingModelStandard) {
-						return rr.BillingModel(), nil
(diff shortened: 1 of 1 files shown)


pb82:
ping @pawelpaszki

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -143,18 +146,45 @@ func (q amsQuotaService) getAvailableBillingModelFromKafkaInstanceType(externalI
...
+				} else {
+					if rr.BillingModel() == string(amsv1.BillingModelStandard) {
+						return rr.BillingModel(), nil
+					} else if rr.BillingModel() == string(amsv1.BillingModelMarketplace) {
+						billingModel = rr.BillingModel()
+						matchedBillingModels++
+					} else if rr.BillingModel() == string(amsv1.BillingModelMarketplaceAWS) {
+						billingModel = rr.BillingModel()
+						matchedBillingModels++
+					}
```
pawelpaszki, line 178:
```suggestion
					} else if rr.BillingModel() == string(amsv1.BillingModelMarketplace) || rr.BillingModel() == string(amsv1.BillingModelMarketplaceAWS) {
						billingModel = rr.BillingModel()
						matchedBillingModels++
					}
```

pawelpaszki:
will be resolved here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1074

PR #1072, opened 2022-06-02
chore: add logging to redhatsso_service.go

JasonPowr:
## Description
This pr adds logging to redhatsso_service.go because Currently, no logging is implemented on the redhat sso service. This will help debug issues with sso integration

jira: https://issues.redhat.com/browse/MGDSTRM-8657

## Verification Steps
N/a

- [ ] Code Review completed

### DIFF
--- a/pkg/services/sso/redhatsso_service.go
+++ b/pkg/services/sso/redhatsso_service.go
@@ -10,6 +10,7 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/redhatsso"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
+	"github.com/golang/glog"
 	serviceaccountsclient "github.com/redhat-developer/app-services-sdk-go/serviceaccounts/apiv1internal/client"
 )
 
@@ -25,10 +26,12 @@ func (r *redhatssoService) RegisterClientInSSO(accessToken string, clusterId str
 }
 
 func (r *redhatssoService) DeRegisterClientInSSO(accessToken string, clientId string) *errors.ServiceError {
+	glog.V(5).Infof("Deregistering client with id: %s", clientId)
 	err := r.client.DeleteServiceAccount(accessToken, clientId)
 	if err != nil {
 		return errors.NewWithCause(errors.ErrorFailedToDeleteSSOClient, err, "failed to delete the sso client")
 	}
+	glog.V(5).Infof("Deregistered client with id: %s", clientId)
 	return nil
 }
 
@@ -41,6 +44,7 @@ func (r *redhatssoService) GetRealmConfig() *keycloak.KeycloakRealmConfig {
 }
 
 func (r *redhatssoService) IsKafkaClientExist(accessToken string, clientId string) *errors.ServiceError {
+	glog.V(5).Infof("Checking if client with id: %s exists", clientId)
 	_, found, err := r.client.GetServiceAccount(accessToken, clientId)
 	if err != nil {
 		return errors.NewWithCause(errors.ErrorFailedToGetSSOClient, err, "failed to get sso client with id: %s", clientId)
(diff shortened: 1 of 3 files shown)


ziccardi (changes requested):
Logs were removed from `redhatsso_service` on purpose
Users of both redhatsso and massso will pass through the `keycloakServiceProxy` proxy object (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d15bc373427e992a5b3ff2316fdad36a998e2845/pkg/services/sso/keycloak_service_proxy.go).

The idea is that this kind of logging should be put into the proxy, so they will be automatically inherited by all the implementation of `KeycloakService`.

I would suggest to move all this logs to the proxy and remove them from here too: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/398f0d1a50be9b58f83c4740064c5799f3a921ff/pkg/services/sso/keycloak_service.go

machi1990:
@ziccardi @JasonPowr @akoserwal I think it might be good to still have logs in the redhat sso client / services layers for things like token retrieval etc what do you think?

ziccardi:
> @ziccardi @JasonPowr @akoserwal I think it might be good to still have logs in the redhat sso client / services layers for things like token retrieval etc what do you think?

My 2 cents: if some log is specific for a given provider, then it should go there, otherwise should go into the proxy, to avoid polluting the implementations.

akoserwal:
+++++


> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d15bc373427e992a5b3ff2316fdad36a998e2845/pkg/services/sso/keycloak_service_proxy.go

+1, please create a Jira to add logging for clients

akoserwal:
> @ziccardi @JasonPowr @akoserwal I think it might be good to still have logs in the redhat sso client / services layers for things like token retrieval etc what do you think?

We can create follow up jiras

machi1990:
I've created a followup task as suggested by @akoserwal, I am happy for this to be progressed in its current state.

ziccardi:
> I think we need to log in to both services. it will critical during the migration. To get even minor details. both services react in a different way. we can't summarize on a layer above

If the log just logs the entry and the exit of the method (as the current log does), there is no added value in logging into the service itself.
OTH, if the log we are going to add will add **custom** detail for the **specific** implementation, then it will need to be added into the implementation.

akoserwal:
> > I think we need to log in to both services. it will critical during the migration. To get even minor details. both services react in a different way. we can't summarize on a layer above
> 
> If the log just logs the entry and the exit of the method (as the current log does), there is no added value in logging into the service itself. OTH, if the log we are going to add will add **custom** detail for the **specific** implementation, then it will need to be added into the implementation.

That's what we want **custom** detail for the **specific** implementation,"

pkg/services/sso/redhatsso_service.go
```diff
@@ -26,11 +26,12 @@ func (r *redhatssoService) RegisterClientInSSO(accessToken string, clusterId str
 }
 
 func (r *redhatssoService) DeRegisterClientInSSO(accessToken string, clientId string) *errors.ServiceError {
+	glog.V(5).Infof("Deregistering client with id: %s in redhat_sso", clientId)
```
ziccardi, line 29: This log is not adding any custom info for redhat_sso. It can be moved to the proxy:
```
glog.V(5).Infof("Deregistering client with id: %s in %s", clientId, r.GetConfig().SelectSSOProvider)
```

pkg/services/sso/redhatsso_service.go
```diff
@@ -26,11 +26,12 @@ func (r *redhatssoService) RegisterClientInSSO(accessToken string, clusterId str
...
 	err := r.client.DeleteServiceAccount(accessToken, clientId)
 	if err != nil {
 		return errors.NewWithCause(errors.ErrorFailedToDeleteSSOClient, err, "failed to delete the sso client")
 	}
-	glog.V(5).Infof("Kafka Client %s deleted successfully", clientId)
+	glog.V(5).Infof("Deregistered client with id: %s in redhat_sso", clientId)
```
ziccardi, line 34: see previous comment

pkg/services/sso/redhatsso_service.go
```diff
@@ -43,43 +44,51 @@ func (r *redhatssoService) GetRealmConfig() *keycloak.KeycloakRealmConfig {
...
 	if !found {
-		return errors.New(errors.ErrorNotFound, "sso client with id: %s not found", clientId)
+		return errors.New(errors.ErrorNotFound, "redhat_sso client with id: %s not found", clientId)
 	}
+	glog.V(5).Infof("redhat_sso client with id: %s found", clientId)
 	return nil
 }
 
 func (r *redhatssoService) CreateServiceAccount(accessToken string, serviceAccountRequest *api.ServiceAccountRequest, ctx context.Context) (*api.ServiceAccount, *errors.ServiceError) {
+	glog.V(5).Infof("Creating service account for redhat_sso")
```
ziccardi, line 61: see previous comment

pkg/services/sso/redhatsso_service.go
```diff
@@ -43,43 +44,51 @@ func (r *redhatssoService) GetRealmConfig() *keycloak.KeycloakRealmConfig {
...
 	serviceAccount, err := r.client.CreateServiceAccount(accessToken, serviceAccountRequest.Name, serviceAccountRequest.Description)
 	if err != nil {
 		return nil, errors.NewWithCause(errors.ErrorFailedToCreateServiceAccount, err, "failed to create service account")
 	}
+	glog.V(5).Infof("Service account for redhat_sso created")
```
ziccardi, line 66: see previous comment

pkg/services/sso/redhatsso_service.go
```diff
@@ -43,43 +44,51 @@ func (r *redhatssoService) GetRealmConfig() *keycloak.KeycloakRealmConfig {
...
 	return convertServiceAccountDataToAPIServiceAccount(&serviceAccount), nil
 }
 
 func (r *redhatssoService) DeleteServiceAccount(accessToken string, ctx context.Context, clientId string) *errors.ServiceError {
+	glog.V(5).Infof("Deleting redhat_sso service account with id: %s", clientId)
```
ziccardi, line 71: see previous comment

JasonPowr:
> I didn't check all the logs, but looks like most of them can be moved to the proxy without any loss of information (see my other comments)

It was decided by @akoserwal and @machi1990 to keep the logs in redhatsso_service.go and remove them from the proxy

ziccardi:
> It was decided by @akoserwal and @machi1990 to keep the logs in redhatsso_service.go and remove them from the proxy

@akoserwal @machi1990 Why do we want to move those logs to the service implementation if they don't provide any info that can't be logged from the proxy? By doing that:

1. we just pollute the service implementation code without any real benefit
2. we duplicate the log across all the implementations
3. we lose the advantage of using the proxy

ziccardi (approved):
We (me, @akoserwal and @machi1990 ) had a long discussion about this.
Here is the recap.

- Pros of having the log into the proxy:
  - Logs would be centralised into the proxy and automatically inherited by all the implementations of the service, without need to duplicate them into each service implementation
  - Being centralised into the proxy, the logs will be coherent among all the service implementations
  - Being the logs just logs like 'Going to do action xxx' and 'Action xxx resulted in error yyy' or similar, the proxy would be able to add a tag to each log so that will be easy to understand what is the service implementation related to log being printed
- Cons of having the log into the proxy:
  - The filename indicated into the logs will be the filename of the proxy implementation
  - The line number into the log will be the line number of a line into the proxy implementation
  - To identify what is the currently selected provider, we will have to look at the tag instead of having the filename

After analysing all the pros and cons, majority has decided that the pros are not so important to overrule the cons.

machi1990:
Thanks @ziccardi for the summary 

Let's get this in.

There is another issue about adding more debug logs to what we current have.

PR #1076, opened 2022-06-03
fix: remove size limit

akoserwal:
Removing size limit check

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/handlers/service_accounts.go
+++ b/internal/kafka/internal/handlers/service_accounts.go
@@ -30,10 +30,6 @@ func (s serviceAccountsHandler) ListServiceAccounts(w http.ResponseWriter, r *ht
 		Action: func() (interface{}, *errors.ServiceError) {
 			ctx := r.Context()
 			Page, Size := s.handleParams(r.URL.Query())
-			// required for redhat sso client, returns a bad request stating this field needs to be larger than 0
-			if Size == 0 {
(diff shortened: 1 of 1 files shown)


JameelB (commented):
@akoserwal this doesn't return a bad request anymore if it's 0 with rhsso client?

akoserwal:
@JameelB I'll create a jira to test it with rhsso client & testing. I think is it creating an issue with the listing service account.

JameelB (approved):
approving since we're not using rhsso client just yet.

ziccardi:
> @JameelB I'll create a jira to test it with rhsso client & testing. I think is it creating an issue with the listing service account.

I don't think it is related. The keycloak_service (massso) is using the `GetClients` method that returns the max allowed number of clients when max is 0: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/398f0d1a50be9b58f83c4740064c5799f3a921ff/pkg/services/sso/keycloak_service.go#L228

Doc: https://pkg.go.dev/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/keycloak#KcClient.GetClients

The fix is correct but, IMO, should be put here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3693437afa71e729154a76fd4fa26f28592b4a36/pkg/client/redhatsso/client.go#L147 to have in redhatsso the same behaviour we had previously with massso.

PR #1078, opened 2022-06-06
fix(connectors): fixes MGDCTRS-1029, fixed namespace status update to avoid updating row when the status is the same

dhirajsb:
## Description
fixes MGDCTRS-1029, fixed namespace status update to avoid updating row when the status is the same

## Verification Steps
Added integration test should pass. 

- [ ] Code Review completed

### DIFF
--- a/internal/connector/test/integration/features/connector-agent-api.feature
+++ b/internal/connector/test/integration/features/connector-agent-api.feature
@@ -251,7 +251,7 @@ Feature: connector agent API
     Then the response code should be 200
     And I store the ".resource_version" selection from the response as ${namespace_version}
 
-    # send a namespace update without status change
+    # send a namespace status update with the same phase, which shouldn't trigger a version change
     When I PUT path "/v1/agent/kafka_connector_clusters/${connector_cluster_id}/namespaces/${connector_namespace_id}/status" with json body:
       """
       {
@@ -262,6 +262,11 @@ Feature: connector agent API
         "conditions": [
           {
             "type": "Ready",
+            "status": "True",
+            "lastTransitionTime": "2018-01-01T00:00:00Z"
+          },
+          {
+            "type": "Dummy Condition",
             "status": "True",
             "lastTransitionTime": "2018-01-01T00:00:00Z"
           },
@@ -285,7 +290,7 @@ Feature: connector agent API
     Then the response code should be 204
     And the response should match ""
 
-    # check that namespace status updates with no change don't change resource version
+    # check that resource version didn't change
     When I GET path "/v1/agent/kafka_connector_clusters/${connector_cluster_id}/namespaces/${connector_namespace_id}"
     Then the response code should be 200
     And the ".resource_version" selection from the response should match "${namespace_version}"
--- a/internal/connector/test/integration/features/connector-agent-api.feature
+++ b/internal/connector/test/integration/features/connector-agent-api.feature
@@ -246,6 +246,50 @@ Feature: connector agent API
     Then the response code should be 204
     And the response should match ""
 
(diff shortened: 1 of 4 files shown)


internal/connector/test/integration/features/connector-agent-api.feature
```diff
@@ -246,6 +246,50 @@ Feature: connector agent API
...
+            "reason": "Testing2",
+            "message": "This is another test failure message"
+          }
+        ]
+      }
+      """
+    Then the response code should be 204
+    And the response should match ""
+
+    # check that namespace status updates with no change don't change resource version
```
lburgazzoli, line 288: Is this the correct expectation  ? 

When updating a ConnectorDeployment status, the resource version does not change so I think the same behavior should be expected here otherwise there's at least a waste poll cycle as the next poll cycle would return the same Namespace (the resource_version changes) and there's a potential endless loop in case the Namespace status update would include any change.

dhirajsb:
@lburgazzoli this changes namespace status update handling to avoid unnecessary revision updates.

internal/connector/test/integration/features/connector-agent-api.feature
dhirajsb: The test checks whether the actual resource version changed or not, which was the root cause of the problem. It's not polling for namespace updates in that test.

lburgazzoli: Yes but from what I understand from the test and the comments, it appears that the resource version is not updated only if the current status and the received status are the same.  SO I expect that, i.e. if a condition changes, then the resource version changes as well, correct ?

lburgazzoli: if any of the conditions changes i.e. a time stamp, then the resource version is bumped right ? 

unless conditions are not stored...

dhirajsb: Are you saying that `LastTransitionTime` constantly changes on every status update? I thought it only changes when the condition changes, as the name suggests. 

Conditions are stored in embedded status columns.

dhirajsb: I looked into Postgres docs and it seems like we could also modify the trigger to only update version when specific columns change. So we might not need to move the status columns out to avoid resource version update.

lburgazzoli: I'm not saying that condition do change all the time but there's no reason why the should not change after the initial namespace creation and in such case, it should not lead to an infinite loop nor in an additional useless updates.

Having the status update triggering a bump to the revision seems to impose limits in the evolution of the APIs i.e. I was thinking to include the current resource usage of a namespace in the status report which of course will change over time.

dhirajsb: I updated namespace resource version update trigger such that status version and condition updates won't cause revision updates. Unfortunately, it still doesn't get rid of agent phase updates (to `ready` and `deleted`) from causing a resource version update. 

We can open another issue to introduce a new `desired_state` property in cluster and namespace, but it's a much bigger change.

lburgazzoli: > We can open another issue to introduce a new `desired_state` property in cluster and namespace, but it's a much bigger change.

I think this is much needed as we got a number of issues related to the lack of a `desired_state` and the mix of ownership already.

dhirajsb: Can we merge this patch for fixing the huge number of unnecessary version updates right now and work on the bigger fix later?

PR #1085, opened 2022-06-07
test: kafka/handlers admin kafkas and cloud providers unit tests

pawelpaszki:
## Description
test: kafka/handlers admin kafkas and cloud providers unit tests

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/handlers/admin_kafka.go
+++ b/internal/kafka/internal/handlers/admin_kafka.go
@@ -20,14 +20,14 @@ import (
 )
 
 type adminKafkaHandler struct {
-	service        services.KafkaService
+	kafkaService   services.KafkaService
 	accountService account.AccountService
 	providerConfig *config.ProviderConfig
 }
 
-func NewAdminKafkaHandler(service services.KafkaService, accountService account.AccountService, providerConfig *config.ProviderConfig) *adminKafkaHandler {
+func NewAdminKafkaHandler(kafkaService services.KafkaService, accountService account.AccountService, providerConfig *config.ProviderConfig) *adminKafkaHandler {
 	return &adminKafkaHandler{
-		service:        service,
+		kafkaService:   kafkaService,
 		accountService: accountService,
 		providerConfig: providerConfig,
 	}
@@ -38,7 +38,7 @@ func (h adminKafkaHandler) Get(w http.ResponseWriter, r *http.Request) {
 		Action: func() (i interface{}, serviceError *errors.ServiceError) {
 			id := mux.Vars(r)["id"]
 			ctx := r.Context()
-			kafkaRequest, err := h.service.Get(ctx, id)
+			kafkaRequest, err := h.kafkaService.Get(ctx, id)
 			if err != nil {
 				return nil, err
 			}
@@ -59,7 +59,7 @@ func (h adminKafkaHandler) List(w http.ResponseWriter, r *http.Request) {
 				return nil, errors.NewWithCause(errors.ErrorMalformedRequest, err, "Unable to list kafka requests: %s", err.Error())
 			}
 
-			kafkaRequests, paging, err := h.service.List(ctx, listArgs)
+			kafkaRequests, paging, err := h.kafkaService.List(ctx, listArgs)
 			if err != nil {
 				return nil, err
 			}
@@ -96,7 +96,7 @@ func (h adminKafkaHandler) Delete(w http.ResponseWriter, r *http.Request) {
 			id := mux.Vars(r)["id"]
 			ctx := r.Context()
 
-			err := h.service.RegisterKafkaDeprovisionJob(ctx, id)
+			err := h.kafkaService.RegisterKafkaDeprovisionJob(ctx, id)
 			return nil, err
 		},
 	}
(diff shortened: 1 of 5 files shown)


internal/kafka/internal/handlers/admin_kafka_test.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	RegisterTestingT(t)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			h := NewAdminKafkaHandler(tt.fields.kafkaService, tt.fields.accountService, tt.fields.providerConfig)
+			req, rw := GetHandlerParams("GET", "/{id}", nil)
+			h.Get(rw, req)
+			Expect(rw.Result().StatusCode).To(Equal(tt.wantStatusCode))
```
ziccardi, line 74: `rw.Result().Body` must be closed

internal/kafka/internal/handlers/admin_kafka_test.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	RegisterTestingT(t)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			h := NewAdminKafkaHandler(tt.fields.kafkaService, tt.fields.accountService, tt.fields.providerConfig)
+			req, rw := GetHandlerParams("GET", tt.args.url, nil)
+			h.List(rw, req)
+			Expect(rw.Result().StatusCode).To(Equal(tt.wantStatusCode))
```
ziccardi, line 160: `rw.Result().Body` must be closed

internal/kafka/internal/handlers/admin_kafka_test.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	RegisterTestingT(t)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			h := NewAdminKafkaHandler(tt.fields.kafkaService, tt.fields.accountService, tt.fields.providerConfig)
+			req, rw := GetHandlerParams("DELETE", tt.args.url, nil)
+			h.Delete(rw, req)
+			Expect(rw.Result().StatusCode).To(Equal(tt.wantStatusCode))
```
ziccardi, line 212: `rw.Result().Body` must be closed

internal/kafka/internal/handlers/admin_kafka_test.go
```diff
@@ -0,0 +1,389 @@
...
+			},
+			wantStatusCode: http.StatusBadRequest,
+		},
+	}
+
+	RegisterTestingT(t)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			h := NewAdminKafkaHandler(tt.fields.kafkaService, tt.fields.accountService, tt.fields.providerConfig)
```
ziccardi, line 209: `tt` is a range variable. It is reused on each iteration. If we run the tests in parallel the value of `tt` will be unpredictable. Better to copy `tt` and use the copy from inside the function.

pawelpaszki: does it apply to all test cases? btw. I don't think we are running the tests in parallel

pawelpaszki: do you think this should be a separate Jira to do it, including adding documentation?

ziccardi: Yeah, I think so. 
I added a point to the tech meeting about adding some linters that will check for these errors.
When we will add the linter, we will have to fix all of these issues as well.

pawelpaszki: ok. thanks. I will push the changes for the other comments shortly

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -20,14 +20,14 @@ import (
...
 
 type adminKafkaHandler struct {
-	service        services.KafkaService
+	kafkaService   services.KafkaService
 	accountService account.AccountService
 	providerConfig *config.ProviderConfig
 }
 
-func NewAdminKafkaHandler(service services.KafkaService, accountService account.AccountService, providerConfig *config.ProviderConfig) *adminKafkaHandler {
+func NewAdminKafkaHandler(kafkaService services.KafkaService, accountService account.AccountService, providerConfig *config.ProviderConfig) *adminKafkaHandler {
```
ziccardi, line 28: This is not related to this PR, but why an exported function (NewAdminKafkaHandler) is returning a non exported type (adminKafkaHandler)?

pawelpaszki: not sure. I can change it if you think it should be exported

ziccardi: I don't think it makes sense that an exported function returns a non exported value: how can code calling this function from outside the package deal with the return value?

I believe visibility should be coherent between functions and return values.

@machi1990 @pawelpaszki  WDYT?

pawelpaszki: yes makes sense to me. @machi1990 - if you agree, I will make this change

machi1990: I think @ziccardi argument makes sense. But I'll defer from doing the change in this PR and create a follow issue to address this, plus do the same for the other handlers as well if it is needed: so far, the handlers have not been really exported, and the route_loader.go (the only place where the handlers are being used) is able to interact with the handlers without any issue.

PR #1088, opened 2022-06-08
ci: added golangci action and enabled bodyclose linter

ziccardi:

- [ ] Code Review completed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,7 @@ on:
       - 'docs/**'
       - 'pkg/api/openapi/docs/**'
       - 'pkg/api/openapi/.openapi-generator-ignore'
-      
+
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     paths-ignore:
@@ -50,7 +50,7 @@ env:
   MAS_SSO_CLIENT_SECRET: kas-fleet-manager
   OSD_IDP_MAS_SSO_CLIENT_ID: kas-fleet-manager
   OSD_IDP_MAS_SSO_CLIENT_SECRET: kas-fleet-manager
-  # Dummy AWS credentials 
+  # Dummy AWS credentials
   AWS_ACCOUNT_ID: aws_accountid
   AWS_ACCESS_KEY: aws_accesskey
   AWS_SECRET_ACCESS_KEY: aws_secretaccesskey
@@ -104,7 +104,11 @@ jobs:
         with:
           go-version: "1.17"
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+      - name: Run Golang-ci linters
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.29
       - name: Setup Keycloak realm config
         run: ./scripts/mas_sso_config.sh
       - name: Cache go module
@@ -125,7 +129,7 @@ jobs:
         run: |
           export GOPATH=$(go env GOPATH)
           export PATH=${PATH}:$GOPATH/bin
-          make verify lint binary test test/integration
+          make verify binary test test/integration
         timeout-minutes: 14
       - name: Upload unit tests code coverage to codecov
         uses: codecov/codecov-action@v2
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,3 +8,6 @@ run:
     - internal/connector/internal/api/admin/private
   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 10m
+linters:
+  enable:
+    - bodyclose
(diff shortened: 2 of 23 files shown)


.github/workflows/ci.yaml
```diff
@@ -125,7 +129,7 @@ jobs:
         run: |
           export GOPATH=$(go env GOPATH)
           export PATH=${PATH}:$GOPATH/bin
-          make verify lint binary test test/integration
+          make verify binary test test/integration
```
machi1990, line 132: Why the move from using `lint` target to using the github action?

.golangci.yml
```diff
@@ -8,3 +8,6 @@ run:
     - internal/connector/internal/api/admin/private
   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 10m
+linters:
+  enable:
+    - bodyclose
```
machi1990, line 13: Is it possible to add more linting rules e.g on using loop variables without pinning them? Or that's not possible?

.github/workflows/ci.yaml
ziccardi: The lint target still works and can be used on the CLI, I though that the github action was a better choice because it is integrated with GITHUB and it is easier/faster by looking at the `File changed` tab to see where the violation is.

Do you see any drawback in using the github action?

.golangci.yml
ziccardi: I can give a look to see if there is any other option, but ATM I see only 2 options:

1. put here `enable-all: true` to enable all the linters
2. list all the linters you want to enable under `enable` as in this PR

.github/workflows/ci.yaml
machi1990: No, I do not see any major drawback.
We might have to remember updating the version so that it aligns to whatever version we are using locally when a change happen. 
I agree seeing the violations directly in the UI  is better than having to see the logs. But ideally, that shouldn't be needed if we (the team) during dev, proactively make sure that the linting checks are intact.

.golangci.yml
machi1990: Thanks, let me know if you find something. 

I'd say option 2 is better since we are a bit in control of want we want to check and we'll be progressively enabling some lints as we go. 
Option 1 might be too hard to make happen due to the massive errors we'll see.

For a start, I was only thinking of enabling the linting so that we fail if someone tries to re-use a loop variable in a way that's likely to cause issues.

machi1990: Thanks.

internal/kafka/test/common/kafka_pollers.go
```diff
@@ -61,6 +64,7 @@ func WaitForKafkaCreateToBeAccepted(ctx context.Context, db *db.ConnectionFactor
 			}
 		}).
 		OnRetry(func(attempt int, maxRetries int) (done bool, err error) {
+			//nolint: bodyclose
```
machi1990, line 67: Is there a reason why we are ignoring lint here?

internal/kafka/test/common/kafka_pollers.go
```diff
@@ -87,11 +91,16 @@ func WaitForKafkaToReachStatus(ctx context.Context, db *db.ConnectionFactory, cl
 			}
 		}).
 		OnRetry(func(attempt int, maxRetries int) (done bool, err error) {
-			kafka, _, err = client.DefaultApi.GetKafkaById(ctx, kafkaId)
+			k, resp, err := client.DefaultApi.GetKafkaById(ctx, kafkaId)
```
machi1990, line 94: why `k`, couldn't `kafka` be used?

.github/workflows/ci.yaml
machi1990: @ziccardi following on to what I said above , it'll be good to have linting on pre-commit hook: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/.githooks/pre-commit. What do you think? Is there a task for this?

internal/kafka/test/common/kafka_pollers.go
ziccardi: I had to replace it with `k` because now I had to use `:=` to create the `resp` variable.
By doing that, however, `kafka` become a new variable in the inner scope and was hiding the outer scope ones.

internal/kafka/test/common/kafka_pollers.go
machi1990: Thanks. It'll be good to document the `WaitForKafkaCreateToBeAccepted` method as such. Current it does not say anything about this.

internal/kafka/test/common/kafka_pollers.go
machi1990: Thanks.

.github/workflows/ci.yaml
ziccardi: I will add a task for that

machi1990 (approved):

machi1990:
Thanks @ziccardi for this

PR #1097, opened 2022-06-15
fix: support rhm and azure cloud providers and check quota on marketp…

pb82:
Support `rhm` and `azure` cloud providers. Also check Quota for marketplace billing models.

Verification steps:

1. Additional unit tests should pass
2. Creation of rhm marketplace instances should succeed when using an organization that has a cloud account linked to it

ping @davidffrench

### DIFF
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -2,12 +2,12 @@ package quota
 
 import (
 	"fmt"
-
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/kafkas/types"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/ocm"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 )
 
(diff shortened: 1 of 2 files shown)


internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -162,7 +179,7 @@ func (q amsQuotaService) getAvailableBillingModelFromKafkaInstanceType(externalI
 	return billingModel, nil
 }
 
-func (q amsQuotaService) getBillingModel(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (string, error) {
+func (q amsQuotaService) GetBillingModel(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (string, error) {
```
machi1990, line 182: Is this change needed?

machi1990: Thanks @pb82 since it is a unit test within the package it should be okay to still keep it a private and test it, what do you think?

machi1990:
Left one minor comment but my knowledge in this are pretty basic so I'll let others review and approve the functionality.
Ping @miguelsorianod and @pawelpaszki as they worked on this recent.

internal/kafka/internal/services/quota/ams_quota_service.go
pb82: you're absolutely correct @machi1990 , i've updated this now.

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -48,6 +48,407 @@ var ocmClientMockWithCloudAccounts = &ocm.ClientMock{
...
+				},
+				kafkaConfig: &defaultKafkaConf,
+			},
+			wantErr:          false,
+			wantBillingModel: string(v1.BillingModelMarketplaceAWS),
+		},
+	}
+
+	RegisterTestingT(t)
+	for _, tt := range tests {
```
pawelpaszki, line 439: I think the same pattern as here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1089/files#diff-ace4f53881679319ea877b48a99b7466398f3187ef68ae0825bb1ba233632168R49 needs to be followed for the `tt` in the tests

pawelpaszki:
one comment added. otherwise LGTM

internal/kafka/internal/services/quota/ams_quota_service_test.go
pb82: good catch, fixed

PR #1100, opened 2022-06-16
test(integration): review instance type test to make it pass for the right reasons

machi1990:
## Description

1. The way the test was setup e.g cluster ids, made the test pass most of the times since the cluster ids contained numbers which made the metric checker util to pass all the time : the checker uses `.Contains()`. 
2. Rewrites the metric checker a bit because the way the checker was written seemed suspicious. Notably https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/b08ee74bbe1c807f52646c59ed3015bd9e122f2e/internal/kafka/test/common/util.go#L245-L255 as it increased the count of found items while the item was not found: see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/b08ee74bbe1c807f52646c59ed3015bd9e122f2e/internal/kafka/test/common/util.go#L247-L249 

## Verification Steps

The integration test should continue to pass

## Checklist (Definition of Done)

- ~~[ ] All acceptance criteria specified in JIRA have been completed~~
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- ~~[ ] Documentation added for the feature~~
- [x] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- ~~[ ] Required metrics/dashboards/alerts have been added (or PR created).~~
- ~~[ ] Required Standard Operating Procedure (SOP) is added.~~
- ~~[ ] JIRA has been created for changes required on the client side~~

PR #1101, opened 2022-06-16
fix(metrics): cluster_status_capacity_used doesn't update when kafka count goes down to 0

machi1990:
## Description

Fixes https://issues.redhat.com/browse/MGDSTRM-8866

This depends on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1100 for the test to pass, so I'll open it as draft. 

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"sync"
 
+	apiErrors "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services/sso"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	"gorm.io/gorm"
@@ -89,7 +90,8 @@ type KafkaService interface {
 	DeprovisionKafkaForUsers(users []string) *errors.ServiceError
 	DeprovisionExpiredKafkas() *errors.ServiceError
 	CountByStatus(status []constants2.KafkaStatus) ([]KafkaStatusCount, error)
-	CountByRegionAndInstanceType() ([]KafkaRegionCount, error)
+	// CountStreamingUnitByRegionAndInstanceType returns streaming unit counts per regions and instance type
+	CountStreamingUnitByRegionAndInstanceType() ([]KafkaStreamingUnitCountPerRegion, error)
 	ListKafkasWithRoutesNotCreated() ([]*dbapi.KafkaRequest, *errors.ServiceError)
 	VerifyAndUpdateKafkaAdmin(ctx context.Context, kafkaRequest *dbapi.KafkaRequest) *errors.ServiceError
 	ListComponentVersions() ([]KafkaComponentVersions, error)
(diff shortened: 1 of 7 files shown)


internal/kafka/internal/services/kafka.go
```diff
@@ -960,52 +962,107 @@ type KafkaStatusCount struct {
...
-	ClusterId     string `gorm:"column:cluster_id"`
+	InstanceType  string
+	ClusterId     string
 	Count         float64
-	CloudProvider string `gorm:"column:cloud_provider"`
+	CloudProvider string
 }
 
-func (k *kafkaService) CountByRegionAndInstanceType() ([]KafkaRegionCount, error) {
+func (kafkaStreamingUnitCountPerRegion KafkaStreamingUnitCountPerRegion) isSame(kafkaPerRegionFromDB *KafkaPerRegionCount) bool {
```
pawelpaszki, line 973: maybe `isSameInstanceType` would make more sene here?

internal/kafka/internal/services/kafka.go
```diff
@@ -960,52 +962,107 @@ type KafkaStatusCount struct {
...
+	var clusters []*ClusterSelection
 	dbConn := k.connectionFactory.New()
+	if err := dbConn.Model(&api.Cluster{}).Scan(&clusters).Error; err != nil {
+		return nil, apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to list clusters")
+	}
 
-	var kafkas []*dbapi.KafkaRequest
+	streamingUnitsCountPerRegion := []KafkaStreamingUnitCountPerRegion{}
+
+	// pre-populate regions cound with zero cound values.
```
pawelpaszki, line 1007:
```suggestion
	// pre-populate regions count with zero count values.
```

internal/kafka/internal/services/kafka.go
machi1990: It checks for several things, instance type being one of them.

pawelpaszki (approved):

internal/kafka/internal/services/kafka.go
```diff
@@ -960,52 +962,107 @@ type KafkaStatusCount struct {
...
+		streamingUnitCount := float64(instSize.CapacityConsumed) * kafkaCountPerRegion.Count
+		for i, streamingUnitCountPerRegion := range streamingUnitsCountPerRegion {
+			if streamingUnitCountPerRegion.isSame(kafkaCountPerRegion) {
+				streamingUnitsCountPerRegion[i].Count += streamingUnitCount
 				resultPresent = true
+				break
 			}
 		}
+
 		if !resultPresent {
```
machi1990, line 1053: This might be useful when a kafka is not assigned to a cluster yet (in the dynamic scaling mode).
The metric will be further revisited then as well if needed for now I left the condition as it was before.

internal/kafka/internal/services/kafka.go
```diff
@@ -960,52 +962,107 @@ type KafkaStatusCount struct {
...
-	CloudProvider string `gorm:"column:cloud_provider"`
+	CloudProvider string
 }
 
-func (k *kafkaService) CountByRegionAndInstanceType() ([]KafkaRegionCount, error) {
+func (kafkaStreamingUnitCountPerRegion KafkaStreamingUnitCountPerRegion) isSame(kafkaPerRegionFromDB *KafkaPerRegionCount) bool {
+	return kafkaStreamingUnitCountPerRegion.CloudProvider == kafkaPerRegionFromDB.CloudProvider &&
+		kafkaStreamingUnitCountPerRegion.ClusterId == kafkaPerRegionFromDB.ClusterId &&
+		kafkaStreamingUnitCountPerRegion.InstanceType == kafkaPerRegionFromDB.InstanceType &&
+		kafkaStreamingUnitCountPerRegion.Region == kafkaPerRegionFromDB.Region
```
JameelB, line 977: Do we need to check cloud provider and region? would `ClusterId` + `InstanceType` be sufficient? `ClusterId` should be unique and `InstanceType` can be checked if the cluster supports multiple instances types :thinking:

internal/kafka/internal/services/kafka.go
```diff
@@ -960,52 +962,107 @@ type KafkaStatusCount struct {
...
+	ClusterID             string
+	Region                string
+	SupportedInstanceType string
+}
+
+func (k *kafkaService) CountStreamingUnitByRegionAndInstanceType() ([]KafkaStreamingUnitCountPerRegion, error) {
+
+	var clusters []*ClusterSelection
 	dbConn := k.connectionFactory.New()
+	if err := dbConn.Model(&api.Cluster{}).Scan(&clusters).Error; err != nil {
```
JameelB, line 1001: Will there be a chance where `ClusterID` may be empty when using dynamic scaling? :thinking:

internal/kafka/internal/services/kafka.go
JameelB: Can we add a comment about that here just as a reminder :)

internal/kafka/internal/services/kafka_test.go
```diff
@@ -2204,32 +2205,112 @@ func Test_KafkaService_CountByStatus(t *testing.T) {
...
+							Id:               "x1",
+							QuotaConsumed:    1,
+							CapacityConsumed: 1,
+						},
+						{
+							Id:               "x2",
+							QuotaConsumed:    2,
+							CapacityConsumed: 2,
+						},
+					},
```
JameelB, line 2229: We've got mocks for these, can we use them here? https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/mocks/supported_instance_types/kafka_instance_size.go

internal/kafka/internal/services/kafka_test.go
```diff
@@ -2204,32 +2205,112 @@ func Test_KafkaService_CountByStatus(t *testing.T) {
...
+				mocket.Catcher.NewMock().
+					WithQuery(`SELECT * FROM "clusters"`).
+					WithReply([]map[string]interface{}{})
+
+				mocket.Catcher.NewMock().WithQueryException().WithExecException()
+			},
+			want: nil,
+		},
+		{
+			name: "should return an empty list when there is no kafka",
```
JameelB, line 2296:
```suggestion
			name: "should return an empty list when there are no data plane clusters",
```

internal/kafka/test/integration/cluster_metrics_test.go
```diff
@@ -0,0 +1,94 @@
...
+		Region:                "us-east-1",
+		CloudProvider:         "aws",
+		Status:                api.ClusterReady,
+		IdentityProviderID:    "some-id",
+		ClusterDNS:            "some-cluster-dns",
+		ProviderType:          api.ClusterProviderStandalone,
+		SupportedInstanceType: "standard",
+		ClientID:              "some-client-id",
+		ClientSecret:          "some-client-secret",
+	}
```
JameelB, line 49: Can we build this using the cluster? https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/mocks/clusters/clusters.go

internal/kafka/test/integration/cluster_metrics_test.go
```diff
@@ -0,0 +1,94 @@
...
+		Status:                  constants.KafkaRequestStatusReady.String(),
+		ClusterID:               cluster.ClusterID,
+		ActualKafkaVersion:      "2.6.0",
+		DesiredKafkaVersion:     "2.6.0",
+		ActualStrimziVersion:    "v.23.0",
+		DesiredStrimziVersion:   "v0.23.0",
+		InstanceType:            types.STANDARD.String(),
+		ReauthenticationEnabled: false,
+		SizeId:                  "x1",
+	}
```
JameelB, line 70: Can we build this using the kafka mocks? https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/mocks/kafkas/kafkas.go

internal/kafka/test/integration/cluster_metrics_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if err := db.Create(kafka).Error; err != nil {
+		Expect(err).NotTo(HaveOccurred())
+	}
```
JameelB, line 74:
```suggestion
	err := db.Create(kafka).Error
	Expect(err).NotTo(HaveOccurred())
```

internal/kafka/test/integration/cluster_metrics_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if err := db.Create(cluster).Error; err != nil {
+		Expect(err).NotTo(HaveOccurred())
+	}
```
JameelB, line 78:
```suggestion
	err := db.Create(cluster).Error
	Expect(err).NotTo(HaveOccurred())
```

internal/kafka/test/integration/cluster_metrics_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// check that there is "1" streaming unit capacity used
+	metricValue := "1"
+	checkMetricsError := common.WaitForMetricToBePresent(h, t, metrics.ClusterStatusCapacityUsed, metricValue, kafka.InstanceType, kafka.ClusterID, kafka.Region, kafka.CloudProvider)
+	Expect(checkMetricsError).NotTo(HaveOccurred())
+
+	// now delete the kafka and make verify that the capacity used goes down from "1" to "0"
+	if err := db.Unscoped().Delete(kafka).Error; err != nil {
+		Expect(err).NotTo(HaveOccurred())
+	}
```
JameelB, line 88:
```suggestion
	err := db.Unscoped().Delete(kafka).Error
	Expect(err).NotTo(HaveOccurred())
```

internal/kafka/internal/services/kafka.go
machi1990: That's true, especially if the cluster has just been accepted. 

We could ignore accepted clusters / clusters that do not have the cluster id yet, but I'll prefer to leave it when we've dynamic scaling in full flow. What do you think?

internal/kafka/internal/services/kafka.go
machi1990: I think cluster Id and instance type should be enough.
I kept the same condition as the one which was there previously

internal/kafka/test/integration/cluster_metrics_test.go
machi1990: Thanks, I was looking at https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/mocks/clusters/clusters.go#L68 and I feel like I missing something. 
Since I need to control most of the things e.g clientIDs, identity provider id, I need to pass it a `modifyFn` which takes in the `*api.cluster` which'll have to built the same way I am doing now. So why go that extra step while I can just do like what I am doing which is the same as passing in the `modifyFn`?

internal/kafka/test/integration/cluster_metrics_test.go
machi1990: same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1101#discussion_r900283720

internal/kafka/internal/services/kafka.go
JameelB: thanks and yep I agree :+1:

internal/kafka/test/integration/cluster_metrics_test.go
JameelB: For me it should allow us to be consistent with a mock object for these tests rather than using different values for things like ids, etc. If we need to add new fields to a struct, we could also just add it to that mock without having to update each tests individually. 

Does it need to be specific values for those fields or can it be some standard default values? :thinking:

internal/kafka/internal/services/kafka_test.go
machi1990: I think the `github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/mocks/supported_instance_types` package isn't importable in the `services` package.

internal/kafka/test/integration/cluster_metrics_test.go
machi1990: The can be any values, I care for the values to be there so avoid for the secrets to be auto created. 

There is something about the build method which I do not understand i.e the modifyFn to take in something that it is building. If someone can pass in the `api.cluster`, then there is no really need to call the builder at all since they know how to build it.

internal/kafka/internal/services/kafka_test.go
JameelB: should be able to import it like this: 
`mocksupportedinstancetypes "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test/mocks/supported_instance_types"`

internal/kafka/test/integration/cluster_metrics_test.go
JameelB: Ah just noticed those are missing in that one. The Cluster mock may be a bad example :thinking: I think the goal was to have a builder to something like this: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/mocks/kafkas/kafkas.go#L152

@ziccardi may be able to fill the gaps here? :)

internal/kafka/internal/services/kafka_test.go
machi1990: I guess I am importing the wrong package :(

JameelB: it was the `/blob/main` bit that was causing issues :)

JameelB (approved):
Verified locally and works as expected. 

Comments above around refactoring the cluster/kafka object builder is an optional enhancement and not a blocker for me as it may be covered in this issue: https://issues.redhat.com/browse/MGDSTRM-7993

internal/kafka/test/integration/cluster_metrics_test.go
machi1990: Thanks. The method was confusing as well as I was reading it wrong.
Having a signature like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/mocks/kafkas/kafkas.go#L152 would be best

internal/kafka/internal/services/kafka_test.go
machi1990: Thanks. 

Yes, it was that.

internal/kafka/test/integration/cluster_metrics_test.go
machi1990: done

machi1990:
> Verified locally and works as expected.
> 
> Comments above around refactoring the cluster/kafka object builder is an optional enhancement and not a blocker for me as it may be covered in this issue: https://issues.redhat.com/browse/MGDSTRM-7993

Thanks @JameelB I've added the builder as well.

PR #1106, opened 2022-06-17
Remove unused jsonstructtags from configs

miguelsorianod:
## Description
Some configuration types in the internal kafka config package had json struct tags but we didn't use those types to serialize/deserialize JSON.

This PR removes the json struct tags for them.

Aside from that, the `Provider` type in the config package had json struct tags but the content of the providers configuration file is expected to be in `yaml` so it was not correct. The PR also changes that type to use `yaml` struct tags. Even if this was not correct it still worked because the Go json serializer/deserializer uses the name of the struct attribute in lowercase as a corresponding json key, and the keys that were specified had the same name (in lowercase).

## Verification Steps

Verify kas fleet manager still works and returns correct information from the endpoints.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/config/aws.go
+++ b/internal/kafka/internal/config/aws.go
@@ -7,18 +7,18 @@ import (
 
 type AWSConfig struct {
 	// Used for OSD Cluster creation with OCM
-	AccountID           string `json:"account_id"`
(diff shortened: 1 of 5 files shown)


machi1990 (approved):
lgtm, thanks @miguelsorianod

PR #1107, opened 2022-06-17
feat: made sso configurable via params

valeriiashapoval:
## Description
(https://issues.redhat.com/browse/MGDSTRM-8565)

## Verification Steps
make sso/setup and make sso/config should be executed both with and without any parameters to see if it works both ways

## Checklist 
- [x] Relevant tests are passing


scripts/mas_sso_setup.sh
```diff
@@ -7,9 +7,8 @@ docker network create mas-sso-network || true
 docker run \
   --name=mas-sso \
   --net mas-sso-network \
-  -p 8180:8080 \
+  -p $KEYCLOAK_PORT_NO:8080 \
   -e DB_VENDOR=h2  \
-  -e KEYCLOAK_USER=admin \
-  -e KEYCLOAK_PASSWORD=admin \
+  -e KEYCLOAK_USER=$KEYCLOAK_USER \
+  -e KEYCLOAK_PASSWORD=$KEYCLOAK_USER \
```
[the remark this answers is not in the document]
miguelsorianod: Another comment about style:
In bash you can also reference variables using ${_varname_} which is tend to prefer due to it protects you in scenarios like when you want to append/prepend content to a var name like with:
```
${myvarname}foo # adds foo to the result of the content of varname
```

scripts/mas_sso_config.sh
```diff
@@ -1,12 +1,10 @@
...
 
-KEYCLOAK_URL=http://localhost:8180
 TOKEN_PATH="/auth/realms/master/protocol/openid-connect/token"
-REALM=rhoas
 
 CLIENT_ID=admin-cli
-USERNAME=admin
-PASS=admin
+
+URL=$KEYCLOAK_URL:$KEYCLOAK_PORT_NO
```
machi1990, line 7: I'd suggest we have a variable that takes in the complete URL and not the host and port.
The advantage of having that is, for a deployed instance more often you do not need the port, just the host is enough. 
```suggestion
URL=$KEYCLOAK_URL
```

where KEYCLOAK_URL is a complete url, with host and port e.g `http://localhost:8180`

pawelpaszki: the port number is also used in the `scripts/mas_sso_setup.sh` script. it might become inconsistent, when we provide full url

pawelpaszki: @machi1990 ^^

machi1990: > the port number is also used in the `scripts/mas_sso_setup.sh` script. it might become inconsistent, when we provide full url

Yes, I think that's only where the port is used and I do not think the url is needed there.
It is okay to have not the same number of variables since it is for different make targets. 

I'll keep the port  to configure `sso/setup` target and `url` to configure the `sso/config` target.
The rest of the variables, username/password can be shared between the two targets.

scripts/mas_sso_config.sh
```diff
@@ -15,23 +13,23 @@ CONNECTOR_OPERATOR_ROLE=connector_fleetshard_operator
 ERR_MESSAGE="Keycloak server not running at localhost:8180. No realm configuration will be applied"
 timeout 120 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8180/auth/realms/master)" != "200" ]]; do echo "Waiting for keycloak server at localhost:8180"; sleep 10; done' || echo $ERR_MESSAGE
 
-RESULT=`curl -sk --data "grant_type=password&client_id=$CLIENT_ID&username=$USERNAME&password=$PASS" $KEYCLOAK_URL$TOKEN_PATH`
+RESULT=`curl -sk --data "grant_type=password&client_id=$CLIENT_ID&username=$KEYCLOAK_USER&password=$KEYCLOAK_PASSWORD" $URL$TOKEN_PATH`
```
ziccardi, line 16: The `$(...)` syntax should be preferred to the old backticks syntax (see [this](http://mywiki.wooledge.org/BashFAQ/082) and [this](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03))
```suggestion
RESULT=$(curl -sk --data "grant_type=password&client_id=$CLIENT_ID&username=$KEYCLOAK_USER&password=$KEYCLOAK_PASSWORD" $URL$TOKEN_PATH)
```

scripts/mas_sso_config.sh
```diff
@@ -15,23 +13,23 @@ CONNECTOR_OPERATOR_ROLE=connector_fleetshard_operator
...
 TOKEN=$(jq -r '.access_token' <<< $RESULT)
 echo $TOKEN
 
-CREATE_REALM_RHOAS=`curl -sk --data-raw '{"enabled":true,"id":"rhoas","realm":"rhoas"}' --header "Content-Type: application/json" --header "Authorization: Bearer $TOKEN" $KEYCLOAK_URL/auth/admin/realms`
+CREATE_REALM_RHOAS=`curl -sk --data-raw '{"enabled":true,"id":"rhoas","realm":"rhoas"}' --header "Content-Type: application/json" --header "Authorization: Bearer $TOKEN" $URL/auth/admin/realms`
```
ziccardi, line 20: See previous comment

scripts/mas_sso_config.sh
```diff
@@ -15,23 +13,23 @@ CONNECTOR_OPERATOR_ROLE=connector_fleetshard_operator
...
 echo $CREATE_REALM_RHOAS
 echo "Realm rhoas"
 
-CREATE_REALM_RHOAS_SRE=`curl -sk --data-raw '{"enabled":true,"id":"rhoas-kafka-sre","realm":"rhoas-kafka-sre"}' --header "Content-Type: application/json" --header "Authorization: Bearer $TOKEN" $KEYCLOAK_URL/auth/admin/realms`
+CREATE_REALM_RHOAS_SRE=`curl -sk --data-raw '{"enabled":true,"id":"rhoas-kafka-sre","realm":"rhoas-kafka-sre"}' --header "Content-Type: application/json" --header "Authorization: Bearer $TOKEN" $URL/auth/admin/realms`
```
ziccardi, line 24: See previous comment

ziccardi (commented):
I added a couple of comments about using `$(...)` syntax instead of the legacy backticks syntax. I didn't comment all the occurrence of the old syntax, but I believe all of them should be updated to the `$(...)` syntax.

Reasons can be found [here](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03) and [here](http://mywiki.wooledge.org/BashFAQ/082)

scripts/mas_sso_config.sh
pawelpaszki: @ziccardi - I saw that the whole script uses old syntax. If you have shellcheck plugin installed, it will complain about many syntax issues. do you think this should be addressed as a part of this PR?

scripts/mas_sso_setup.sh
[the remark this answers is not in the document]
valeriiashapoval: > Another comment about style: In bash you can also reference variables using ${_varname_} which I tend to prefer due to it protects you in scenarios like when you want to append/prepend content to a var name like with:
> 
> ```
> ${myvarname}foo # adds foo to the result of the content of myvarname
> ```

Thank you Miguel for pointing it out! Noted and changing it. @miguelsorianod

scripts/mas_sso_config.sh
ziccardi: hmm.. probably would worth creating an issue about fixing all those issues then.

scripts/mas_sso_config.sh
valeriiashapoval: > > the port number is also used in the `scripts/mas_sso_setup.sh` script. it might become inconsistent, when we provide full url
> 
> Yes, I think that's only where the port is used and I do not think the url is needed there. It is okay to have not the same number of variables since it is for different make targets.
> 
> I'll keep the port to configure `sso/setup` target and `url` to configure the `sso/config` target. The rest of the variables, username/password can be shared between the two targets.

Thank you Manyanda, changing it in the sso/config! @machi1990 @pawelpaszki

Makefile
```diff
@@ -548,7 +548,7 @@ sso/setup:
 .PHONY: sso/setup
 
 sso/config:
-	KEYCLOAK_URL=$(KEYCLOAK_URL) KEYCLOAK_PORT_NO=$(KEYCLOAK_PORT_NO) KEYCLOAK_USER=$(KEYCLOAK_USER) \
+	KEYCLOAK_URL=$(KEYCLOAK_URL) KEYCLOAK_USER=$(KEYCLOAK_USER) \
```
machi1990, line 551: This line:
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/41ca4d657372a66cb53bcb5b80ef91a12fcb5300/Makefile#L541

will have to be `KEYCLOAK_URL ?= http://localhost:8180`

machi1990 (approved):
lgtm

I'll let others have a look as well.

Thanks @valeriiashapoval

pawelpaszki:
@machi1990 - with current changes, should [this block](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/.github/workflows/ci.yaml#L112-L113) be updated to 

```
      - name: Setup Keycloak realm config
        run: make sso/config
```

?

machi1990 (commented):
Thanks @valeriiashapoval 

let's squash when merging as well @pawelpaszki

PR #1129, opened 2022-06-24
test: increase coverage in pkg/services

JasonPowr:
## Description
Increase unit test coverage in pkg/services
jira: https://issues.redhat.com/browse/MGDSTRM-7988

## Verification Steps
1. Run `make test`
2. Run `make test/html/coverage/report`
4. Should be able to see the increased test coverage 

- [ ] Code Review completed

### DIFF
--- a/pkg/services/types_test.go
+++ b/pkg/services/types_test.go
@@ -1,9 +1,11 @@
 package services
 
 import (
+	"net/url"
 	"testing"
 
 	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
 )
 
 func makeParams(orderByAry []string) map[string][]string {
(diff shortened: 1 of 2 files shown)


pawelpaszki:
existing changes look good! @JasonPowr - do you think some additional tests could be added to the `types_test.go` to increase the coverage for `types.go`?

JasonPowr:
> existing changes look good! @JasonPowr - do you think some additional tests could be added to the `types_test.go` to increase the coverage for `types.go`?

Thanks @pawelpaszki I do However I was finding it difficult to increase the coverage in `types.go`. So I left it be since it was high enough as is. If you could take a look that would be great.

pawelpaszki:
> > existing changes look good! @JasonPowr - do you think some additional tests could be added to the `types_test.go` to increase the coverage for `types.go`?
> 
> Thanks @pawelpaszki I do However I was finding it difficult to increase the coverage in `types.go`. So I left it be since it was high enough as is. If you could take a look that would be great.

Example of additional test for `NewListArguments` would be: 

```
func Test_NewListArguments(t *testing.T) {
	page := "5"
	overriddenListArgs := &ListArguments{
		Page:   5,
		Size:   100,
		Search: "",
	}
	type args struct {
		params url.Values
	}

	tests := []struct {
		name string
		args args
		want *ListArguments
	}{
		{
			name: "should override default page param",
			args: args{
				params: url.Values{
					"page": []string{page},
				},
			},
			want: overriddenListArgs,
		},
	}

	RegisterTestingT(t)

	for _, testcase := range tests {
		tt := testcase
			t.Parallel()
			Expect(NewListArguments(tt.args.params)).To(Equal(tt.want))
		})
	}
}
```

Also - if you could make changes to the `pkg/services/types_test.go`, so that it conforms to the test of tests (e.g. RegisterTesting outside of the for loop, not inside; use parallel testing and reassign tt, like this: `tt := testcase`)?

pkg/services/types_test.go
```diff
@@ -85,18 +86,125 @@ func Test_ValidateOrderBy(t *testing.T) {
...
+			name: "should return an error if there are too many keywords in orderBy",
+			fields: fields{
+				Page:    1,
+				Size:    100,
+				OrderBy: []string{"region desc name owner"},
+			},
+			args: args{
+				acceptedOrderByParams: getValidTestParams(),
+			},
+			wantErr: true,
```
ziccardi, line 191: We don't have test cases where `wantErr` is false. Either remove it or add a test case with `wantErr == false`

pkg/services/types_test.go
```diff
@@ -85,18 +86,125 @@ func Test_ValidateOrderBy(t *testing.T) {
...
+			t.Parallel()
+			la := &ListArguments{
+				Page:     tt.fields.Page,
+				Size:     tt.fields.Size,
+				Preloads: tt.fields.Preloads,
+				Search:   tt.fields.Search,
+				OrderBy:  tt.fields.OrderBy,
 			}
+			err := la.Validate(tt.args.acceptedOrderByParams)
+			g.Expect(err).To(HaveOccurred())
```
ziccardi, line 207: This is ignoring the `wantErr` value

pkg/services/types_test.go
```diff
@@ -85,18 +86,121 @@ func Test_ValidateOrderBy(t *testing.T) {
...
+			t.Parallel()
 			la := NewListArguments(tt.params)
 			err := la.Validate(tt.validParams)
 			if tt.wantErr {
-				Expect(err).To(HaveOccurred())
+				g.Expect(err).To(HaveOccurred())
 			} else {
-				Expect(err).NotTo(HaveOccurred())
+				g.Expect(err).NotTo(HaveOccurred())
+			}
```
ziccardi, line 100:
```go
g.Expect(err != nil).To(Equal(tt.wantErr))
```

pawelpaszki:
changes look good to me. I will leave @ziccardi to approve the PR. great work!

pkg/services/types_test.go
```diff
@@ -85,18 +86,117 @@ func Test_ValidateOrderBy(t *testing.T) {
...
+			t.Parallel()
+			la := &ListArguments{
+				Page:     tt.fields.Page,
+				Size:     tt.fields.Size,
+				Preloads: tt.fields.Preloads,
+				Search:   tt.fields.Search,
+				OrderBy:  tt.fields.OrderBy,
 			}
+			err := la.Validate(tt.args.acceptedOrderByParams)
+			g.Expect(err).To(HaveOccurred())
```
ziccardi, line 199: All the tests here are validating that errors are detected. Can we add a test to check that valid parameters are correctly validated?

JasonPowr: I believe that the correct parameters are being tested when the function is being called here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1129/files#diff-68ad7f79ef378bd62790c95989296d431e4221a4db8601f778b0cf5ca83d7214R95 .It was just these three errors where missing from the coverage.

ziccardi: The test you are referring to is validating the order by arguments and, incidentally, validates the other args as well.

Each test should be auto-contained, so if this testing if argument validation works, it should test that both valid and invalid arguments are correctly evaluated, without relying on other tests.

The other test could change and it could stop having the side effect of validating other args than order by in the future.

pkg/services/util_test.go
```diff
@@ -0,0 +1,217 @@
...
+		{
+			name: "should return conflict error",
+			args: args{
+				resourceType: "resourceType",
+				err:          errors.GeneralError("violates unique constraint"),
+			},
+			want: errors.Conflict("This resourceType already exists"),
+		},
+		{
+			name: "should return create error",
```
pawelpaszki, line 164:
```suggestion
			name: "should return general error",
```

pawelpaszki (approved):

PR #1130, opened 2022-06-24
fix(openapi): Edit of the KFM oas file + additional rules for linter to ignore

dimakis:
## Description

Edit of the kas-fleet-manager.yaml openapi spec (OAS) file. The file now passes linter. 

~- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)~
- [ ] Code Review completed


openapi/kas-fleet-manager.yaml
```diff
@@ -1413,8 +1431,8 @@ components:
         values:
           type: array
           items:
-            $ref: '#/components/schemas/values'
-    values:
+            $ref: '#/components/schemas/Values'
+    Values:
```
dimakis, line 1435: does changing the case of these values create any /much chaos in the repo?

dimakis:
I cannot seem to find the 'request review' button 😮‍💨  @wtrocki @machi1990 if ye get a chance can ye look over this?

openapi/kas-fleet-manager.yaml
```diff
@@ -4,6 +4,17 @@ info:
   title: Kafka Service Fleet Manager
   version: 1.10.0
   description: Kafka Service Fleet Manager is a Rest API to manage Kafka instances.
+  license:
+    name: 'Apache 2.0'
+    url: 'https://www.apache.org/licenses/LICENSE-2.0'
+  contact: 
+    name: 'Red Hat Hybrid Cloud Console'
+    url: 'rhosak-support@redhat.com'
```
machi1990, line 12:
```suggestion
    email: 'rhosak-support@redhat.com'
```

.spectral.yaml
```diff
@@ -1,2 +1,6 @@
 extends: '@rhoas/spectral-ruleset'
-
+rules:
+  no-$ref-siblings: off
+  rhoas-response-media-type: off
+  rhoas-error-schema: off
+  oas3-valid-schema-example: off
```
machi1990, line 6: What's the consequence of turning these off?

machi1990:
> The file now passes linter.

Awesome. Do you think we are at a point where we could enabled spectral validation as part of CI workflow?

Can you run `make openapi/generate code/fix` and commit the changes as well. 
That will enable the full CI run and we'll see how much of the generated code are usable as it.
Seeing the changes, I do not expect that the DTOs have changed much.

openapi/kas-fleet-manager.yaml
```diff
@@ -1443,8 +1461,6 @@ components:
               items:
                 allOf:
                   - $ref: "#/components/schemas/InstantQuery"
-    MetricsFederationResult:
```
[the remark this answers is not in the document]
machi1990: Good catch. This can be removed.

What do you think @JameelB?

openapi/kas-fleet-manager.yaml
machi1990: It should be fine. But let's run `make openapi/generate code/fix` and push the changes to see how far CI cries :-D

openapi/kas-fleet-manager.yaml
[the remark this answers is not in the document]
JameelB: yep this can be removed, good catch!

.spectral.yaml
dimakis: our rhoas-error-schema appears to be broken anyway so as far as I can see every API this has ran against has failed. 

The oas3-valid-schema-example appears to error our when the value needed is actually included but is nested or the value is included in the doc but one must navigate to it via the $ref. The rule isn't capable of reading nested items or $refs so I proposed in the google doc I attached that we disable this for now and write a custom rule when we upgrade to spectral v6 (which I'll be doing shortly anyway) to check for these scenarios. This happens quite frequently throughout the doc

machi1990: Thank you That makes sense. Do you mind adding in comment for now, briefly saying why it is disabled and when it can be enabled?

dimakis: The error schema may have to be rethought. This is why I was asking about the errors earlier. I'm not looking to make massive changes for the sake of the neighbors. Especially breaking changes, I'll be looking into this further as we upgrade to v6 and look to maybe create some new rules but edit some that are currently unsatisfactory

openapi/kas-fleet-manager.yaml
```diff
@@ -24,7 +35,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/VersionMetadata'
           description: Version metadata
-      summary: Returns the version metadata
+      description: Returns the kafka version metadata
```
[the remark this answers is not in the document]
dimakis: great suggestion. thanks. if there are any more which could be more concise be sure to holler!

dimakis:
> 



> > The file now passes linter.
> 
> Awesome. Do you think we are at a point where we could enabled spectral validation as part of CI workflow?
> 
> Can you run `make openapi/generate code/fix` and commit the changes as well. That will enable the full CI run and we'll see how much of the generated code are usable as it. Seeing the changes, I do not expect that the DTOs have changed much.

it fails for me on

.spectral.yaml
machi1990: > The error schema may have to be rethought. 

I think just changing the way the schema is declared should be fine in my opinion as long as we do not change or remove the properties. e.g a change like below to the schema should be okay.
```yaml
 Error:
      type: object
      required: [id, kind, href, code, reason]
      properties:
        id:
         type: string
        kind:
         type: string
        code:
         type: string
        reason:
         type: string
        href:
         type: string
        operation_id:
         type: string
```

machi1990:
> > 
> 
> > > The file now passes linter.
> > 
> > 
> > Awesome. Do you think we are at a point where we could enabled spectral validation as part of CI workflow?
> > Can you run `make openapi/generate code/fix` and commit the changes as well. That will enable the full CI run and we'll see how much of the generated code are usable as it. Seeing the changes, I do not expect that the DTOs have changed much.
> 
> it fails for me on

The command `make openapi/generate code/fix` fails for you? what error is it showing?

dimakis:
> > The file now passes linter.
> 
> Awesome. Do you think we are at a point where we could enabled spectral validation as part of CI workflow?
 so long as it passes the tests yes that was my next step to enable on CI
> 
> Can you run `make openapi/generate code/fix` and commit the changes as well. That will enable the full CI run and we'll see how much of the generated code are usable as it. Seeing the changes, I do not expect that the DTOs have changed much.

I do not see any errors when I run that command. I did initially until I replaced 'url' with 'email'

it still appears to fail CI though, although I'm not sure it ran again after force push. wasn't there a bot command to rerun checks?

.spectral.yaml
dimakis: I left a detailed reply here, has it just vanished or did you get to see it @machi1990 ?

dimakis:
> We have a make target that runs the validation for both the public and private admin endpoints.
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c3e902078616fd8c192d428b14ee923cdaefd893/Makefile#L154-L155
> 

I've tried this and it fails to run locally for me. That was another something I was going to suggest is updated if it fails for others too?

> Do you think the file `openapi/kas-fleet-manager-private-admin.yaml` could get the needed improvements as well so that we can enable that validation target in CI for both files?

I'm concentrating on creating APIs which produce consistent, non breaking (hopefully 😅 ) SDKs first but yes that shouldn't be a major issue 👍

machi1990:
> > > The file now passes linter.
> > 
> > 
> > Awesome. Do you think we are at a point where we could enabled spectral validation as part of CI workflow?
> > so long as it passes the tests yes that was my next step to enable on CI
> 
> @machi1990
> 
> I do not see any errors when I run that command. I did initially until I replaced 'url' with 'email'
> 
> > Can you run `make openapi/generate code/fix` and commit the changes as well. That will enable the full CI run and we'll see how much of the generated code are usable as it. Seeing the changes, I do not expect that the DTOs have changed much.
> 
> it still appears to fail CI though, although I'm not sure it ran again after force push. wasn't there a bot command to rerun checks?

So CI fails because the generated files have not been committed and force pushed.

machi1990:
> I've tried this and it fails to run locally for me. That was another something I was going to suggest is updated if it fails for others too?

What fails for you when you run it locally?

.spectral.yaml
dimakis: I have tried that and it hasn't worked unfortunately

> > > > The file now passes linter.
> > > 
> > > 
> > > Awesome. Do you think we are at a point where we could enabled spectral validation as part of CI workflow?
> > > so long as it passes the tests yes that was my next step to enable on CI
> > 
> > 
> > @machi1990
> > I do not see any errors when I run that command. I did initially until I replaced 'url' with 'email'
> > > Can you run `make openapi/generate code/fix` and commit the changes as well. That will enable the full CI run and we'll see how much of the generated code are usable as it. Seeing the changes, I do not expect that the DTOs have changed much.
> > 
> > 
> > it still appears to fail CI though, although I'm not sure it ran again after force push. wasn't there a bot command to rerun checks?
> 
> So CI fails because the generated files have not been committed and force pushed.

so the files like this:
`internal/kafka/internal/api/public/model_cloud_region.go` 
which are generated after `make openapi/generate code/fix` is run
need to be committed too? that does make sense

machi1990:
> The make target `make openapi/spec/validate`

Sorry, I should've been clearer. I meant to ask what errors do you get when you run that command?

.spectral.yaml
machi1990: Yes.

openapi/kas-fleet-manager.yaml
```diff
@@ -921,15 +932,22 @@ components:
             $ref: "#/components/schemas/ObjectReference"
     Error:
       allOf:
-        - $ref: "#/components/schemas/ObjectReference"
         - type: object
+          required: [id, kind, href, code, reason]
```
wtrocki, line 936: I think we can revert this part as rule is disabled

wtrocki:
Issues were that:
- Readme is missing moq
- Go version needs to be bellow 18

I think we need gvm now since we are moving to go 18 on CLI and some other projects are still working with 16-17. 
https://github.com/moovweb/gvm

machi1990:
> Issues we god
> 
> * Readme is missing requirement to install moq

moq is auto installed so you do not need to install it: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/Makefile#L88-L101 

I also think for the `make openapi/generate` command it is not really needed. Its only needed when you run `make generate`

> * Go version needs to be bellow 18

Yes, we run on go 1.17.
 
> I think we need gvm now since we are moving to go 18 on CLI and some other projects are still working with 16-17. https://github.com/moovweb/gvm

Yes. I'll recommend this, it's what I use as well.

openapi/kas-fleet-manager.yaml
machi1990: let's see if we can find ways to come with a better schema that works with the error check enabled.
I'll prefer if we can deal with quick wins right away rather than disabling the checks.

internal/kafka/internal/api/admin/private/model_error.go
```diff
@@ -11,10 +11,10 @@ package private
 
 // Error struct for Error
 type Error struct {
-	Id          string `json:"id,omitempty"`
-	Kind        string `json:"kind,omitempty"`
-	Href        string `json:"href,omitempty"`
 	Code        string `json:"code,omitempty"`
 	Reason      string `json:"reason,omitempty"`
 	OperationId string `json:"operation_id,omitempty"`
+	Id          string `json:"id"`
```
wtrocki, line 17: Wondering why Kafka models got affected. Maybe generator version changed?

internal/kafka/internal/api/admin/private/api/openapi.yaml
```diff
@@ -349,6 +349,10 @@ components:
           type: string
         href:
           type: string
+      required:
```
wtrocki, line 352: Btw. I think this is fine but for future this would be breaking change. We need to have tooling for detecting breaking changes to assist API teams.

wtrocki:
This looks very good! We need core team to approve this

machi1990 (changes requested):
Can we address the issue in `openapi/kas-fleet-manager-private-admin.yaml` as well? Once those are fixed, we'll be in a position to add the spectral validation in CI, which will help a lot in detecting issues.

Additionally, I'd love to hear what are the plans in regard to extending the ruleset especially the errors checks which are disabled in the current state of the PR.

wtrocki:
> Can we address the issue in openapi/kas-fleet-manager-private-admin.yaml as well?

My answer is yes, but not initially.
Our target is to get public API passing and on full CICD flow asap. Private apis should probably have different set of rules than public ones etc. 

> Additionally, I'd love to hear what are the plans in regard to extending the ruleset 

Yes! that is whole plan - add rules to cicd and then add more rules.
After we have all services covered by rules we will be improving and extending our ruleset while still making sure that all existing api pass that.

machi1990:
> I think we might have cicd issue here - rhoas ruleset should be working only on public kas and connector api. then we can run plain spectral on private ones.

The mention admin endpoint follows the same API guidelines and it is one of the reason why it's included in the in the spec validate target. It's only a few errors that are printed on it - I think 2, which are already addressed by some of the changes in this PR so I'd say fixing them will be quick win as well.

machi1990:
@dimakis @wtrocki 
what spectral versions are you using?
I am seeing CI is failing with 
```
spectral lint openapi/kas-fleet-manager.yaml openapi/kas-fleet-manager-private-admin.yaml
ENOENT: no such file or directory, open '/home/runner/work/kas-fleet-manager/kas-fleet-manager/@rhoas/spectral-ruleset'
```

And it is running with `6.0.1`

machi1990:
Locally, I've a pretty old version `4.x` or `5.x` installed from a year ago and that's passing.

machi1990:
> Locally, I've a pretty old version `4.x` or `5.x` installed from a year ago and that's passing.

Fell back to using `5.9.2` but it is a deprecated version. Let's see what CI says.

dimakis:
@machi1990 Only getting back around to this now. The latest push sorted the issue with the `rhoas-error-schema`, the rule does not need to be dropped now as per your earlier comment comment.

dimakis:
@machi1990  yea, once we get the current apis passing I was going to set about migrating to v6. The problem is a lot if not all of our custom rules need to be rewritten to work with v6.

machi1990:
Thank @dimakis and @wtrocki for the amazing work. 

The second commit adds the validation in CI.

dimakis:
> Thank @dimakis and @wtrocki for the amazing work.
> 
> The second commit adds the validation in CI.

Thanks for the help reviewing all 😄

machi1990:
> Thanks for the help reviewing all smile

No worries. Feel free to promote the PR for review whenever you feel like it is no longer a `WIP`

machi1990:
@JameelB would you like to have a final look as well?
Thanks

dimakis:
@JameelB can I get you to have a look again please?

.github/workflows/ci.yaml
```diff
@@ -105,6 +105,14 @@ jobs:
           go-version: "1.17"
       - name: Check out code into the Go module directory
         uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          base: main
+          list-files: json
+          filters: |
+            openapi:
```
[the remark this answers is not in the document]
machi1990: Hi @dimakis, no it has to be openapi. The GH actions appends in the `_files` prefix.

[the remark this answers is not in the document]
dimakis: thats great to know. Thanks 😀

.spectral.yaml
```diff
@@ -1,2 +1,6 @@
 extends: '@rhoas/spectral-ruleset'
+rules:
+  no-$ref-siblings: off
```
JameelB, line 3: what was the reason this was disabled?

JameelB (commented):
Great job on this @dimakis! Just a few questions from me.

internal/kafka/internal/api/admin/private/api_default.go
```diff
@@ -4,6 +4,7 @@
  * The admin APIs for the fleet manager of Kafka service
  *
  * API version: 0.1.0
+ * Contact: rhosak-support@redhat.com
```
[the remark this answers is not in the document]
machi1990: It should be possible. We'll need to play a little bit with the template used for the code generation: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/openapi/templates/model.mustache

.spectral.yaml
dimakis: it is no longer a thing in OAS v3.1
They've completely got rid of it. May have something to do with the json schema version support but I'm not 100%, either way it doesn't appear to have any effect. 
@wtrocki may be able to give a better explanation

wtrocki: Instead of explanation I will come up with question. Can we update your openapi version to 3.1? From what I see it would not affect much

JameelB: I don't see why not :thinking: can you confirm @machi1990?

Thanks for taking on this work :)

dimakis: @wtrocki I was going to bump up the API versions across the board but I think they should be in separate PRs, they may have breaking changes etc. even if not, they should be seperated into their own PRs imho

JameelB: agreed for bumping it in separate PR. Thanks for the explanation! :)

JameelB (approved):

.spectral.yaml
machi1990: > I don't see why not thinking can you confirm @machi1990?

It should be okay but as already suggested maybe something to do in a separate PR.

PR #1132, opened 2022-06-24
Added  tests for the keycloak service proxy

ziccardi:
## Description
[keycloak_kafka_test.go](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/internal/services/keycloak_kafka_test.go) was duplicating most of the tests already done in [keycloak_service_test](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/services/sso/keycloak_service_test.go).
This PR deletes the [keycloak_kafka_test.go](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/internal/services/keycloak_kafka_test.go) and introduces the new `keycloak_service_proxy_test.go` that tests the service proxy.

With this change, the coverage doesn't change much, but we move the tests in the right package (before we had the test in the `service` package testing objects in the `sso` package) and we can remove methods we created just to expose internals to the tests.

- [ ] Code Review completed


pkg/services/sso/keycloak_service_builder.go
```diff
@@ -101,12 +101,20 @@ func build(providerName string, keycloakConfig *keycloak.KeycloakConfig, realmCo
 	if providerName == keycloak.MAS_SSO ||
 		realmConfig != nil {
 		_, realmConfig := arrays.FindFirst(notNilPredicate, realmConfig, keycloakConfig.KafkaRealm)
-		return newKeycloakService(keycloakConfig, realmConfig.(*keycloak.KeycloakRealmConfig))
```
machi1990, line 104: is this func used anywhere? If not it can be deleted.

pkg/services/sso/keycloak_service_proxy.go
```diff
@@ -10,13 +10,11 @@ import (
 	"github.com/openshift-online/ocm-sdk-go/authentication"
 )
 
-type tokenProvider interface {
-	GetToken() (string, error)
-}
+type tokenProvider func() (string, error)
```
machi1990, line 13: Why this change?

pkg/services/sso/keycloak_service_builder.go
ziccardi: Good point. Was used by one test, but I replaced it with the builder. Removed.

pkg/services/sso/keycloak_service_proxy.go
ziccardi: Because the `tokenProvider` is just one method interface: I find creating an interface to pass an object with just one method to be much _javish_ and a bit overkill provided that in go we can pass functions.

machi1990: Thanks for the explanation.

PR #1135, opened 2022-06-27
fix: update naming conventions for documentation and sdks

wtrocki:
Content, API and UX team trying to make our API more client facing by documenting them in the console.redhat.com

For that API we cannot use internal names like "kas", "fleet" etc. as those would be confusing for users.


wtrocki:
CC @jgiardino @dimakis

wtrocki:
See SRS fleet manager PR as well: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1135/files

wtrocki:
I will regenerate codebase after review to avoid conflicts. 

CC @machi1990

openapi/kas-fleet-manager.yaml
```diff
@@ -1,9 +1,9 @@
 openapi: 3.0.0 # need this as first line to allow some IDEs to know this is an openapi document.
 # All endpoints defined here will be discoverable by all users. If there is a need to keep some endpoints "private", add them to the "managed-services-api-private.yaml" file
 info:
-  title: Kafka Service Fleet Manager
```
machi1990, line 4: The name  `Kafka Service Fleet Manager` was a suggested name back then and this is where `kas-fleet-manager` came from. Has this been changed?

machi1990:
> I will regenerate codebase after review to avoid conflicts.
> 
> CC @machi1990

The changes lgtm and make sense: the endpoints already have `kafkas_mgmt` in them so it only makes sense to change the title as well.
But I'd love to hear thoughts from others as well and if `Kafka Management API` is the agreed name.
/cc @pb82 @davidffrench @emmanuelbernard any thoughts on this?

openapi/kas-fleet-manager.yaml
wtrocki: Yes. I agree. This is suggested name for "project name". Internal architecture and all internal communication.
For our customers and users however that might be quite cryptic term etc.

wtrocki:
CC @jbyrne-redhat

valdar (approved):

wtrocki:
@davidffrench looking for favour here to enable us to unblock epic. Rename affects nothing but API UI

davidffrench:
Sorry, I normally miss all GitHub notifications. Always best to ping me on chat or email.

I hate to be a pain here but I am not on board with changing the name to `Kafka Management API`. I am not sure I grasp the need. The term fleet manager is becoming significantly more common in Red Hat with all new managed services.

wtrocki:
To clarify:
We are not changing name of project etc. we are only looking to change name in OpenAPI file that has no impact to anything else than API UI. This file is dedicated to be used by end users and we want to have the same standard. 

This is to help our customers my documenting our API under some naming patterns so they understand our APIs

davidffrench:
Thanks for the clarification @wtrocki . I agree with you that the API endpoints exposed through this openapi spec are specific to Kafka Management, so I am not opposed to this change.

The overall service has additional responsibilities such as fleet management etc.

wtrocki:
Rebased :D

PR #1144, opened 2022-06-28
feat: dynamic scaling: add capacity_reservation to config

JasonPowr:
## Description
Adding `capacity_reservation` to `config/dynamic-scaling-configuration.yaml`. To enable the modification of the capacity reservation for each streaming unit via the GitOps flow in AppInterface.

jira: https://issues.redhat.com/browse/MGDSTRM-8736

- [x] Code Review completed


internal/kafka/internal/config/dynamic_scaling_config.go
```diff
@@ -52,14 +52,19 @@ func (c *DynamicScalingConfig) validate() error {
...
+	ComputeNodesConfig  *DynamicScalingComputeNodesConfig `yaml:"compute_nodes_config"`
+	CapacityReservation int                               `yaml:"capacity_reservation"`
 }
 
 func (c *InstanceTypeDynamicScalingConfig) validate() error {
 	if c.ComputeNodesConfig == nil {
 		return fmt.Errorf("compute_nodes_config is mandatory")
 	}
 
+	if c.CapacityReservation <= 0 {
```
machi1990, line 64: I am wondering if we should allow for this to be set to `0` to express the `no capacity reservation` intent.
What do you think? @JasonPowr @JameelB @VanillaSpoon

config/dynamic-scaling-configuration.yaml
```diff
@@ -1,7 +1,17 @@
...
+#
+#The following properties must be defined for each Kafka instance type:
+#   - capacity_reservation: the number of reservation to be made on each data plane cluster
+#
+#The following properties must be defined for each instance types compute node configuration:
+#   - max_compute_nodes: the maximum number of nodes needed to create an appropriate machine pool with the configuration needed by the autoscaler
+
 ---
 developer:
+  capacity_reservation: 3
```
miguelsorianod, line 11: When reviewing this it didn't seem clear to me what `capacity_reservation` is. Nodes? kafka instances? pause pods?

Maybe that's an indicative of needing a more specific name?

cc @VanillaSpoon @machi1990

machi1990: Good point. `reserved_streaming_units` or `reserved_kafka_streaming_units`?

miguelsorianod: Streaming units are limited to the kafka domain (at least for now) so both seem correct to me.

machi1990: /cc @JasonPowr see the above suggestions for the naming of the field.

internal/kafka/internal/config/dynamic_scaling_config.go
machi1990: I am leaning toward making it optional i.e not mandatory. If it is set to `0` or left out (thus defaults to `0`) we'd consider it as no reservation intent. 

Just to avoid making it a nullable field and to have to handle `config !=nil` where it'll be used.

machi1990 (changes requested):

VanillaSpoon:
Hey @JasonPowr, how would you suggest I test this?

JasonPowr:
> 

I suppose the best way to test it would be just to ensure that the validation works correctly :)

JasonPowr:
> Hey @JasonPowr, how would you suggest I test this?

I suppose that the best way to test it would be just to ensure that the validation works correctly :)

templates/service-template.yml
```diff
@@ -165,7 +165,7 @@ parameters:
 - name: DYNAMIC_SCALING_CONFIG
   displayName: Dynamic Scaling configuration
   description: "YAML content containing a map of the dynamic scaling configuration for each instance type"
-  value: "{developer: {compute_nodes_config: {max_compute_nodes: 3}}, standard: {compute_nodes_config: {max_compute_nodes: 9}}}"
```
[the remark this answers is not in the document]
JasonPowr: @JameelB Done that there now :)

docs/feature-flags.md
```diff
@@ -132,6 +132,7 @@ This lists the feature flags and their sub-configurations to enable/disable and
         - `providers-config-file` [Required]: The path to the file containing a list of supported cloud providers that the service can provision dataplane clusters to (default: `'config/provider-configuration.yaml'`, example: [provider-configuration.yaml](../config/provider-configuration.yaml)).
         - `cluster-compute-machine-type` [Optional]: The compute machine type to be used for provisioning a new dataplane cluster (default: `m5.2xlarge`).
         - `cluster-openshift-version` [Optional]: The OpenShift version to be installed on the dataplane cluster (default: `""`, empty string indicates that the latest stable version will be used). 
+        - `DYNAMIC_SCALING_CONFIG`: The configuration file that contains information about each Kafka instance types, dynamic scaling configuration. Defaults to `"{developer: {reserved_streaming_units: 3, compute_nodes_config: {max_compute_nodes: 3}}, standard: {reserved_streaming_units: 10, compute_nodes_config: {max_compute_nodes: 9}}}"`
```
JameelB, line 135: Instead of the service template parameter, this should document the `dynamic-scaling-config-file` flag instead. The default value would be `config/dynamic-scaling-configuration.yaml` and you can point to the dynamic-scaling-configuration.yaml file as an example. I believe this is required when `auto` scaling is enabled.

See the description of `providers-config-file` above as an example.

config/dynamic-scaling-configuration.yaml
```diff
@@ -1,7 +1,17 @@
...
+#The following properties must be defined for each instance types compute node configuration:
+#   - max_compute_nodes: the maximum number of nodes needed to create an appropriate machine pool with the configuration needed by the autoscaler
+
 ---
 developer:
+  reserved_streaming_units: 3 
   compute_nodes_config:
     max_compute_nodes: 3
 standard:
+  reserved_streaming_units: 10
```
JameelB, line 15: wdyt of changing this to 1 or maybe 2 by default? I don't think we need to reserve 10 for local dev. Same with developer. We should try and minimise the amount of nodes to scale up to for local dev.

cc @machi1990

templates/service-template.yml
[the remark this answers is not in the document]
JameelB: thanks Jason! :)

config/dynamic-scaling-configuration.yaml
machi1990: This make sense, thank you for catching this.

Also considering the max_nodes is `3` that's won't be enough to fit `10` streaming units.

JameelB (approved):
Great job on this @JasonPowr! Looks good to me :)

config/dynamic-scaling-configuration.yaml
JameelB: Thanks! See that it's updated now, closing as resolved :)

PR #1149, opened 2022-06-29
fix(metrics): reset capacity used and capacity available metrics for KafkaManager reconciler

machi1990:
## Description

The two metrics were calculated in the KafkaManager reconciler but the
metrics were reset from the clusters reconciler.

By moving this to the correct resets methods, the metrics will get properly reset when the leader changes for the KafkaManager reconciler.
Which in turn will ensures that the old leader pod does not show stale data when it's scraped

## Verification Steps

Code review should be enough

- ~~[ ] All acceptance criteria specified in JIRA have been completed~~
- [ ] Code Review completed

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -717,6 +717,8 @@ func init() {
 func ResetMetricsForKafkaManagers() {
 	kafkaStatusSinceCreatedMetric.Reset()
 	KafkaStatusCountMetric.Reset()
+	clusterStatusCapacityUsedMetric.Reset()
+	clusterStatusCapacityAvailableMetric.Reset()
 }
 
 // ResetMetricsForClusterManagers will reset the metrics for the ClusterManager background reconciler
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -726,8 +728,6 @@ func ResetMetricsForClusterManagers() {
 	clusterStatusCountMetric.Reset()
 	kafkaPerClusterCountMetric.Reset()
 	clusterStatusCapacityMaxMetric.Reset()
-	clusterStatusCapacityUsedMetric.Reset()
-	clusterStatusCapacityAvailableMetric.Reset()
```
machi1990, line 730: These three metrics could probably be calculated in one reconciler. 
There will be some work around these metrics for dynamic scaling tasks. We could use the opportunity to revisit them and calculate them in one reconciler.

pawelpaszki (approved):

JameelB (approved):
Good catch @machi1990!

machi1990:
Thanks for the review @pawelpaszki and @JameelB

PR #1155, opened 2022-06-30
Revert "fix MGDCTRS-997: Admin API to list connectors with new revisions"

dhirajsb:
This reverts commit 66ae69087ac29e6abb21a06fa7917936b7f0804c.

## Description
Reverting a major DB change just before going into code freeze for SSO migration. 

## Verification Steps
CI should work without the DB migration introduced in the reverted migration. 

~~- [ ] All acceptance criteria specified in JIRA have been completed~~
- [ ] Code Review completed


dhirajsb:
@lburgazzoli @valdar we have to revert the big DB change introduced by the recent connectors change, unless we want to go into code freeze with that change in stage and prod. 
We need to pull in several sso related changes that were done recently, and I can't think of a way to cherry-pick just those changes into c-f-m.

dhirajsb:
Actually, I think there is a solution to this problem. We can just bump c-f-m upto the latest k-f-m production commit, which doesn't have the latest connector change. That way we can pull in all the sso changes we need without having to do this revert.

PR #1162, opened 2022-07-04
feat: Add reserved managed kafka generation

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-8738.

Implements the addition of the concept of reserved managed kafka instances. A reserved kafka instance is a ManagedKafka that contains the label `bf2.org/deployment` with the value set to `reserved`.
The agent-cluster's 'kafkas' endpoint now returns a list of reserved managed kafka instances based on dynamic scaling configuration attributes when dynamic scaling is enabled.

Additional remarks:
* The number of reserved ManagedKafka instances that is returned depends on the `reserved_streaming_units` attribute defined for each instance type in the dynamic scaling configuration (implemented in PR https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1144). Only the supported instance types in the cluster are evaluated.
* Each reserved ManagedKafka instance has its own dedicated namespace in the data plane cluster
* a reserved ManagedKafka instance naming schema is: `reserved-kafka-<instance_type>-<kafka_number>` where kafka_number goes from 1..<number_of_reserved_streaming_units_for_the_given_instance_type>
* The K8s namespace of a ManagedKafka instance is the same as its name. The K8s object name is also the same as its name
* If fleet manager is not in autoscaling mode an empty list is returned
* If the data plane cluster is not in 'ready' state an empty list is returned
* If the data plane cluster is 'ready' but there are no 'ready' strimzi versions an error is returned

- [ ] Code Review completed


internal/kafka/internal/services/kafka.go
```diff
@@ -776,6 +787,59 @@ func (k *kafkaService) GetManagedKafkaByClusterID(clusterID string) ([]managedka
...
+	}
+
+	strimziVersions, err := cluster.GetAvailableAndReadyStrimziVersions()
+	if err != nil {
+		return nil, errors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to generate reserved managed kafkas for clusterID %s: error finding ready strimzi versions", clusterID)
+	}
+	if len(strimziVersions) == 0 {
+		return nil, apiErrors.GeneralError("failed to generate reserved managed kafkas for clusterID %s: no ready strimzi versions found")
+	}
+	latestStrimziVersion := strimziVersions[len(strimziVersions)-1]
```
machi1990, line 816: We could have a `GetLatestStrimziVersion` as well which we could use here as:
```go
latestStrimziVersion, strimziVersionErr := cluster.GetLatestStrimziVersion()
```

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1214,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 	return managedKafkaCR, nil
 }
 
+// buildReservedManagedKafkaCR buklds a Reserved Managed Kafka CR.
+// The ID, K8s object ID, K8s namespace and PlacementID are all set to
+// the provided kafkaID.
+func (k *kafkaService) buildReservedManagedKafkaCR(kafkaID string, instanceType string, desiredStrimziVersion api.StrimziVersion, kafkaConfig *config.KafkaConfig) (*managedkafka.ManagedKafka, *errors.ServiceError) {
+	// Reserved instances always make use of the streaming base unit which is x1.
```
machi1990, line 1221: okay for this to be hardcoded for now.

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1214,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
...
+			Endpoint: managedkafka.EndpointSpec{
+				BootstrapServerHost: fmt.Sprintf("%s-dummyhost", kafkaID),
+			},
+			Versions: managedkafka.VersionsSpec{
+				Kafka:    desiredKafkaVersion.Version,
+				Strimzi:  desiredStrimziVersion.Version,
+				KafkaIBP: desiredKafkaIBPVersion.Version,
+			},
+			Deleted: false,
+			Owners:  []string{}, // TODO is this enough?
```
machi1990, line 1277: From what I see, this is enough. 
The ManagedKafkaCR is created and there are no errors.

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1214,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kafkaID,
+			Namespace: kafkaID,
+			Annotations: map[string]string{
+				"bf2.org/id":          kafkaID,
+				"bf2.org/placementId": kafkaID,
+			},
+			Labels: labels,
+		},
+		Spec: managedkafka.ManagedKafkaSpec{
```
machi1990, line 1258: Hi @shawkins we are not setting the `ManagedKafka.Spec.OAuth` since this is optional and not at all suspected to cause an deserialization issues. Can you confirm that this is okay?

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -519,7 +519,7 @@ func (c *ClusterManager) reconcileDynamicCapacityInfo(cluster api.Cluster) error
 			return nil
 		}
 
-		supportedInstanceTypes := strings.Split(cluster.SupportedInstanceType, ",")
+		supportedInstanceTypes := cluster.GetSupportedInstanceTypes()
```
machi1990, line 522: love this!

openapi/kas-fleet-manager-private.yaml
```diff
@@ -258,7 +258,8 @@ components:
                       type: string
                     bf2.org/kafkaInstanceProfileQuotaConsumed:
                       type: string
-
+                    bf2.org/deployment:
```
machi1990, line 261: @shawkins as described in the epic brief:
The label `bf2.org/deployment` will be set for `reserved` kafkas only. For real workload this label won't be set

internal/kafka/internal/services/kafka.go
miguelsorianod: In any case it would be a `

> should we ignore querying the database for the kafkas whose id begins with `reserved-kafka-` and only log the status info of the rejected kafkas? This newly added log could later on be replaced by a metric to see how many reserved streaming units are there and are in a `ready` or `rejected` status.


> should we ignore querying the database for the kafkas whose id begins with `reserved-kafka-` and only log the status info of the rejected kafkas? This newly added log could later on be replaced by a metric to see how many reserved streaming units are there and are in a `ready` or `rejected` status.

That is a good question. This method also persists information of the kafka in the DB.
With the current implementation an error will be logged and the iteration continue:
```
		if getErr != nil {
			glog.Error(errors.Wrapf(getErr, "failed to get kafka cluster by id %s", ks.KafkaClusterId))
			continue
		}
```

In any case if we were to do any comparison the ID comparison seems a little bit fragile. If we were to improve that maybe it would be a good idea for the data plane to send us back if it is a reserved kafka or not? (as we do when we provide that info to them)

miguelsorianod: In this case it would be a `GetLatestAvailableAndReadystrimziVersion` method.
I will implement `GetLatestAvailableAndReadystrimziVersion` and `GetLatestAvailableReadystrimziVersion` methods.

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1214,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
...
+	// For now we hardcode it but there might be a better alternative to it.
+	streamingBaseUnit := "x1"
+	kafkaInstanceSize, err := kafkaConfig.GetKafkaInstanceSize(instanceType, streamingBaseUnit)
+	if err != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "unable to list kafka request")
+	}
+	labels := map[string]string{
+		"bf2.org/kafkaInstanceProfileQuotaConsumed": strconv.Itoa(kafkaInstanceSize.QuotaConsumed),
+		"bf2.org/kafkaInstanceProfileType":          instanceType,
+		"bf2.org/deployment":                        "reserved",
```
JameelB, line 1231: shall we make `reserved` into a constant?

machi1990:
> In any case if we were to do any comparison the ID comparison seems a little bit fragile. 

I agree. We are eliminating the reserved ones based on the id format. It should be okay if we do not change the `id` format of the reserved and real kafkas. For real kafkas, I can't see of a change anytime soon, but for reserved kafka I do not know so if that changes we might need to "remember" updating the status update method as well.

> If we were to improve that maybe it would be a good idea for the data plane to send us back if it is a reserved kafka or not? (as we do when we provide that info to them)

I do not know if the data plane sends back the labels. A quick check on the managed kafka cr, there is nothing in the `.Status` that indicates this.

JameelB (commented):
Looks good to me!

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1214,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 	return managedKafkaCR, nil
 }
 
+// buildReservedManagedKafkaCR buklds a Reserved Managed Kafka CR.
+// The ID, K8s object ID, K8s namespace and PlacementID are all set to
+// the provided kafkaID.
+func (k *kafkaService) buildReservedManagedKafkaCR(kafkaID string, instanceType string, desiredStrimziVersion api.StrimziVersion, kafkaConfig *config.KafkaConfig) (*managedkafka.ManagedKafka, *errors.ServiceError) {
+	// Reserved instances always make use of the streaming base unit which is x1.
+	// For now we hardcode it but there might be a better alternative to it.
```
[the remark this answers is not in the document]
machi1990: We won't be doing the above, at least in this first version.
Partly because, we need to have the support for higher sizes and for it to be in stable maturity state.

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1214,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
...
+				IngressPerSec:               kafkaInstanceSize.IngressThroughputPerSec.String(),
+				EgressPerSec:                kafkaInstanceSize.EgressThroughputPerSec.String(),
+				TotalMaxConnections:         kafkaInstanceSize.TotalMaxConnections,
+				MaxDataRetentionSize:        kafkaInstanceSize.MaxDataRetentionSize.String(),
+				MaxPartitions:               kafkaInstanceSize.MaxPartitions,
+				MaxDataRetentionPeriod:      kafkaInstanceSize.MaxDataRetentionPeriod,
+				MaxConnectionAttemptsPerSec: kafkaInstanceSize.MaxConnectionAttemptsPerSec,
+			},
+			Endpoint: managedkafka.EndpointSpec{
+				BootstrapServerHost: fmt.Sprintf("%s-dummyhost", kafkaID),
```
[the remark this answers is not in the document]
machi1990: Yes. Unfortunately it is need to bypass null checks in FSO side.
It is one of the required fields. Without this dummy value the endpoints looks like `endpoint: {}` and the FSO will see bootstrap server host as `null`.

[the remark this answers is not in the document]
JameelB: ah I see, thanks for confirmation @machi1990 :)

pkg/api/cluster_types.go
```diff
@@ -277,6 +278,30 @@ func (s *StrimziVersion) DeepCopy() *StrimziVersion {
 	return &res
 }
 
+// GetLatestKafkaVersion returns the latest (most recent) Kafka version
```
machi1990, line 281: @VanillaSpoon these newly added method here could be of use in the re-assignment sub-task

internal/kafka/internal/services/kafka.go
miguelsorianod: The methods have been added

miguelsorianod:
> > In any case if we were to do any comparison the ID comparison seems a little bit fragile.
> 
> I agree. We are eliminating the reserved ones based on the id format. It should be okay if we do not change the `id` format of the reserved and real kafkas. For real kafkas, I can't see of a change anytime soon, but for reserved kafka I do not know so if that changes we might need to "remember" updating the status update method as well.
> 
> > If we were to improve that maybe it would be a good idea for the data plane to send us back if it is a reserved kafka or not? (as we do when we provide that info to them)
> 
> I do not know if the data plane sends back the labels. A quick check on the managed kafka cr, there is nothing in the `.Status` that indicates this.

What I was proposing is for them to send that information in case it is not being sent

internal/kafka/internal/services/kafka.go
miguelsorianod: :+1: . I thought about it but I didn't do it. It does no harm so I added it as a constant and I also added the label key as a constant

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1213,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
...
+	}
+	labels := map[string]string{
+		"bf2.org/kafkaInstanceProfileQuotaConsumed":    strconv.Itoa(kafkaInstanceSize.QuotaConsumed),
+		"bf2.org/kafkaInstanceProfileType":             instanceType,
+		managedkafka.ManagedKafkaBf2DeploymentLabelKey: managedkafka.ManagedKafkaBf2DeploymentLabelValueReserved,
+	}
+
+	desiredKafkaVersion := desiredStrimziVersion.GetLatestKafkaVersion()
+	if desiredKafkaVersion == nil {
+		return nil, errors.GeneralError("No available Kafka versions in Strimzi version %s", desiredStrimziVersion.Version)
```
ziccardi, line 1235: Errors should not start with uppercase

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1213,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
...
+	}
+	desiredKafkaIBPVersion := desiredStrimziVersion.GetLatestKafkaIBPVersion()
+	if desiredKafkaIBPVersion == nil {
+		return nil, errors.GeneralError("No available Kafka IBP versions in Strimzi version %s", desiredStrimziVersion.Version)
```
ziccardi, line 1239: Errors should not start with uppercase

pkg/api/cluster_types_test.go
```diff
@@ -84,7 +85,79 @@ func TestGetAvailableStrimziVersions(t *testing.T) {
...
+
+	for _, testcase := range tests {
+		tt := testcase
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			res, err := tt.cluster().GetLatestAvailableStrimziVersion()
+			gotErr := err != nil
+			if gotErr != tt.wantErr {
+				t.Errorf("wantErr: %v got: %v", tt.wantErr, err)
+			}
```
ziccardi, line 154:
```suggestion
			Expect(err != nil).To(Equal(tt.wantErr), "wantErr: %v got: %v", tt.wantErr, err)
```

pkg/api/cluster_types_test.go
```diff
@@ -84,7 +85,79 @@ func TestGetAvailableStrimziVersions(t *testing.T) {
...
+				return &res
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, testcase := range tests {
+		tt := testcase
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
```
ziccardi, line 149: Move this out of the for loop

pkg/api/cluster_types_test.go
```diff
@@ -84,7 +85,79 @@ func TestGetAvailableStrimziVersions(t *testing.T) {
...
+			res, err := tt.cluster().GetLatestAvailableStrimziVersion()
+			gotErr := err != nil
+			if gotErr != tt.wantErr {
+				t.Errorf("wantErr: %v got: %v", tt.wantErr, err)
+			}
+			g.Expect(res).To(Equal(tt.want))
```
ziccardi, line 155: If we expect an error to occur, we should check that the error is what we were expecting and not another, unrelated error. I think a check should be added here.

pkg/api/cluster_types_test.go
```diff
@@ -157,6 +230,75 @@ func TestGetAvailableAndReadyStrimziVersions(t *testing.T) {
...
+				return &res
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, testcase := range tests {
+		tt := testcase
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
```
ziccardi, line 294: Move this out of the for loop

pkg/api/cluster_types_test.go
ziccardi, line 149: Can we run the test in parallel? If yes, please add `t.Parallel()` here

pkg/api/cluster_types_test.go
```diff
@@ -157,6 +230,75 @@ func TestGetAvailableAndReadyStrimziVersions(t *testing.T) {
...
+			res, err := tt.cluster().GetLatestAvailableAndReadyStrimziVersion()
+			g.Expect(err != nil).To(Equal(tt.wantErr))
```
ziccardi, line 296: Please, add a test to check that the error we received was the error we were expecting

internal/kafka/internal/services/kafka.go
miguelsorianod: :+1: fixed. I understand the linter should catch this at some point. I still hadn't look at linters.

internal/kafka/internal/services/kafka.go
miguelsorianod: :+1: fixed

pkg/api/cluster_types_test.go
miguelsorianod: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1162#discussion_r913091719

pkg/api/cluster_types_test.go
miguelsorianod: I don't think this test needs to check specific error types. We are only interested on whether there's been an error

pkg/api/cluster_types_test.go
ziccardi: Oh.. nice finding. Then I guess we have to fix a number of our tests.

pkg/api/cluster_types_test.go
miguelsorianod: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1162#discussion_r913094285. I don't think in all tests we receive errors we necessarily need to check for the error types (unless the test is testing the error types received)

pkg/api/cluster_types_test.go
miguelsorianod: I think it probably can be done. This makes me wonder what's the strateg to follow for this however. It's hard to enforce. It also makes me wodner whether we really need to worry about this until performance becomes a real concern.

pkg/api/cluster_types_test.go
```diff
@@ -601,6 +743,98 @@ func Test_CompareSemanticVersionsMajorAndMinor(t *testing.T) {
...
+				return &StrimziVersion{}
+			},
+			want: nil,
+		},
+	}
+
+	for _, testcase := range tests {
+		tt := testcase
+		t.Run(tt.name, func(test *testing.T) {
+			g := NewWithT(t)
```
ziccardi, line 830: Can we add `test.Parallel()` here?

pkg/api/cluster_types_test.go
miguelsorianod: :+1: . This was written this way because it was a copy-paste of the already existing test `Test_Cluster_GetAvailableStrimziVersions`. I will update it for both.

ziccardi (commented):
Awesome job. I just left few comments:

1. When we expect in a test a function to end with en error, I believe we should check what error we received, to avoid having the test succeed because some totally unrelated error has happened
2. When it is possible, I believe we should run the tests in parallel (t.Parallel())

pkg/api/cluster_types_test.go
miguelsorianod: updated

miguelsorianod:
> When we expect in a test a function to end with en error, I believe we should check what error we received, to avoid having the test succeed because some totally unrelated error has happened

I don't think that's the case to all tests. It totally depends on what are you interested in testing. Also the case you described does not apply. If you check for an error (any) it will fail if there's any error (even unrelated ones)

pkg/api/cluster_types_test.go
ziccardi: > until performance becomes a real concern

I usually agree about this when the change is big, but since the change is just one row, I really prefer to avoid waiting to have a problem to add a possible solution.

Moreover, running the tests in parallel has the added value of testing that the tested code works as expected when invoked in parallel too.

miguelsorianod:
> When it is possible, I believe we should run the tests in parallel (t.Parallel())

This would be the ideal scenario but it makes me think that the cognitive load to evaluate this is potentially big. I am wondering if we should only worry about this when we know the test is time consuming and we are hitting performance problems.

pkg/api/cluster_types_test.go
miguelsorianod: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1162#issuecomment-1173934813

miguelsorianod: > Moreover, running the tests in parallel has the added value of testing that the tested code works as expected when invoked in parallel too.

That is a good point

ziccardi:
> I don't think that's the case to all tests. It totally depends on what are you interested in testing. Also the case you described does not apply. If you check for an error (any) it will fail if there's any error (even unrelated ones). It is also infeasible to test all possible error combinations. Specially when the method you are testing calls other methods each of those possibily returning different errors at different levels.

I don't agree. The role of a good Unit test is to be repeatable and predictable.
If you expect a function to end with an error in a test, that error must be exactly one and you should be able to know what error you are expecting.

If you are not able to predict what is going to happen, then I think the unit test fails its mandate.

miguelsorianod:
> > I don't think that's the case to all tests. It totally depends on what are you interested in testing. Also the case you described does not apply. If you check for an error (any) it will fail if there's any error (even unrelated ones). It is also infeasible to test all possible error combinations. Specially when the method you are testing calls other methods each of those possibily returning different errors at different levels.
> 
> I don't agree. The role of a good Unit test is to be repeatable and predictable. If you expect a function to end with an error in a test, that error must be exactly one and you should be able to know what error you are expecting.
> 
> If you are not able to predict what is going to happen, then I think the unit test fails its mandate.

My comment still applies. I can't see a reliable, feasible and maintainable way to test this, aside from not being interested on testing specific errors in this test (or others)

ziccardi:
> My comment still applies. I can't see a reliable, feasible and maintainable way to test this, aside from not being interested on testing specific errors in this test (or others). If the type of the error being returned was relevant to the consumer of the method then I would agree testing the type would be desirable, but that's not always the case. There is no contract in the method about what errors and when those happen.

If you are expecting an error to happen, that means that you exactly know what is wrong and why. It doesn't matter the contract. If I call a method with a wrong username, I expect to receive a wrong username error, not a 'malformed URL' error because, maybe, I sent a wrong parameter from inside my test.

This is my opinion.

miguelsorianod:
> If you are expecting an error to happen, that means that you exactly know what is wrong and why. It doesn't matter the contract.

I don't think that's accurate. The contract does matter. If the methods your method is consuming do not specify the set of errors returned in the contract and you don't check all of them then it becomes infeasible for your method to have a contract about the methods it will return unless you overwrite all the errors returned with your own errors.

> If I call a method with a wrong username, I expect to receive a wrong username error, not a 'malformed URL' error because, maybe, I sent a wrong parameter from inside my test.
I don't necessarily agree with this. It all depends on what are you interested in testing. The specific error or an error occurrence.

ziccardi:
> I don't think that's accurate. The contract does matter. If the methods your method is consuming do not specify the set of errors returned in the contract and you don't check all of them then it becomes infeasible for your method to have a contract about the methods it will return unless you overwrite all the errors returned with your own errors.

In a given situation, the error returned will be just one, and that is what the unit test does: it runs method in a controlled environment to make the outcome predictable, so you just have to check one error.

> I don't necessarily agree with this. It all depends on what are you interested in testing. The specific error or an error occurrence.

I believe that is inaccurate. Suppose I test a login function passing a wrong username. The login function is:
```
func login(url string, username string, pwd string)
```

and I make an error in my test, passing `http:/myloginurl.com, baduser, badpassword`.

If I just test for the function to return an error, my test will succeed, but it didn't test what it was supposed to.
I strongly believe that if in a test I expect a function to return an error, then I know what error I'm expecting, otherwise means I have no control of the environment my test is running into.

However, as I said, this is just my opinion.

pkg/api/cluster_types_test.go
miguelsorianod: I added `Parallel` calls to the subtests in the newly written tests.
I still wonder about the cost of adding this at cognitive load level vs the real need for parallel tests.

machi1990 (approved):
lgtm

Thanks @miguelsorianod 

I've asked @JameelB and @ziccardi to have another look as well.

internal/kafka/internal/services/kafka.go
```diff
@@ -1150,6 +1211,73 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 	return managedKafkaCR, nil
 }
 
+// buildReservedManagedKafkaCR buklds a Reserved Managed Kafka CR.
```
pawelpaszki, line 1214:
```suggestion
// buildReservedManagedKafkaCR builds a Reserved Managed Kafka CR.
```

miguelsorianod: fixed :+1:

internal/kafka/internal/handlers/data_plane_kafka_test.go
```diff
@@ -3,15 +3,19 @@ package handlers
...
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/private"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
 	v1 "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api/managedkafkas.managedkafka.bf2.org/v1"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/gorilla/mux"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/onsi/gomega"
```
ziccardi, line 18: why this?

miguelsorianod: using this to call `NewWithT`. Given that we are using an object now to call the expectations I felt it is clear to just reference the package name instead of importing the package attributes into the namespace to do it.

pkg/api/cluster_types.go
```diff
@@ -277,6 +278,36 @@ func (s *StrimziVersion) DeepCopy() *StrimziVersion {
...
+// in the StrimziVersion. If there are no Kafka versions nil is returned.
+// This method does not perform any sorting of the Kafka versions, it simply
+// returns the latest stored element. Use the StrimziVersionsDeepSort method
+// to get a semantically sorted StrimziVersion beforehand
+func (s *StrimziVersion) GetLatestKafkaVersion() *KafkaVersion {
+	var kafkaVersion *KafkaVersion
+
+	if len(s.KafkaVersions) == 0 {
+		return kafkaVersion
```
ziccardi, line 290:
```suggestion
		return nil
```

miguelsorianod: `kafkaVersion` is `nil`.

pkg/api/cluster_types.go
```diff
@@ -277,6 +278,36 @@ func (s *StrimziVersion) DeepCopy() *StrimziVersion {
...
+// returns the latest stored element. Use the StrimziVersionsDeepSort method
+// to get a semantically sorted StrimziVersion beorehand
+func (s *StrimziVersion) GetLatestKafkaIBPVersion() *KafkaIBPVersion {
+	var kafkaIBPVersion *KafkaIBPVersion
+
+	if len(s.KafkaIBPVersions) == 0 {
+		return kafkaIBPVersion
+	}
+
+	return &s.KafkaIBPVersions[len(s.KafkaIBPVersions)-1]
```
ziccardi, line 308:
```suggestion
	if count := len(s.KafkaIBPVersions); count == 0 {
		return nil
	} else {
		return &s.KafkaIBPVersions[count-1]
	}

	return &s.KafkaIBPVersions[len(s.KafkaIBPVersions)-1]
```

pkg/api/cluster_types.go
miguelsorianod: updated

pkg/api/cluster_types.go
```diff
@@ -277,6 +278,36 @@ func (s *StrimziVersion) DeepCopy() *StrimziVersion {
...
+// returns the latest stored element. Use the StrimziVersionsDeepSort method
+// to get a semantically sorted StrimziVersion beforehand
+func (s *StrimziVersion) GetLatestKafkaVersion() *KafkaVersion {
+	var kafkaVersion *KafkaVersion
+
+	if len(s.KafkaVersions) == 0 {
+		return kafkaVersion
+	}
+
+	return &s.KafkaVersions[len(s.KafkaVersions)-1]
```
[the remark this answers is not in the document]
miguelsorianod: :+1: I am going to change it. The `else` condition is not needed.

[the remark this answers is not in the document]
miguelsorianod: updated

pkg/api/cluster_types.go
miguelsorianod: updated https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1162#discussion_r914549479

ziccardi (approved):
Just one comment: I think that if we expect an error to occur during the tests, we should check that the error is the one we are expecting, to avoid it being an unrelated error and having the test pass without really testing what it was supposed to do.

Other than that, great job! Thanks @miguelsorianod !

miguelsorianod:
> Just one comment: I think that if we expect an error to occur during the tests, we should check that the error is the one we are expecting, to avoid it being an unrelated error and having the test pass without really testing what it was supposed to do.
> 
> Other than that, great job! Thanks @miguelsorianod !

Thanks for the review :+1: 

Regarding testing and error checking, I am not convinced about it for the mentioned reasons above, but I am more than willing to explore this concern together at some point, probably with some sample case to see what are the implications.

PR #1167, opened 2022-07-06
feat: take cluster max streaming unit into account for first ready cluster strategy

JameelB:
## Description

The following changes modifies the FirstReadyCluster placement strategy based on **Option 3** as described below. 

The total streaming unit used for each cluster that matches the criteria will need to be calculated in order to check if the cluster still has remaining capacity (total streaming unit used by existing Kafkas assigned to the cluster and the new Kafka Request does not go over the current cluster max streaming unit, "maxUnit").

### Why Use MaxUnit? 
There are multiple options to determine the capacity of a data plane cluster for kafka placement. 

#### Option 1: Determine using the cluster status
Dataplane cluster with no capacity, can no longer be scaled up, should be marked as 'full' by Fleetshard. This means, there would be no need to change the logic in FirstReadyCluster placement strategy as it will only match clusters with a 'Ready' status.

No additional calls to the database and computation is needed. However, the Fleetshard resync interval is only every 60s so there may be a slight delay in the status update. If Fleetshard is unable to communicate with the Fleet Manager, it will continue assigning Kafkas to this cluster as it is still deemed as 'ready'. 

An additional point made by @machi1990 below, the status 'full' is applied for the cluster as a whole. We need to accommodate for each of its supported instance type as they may not have the same remaining capacity left at the same time. 

#### Option 2: Use RemainingUnit capacity information
This information is available in the cluster object, under `dynamic_capacity_info`, updated by the Fleetshard sync. This represents the available streaming unit that can still be used by a Kafka instance in the cluster. It is based on the MaxUnit and the Kafka instances provisioned in the cluster so this value will change quite often.

With this, no additional calls to the database will be needed. We can simply check if remainingUnit is still larger than 0. However, the same scenarios for Option 1 still applies with the delay and if Fleetshard is unable to communicate with the Fleet Manager. 

#### Option 3: Calculate capacity based on existing Kafka Requests and MaxUnit capacity information
MaxUnit is available in the cluster object, under `dynamic_capacity_info`, updated by the Fleetshard sync. This value is based on the MaxNodes that is applied to a data plane cluster, which doesn't change that often. Even if Fleetshard is unable to communicate with the Fleet Manager/there is a delay, it will only assign Kafkas to the cluster as per the last MaxUnit received. 

The disadvantage here is that we would need to do an additional call to the database and computation to get the streaming unit count for existing Kafka Requests based on the cluster criteria. But with the additional checks based on the MaxUnit, which does not change often, this would allow us to minimise the amount of rejected KafkaRequests due to the cluster having no remaining capacity left.

## Verification Steps
TBC

- [ ] Code Review completed


internal/kafka/internal/services/kafka.go
```diff
@@ -914,6 +914,18 @@ func (kafkaStreamingUnitCountPerRegion KafkaStreamingUnitCountPerRegion) isSame(
 		kafkaStreamingUnitCountPerRegion.Region == kafkaPerRegionFromDB.Region
 }
 
+type KafkaStreamingUnitCountPerRegionList []KafkaStreamingUnitCountPerRegion
+
+func (ksuList KafkaStreamingUnitCountPerRegionList) GetStreamingUnitCountForCluster(clusterId string) float64 {
+	for _, kafkaStreamingUnitCountPerRegion := range ksuList {
+		if kafkaStreamingUnitCountPerRegion.ClusterId == clusterId {
+			return kafkaStreamingUnitCountPerRegion.Count
+		}
```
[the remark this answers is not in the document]
machi1990: `MaxUnits` is divided per instance type as well. 

If a cluster supports both `standard` and `developer`, there'll be a max unit for `standard` and another max unit for `developer`.

> What I really need is to get the streaming unit count per cluster, this method may return multiple results for each cluster depending if it supports multiple instance types. Will need to change this method to accommodate for that.

+1. We need to find the corresponding streaming unit count of an instance type in a given cluster.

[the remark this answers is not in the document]
JameelB: I completely forgot that I've added this to get the maxUnit :joy: Thanks for confirmation @machi1990 

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/0dc42c83e569b1ebd48301d4932fffa7f76ea947/internal/kafka/internal/services/cluster_placement_strategy.go#L68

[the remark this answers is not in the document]
JameelB: updated there :)

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -38,12 +40,39 @@ func (f *FirstReadyCluster) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cluster
...
+
+	// Get total number of streaming unit used per cluster that matches with criteria
+	streamingUnitCountPerRegionList, countStreamingUnitErr := f.KafkaService.CountStreamingUnitByRegionAndInstanceType()
+	if countStreamingUnitErr != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, countStreamingUnitErr, "failed to get count of streaming units by region and instance type")
+	}
+
+	streamingUnitsUsed := make(map[string]int32, len(clusters))
+
+	for _, cluster := range clusters {
```
machi1990, line 56: I think, this loop in line 56 and the one in line 62 can be combined.

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -38,12 +40,39 @@ func (f *FirstReadyCluster) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cluster
...
+		capacityInfo, getCapacityInfoErr := cluster.RetrieveDynamicCapacityInfo()
+		if getCapacityInfoErr != nil {
+			return nil, errors.NewWithCause(errors.ErrorGeneral, getCapacityInfoErr, "failed to retrieve capacity information for cluster %s", cluster.ClusterID)
+		}
+		maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
+
+		if streamingUnitsUsed[cluster.ClusterID] < maxStreamingUnits {
+			return cluster, nil
+		}
```
machi1990, line 71: Let's also include the size of the kafka as well. 
```go
		if (streamingUnitsUsed[cluster.ClusterID] + capacityConsumedOfTheNewKafka)  < maxStreamingUnits {
			return cluster, nil
		}
```
```suggestion
		if streamingUnitsUsed[cluster.ClusterID] < maxStreamingUnits {
			return cluster, nil
		}
```

machi1990: Something like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/fea39a658464424833b7742a13f4320b0f213537/internal/kafka/internal/services/cluster_placement_strategy.go#L131

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -38,12 +41,39 @@ func (f *FirstReadyCluster) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cluster
...
+	for _, cluster := range clusters {
+		currentStreamingUnitsUsed := streamingUnitCountPerRegionList.GetStreamingUnitCountForClusterAndInstanceType(cluster.ClusterID, kafka.InstanceType)
+
+		capacityInfo, getCapacityInfoErr := cluster.RetrieveDynamicCapacityInfo()
+		if getCapacityInfoErr != nil {
+			return nil, errors.NewWithCause(errors.ErrorGeneral, getCapacityInfoErr, "failed to retrieve capacity information for cluster %s", cluster.ClusterID)
+		}
+		maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
+
+		if currentStreamingUnitsUsed+instanceSize.CapacityConsumed < int(maxStreamingUnits) {
```
machi1990, line 70:
```suggestion
		if currentStreamingUnitsUsed+instanceSize.CapacityConsumed <= int(maxStreamingUnits) {
```
I think, this needs to be `<=`. e.g if maxUnit is `2`, capacityUsed is `1` and `capacityConsumed` is `1` as well, we want to accept the kafka as it can be placed onto the cluster. With `<` we'll end up in a situation where we do not accept this kafka onto this cluster.

internal/kafka/internal/services/kafka.go
```diff
@@ -931,7 +943,7 @@ type ClusterSelection struct {
 	SupportedInstanceType string
 }
 
-func (k *kafkaService) CountStreamingUnitByRegionAndInstanceType() ([]KafkaStreamingUnitCountPerRegion, error) {
+func (k *kafkaService) CountStreamingUnitByRegionAndInstanceType() (KafkaStreamingUnitCountPerRegionList, error) {
```
JameelB, line 946: This returns all Kafkas, no matter the status, as long instance type, size id and cluster id is set. Since I'm using this in the FirstReadyCluster strategy, this means Kafkas that have already been deleted from the Data Plane is still counted as taking up capacity in that cluster. Only until this Kafka is deleted from the Fleet Manager will that capacity be freed.

For context, once the Kafka has been deleted in the Data Plane, Fleetshard will update its status from `deprovision` to `deleting`. In the `deleting` phase, only quota from AMS needs to be removed then it's deleted in the Fleet Manager.

Would these Kafkas in a *deleting* state be okay to include in the capacity used? This function is also used for calculating the capacity_used metrics. What should be considered as a *used* cluster capacity?

machi1990:
> Option 1: Determine using the cluster status

+/- 0.5. 

We can't reliable do this as the status will have to be per instance type e.g you can have a situation where a cluster supports multiple instance types, and only one of the instance type is `full` (we can no longer accept a kafka of this instance type in this cluster). If we mark the cluster as full, then we'll also stop allocating other instance types in this cluster even though they can be accepted.  And then if we do not mark the cluster as full and not consider max units info, we might end up in a situation where we accept a Kafka with an instance type that's full for the given cluster. 

I feel like this option works best with `Option 3` that uses max units. I.e if we detect that all the instances types of a cluster are full, then we can mark the cluster as full. And if there is at least space in one of the instance type, we could mark the cluster as ready. Marking the cluster as full will reduce the number of clusters to process as only `ready` cluster will be retrieved.

> Option 2: Use RemainingUnit capacity information

-1 as well in using it. The volatility nature of this information and the delay it takes for us to get an update value (correct info at a given time `T`) makes me think we shouldn't use it for cluster placement. If we do, we might potentially find ourselves doing a lot of kafka re-assignment because we've overprovisioned the cluster.

> Option 3: Calculate capacity based on existing Kafka Requests and MaxUnit capacity information

+1 on going for this approach. It could work along side Option 1 to limit the `clusters` results set: i.e we could add Option 1 to this at some point in the future to improve perf.

internal/kafka/internal/services/kafka.go
machi1990: Very good point. 

What is done in this PR (including the `deleting` and `deprovision`) is no different from what was done for the manual cluster placement strategy which calls https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/7c9fb1601a64322f38936aaadb70d1c9876b3b7f/internal/kafka/internal/services/clusters.go#L392 at least we do not have an inconsistency ;-) 

But you are making a good case here and we probably should stop counting the `deleting` kafka in capacity used in a cluster since they've been removed from the data plane already.

What do others think? @miguelsorianod @VanillaSpoon @JameelB 

Also cc @pb82 and @davidffrench

miguelsorianod: +1 to not count clusters in `deleting` state as occupying resources

pb82: +1 in that case

internal/kafka/internal/services/clusters.go
```diff
@@ -665,3 +668,120 @@ func (c clusterService) IsStrimziKafkaVersionAvailableInCluster(cluster *api.Clu
...
+		Model(&api.Cluster{}).
+		Where("cluster_id != ''")
+
+	if err := dbConn.Scan(&clusters).Error; err != nil {
+		return nil, apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to list clusters")
+	}
+
+	streamingUnitsCountPerRegion := KafkaStreamingUnitCountPerClusterList{}
+
+	// pre-populate regions count with zero count values.
```
miguelsorianod, line 731: I am not sure I understand this comment. Could you elaborate on that? by default the zero-value of `int32` is 0

internal/kafka/internal/services/clusters.go
```diff
@@ -665,3 +668,120 @@ func (c clusterService) IsStrimziKafkaVersionAvailableInCluster(cluster *api.Clu
...
+	dbConn = c.connectionFactory.New()
+	var kafkasPerRegion []*KafkaPerClusterCount
+	if err := dbConn.Model(&dbapi.KafkaRequest{}).
+		Select("cloud_provider, region, count(1) as Count, size_id, cluster_id, instance_type").
+		Group("size_id, cluster_id, cloud_provider, region, instance_type").
+		Scan(&kafkasPerRegion).Error; err != nil {
+		return nil, apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to perform count query on kafkas table")
+	}
+
+	for _, kafkaCountPerRegion := range kafkasPerRegion {
```
miguelsorianod, line 771: Wouldn't the name of the iterator variable be `kafkaCountPerCluster`? I understand each element of `kafkasPerRegion` would be a struct containing the count of kafkas in a specific cluster in a region

internal/kafka/internal/services/clusters.go
```diff
@@ -665,3 +668,120 @@ func (c clusterService) IsStrimziKafkaVersionAvailableInCluster(cluster *api.Clu
...
+
+	streamingUnitsCountPerRegion := KafkaStreamingUnitCountPerClusterList{}
+
+	// pre-populate regions count with zero count values.
+	// This is useful and it ensures that the count drops to 0 for each instance type and region when all the kafkas are removed
+	for _, clusterSelection := range clusters {
+		supportedInstanceTypes := strings.Split(clusterSelection.SupportedInstanceType, ",")
+
+		c := &api.Cluster{DynamicCapacityInfo: clusterSelection.DynamicCapacityInfo}
+		clusterDynamicCapacityInfo, _ := c.RetrieveDynamicCapacityInfo()
```
miguelsorianod, line 737: The error shouldn't be ignored

machi1990: An error won't occur here. 
We'll be parsing a valid json at this point as we'd have stored a valid json.
The `RetrieveDynamicCapacityInfo()` method could also be made to not return the error at all.

machi1990: +1, I'd agree on that.

machi1990 (approved):

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test/mocks/kasfleetshardsync"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/metrics"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/test/mocks"
+
+	"github.com/onsi/gomega"
+)
+
+// Test Kafka creation with manual scaling enabled
+func TestKafkaCreate_ManualScaling(t *testing.T) {
```
machi1990, line 29: I've approved based on the integration test running in CI.
I've not tried running them when the env is `development`
Is this something that you've done? @JameelB

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+			if testcase.setup != nil {
+				testcase.setup()
+			}
+
+			testcase.assert()
+
+			if testcase.cleanup != nil {
+				testcase.cleanup()
+			}
+		})
```
machi1990, line 235: I really love these test arrangement.

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+	// run mock fleetshard sync
+	mockKasFleetshardSyncBuilder := kasfleetshardsync.NewMockKasFleetshardSyncBuilder(h, t)
+	mockKasFleetshardSync := mockKasFleetshardSyncBuilder.Build()
+	mockKasFleetshardSync.Start()
+	defer mockKasFleetshardSync.Stop()
+
+	// set up data plane cluster
+	clusterID, getClusterErr := common.GetOSDClusterIDAndWaitForStatus(h, t, api.ClusterReady)
+	g.Expect(getClusterErr).ToNot(gomega.HaveOccurred(), "failed to get dataplane cluster details: %v", getClusterErr)
+	g.Expect(clusterID).ToNot(gomega.BeEmpty(), "no dataplane cluster found")
```
machi1990, line 55: This might not be needed as we can right away use the manual list to put the cluster in the database in a given status.
This can save us some test time

machi1990: i.e just keep what's done in line 68 with some hardcoded values. 

Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1167/files#r924748354

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+
+	h, client, teardown := kafkatest.NewKafkaHelperWithHooks(t, ocmServer, nil)
+	defer teardown()
+
+	mockKasFleetshardSyncBuilder := kasfleetshardsync.NewMockKasFleetshardSyncBuilder(h, t)
+	mockKasfFleetshardSync := mockKasFleetshardSyncBuilder.Build()
+	mockKasfFleetshardSync.Start()
+	defer mockKasfFleetshardSync.Stop()
+
+	clusterID, getClusterErr := common.GetRunningOsdClusterID(h, t)
```
machi1990, line 457: we could instead directly seed the cluster in the database with a `ready` status to save some tests time

internal/kafka/internal/services/kafka.go
machi1990: @JameelB can we create a follow up for this discussion so that we do not forget about it? Thanks

internal/kafka/test/integration/kafka_create_test.go
machi1990: Thanks and :crossed_fingers:

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+				},
+			}
+		}
+	})
+	defer teardown()
+
+	// set up data plane cluster
+	// The mock fleetshard sync only updates the cluster if it's in a non-ready state. Only wait until the cluster has status
+	// waiting_for_kas_fleetshard_operator to ensure that the mock fleetshard sync updates the dynamic capacity info correctly.
+	clusterID, getClusterErr := common.GetOSDClusterIDAndWaitForStatus(h, t, api.ClusterWaitingForKasFleetShardOperator)
```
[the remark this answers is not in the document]
JameelB: I don't think this is coupled with the cluster_mgr_test. If there wasn't any cluster available, this GetOSDCluster should create a new one.

Alternative is to create a cluster, with a `waiting_for_fleetshard_operator` status, by manually creating it with db client. It would save time by bypassing cluster reconcile up to this point. Would this be preferred?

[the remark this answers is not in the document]
machi1990: > I don't think this is coupled with the cluster_mgr_test. If there wasn't any cluster available, this GetOSDCluster should create a new one.

Where I see the coupling is that this method call will wait for the cluster to be almost fully terrraformed thus following the reconciliation logic similar to cluster_mgr_test. This method call will then either reuses the cluster from the cluster_mgr_test or spins up a new one and starts to terraform it until it reaches that stataus. 

> Alternative is to create a cluster, with a waiting_for_fleetshard_operator status, by manually creating it with db client. It would save time by bypassing cluster reconcile up to this point. Would this be preferred?

+1. The terraforming process takes a bit of time and thus my preference of directly seeding the cluster in the database with the given status. To ensure that we won't have any issues with the seeded cluster not being available in OCM, the seeded cluster ProviderType has to be `standalone`.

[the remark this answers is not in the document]
JameelB: sounds good :+1:

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+	mockKasfFleetshardSync.Start()
+	defer mockKasfFleetshardSync.Stop()
+
+	clusterID, getClusterErr := common.GetRunningOsdClusterID(h, t)
+	if getClusterErr != nil {
+		t.Fatalf("Failed to retrieve cluster details: %v", getClusterErr)
+	}
+	if clusterID == "" {
+		panic("No cluster found")
+	}
```
machi1990, line 463:
```suggestion
	db := h.DBFactory().New()
	cluster := clusterMocks.BuildCluster(func(cluster *api.Cluster) {
		cluster.Meta = api.Meta{
			ID: api.NewID(),
		}
		cluster.ProviderType = api.ClusterProviderStandalone
		cluster.SupportedInstanceType = api.AllInstanceTypeSupport.String()
		cluster.ClientID = "some-client-id"
		cluster.ClientSecret = "some-client-secret"
		cluster.ClusterID = "some-cluster-id"
		cluster.Region = mocks.MockCluster.Region().ID()
		cluster.CloudProvider = mocks.MockCluster.CloudProvider().ID()
		cluster.Status = api.ClusterReady
		cluster.ProviderSpec = api.JSON{}
		cluster.ClusterSpec = api.JSON{}
	})

	err := db.Create(cluster).Error
	g.Expect(err).NotTo(gomega.HaveOccurred())
```

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+		}
+	})
+	defer teardown()
+
+	// set up data plane cluster
+	// The mock fleetshard sync only updates the cluster if it's in a non-ready state. Only wait until the cluster has status
+	// waiting_for_kas_fleetshard_operator to ensure that the mock fleetshard sync updates the dynamic capacity info correctly.
+	clusterID, getClusterErr := common.GetOSDClusterIDAndWaitForStatus(h, t, api.ClusterWaitingForKasFleetShardOperator)
+	g.Expect(getClusterErr).ToNot(gomega.HaveOccurred(), "failed to get dataplane cluster details: %v", getClusterErr)
+	g.Expect(clusterID).ToNot(gomega.BeEmpty(), "no dataplane cluster found")
```
machi1990, line 269:
```suggestion
	db := h.DBFactory().New()
	cluster := clusterMocks.BuildCluster(func(cluster *api.Cluster) {
		cluster.Meta = api.Meta{
			ID: api.NewID(),
		}
		cluster.ProviderType = api.ClusterProviderStandalone
		cluster.SupportedInstanceType = api.StandardTypeSupport.String()
		cluster.ClientID = "some-client-id"
		cluster.ClientSecret = "some-client-secret"
		cluster.ClusterID = "some-cluster-id"
		cluster.Region = mocks.MockCluster.Region().ID()
		cluster.CloudProvider = mocks.MockCluster.CloudProvider().ID()
		cluster.Status = api.ClusterWaitingForKasFleetShardOperator
		cluster.ProviderSpec = api.JSON{}
		cluster.ClusterSpec = api.JSON{}
		cluster.DynamicCapacityInfo = api.JSON([]byte(`{"standard":{"max_nodes":1}}`))
	})

	clusterErr := db.Create(cluster).Error
	g.Expect(clusterErr).NotTo(gomega.HaveOccurred())
```

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+
+	var enableManualScaling bool
+	clusterList := config.ClusterList{}
+	h, client, teardown := kafkatest.NewKafkaHelperWithHooks(t, ocmServer, func(d *config.DataplaneClusterConfig) {
+		if enableManualScaling {
+			// Enable manual scaling and register data plane cluster to cluster config
+			d.DataPlaneClusterScalingType = config.ManualScaling
+			d.ClusterConfig = config.NewClusterConfig(clusterList)
+		}
+	})
```
machi1990, line 43:
```suggestion
	clusterID := "some-cluster-id"
	clusterList := config.ClusterList{config.ManualCluster{
		Name:                  "test-cluster",
		ClusterId:             clusterID,
		CloudProvider:         mocks.MockCluster.CloudProvider().ID(),
		Region:                mocks.MockCluster.Region().ID(),
		MultiAZ:               true,
		Schedulable:           true,
		KafkaInstanceLimit:    2,
		ProviderType:          api.ClusterProviderStandalone,
		ClusterDNS:            "somec-cluster-dns",
		Status:                api.ClusterStatus(api.ClusterWaitingForKasFleetShardOperator.String()),
		SupportedInstanceType: api.AllInstanceTypeSupport.String(),
	}}

	h, client, teardown := kafkatest.NewKafkaHelperWithHooks(t, ocmServer, func(d *config.DataplaneClusterConfig) {
		// Enable manual scaling and register data plane cluster to cluster config
		d.DataPlaneClusterScalingType = config.ManualScaling
		d.ClusterConfig = config.NewClusterConfig(clusterList)
	})
```

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -0,0 +1,528 @@
...
+		KafkaInstanceLimit:    2,
+		Status:                cluster.Status,
+		SupportedInstanceType: cluster.SupportedInstanceType,
+	})
+	enableManualScaling = true
+
+	err = h.Env.CreateServices()
+	g.Expect(err).ToNot(gomega.HaveOccurred(), "unable to initialize testing environment: %v", err)
+
+	h.Env.Start()
```
machi1990, line 84:
```suggestion
	// ensures that the data plane is ready
	_, getClusterErr := common.WaitForClusterStatus(h.DBFactory(), &kafkatest.TestServices.ClusterService, clusterID, api.ClusterReady)
	g.Expect(getClusterErr).ToNot(gomega.HaveOccurred(), "failed to get dataplane cluster details: %v", getClusterErr)
```

machi1990 (commented):
Some suggested changes to direcly seed into the database.

I tried the tests against a integration env and they all passed.

Let me know what you think.

internal/kafka/test/integration/kafka_create_test.go
JameelB: Thanks, I've done the exact same :D

internal/kafka/test/integration/kafka_create_test.go
JameelB: done the same here too, didn't need to provide a cluster dns though. it still passed :thinking:

internal/kafka/test/integration/kafka_create_test.go
JameelB: i've done the same here

JameelB:
Thanks for the suggestions @machi1990! I've done the same and i'm currently running the tests on dev env with those changes. Takes awhile :) 

I'll push my changes once done, there was a test that failed which I'll need to investigate.

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -18,6 +18,8 @@ func NewClusterPlacementStrategy(clusterService ClusterService, dataplaneCluster
 	var clusterSelection ClusterPlacementStrategy
 	if dataplaneClusterConfig.IsDataPlaneManualScalingEnabled() {
 		clusterSelection = &FirstSchedulableWithinLimit{dataplaneClusterConfig, clusterService, kafkaConfig}
+	} else if dataplaneClusterConfig.IsDataPlaneAutoScalingEnabled() {
+		clusterSelection = &FirstReadyWithCapacity{clusterService, kafkaConfig}
 	} else {
 		clusterSelection = &FirstReadyCluster{clusterService}
 	}
```
ziccardi, line 25: Instead of `if ... else if ... else` I think a switch is more readable

```suggestion
	switch {
		case dataplaneClusterConfig.IsDataPlaneManualScalingEnabled():
			clusterSelection = &FirstSchedulableWithinLimit{dataplaneClusterConfig, clusterService, kafkaConfig}
		case dataplaneClusterConfig.IsDataPlaneAutoScalingEnabled():
			clusterSelection = &FirstReadyWithCapacity{clusterService, kafkaConfig}
		default:
			clusterSelection = &FirstReadyCluster{clusterService}
	}
```

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -129,3 +131,52 @@ func (f *FirstSchedulableWithinLimit) findClusterKafkaInstanceCount(clusterIDs [
...
 }
+
+// FirstReadyWithCapacity finds and returns the first cluster in a Ready status with remaining capacity
+type FirstReadyWithCapacity struct {
+	ClusterService ClusterService
+	KafkaConfig    *config.KafkaConfig
+}
+
+func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cluster, *errors.ServiceError) {
```
ziccardi, line 141: we should return `error` instead of `*errors.ServiceError`

internal/kafka/test/integration/kafka_create_test.go
machi1990: +1. I think registering the cluster manually during startup removed the need to Stop() and Start() the services.

internal/kafka/test/integration/kafka_create_test.go
machi1990: Already addressed

internal/kafka/internal/services/cluster_placement_strategy.go
JameelB: when this is changed to `error`, should the underlying error types returned from the functions these strategies call also be changed to type `error`?

ziccardi: All functions should return `error` unless there is a real strong reason to do otherwise IMO.

It is caller duty to cast the `error` interface to the concrete type if needed.

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -129,3 +131,52 @@ func (f *FirstSchedulableWithinLimit) findClusterKafkaInstanceCount(clusterIDs [
...
+		Status:                api.ClusterReady,
+		SupportedInstanceType: kafka.InstanceType,
+	}
+
+	clusters, findAllClusterErr := f.ClusterService.FindAllClusters(criteria)
+	if findAllClusterErr != nil {
+		return nil, findAllClusterErr
+	}
+
+	if len(clusters) != 0 {
```
ziccardi, line 156: With the change suggested above, this `if` can be removed.

internal/kafka/internal/services/cluster_placement_strategy.go
JameelB: I think this may be a substantial change. I'm happy to follow up on this in a separate PR, to make it easier for people to review this one. wdyt?

internal/kafka/internal/services/data_plane_cluster.go
```diff
@@ -112,10 +112,7 @@ func (d *dataPlaneClusterService) setClusterStatus(cluster *api.Cluster, status
 		return err
 	}
 
-	dynamicCapacityInfo, err := cluster.RetrieveDynamicCapacityInfo()
-	if err != nil {
-		return err
-	}
+	dynamicCapacityInfo := cluster.RetrieveDynamicCapacityInfo()
```
ziccardi, line 115: see previous comment

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -505,10 +505,7 @@ func (c *ClusterManager) reconcileDynamicCapacityInfo(cluster api.Cluster) error
 	updatedDynamicCapacityInfo := map[string]api.DynamicCapacityInfo{}
 
 	if c.DataplaneClusterConfig.IsDataPlaneAutoScalingEnabled() {
-		previousDynamicCapacityInfo, err := cluster.RetrieveDynamicCapacityInfo()
-		if err != nil {
-			return err
-		}
+		previousDynamicCapacityInfo := cluster.RetrieveDynamicCapacityInfo()
```
ziccardi, line 508: Why do we ignore the error?

internal/kafka/internal/workers/clusters_mgr_test.go
```diff
@@ -3849,10 +3849,7 @@ func TestClusterManager_reconcileDynamicCapacityInfo(t *testing.T) {
 				},
 				clusterService: &services.ClusterServiceMock{
 					UpdateFunc: func(cluster api.Cluster) *apiErrors.ServiceError {
-						capacityInfoToBePersisted, err := cluster.RetrieveDynamicCapacityInfo()
-						if err != nil {
-							t.Error(err)
-						}
+						capacityInfoToBePersisted := cluster.RetrieveDynamicCapacityInfo()
```
ziccardi, line 3852: see other comments on this

internal/kafka/internal/workers/clusters_mgr_test.go
```diff
@@ -3924,10 +3921,7 @@ func TestClusterManager_reconcileDynamicCapacityInfo(t *testing.T) {
...
```
ziccardi, line 3924: see other comments on this

internal/kafka/test/integration/cluster_mgr_test.go
```diff
@@ -110,8 +110,7 @@ func TestClusterManager_SuccessfulReconcile(t *testing.T) {
 	g.Expect(checkReadyErr).NotTo(gomega.HaveOccurred(), "Error waiting for cluster to be ready: %s %v", cluster.ClusterID, checkReadyErr)
 
 	// check that the cluster has dynamic capacity info persisted
-	dynamicCapacityInfo, err := cluster.RetrieveDynamicCapacityInfo()
-	g.Expect(err).ToNot(gomega.HaveOccurred(), "Error retrieving dynamic capacity info")
+	dynamicCapacityInfo := cluster.RetrieveDynamicCapacityInfo()
```
ziccardi, line 113: see other comments on this

internal/kafka/internal/services/cluster_placement_strategy.go
ziccardi: I'd expect that most of the places where this function is called won't do much more than

```go
if err != nil {
   return nil
}
```

Unless there are many places that are directly referring to the `ServiceError` type when invoking this methods, the change shouldn't be too far from just changing the return to `error`.

That being said, I'm ok with doing this in a follow up PR: the reason I pointed it out here was that I think we should start following the _return error rule_ in the newly added functions.

internal/kafka/internal/services/data_plane_cluster.go
```diff
@@ -53,7 +53,7 @@ func (d *dataPlaneClusterService) GetDataPlaneClusterConfig(ctx context.Context,
 
 	dynamicCapacityInfo := map[string]api.DynamicCapacityInfo{}
 	if d.DataplaneClusterConfig.IsDataPlaneAutoScalingEnabled() {
-		dynamicCapacityInfo, _ = cluster.RetrieveDynamicCapacityInfo()
+		dynamicCapacityInfo = cluster.RetrieveDynamicCapacityInfo()
```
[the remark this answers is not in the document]
ziccardi: Ok, but then there are 2 choices IMO:

1. Change the `RetrieveDynamicCapacityInfo()` so that no errors are returned
2. Explicitly ignore it by assigning the error to `_`

internal/kafka/internal/services/clusters.go
ziccardi: +1 agree

internal/kafka/internal/services/data_plane_cluster.go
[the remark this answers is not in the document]
ziccardi: Just saw not that you proposed to change `RetrieveDynamicCapacityInfo` in the linked comment. Ignore this.

JameelB:
I've addressed your comments @machi1990 @ziccardi, thanks for reviewing! :pray: 

The integration tests using dev env has passed. The test `TestDataPlaneEndpoints_ReassignRejectedKafkaDueToInsufficientResources` timed out waiting on a kafka to be re-assigned to a cluster:
```
failed to set kafka per cluster count metrics: KAFKAS-MGMT-8: clusterID is undefined
failed to reconcile provisioning kafka cbc24hsliprpc131mtb0: Region us-east-1 cannot accept instance type: developer at this moment for kafka cbc24hsliprpc131mtb0
Waiting for kafka named 'test-kafka1' to have a ClusterID
```

I couldn't reproduce it though when I tried running it again (on both integration and development envs). This test also passed in the nightly run this morning. May be safe to ignore for now but will keep an eye on it in the future.

internal/kafka/internal/services/cluster_placement_strategy.go
JameelB: May not too bad in the service itself but I think most of the changes will be on the tests.

machi1990:
> I've addressed your comments @machi1990 @ziccardi, thanks for reviewing! pray

Thanks.

> failed to set kafka per cluster count metrics: KAFKAS-MGMT-8: clusterID is undefined

This should be an issue with the way the kafkas per cluster count metric is setup. I also wonder whether this metric is needed (it could still be used somewhere) but I think it is better to use the cluster_capacity_used metric in place and we can remove the metric from the code.
 
> failed to reconcile provisioning kafka cbc24hsliprpc131mtb0: Region us-east-1 cannot accept instance type: developer at this moment for kafka cbc24hsliprpc131mtb0
> Waiting for kafka named 'test-kafka1' to have a ClusterID

This seems like an issue which was fixed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1194 where the clusters in question didn't move to `ready` state causing the kafkas to not be re-assigned. 
 
> I couldn't reproduce it though when I tried running it again (on both integration and development envs). This test also passed in the nightly run this morning. May be safe to ignore for now but will keep an eye on it in the future.

+1, I agree maybe there are still some edge cases on that test to fix and we could investigate further when the issue re-occurs again.

pkg/api/cluster_types.go
```diff
@@ -490,18 +490,14 @@ func (cluster *Cluster) SetDynamicCapacityInfo(dynamicCapacityInfo map[string]Dy
...
 
 // RetrieveDynamicCapacityInfo returns the dynamic scaling info per instance type
-func (cluster *Cluster) RetrieveDynamicCapacityInfo() (map[string]DynamicCapacityInfo, error) {
+func (cluster *Cluster) RetrieveDynamicCapacityInfo() map[string]DynamicCapacityInfo {
 	dynamicCapacityInfo := map[string]DynamicCapacityInfo{}
-	if cluster.DynamicCapacityInfo == nil {
-		return dynamicCapacityInfo, nil
+	if cluster.DynamicCapacityInfo != nil {
+		// ignore error returned by Unmarshal as the json stored in the cluster object will always be a valid DynamicCapacityInfo json object.
+		_ = json.Unmarshal(cluster.DynamicCapacityInfo, &dynamicCapacityInfo)
```
ziccardi, line 497: I'm ok with the function not returning any error, but I believe we shouldn't ignore JSON errors.
I prefer to log it if an error occurs: it will never happen, but in case we have at least some log about it.

JameelB: makes sense, i'll log it here instead :+1:

machi1990 (approved):

machi1990:
Changes looks good to me.
This can be merged once @ziccardi comments are addressed.

internal/kafka/internal/services/kafka.go
JameelB: jira created: https://issues.redhat.com/browse/MGDSTRM-9252

ziccardi (approved):
lgtm. Awesome job!

PR #1171, opened 2022-07-07
feat(cluster-metrics): add cluster capacity metrics for when dynamic scaling is turned on

machi1990:
## Description

Fixes https://issues.redhat.com/browse/MGDSTRM-7480 and
https://issues.redhat.com/browse/MGDSTRM-7481

This relies on the `MaxUnits` information which will be present in the database when dynamic scaling is turned on and FSO reports it back to us.

## Verification Steps

Unit and integration tests passing

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -964,8 +964,9 @@ type KafkaStreamingUnitCountPerRegion struct {
 	Region        string
 	InstanceType  string
 	ClusterId     string
-	Count         float64
+	Count         int32
 	CloudProvider string
+	MaxUnits      int32
 }
 
 func (kafkaStreamingUnitCountPerRegion KafkaStreamingUnitCountPerRegion) isSame(kafkaPerRegionFromDB *KafkaPerRegionCount) bool {
@@ -980,7 +981,7 @@ type KafkaPerRegionCount struct {
 	Region        string
 	InstanceType  string
 	ClusterId     string
-	Count         float64
+	Count         int32
 	CloudProvider string
 	SizeId        string
 }
@@ -990,14 +991,17 @@ type ClusterSelection struct {
 	ClusterID             string
 	Region                string
 	SupportedInstanceType string
+	DynamicCapacityInfo   api.JSON
 }
 
 func (k *kafkaService) CountStreamingUnitByRegionAndInstanceType() ([]KafkaStreamingUnitCountPerRegion, error) {
 
 	var clusters []*ClusterSelection
-	dbConn := k.connectionFactory.New()
-	// TODO ignore clusters that do not have a cluster_id set e.g the one in accepted state
-	if err := dbConn.Model(&api.Cluster{}).Scan(&clusters).Error; err != nil {
+	dbConn := k.connectionFactory.New().
+		Model(&api.Cluster{}).
+		Where("cluster_id != ''")
+
+	if err := dbConn.Scan(&clusters).Error; err != nil {
 		return nil, apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to list clusters")
 	}
 
@@ -1005,8 +1009,11 @@ func (k *kafkaService) CountStreamingUnitByRegionAndInstanceType() ([]KafkaStrea
 
 	// pre-populate regions count with zero count values.
 	// This is useful and it ensures that the count drops to 0 for each instance type and region when all the kafkas are removed
-	for _, cluster := range clusters {
-		supportedInstanceTypes := strings.Split(cluster.SupportedInstanceType, ",")
+	for _, clusterSelection := range clusters {
+		supportedInstanceTypes := strings.Split(clusterSelection.SupportedInstanceType, ",")
+
+		c := &api.Cluster{DynamicCapacityInfo: clusterSelection.DynamicCapacityInfo}
+		clusterDynamicCapacityInfo, _ := c.RetrieveDynamicCapacityInfo()
 
 		for _, supportedInstanceType := range supportedInstanceTypes {
 			instanceType := strings.TrimSpace(supportedInstanceType)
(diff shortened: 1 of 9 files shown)


internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -200,10 +199,6 @@ func (c *ClusterManager) processMetrics() []error {
 		return []error{errors.Wrapf(err, "failed to set kafka per cluster count metrics")}
 	}
 
-	if err := c.setClusterStatusMaxCapacityMetrics(); err != nil {
```
machi1990, line 203: This has moved into `kafkas_mgr.go`

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -1273,37 +1268,6 @@ func (c *ClusterManager) reconcileClusterIdentityProvider(cluster api.Cluster) e
 	return nil
 }
 
-func (c *ClusterManager) setClusterStatusMaxCapacityMetrics() error {
```
machi1990, line 1276: moved into `kafkas_mgr.go`

pkg/metrics/metrics.go
```diff
@@ -727,7 +728,6 @@ func ResetMetricsForClusterManagers() {
 	clusterStatusSinceCreatedMetric.Reset()
 	clusterStatusCountMetric.Reset()
 	kafkaPerClusterCountMetric.Reset()
-	clusterStatusCapacityMaxMetric.Reset()
```
machi1990, line 730: moved as part of kafkas_mgr.go hence the removal from here

internal/kafka/internal/services/kafka.go
```diff
@@ -1044,23 +1057,9 @@ func (k *kafkaService) CountStreamingUnitByRegionAndInstanceType() ([]KafkaStrea
...
 			if streamingUnitCountPerRegion.isSame(kafkaCountPerRegion) {
 				streamingUnitsCountPerRegion[i].Count += streamingUnitCount
-				resultPresent = true
 				break
 			}
 		}
-
-		// We should never reach here in manual scaling
-		// TODO review if this is needed when dynamic scaling is implemented
-		if !resultPresent {
```
machi1990, line 1054: This has been removed as a consideration in this work. 
For us to arrive here, it'll mean the kafka is not assigned to the cluster yet: 
- either, accepted but not assigned to a cluster 
- or accepted assigned to a cluster `A`, got rejected in the data plane, removed from cluster `A` and it'll be re-assigned onto another cluster at the next reconciler run


For all the above case, we'll have a different and dedicated metric for Kafkas that are not assigned to a cluster yet. 
This will be done as part of the work that's allows for a kafka to be accepted even when we do not have a ready data plane yet but regions limits have not been reached.

internal/kafka/internal/workers/clusters_mgr_test.go
```diff
@@ -294,19 +272,18 @@ func TestClusterManager_processMetrics(t *testing.T) {
...
+			g := NewWithT(t)
 			c := &ClusterManager{
 				ClusterManagerOptions: ClusterManagerOptions{
 					ClusterService:         tt.fields.clusterService,
 					DataplaneClusterConfig: tt.fields.dataplaneClusterConfig,
 					SupportedProviders:     tt.fields.supportedProviders,
 				},
 			}
-			Expect(len(c.processMetrics()) > 0).To(Equal(tt.wantErr))
+			g.Expect(len(c.processMetrics()) > 0).To(Equal(tt.wantErr))
```
ziccardi, line 286: Mind adding a comment to explain why `len(c.processMetrics()) > 0` is related to `tt.wantErr` ?

ziccardi: That's perfect! Thank you!

ziccardi: It is outside of the scope of this PR, but I was wondering if it wouldn't be better/more standard to return just one `error` containing a method (for example: `GetCause error[]`) that will return the list of errors.

machi1990: I think the idea behind return the error array was to able to iterate through all of them and capture them in sentry / logs.
This was driven by the `.Reconcile() []error` method signature. 
What will be the gain by encapsulating the errors in an error container with a `GetCause()`?

VanillaSpoon:
Hey,
lgtm, I'll wait for others to have a look too :)

Thanks for adding comments with explanations!

machi1990:
> Hey, lgtm, I'll wait for others to have a look too :)
> 
> Thanks for adding comments with explanations!

Thanks for the review @VanillaSpoon

JasonPowr:
LGTM :)

machi1990:
> LGTM :)

Thanks for the review :-)

internal/kafka/internal/workers/clusters_mgr_test.go
ziccardi: Because when I invoke a method, I expect the `error` to be `nil` when no error has occurred, I don't expect it to be an empty array.

And when the error occurs I expect it to be a single object

internal/kafka/internal/services/kafka.go
```diff
@@ -966,6 +966,7 @@ type KafkaStreamingUnitCountPerRegion struct {
 	ClusterId     string
 	Count         float64
 	CloudProvider string
+	MaxUnits      float64
```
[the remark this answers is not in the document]
machi1990: Good point. This was just my attempting to avoid cascading casting all over as the metrics needs a `float64`. It is the same type for the `Count` field. Do you think we can make them as int at metric level?

internal/kafka/internal/workers/clusters_mgr_test.go
ziccardi: That's why I proposed an error with a `GetCause`  method. If no error is found, it would return `nil`. If an error is found it would return en error `Failure processing metrics` and the `GetCause` would contain the list of errors.

ziccardi: About the `Log`, I would avoid that method: I prefer each object to have only one responsibility. We can however make a `SentryLogger` that, when a `CompoundError` is received, would logs each of the embedded errors.

machi1990: It's probably not worth creating another struct. We could extend the `ServiceError` to suit the multiple causes use case: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/622303cad34fa07ba2aa29841104e6f239c24706/pkg/errors/errors.go#L312 which we've at the moment to allow taking in a list of error causes instead of just one: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/622303cad34fa07ba2aa29841104e6f239c24706/pkg/errors/errors.go#L320

And it sees how everything fits together with the reconciler. As you mentioned, outside the scope of this PR but I am willing to chime in a followup discussion that will be opened.

internal/kafka/internal/services/kafka.go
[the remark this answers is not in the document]
JameelB: that looks good to me :) Thanks!

JameelB (approved):

PR #1172, opened 2022-07-07
test: increase coverage for pkg/workers

JasonPowr:
## Description
Increase coverage for pkg/workers
jira: https://issues.redhat.com/browse/MGDSTRM-7990

## Verification Steps
1. Run `make test`
2. Run `make test/html/coverage/report `
3. Should be able to see the increased code coverage

- [ ] Code Review completed

### DIFF
--- a/pkg/workers/leader_election_mgr_test.go
+++ b/pkg/workers/leader_election_mgr_test.go
@@ -3,10 +3,12 @@ package workers
 import (
 	"errors"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/onsi/gomega"
 	mocket "github.com/selvatico/go-mocket"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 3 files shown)


pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,365 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+			s := &LeaderElectionManager{
+				workers:                                tt.fields.workers,
+				connectionFactory:                      tt.fields.connectionFactory,
+				tearDown:                               tt.fields.tearDown,
+				leaderElectionReconcilerRepeatInterval: tt.fields.leaderElectionReconcilerRepeatInterval,
+				leaderLeaseExpirationTime:              tt.fields.leaderLeaseExpirationTime,
+				workerGrp:                              sync.WaitGroup{},
+			}
+			s.Start()
+		})
```
machi1990, line 299: There is no assertion here, is this desired? If so, why?

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,365 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+			tt.setupFn()
+			s := &LeaderElectionManager{
+				workers:                                tt.fields.workers,
+				connectionFactory:                      tt.fields.connectionFactory,
+				tearDown:                               tt.fields.tearDown,
+				leaderElectionReconcilerRepeatInterval: tt.fields.leaderElectionReconcilerRepeatInterval,
+				leaderLeaseExpirationTime:              tt.fields.leaderLeaseExpirationTime,
+				workerGrp:                              sync.WaitGroup{},
+			}
+			s.startWorkers()
```
machi1990, line 406: There is no assertion here, why?

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,365 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			tt.setupFn()
+			s := &LeaderElectionManager{
+				workers:           tt.fields.workers,
+				connectionFactory: tt.fields.connectionFactory,
+			}
+			if got := s.isWorkerLeader(tt.args.worker); got != tt.want {
+				t.Errorf("LeaderElectionManager.isWorkerLeader() = %v, want %v", got, tt.want)
+			}
```
machi1990, line 548: let's use `gomega` for assertion

pkg/workers/reconciler_config_test.go
```diff
@@ -0,0 +1,58 @@
...
+		LeaderLeaseExpirationTime              time.Duration
+		LeaderElectionReconcilerRepeatInterval time.Duration
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		wantErr bool
+	}{
+		{
+			name: "should return nil if it is able to read the files successfully",
```
machi1990, line 36:
```suggestion
			name: "should always return nil ",
```

pkg/workers/worker_interface_test.go
```diff
@@ -0,0 +1,268 @@
...
+		fields fields
+		want   bool
+	}{
+		{
+			name: "should return true if the worker is running",
+			fields: fields{
+				isRunning: true,
+			},
+			want: true,
+		},
```
machi1990, line 149:
```suggestion
		{
			name: "should return true if the worker is running",
			fields: fields{
				isRunning: true,
			},
			want: true,
		},
		{
			name: "should return false if the worker is not running",
			fields: fields{
				isRunning: false,
			},
			want: false,
		},
```

pkg/workers/worker_interface_test.go
```diff
@@ -0,0 +1,268 @@
...
+	type args struct {
+		w Worker
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+	}{
+		{
+			name: "should start then stop the worker",
```
machi1990, line 221: Why start and then stop the worker? Shouldn't this be testing start and stop methods separately?

pkg/workers/worker_interface_test.go
```diff
@@ -0,0 +1,268 @@
...
+			b := &BaseWorker{
+				Id:         tt.fields.Id,
+				WorkerType: tt.fields.WorkerType,
+				Reconciler: tt.fields.Reconciler,
+				isRunning:  tt.fields.isRunning,
+				imStop:     tt.fields.imStop,
+			}
+			b.StartWorker(tt.args.w)
+			b.StopWorker(tt.args.w)
+		})
```
machi1990, line 266: No any assertions are performed here, is this desired?

pkg/workers/leader_election_mgr_test.go
JasonPowr: Mainly because the method does not return anything to check, So I wasnt sure maybe I could check wantErr?

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,365 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+				workers:                                []Worker{},
+				connectionFactory:                      &db.ConnectionFactory{},
+				tearDown:                               nil,
+				leaderElectionReconcilerRepeatInterval: 15 * time.Second,
+				leaderLeaseExpirationTime:              1 * time.Minute,
+				workerGrp:                              sync.WaitGroup{},
+			},
+		},
+	}
+	g := NewWithT(t)
```
ziccardi, line 223: NewWithT is meant to be called for each subtest. Move this into `t.Run` please

pkg/workers/worker_interface_test.go
machi1990: +1, let's separate them.

pkg/workers/reconciler_config_test.go
```diff
@@ -0,0 +1,58 @@
...
+func Test_AddFlags(t *testing.T) {
+	emptyFlagSet := &pflag.FlagSet{}
+	config := NewReconcilerConfig()
+	config.AddFlags(emptyFlagSet)
+
+	g := NewWithT(t)
+
+	flag, err := emptyFlagSet.GetDuration("reconciler-repeat-interval")
+
+	g.Expect(err).To(BeNil())
```
ziccardi, line 20:
```suggestion
	g.Expect(err).NotTo(HaveOccurred())
```

pkg/workers/reconciler_config_test.go
```diff
@@ -0,0 +1,58 @@
...
+			name: "should return nil if it is able to read the files successfully",
+			fields: fields{
+				ReconcilerRepeatInterval:               30 * time.Second,
+				LeaderLeaseExpirationTime:              1 * time.Minute,
+				LeaderElectionReconcilerRepeatInterval: 15 * time.Second,
+			},
+			wantErr: false,
+		},
+	}
+	g := NewWithT(t)
```
ziccardi, line 45: Move this into `t.Run(...`

pkg/workers/leader_election_mgr_test.go
ziccardi: Moreover, you could pass some `mock worker` instead of an empty array and check that the worker method are invoked with the correct params

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,407 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+			s := &LeaderElectionManager{
+				workers:                                tt.fields.workers,
+				connectionFactory:                      tt.fields.connectionFactory,
+				tearDown:                               tt.fields.tearDown,
+				leaderElectionReconcilerRepeatInterval: tt.fields.leaderElectionReconcilerRepeatInterval,
+				workerGrp:                              sync.WaitGroup{},
+			}
+			s.Start()
+			g.Expect(tt.fields.workers[0].GetID()).To(Equal("01"))
+			g.Expect(tt.fields.workers[0].GetWorkerType()).To(Equal("cluster"))
```
ziccardi, line 299: Here you are just calling the mock function you just created. I don't think this is what you wanted to do.
I think would be better to add something like: 
```suggestion
			g.Expect(tt.fields.workers[0].calls.GetID).To(HaveLen(1))
			g.Expect(tt.fields.workers[0].calls.GetWorkerType).To(HaveLen(1))
```

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,385 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+				workers:                                tt.fields.workers,
+				connectionFactory:                      tt.fields.connectionFactory,
+				tearDown:                               tt.fields.tearDown,
+				leaderElectionReconcilerRepeatInterval: tt.fields.leaderElectionReconcilerRepeatInterval,
+				workerGrp:                              sync.WaitGroup{},
+			}
+			s.Start()
+			workerMock := tt.fields.workers[0].(*WorkerMock)
+			g.Expect(workerMock.calls.GetID).To(HaveLen(2))
+			g.Expect(workerMock.calls.GetWorkerType).To(HaveLen(1))
```
machi1990, line 300: we also need to check if `.Start()` was called.

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,387 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+	for _, testcase := range tests {
+		tt := testcase
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			s := &LeaderElectionManager{
+				workers:   tt.fields.workers,
+				tearDown:  tt.fields.tearDown,
+				workerGrp: sync.WaitGroup{},
+			}
+			s.Stop()
```
machi1990, line 356: let's add some assertion after this call as well

e.g check if `.Stop()` was called etc

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,385 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+		leaderElectionReconcilerRepeatInterval time.Duration
+		workerGrp                              *sync.WaitGroup
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		setupFn func()
+	}{
+		{
+			name: "should start the worker if the worker is a leader and is not running",
```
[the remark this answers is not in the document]
JasonPowr: I attempted to add more test cases for the tests in this file  but every time I tried I kept running into issues with the sync.WaitGroup{}. It would constantly panic with the error `panic: sync: negative WaitGroup counter`. I was able to get around this by using .add(). But doing it this way broke the linter eg `copylocks: literal copies lock value from *tt.fields.workerGrp: sync.WaitGroup contains sync.noCopy`

pkg/workers/leader_election_mgr_test.go
pawelpaszki: Even though we will only confirm what we set in the mocks, `IsRunning()` can be checked like this:

```
	for _, testcase := range tests {
		tt := testcase
			g := NewWithT(t)
			t.Parallel()
			s := &LeaderElectionManager{
				workers:   tt.fields.workers,
				tearDown:  tt.fields.tearDown,
				workerGrp: sync.WaitGroup{},
			}
			if (len(s.workers)) > 0 {
				g.Expect(s.workers[0].IsRunning()).To(BeTrue())
			}
			s.Stop()
		})
	}
```

I find the workers stuff tricky to test and I am not familiar with this part of kfm code

ziccardi: @pawelpaszki Not sure `g.Expect(s.workers[0].IsRunning()).To(BeTrue())`: that would simply invoke the mock `IsRunningFunc` method that we defined here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1172/files/13feb35a18ff857e20f695a7eaafe25f1aed2cc2#diff-f95dcc5a7c0365d81ab0b0499b14bb8222240cde002620e604fe4e02cb5f9d57R335

I believe the test here should check that the mock methods are correctly invoked. Maybe something like:
```go
g.Expect(s.workers[0].calls.Stop).To(HaveLen(1))
```

pawelpaszki: makes sense @ziccardi

pkg/workers/leader_election_mgr_test.go
```diff
@@ -3,10 +3,12 @@ package workers
 import (
 	"errors"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	. "github.com/onsi/gomega"
```
JameelB, line 11: After the tech discussion we decided not to use `.` import alias for gomega any longer. Let's change each assert methods to use `gomega.` instead.

machi1990:
ping @ziccardi @pawelpaszki can you have another look at this PR?

machi1990:
lgtm, thanks @JasonPowr 

let's wait for a review from @ziccardi and @pawelpaszki before merging

pkg/workers/leader_election_mgr_test.go
```diff
@@ -185,3 +187,465 @@ func TestLeaderElectionManager_acquireLeaderLease(t *testing.T) {
...
+				leaderElectionReconcilerRepeatInterval: tt.fields.leaderElectionReconcilerRepeatInterval,
+				workerGrp:                              *tt.fields.workers[0].GetSyncGroup(),
+			}
+			if tt.fields.tearDown != nil {
+				s.Start()
+				s.Stop()
+				workerMock := tt.fields.workers[0].(*WorkerMock)
+				g.Expect(workerMock.calls.Stop).To(gomega.HaveLen(1))
+			}
+			s.Stop()
```
ziccardi, line 508: If `tearDown` is nil, no tests are performed, making the first test ("should return if teardown is nil") meaningless

pkg/workers/worker_interface_test.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	for _, testcase := range tests {
+		tt := testcase
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			t.Parallel()
+			b := &BaseWorker{
+				Id: tt.fields.Id,
+			}
+			g.Expect(b.GetID()).To(gomega.Equal(tt.want))
```
ziccardi, line 39: Most of the tests here show that a getter returns the value set a couple of lines above.
I'm not sure it makes sense to test getter/setters.
IMO, the only useful tests here are the tests for `StartWorker` and `StopWorker`

@JasonPowr @machi1990 @pawelpaszki WDYT?

pawelpaszki: I am on two minds here. getters/ setters should probably be tested if they are complex enough (rather then just setting/ getting values from some constants). I don't have a strong preference TBH. since the tests area already written I guess there is no harm in keeping them. WDYT?

ziccardi (commented):
LGTM. Great job!

ziccardi (approved):

pawelpaszki (approved):

PR #1174, opened 2022-07-08
feat: report a user friendly error message on data plane failed states

valeriiashapoval:
## Description
https://issues.redhat.com/browse/MGDSTRM-6982 - report a user friendly error message on data plane failed states

- [X] CI and all relevant tests are passing
- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/data_plane_kafka.go
+++ b/internal/kafka/internal/services/data_plane_kafka.go
@@ -89,11 +89,7 @@ func (d *dataPlaneKafkaService) UpdateDataPlaneKafkaService(ctx context.Context,
 			// when getStatus returns statusError we know that the ready
 			// condition will be there so there's no need to check for it
 			readyCondition, _ := ks.GetReadyCondition()
-			if strings.Contains(strings.ToLower(readyCondition.Message), "secret") {
-				e = d.setKafkaClusterFailed(kafka, "failed to update kafka")
-			} else {
-				e = d.setKafkaClusterFailed(kafka, readyCondition.Message)
-			}
+			e = d.setKafkaClusterFailed(kafka, readyCondition.Message)
 		case statusDeleted:
 			e = d.setKafkaClusterDeleting(kafka)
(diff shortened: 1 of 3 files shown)


miguelsorianod:
My understanding is that the issue is that the `FailedReason` attribute of the Kafka request ends up being shown in the UI side, and sometimes we show information that might be sensitive or not really relevant to the end user.
This could be the case when information about a Secret is shown but that's not the only case. For that reason I think the set of changes should be more generic instead of checking a specific error message content.

I would take the approach in the following way:
In the `setKafkaClusterFailed` set `FailedReason` to a literal string like "Kafka <id> status reported as failed from the data plane" and then log with the logger the real error message.

miguelsorianod:
cc @davidffrench to confirm if what I wrote in my previous comment (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1174#issuecomment-1178847859) matches the expectations of the original issue.
The implication is that the end-user will not see nor be able to see details on why it failed when the data plane reports the kafka with the ready condition to value error

miguelsorianod:
I also observed that in the PR there are several commits but they seem to all be related to the same functionality. Please squash them once the functionality is ready. Thanks.

machi1990:
> My understanding is that the issue is that the `FailedReason` attribute of the Kafka request ends up being shown in the UI side, and sometimes we show information that might be sensitive or not really relevant to the end user. This could be the case when information about a Secret is shown but that's not the only case. For that reason I think the set of changes should be more generic instead of checking a specific error message content.
> 
> I would take the approach in the following way: In the `setKafkaClusterFailed` set `FailedReason` to a literal string like "Kafka status reported as failed from the data plane" and then log with the logger the real error message.

+1

machi1990:
> Thanks @miguelsorianod @machi1990 ! Just added the changes.

Thanks @valeriiashapoval. It seems like the tests are failing for a related change in this PR. Can you have a look?

PR #1163, opened 2022-07-04
feat: Remove assigned cluster from kafka

VanillaSpoon:
## Description
Remove assignments from kafka to allow kafka to be reassigned to a new Cluster, if it has been rejected by a previous cluster due to insufficient cluster resources.

Assignments Removed: 
- cluster_id
- bootstrap_server_host
- desired_strimzi_version
- desired_kafka_version
- desired_kafka_ibp_version

[MGDSTRM-9031](https://issues.redhat.com/browse/MGDSTRM-9031)

## Verification Steps
Unit and integration tests added.

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has been created for changes required on the client side

PR #1156, opened 2022-07-01
feat: Check if kafka fails to be provisioned due to cluster fullness

VanillaSpoon:
## Description
Here we are checking the DataplaneKafkaStatus for why a kafka has been rejected.
If the kafka is noted as rejected due to `Cluster has insufficient resources` the status returned will be "statusReassigningCluster"

[MGDSTRM-9030](https://issues.redhat.com/browse/MGDSTRM-9030)

## Verification Steps
Unit tests have been added 

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has been created for changes required on the client side

PR #1181, opened 2022-07-12
feat: Re-assign Kafka to another OSD cluster when Rejected with ClusterFull reason

VanillaSpoon:
## Description
Here we Re-assign a Kafka to another OSD cluster when Rejected with ClusterFull reason:

Combined with previous prs, we: 
- Check if kafka fails to be provisioned due to insufficient cluster resources  [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1156#issue-1291093659)
- Remove the assigned cluster from the kafka [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1163#issue-1293184236)
- And with this pr we reassign the kafka to a new cluster.

[MGDSTRM-9034](https://issues.redhat.com/browse/MGDSTRM-9034)

## Verification Steps
Unit and Integration tests passing.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -109,6 +109,7 @@ type KafkaService interface {
 	// GetAvailableSizesInRegion returns a list of ids of the Kafka instance sizes that can still be created according to the specified criteria
 	GetAvailableSizesInRegion(criteria *FindClusterCriteria) ([]string, *errors.ServiceError)
 	ValidateBillingAccount(externalId string, instanceType types.KafkaInstanceType, billingCloudAccountId string, marketplace *string) *errors.ServiceError
+	AssignBootstrapServerHost(kafkaRequest *dbapi.KafkaRequest) error
 }
 
 var _ KafkaService = &kafkaService{}
@@ -392,24 +393,11 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
 }
 
 func (k *kafkaService) PrepareKafkaRequest(kafkaRequest *dbapi.KafkaRequest) *errors.ServiceError {
-	truncatedKafkaIdentifier := buildTruncateKafkaIdentifier(kafkaRequest)
-	truncatedKafkaIdentifier, replaceErr := replaceHostSpecialChar(truncatedKafkaIdentifier)
-	if replaceErr != nil {
-		return errors.NewWithCause(errors.ErrorGeneral, replaceErr, "generated host is not valid")
-	}
-
-	clusterDNS, err := k.clusterService.GetClusterDNS(kafkaRequest.ClusterID)
-	if err != nil {
-		return errors.NewWithCause(errors.ErrorGeneral, err, "error retrieving cluster DNS")
-	}
-
 	kafkaRequest.Namespace = fmt.Sprintf("kafka-%s", strings.ToLower(kafkaRequest.ID))
-	clusterDNS = strings.Replace(clusterDNS, constants2.DefaultIngressDnsNamePrefix, constants2.ManagedKafkaIngressDnsNamePrefix, 1)
-	kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, clusterDNS)
 
-	if k.kafkaConfig.EnableKafkaExternalCertificate {
-		// If we enable KafkaTLS, the bootstrapServerHost should use the external domain name rather than the cluster domain
-		kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, k.kafkaConfig.KafkaDomainName)
+	err := k.AssignBootstrapServerHost(kafkaRequest)
+	if err != nil {
+		return errors.NewWithCause(errors.ErrorGeneral, err, "error assigning bootstrap server host to kafka %s", kafkaRequest.ID)
 	}
 
 	if k.keycloakService.GetConfig().EnableAuthenticationOnKafka {
@@ -1324,4 +1312,28 @@ func buildResourceRecordChange(recordName string, clusterIngress string, action 
 	}
 
 	return resourceRecordChange
+}
+
+func (k *kafkaService) AssignBootstrapServerHost(kafkaRequest *dbapi.KafkaRequest) error {
+	truncatedKafkaIdentifier := buildTruncateKafkaIdentifier(kafkaRequest)
+	truncatedKafkaIdentifier, replaceErr := replaceHostSpecialChar(truncatedKafkaIdentifier)
+	if replaceErr != nil {
+		return errors.NewWithCause(errors.ErrorGeneral, replaceErr, "generated host is not valid")
+	}
+
+	clusterDNS, err := k.clusterService.GetClusterDNS(kafkaRequest.ClusterID)
+	if err != nil {
+		return errors.NewWithCause(errors.ErrorGeneral, err, "error retrieving cluster DNS")
+	}
+
+	clusterDNS = strings.Replace(clusterDNS, constants2.DefaultIngressDnsNamePrefix, constants2.ManagedKafkaIngressDnsNamePrefix, 1)
+
+	if k.kafkaConfig.EnableKafkaExternalCertificate {
+		// If we enable KafkaTLS, the bootstrapServerHost should use the external domain name rather than the cluster domain
+		kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, k.kafkaConfig.KafkaDomainName)
+	} else {
+		kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, clusterDNS)
+	}
+
+	return nil
 }
--- a/internal/kafka/internal/services/kafka_test.go
+++ b/internal/kafka/internal/services/kafka_test.go
@@ -400,6 +400,7 @@ func Test_kafkaService_PrepareKafkaRequest(t *testing.T) {
 		clusterService    ClusterService
 		keycloakService   sso.KeycloakService
 		kafkaConfig       *config.KafkaConfig
+		kafkaService      KafkaService
 	}
 	type args struct {
 		kafkaRequest *dbapi.KafkaRequest
@@ -436,6 +437,11 @@ func Test_kafkaService_PrepareKafkaRequest(t *testing.T) {
 						return &api.ServiceAccount{}, nil
 					},
 				},
+				kafkaService: &KafkaServiceMock{
+					AssignBootstrapServerHostFunc: func(kafkaRequest *dbapi.KafkaRequest) error {
+						return nil
+					},
+				},
 				kafkaConfig: &config.KafkaConfig{},
 			},
 			args: args{
@@ -450,10 +456,9 @@ func Test_kafkaService_PrepareKafkaRequest(t *testing.T) {
 		{
 			name: "failed clusterDNS retrieval",
 			fields: fields{
-				connectionFactory: db.NewMockConnectionFactory(nil),
 				clusterService: &ClusterServiceMock{
 					GetClusterDNSFunc: func(string) (string, *errors.ServiceError) {
-						return "", errors.New(errors.ErrorBadRequest, "")
+						return "", errors.GeneralError("test")
 					},
 				},
 				keycloakService: &sso.KeycloakServiceMock{
@@ -463,6 +468,11 @@ func Test_kafkaService_PrepareKafkaRequest(t *testing.T) {
 								ClientID: "test",
 							},
 						}
+					},
+				},
+				kafkaService: &KafkaServiceMock{
+					AssignBootstrapServerHostFunc: func(kafkaRequest *dbapi.KafkaRequest) error {
+						return nil
 					},
 				},
 				kafkaConfig: &config.KafkaConfig{},
@@ -488,6 +498,11 @@ func Test_kafkaService_PrepareKafkaRequest(t *testing.T) {
 								ClientID: "test",
 							},
 						}
+					},
+				},
+				kafkaService: &KafkaServiceMock{
+					AssignBootstrapServerHostFunc: func(kafkaRequest *dbapi.KafkaRequest) error {
+						return nil
 					},
 				},
 				kafkaConfig: &config.KafkaConfig{},
@@ -525,6 +540,11 @@ func Test_kafkaService_PrepareKafkaRequest(t *testing.T) {
 						return nil, errors.FailedToCreateSSOClient("failed to create the sso client")
 					},
 				},
+				kafkaService: &KafkaServiceMock{
+					AssignBootstrapServerHostFunc: func(kafkaRequest *dbapi.KafkaRequest) error {
+						return nil
+					},
+				},
 				kafkaConfig: &config.KafkaConfig{},
 			},
 			args: args{
@@ -551,6 +571,11 @@ func Test_kafkaService_PrepareKafkaRequest(t *testing.T) {
 							},
 							EnableAuthenticationOnKafka: true,
 						}
+					},
+				},
+				kafkaService: &KafkaServiceMock{
+					AssignBootstrapServerHostFunc: func(kafkaRequest *dbapi.KafkaRequest) error {
+						return nil
 					},
 				},
 				kafkaConfig: &config.KafkaConfig{},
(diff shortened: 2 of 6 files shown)


internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +67,77 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
 	}
 
 	return encounteredErrors
 }
+func (k *ProvisioningKafkaManager) reassignProvisioningKafka(kafka *dbapi.KafkaRequest) error {
+	cluster, e := k.clusterPlacementStrategy.FindCluster(kafka)
+	if e != nil || cluster == nil {
+		msg := fmt.Sprintf("No available cluster found for '%s' Kafka instance in region: '%s'", kafka.InstanceType, kafka.Region)
+		logger.Logger.Infof(msg)
+		return errors.New(fmt.Sprintf("Region %s cannot accept instance type: %s at this moment", kafka.Region, kafka.InstanceType))
```
machi1990, line 89: Maybe we can add the kafka id in the log/error message as well?

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +67,77 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
+	}
+
+	clusterDNS = strings.Replace(clusterDNS, constants2.DefaultIngressDnsNamePrefix, constants2.ManagedKafkaIngressDnsNamePrefix, 1)
+
+	kafka.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, clusterDNS)
+
+	if k.kafkaConfig.EnableKafkaExternalCertificate {
+		// If we enable KafkaTLS, the bootstrapServerHost should use the external domain name rather than the cluster domain
+		kafka.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, k.kafkaConfig.KafkaDomainName)
+	}
```
machi1990, line 114: A possible improvement would be to have a method in the `KafkaService` that knows how to assign the BootStrapServerHost e.g
```
func (kafkaService *KafkaService) AssignBootstrapServerHost(kafkaRequest *KafkaRequest) error {

} 
```

machi1990 (changes requested):

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
machi1990: This method can be used here and in the `KafkaService.Prepare()` implementation

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +67,77 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
+
+	latestStrimziVersion, latestStrimziVersionErr := cluster.GetLatestAvailableAndReadyStrimziVersion()
+	if latestStrimziVersionErr != nil {
+		return errors.Wrapf(latestStrimziVersionErr, "error finding ready strimzi versions for kafka %s", kafka.ID)
+	}
+	if latestStrimziVersion == nil {
+		return errors.Wrapf(err, "could not find latest strimzi version for kafka %s", kafka.ID)
```
ziccardi, line 121: I believe `err` is nil here. It is last assigned here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6f58345f7aaebd09dc13ccff48c410f97646c6a1/internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go#L102

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +67,77 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
+	}
+
+	kafka.DesiredStrimziVersion = latestStrimziVersion.Version
+
+	desiredKafkaVersion := latestStrimziVersion.GetLatestKafkaVersion()
+	if desiredKafkaVersion == nil {
+		return errors.Wrapf(err, "failed to get Kafka version for kafka %s", kafka.ID)
```
ziccardi, line 128: See previous comment

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +67,77 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
+	}
+	kafka.DesiredKafkaVersion = desiredKafkaVersion.Version
+
+	desiredKafkaIBPVersion := latestStrimziVersion.GetLatestKafkaIBPVersion()
+	if desiredKafkaIBPVersion == nil {
+		return errors.Wrapf(err, "failed to get Kafka IBP version for kafka %s", kafka.ID)
```
ziccardi, line 134: See previous comment

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +67,77 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
+	}
+	kafka.DesiredKafkaIBPVersion = desiredKafkaIBPVersion.Version
+
+	updateErr := k.kafkaService.Update(kafka)
+	if updateErr != nil {
+		return errors.Wrapf(err, "Failed to update kafka %s in provisioning state", kafka.ID)
```
ziccardi, line 140: See previous comment

machi1990: This should use `updateErr` in place of `err`

machi1990:
Left a few comments and +1 on @ziccardi 

Other than that the change lgtm

internal/kafka/internal/services/util.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	truncatedNameLen = 10
+	TruncatedNameLen = 10
```
miguelsorianod, line 15: This name was already quite generic: what is the purpose of `truncatedNameLen`? what does it define?)
Now that the constant is becoming public I think at least documenting it would be a good idea

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
miguelsorianod: :+1:

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -17,18 +21,24 @@ import (
...
 		BaseWorker: workers.BaseWorker{
 			Id:         uuid.New().String(),
 			WorkerType: "provisioning_kafka",
 			Reconciler: reconciler,
 		},
-		kafkaService: kafkaService,
+		kafkaService:             kafkaService,
+		clusterPlacementStrategy: clusterPlacementStrategy,
+		clusterService:           clusterService,
+		kafkaConfig:              kafkaConfig,
```
machi1990, line 41:
```suggestion
```

these two fields are not used anymore and can be removed

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
machi1990: Thanks for addressing the feedback. I've left some cleanup comments of the variables that are no longer used.

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +66,60 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
+	desiredKafkaIBPVersion := latestStrimziVersion.GetLatestKafkaIBPVersion()
+	if desiredKafkaIBPVersion == nil {
+		return errors.New(fmt.Sprintf("failed to get Kafka IBP version for kafka %s", kafka.ID))
+	}
+	kafka.DesiredKafkaIBPVersion = desiredKafkaIBPVersion.Version
+
+	updateErr := k.kafkaService.Update(kafka)
+	if updateErr != nil {
+		return errors.New(fmt.Sprintf("Failed to update kafka %s in provisioning state", kafka.ID))
+	}
```
machi1990, line 123:
```suggestion
	if updateErr != nil {
		return errors.Wrapf(updateErr, "Failed to update kafka %s in provisioning state", kafka.ID))
	}
```

machi1990 (changes requested):

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -56,10 +66,60 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
...
 		metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusProvisioning, kafka.ID, kafka.ClusterID, time.Since(kafka.CreatedAt))
 	}
 
 	return encounteredErrors
 }
+func (k *ProvisioningKafkaManager) reassignProvisioningKafka(kafka *dbapi.KafkaRequest) error {
+	cluster, e := k.clusterPlacementStrategy.FindCluster(kafka)
+	if e != nil || cluster == nil {
+		msg := fmt.Sprintf("No available cluster found for '%s' Kafka instance %s in region: '%s'", kafka.InstanceType, kafka.ID, kafka.Region)
+		logger.Logger.Infof(msg)
```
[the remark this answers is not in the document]
VanillaSpoon: Thanks for the great feedback and information @machi1990 :)

internal/kafka/internal/services/kafka.go
```diff
@@ -1319,3 +1307,27 @@ func buildResourceRecordChange(recordName string, clusterIngress string, action
...
+	}
+
+	clusterDNS = strings.Replace(clusterDNS, constants2.DefaultIngressDnsNamePrefix, constants2.ManagedKafkaIngressDnsNamePrefix, 1)
+
+	kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, clusterDNS)
+
+	if k.kafkaConfig.EnableKafkaExternalCertificate {
+		// If we enable KafkaTLS, the bootstrapServerHost should use the external domain name rather than the cluster domain
+		kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, k.kafkaConfig.KafkaDomainName)
+	}
```
ziccardi, line 1330: if `k.kafkaConfig.EnableKafkaExternalCertificate` is `true` this assignment is ineffective.

Better to change this to
 ```suggestion
	if k.kafkaConfig.EnableKafkaExternalCertificate {
		// If we enable KafkaTLS, the bootstrapServerHost should use the external domain name rather than the cluster domain
		kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, k.kafkaConfig.KafkaDomainName)
	} else {
		kafkaRequest.BootstrapServerHost = fmt.Sprintf("%s.%s", truncatedKafkaIdentifier, clusterDNS)
	}
```

File: internal/kafka/internal/services/kafka.go [hunk 12]
```diff
@@ -1319,3 +1307,27 @@ func buildResourceRecordChange(recordName string, clusterIngress string, action
 
 	return resourceRecordChange
 }
+
+func (k *kafkaService) AssignBootstrapServerHost(kafkaRequest *dbapi.KafkaRequest) error {
```
ziccardi, line 1311: Usually a method that changes its input is not a good practice.
Instead of having this method setting the `BootstrapServerHost` attribute of `dbapi.KafkaRequest`, I think would be better to have a `func (k *kafkaService) FindBootstrapServerHost(kafkaRequest dbapi.KafkaRequest) (string, error)` method.
Then the caller would do something like:
```go
	if bootstrapServerHost, err := sevice.FindBootstrapServerHost(kafkaRequest); err != nil {
		// deal with the error
		return nil
	} else {
		kafkaRequest.BootstrapServerHost = bootstrapServerHost
	}
```
What is the advantage of this approach? If a user reads the code, he will know exactly what happened to the `kafkaRequest` object. If the function takes a `pointer` to `kafkaRequest`, then who reads the code will need to inspect the called function has well to search for any side effect.

machi1990: The side effect is intended hence the naming as well `AssignXXX` and passing the pointer. 
The caller, in this case `Prepare..` and the provisiniong reconciler, calls this method passing in a pointer expecting the bootstrap server host to be assigned if no error happened.

ziccardi: If the method is to be used by `Prepare..` then I believe it shouldn't be exported.

In both cases I don't see any reason/benefit for not using the proposed pattern: the current one is open to make any change to the received parameters and who reads the code needs to enter the called function to be sure of what is changed.

machi1990: It is not used only in the `Prepare`. It is also used in the reconciler hence the need for it to be exported. 

> In both cases I don't see any reason/benefit for not using the proposed pattern

I agree with having pure functions and methods being side effects safe when they need to. But in this case, I do not think it is needed because this is method is a "setter" as its name indicates.

> the current one is open to make any change to the received parameters and who reads the code needs to enter the called function to be sure of what is changed

The intended behaviour is to assign the bootstrap server host to the kafka just like what the method name says. If someone for "I do not know what reason" change / assigns other things, then the'll need a dedicated method or something else for that.

> who reads the code needs to enter the called function to be sure of what is changed, how it is changed, what the function does when an error occurs, etc. etc.

The same argument can be made with any method. If you are interested in the implementation details of something, you'll need to look it up. But if the method speaks for itself and only does what it does, you care less about the details.

ziccardi: > The same argument can be made with any method. If you are interested in the implementation details

That's not accurate: if the method takes a pointer, than it is likely to change the parameter, so I need to see what the method does to know how it changes MY variable. If it takes the param by value, then I don't need to check what the function does to know exactly what will be MY state after the function call.

> The intended behaviour is to assign the bootstrap server host to the kafka just like what the method name says.

Yes. And I don't think it is a good practice. Look at the `append` function for slices. Why do we need to do:
```go
myslice = append(myslice, ....)
```
? `append` is called `append` because it APPENDs values to the slice, so it could simply modify its input.
However, that's not the case: `append` takes a **copy** of the slice and returns a modified slice.

machi1990: > > The same argument can be made with any method. If you are interested in the implementation details
> 
> That's not accurate: if the method takes a pointer, than it is likely to change the parameter, so I need to see what the method does to know how it changes MY variable. If it takes the param by value, then I don't need to check what the function does to know exactly what will be MY state after the function call.

If you are interested in the implementation detail of something, you need to check the details no matter how the method is declared. This is what I was saying. 

> Yes. And I don't think it is a good practice. 

I think it all depends on the context of what the method intended behavior is. A side effect for me, will be if the method said that it is not going to modify the value of the received parameter and ended up modifying it or saying that it is only modifying this one value and ended up modifying another value on top of it. But this is not the case here and if someone changes the implementation to start assigning something else, then they'll need a dedicated method to assign `YYY` (whatever the YYY is) and not this one.

The kafka boostrap server host needs to assigned as part of Kafka preparing step or kafka reassignment step. The two callers of this method call it with the intention of assigning it and nothing else. 

> Look at the append function for slices. 

There are also methods in the core lib that takes in a pointer as a param and assigns value to the received param...

ziccardi: > If you are interested in the implementation detail of something

You are missing my point. If the function takes a pointer, I'm **forced** to be interested in the implementation details of the called function to understand how it changes my state

> A side effect for me, will be if the method said that it is not going to modify the value of the received parameter and ended up modifying it or saying that it is only modifying this one value

Again I don't believe again this is correct. The function could contain logic about how to change the value. It could do something like 'if this error happens, assign this, in this other error assign that, otherwise return an error' (this is just an example) or anything else. Without looking at the called function, you can't know what the function does.

Moreover, you will always end looking at the function code because when some error happens, you can't trust the function not changing something else just because it is called 'assignxxx'

> There are also methods in the core lib that takes in a pointer as a param and assigns value to the received param...

I can't find any on top of my mind that are even close to this kind of function.

Anyway: I believe we could discuss for hours without finding an agreement.

This is my conclusion: with the proposed approach there will be the benefit of removing all the side effects and everything we discussed doesn't apply anymore.

OTOH I don't see any benefit in the current approach.

machi1990: > This is my conclusion: with the proposed approach there will be the benefit of removing all the side effects and everything we discussed doesn't apply anymore.

With the proposed approach you are not removing the assignement, you are moving it one level up to the caller.

> OTOH I don't see any benefit in the current approach.

With the current approach, the assignment logic is taken care of by the implementation of the API contract.

> I can't find any on top of my mind that are even close to this kind of function.

The json unmarshalling function for example.

machi1990: > Anyway: I believe we could discuss for hours without finding an agreement.

:-) Yes. I believe there is value in what you are proposing but I feel like there is little of it in the context of this function: or some of the functions like `Prepare()` and `RegisterKafkaJob()` here that also modifies the kafka object. This is my opinion.

I'll leave it to @VanillaSpoon to decide

ziccardi: > With the proposed approach you are not removing the assignement, you are moving it one level up to the caller.

With the proposed approach, the assignment is moved to the **owner** of the object. If another function has other side effects, that is another bad practice IMO.

> With the current approach, the assignment logic is taken care of by the implementation of the API contract.

I don't follow you.

> The json unmarshalling function for example.

I believe that method has nothing in common with our method: deserializers must be generic and have no way to know what is the object to be built, so they have to receive the instance to be populated. That's not our case at all.

Anyway:  I really believe there are no benefits at all in using the current approach. However, it is just my opinion: feel free to ignore it.

internal/kafka/internal/services/kafka_test.go
```diff
@@ -3797,3 +3822,79 @@ func Test_kafkaService_ListKafkasWithRoutesNotCreated(t *testing.T) {
...
+	for _, testcase := range tests {
+		test := testcase
+		t.Run(test.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			k := &kafkaService{
+				clusterService: test.fields.clusterService,
+				kafkaConfig:    test.fields.kafkaConfig,
+			}
+			err := k.AssignBootstrapServerHost(test.args.kafkaRequest)
+			g.Expect(err != nil).To(gomega.Equal(test.wantErr))
```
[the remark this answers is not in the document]
miguelsorianod: That's how `gomega` library is designed.
The expectations are at NewWithT type level whereas assertions are at package level

machi1990 (approved):

internal/kafka/internal/services/kafka.go
VanillaSpoon: Hey @ziccardi 
Thanks, great spot :)

internal/kafka/internal/services/kafka.go
VanillaSpoon: Thanks for both your inputs @ziccardi  and @machi1990 
I took some time to read more into the two options, It was a great learning experience to look at it from both angles. I will leave the approach as it is for now, but I will return to it if there is confusion in the future.

PR #1185, opened 2022-07-14
Ignore the not found error while dleting internal service account

Rajagopalan-Ranganathan:
JIRA Link: https://issues.redhat.com/browse/MGDSRVS-347

## Description
Issue was reported by sso.redhat.com that we are trying to contonously delete
an non existent service account. The issue was found that, the internal service account
deletion logic treats not found as valid errors and contnously keeps retrying to delete it
in a reconciliation loop. Code changes made to ignore not found error and log the same and proceed with the deprovisioning. Unit test case added to test the same.

## Checklist (Definition of Done)

- [X] All acceptance criteria specified in JIRA have been completed
- [X] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has been created for changes required on the client side

PR #1192, opened 2022-07-19
fix: removing error returned if service acc not found

akoserwal:
Remove the error returned. Reason if the service account is already deleted or failed to collect. it should block deletion of Kafka

This is only required for temporary purposes until all legacy canary- clients get deleted.

https://issues.redhat.com/browse/MGDSTRM-9224

- [ ] Code Review completed

### DIFF
--- a/pkg/services/sso/redhatsso_service.go
+++ b/pkg/services/sso/redhatsso_service.go
@@ -299,7 +299,8 @@ func (r *redhatssoService) DeleteServiceAccountInternal(accessToken string, clie
 		for {
 			accounts, err := r.client.GetServiceAccounts(accessToken, first, max)
 			if err != nil {
-				return errors.NewWithCause(errors.ErrorGeneral, err, "failed to collect internal service accounts")
+				glog.Errorf("failed to get internal service account: %s", err)
+				return nil
 			}
 			if len(accounts) == 0 {
 				return nil


machi1990:
Hi @ziccardi can you have a look for this PR as well? You might have better context than myself.

machi1990:
> Remove the error returned. Reason if the service account is already deleted or failed to collect. it should block deletion of Kafka

@akoserwal I thought this was fixed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1185? or a different error is returned and we need to ignore that as well?

akoserwal:
> > Remove the error returned. Reason if the service account is already deleted or failed to collect. it should block deletion of Kafka
> 
> @akoserwal I thought this was fixed by #1185? or a different error is returned and we need to ignore that as well?

if we fail to list the account and the client is already deleted. This will block the deletion of Kafka cluster

miguelsorianod:
Shouldn't we at least check that the received error is the one about not existing, if it is the case that that is returned here too?
The call could fail due to multiple reasons not only due to not existing anymore.

Aside from that, the call is to `GetServiceAccounts` which seems a list method, not to get of a specific account.

machi1990:
> if we fail to list the account and the client is already deleted. This will block the deletion of Kafka cluster

If a client does not exist, does the list service accounts method returns empty list or it returns an error? 

If it returns an empty list and an error is only returned temporarily (5xx for example), then the Kafka will be blocked temporarily and eventually we'll land in  
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/2a1b421a5973e03e5dfc7c485c2fe33af7817c26/pkg/services/sso/redhatsso_service.go#L306 which consider the internal SA as being deleted.

PR #1201, opened 2022-07-22
refactor:return error type for FindCluster in cluster placement strategy interface

JameelB:
## Description
Changes the following methods to return an `error` type instead of the KAS Fleet Manager `ServiceError`. This is based on what was agreed upon in the Control Plane team's technical discussion. Functions should not return `ServiceError` unless there is a strong need for it which wasn't the case for these functions.

Cluster Service
- FindAllClusters: will now return database error as is. This is only called in the cluster placement strategy methods which now returns an `error` type.
- FindCluster: will now return database error as is. This is called in the following places:
   - Cluster placement strategy: This now returns an `error` type.
   - Cluster manager: in the reconcileDeprovisioningCluster, this method already returns an `error` type.
- FindKafkaInstanceCount: will return database error or GetKafkaInstanceSize error as is. This is called in the following places:
   - Cluster placement strategy: This now returns an `error` type.
   - Cluster manager: in the reconcileClusterWithManualConfig and setKafkaPerClusterCountMetrics, both of these methods already returns an `error` type.
- FindStreamingUnitCountByClusterAndInstanceType: Returns database and GetKafkaInstanceSize errors as is. This method already returns an `error` type. This is used for the cluster placement strategy and in the Kafka manager (setClusterStatusCapacityMetrics) which already returns an `error` type.

Cluster Placement Strategy
- FindCluster: interface was changed to return an `error` instead of a `ServiceError`. There wasn't any strong need for it to return a `ServiceError` as the errors it returns will mostly be caused by database errors or configuration error.

Follows up from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1167#discussion_r925801278

## Verification Steps
- [ ] CI Checks passing

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/cluster_placement_strategy.go
+++ b/internal/kafka/internal/services/cluster_placement_strategy.go
@@ -123,7 +123,7 @@ func searchClusterObjInArray(clusters []*api.Cluster, clusterId string) *api.Clu
 // findClusterKafkaInstanceCount searches DB for the number of Kafka instance associated with each OSD Clusters
 func (f *FirstSchedulableWithinLimit) findClusterKafkaInstanceCount(clusterIDs []string) (map[string]int, *errors.ServiceError) {
 	if instanceLst, err := f.ClusterService.FindKafkaInstanceCount(clusterIDs); err != nil {
-		return nil, errors.NewWithCause(err.Code, err, "failed to find kafka instance count for clusters '%v'", clusterIDs)
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to find kafka instance count for clusters '%v'", clusterIDs)
 	} else {
 		clusterWithinLimitMap := make(map[string]int)
 		for _, c := range instanceLst {
--- a/internal/kafka/internal/services/cluster_placement_strategy.go
+++ b/internal/kafka/internal/services/cluster_placement_strategy.go
@@ -74,7 +74,7 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
 	//#1
 	clusterObj, err := f.ClusterService.FindAllClusters(criteria)
 	if err != nil {
-		return nil, err
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to find all clusters with criteria: %v", criteria)
 	}
 
 	dataplaneClusterConfig := f.DataplaneClusterConfig.ClusterConfig
@@ -150,29 +150,31 @@ func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cl
 	}
 
 	clusters, findAllClusterErr := f.ClusterService.FindAllClusters(criteria)
-	if findAllClusterErr != nil || len(clusters) == 0 {
-		return nil, findAllClusterErr
+	if findAllClusterErr != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, findAllClusterErr, "failed to find clusters with criteria: %v", criteria)
 	}
 
-	// Get total number of streaming unit used per region and instance type
-	streamingUnitCountPerRegionList, countStreamingUnitErr := f.ClusterService.FindStreamingUnitCountByClusterAndInstanceType()
-	if countStreamingUnitErr != nil {
-		return nil, errors.NewWithCause(errors.ErrorGeneral, countStreamingUnitErr, "failed to get count of streaming units by region and instance type")
-	}
+	if len(clusters) != 0 {
+		// Get total number of streaming unit used per region and instance type
+		streamingUnitCountPerRegionList, countStreamingUnitErr := f.ClusterService.FindStreamingUnitCountByClusterAndInstanceType()
+		if countStreamingUnitErr != nil {
+			return nil, errors.NewWithCause(errors.ErrorGeneral, countStreamingUnitErr, "failed to get count of streaming units by region and instance type")
+		}
 
-	instanceSize, getInstanceSizeErr := f.KafkaConfig.GetKafkaInstanceSize(kafka.InstanceType, kafka.SizeId)
-	if getInstanceSizeErr != nil {
-		return nil, errors.NewWithCause(errors.ErrorGeneral, getInstanceSizeErr, "failed to get instance size for kafka '%s'", kafka.ID)
-	}
+		instanceSize, getInstanceSizeErr := f.KafkaConfig.GetKafkaInstanceSize(kafka.InstanceType, kafka.SizeId)
+		if getInstanceSizeErr != nil {
+			return nil, errors.NewWithCause(errors.ErrorGeneral, getInstanceSizeErr, "failed to get instance size for kafka '%s'", kafka.ID)
+		}
 
-	// Find first ready cluster that has remaining capacity (total streaming unit used by existing and requested kafka is within the cluster maxUnit limit)
-	for _, cluster := range clusters {
-		currentStreamingUnitsUsed := streamingUnitCountPerRegionList.GetStreamingUnitCountForClusterAndInstanceType(cluster.ClusterID, kafka.InstanceType)
-		capacityInfo := cluster.RetrieveDynamicCapacityInfo()
-		maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
+		// Find first ready cluster that has remaining capacity (total streaming unit used by existing and requested kafka is within the cluster maxUnit limit)
+		for _, cluster := range clusters {
+			currentStreamingUnitsUsed := streamingUnitCountPerRegionList.GetStreamingUnitCountForClusterAndInstanceType(cluster.ClusterID, kafka.InstanceType)
+			capacityInfo := cluster.RetrieveDynamicCapacityInfo()
+			maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
 
-		if currentStreamingUnitsUsed+instanceSize.CapacityConsumed <= int(maxStreamingUnits) {
-			return cluster, nil
+			if currentStreamingUnitsUsed+instanceSize.CapacityConsumed <= int(maxStreamingUnits) {
+				return cluster, nil
+			}
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


JameelB:
please have a look when you get a chance @ziccardi @machi1990 @miguelsorianod @VanillaSpoon

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -74,7 +74,7 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
 	//#1
 	clusterObj, err := f.ClusterService.FindAllClusters(criteria)
 	if err != nil {
-		return nil, err
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to find all clusters with criteria: %v", criteria)
```
VanillaSpoon: Hey Jameel, there are a few errors changed back such as:
```diff
-		return nil, err
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to find all clusters with criteria: %v", criteria)
```

and then back [in another commit](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1201/commits/3a6a8673a46df5fb6275b2ca8cf6bfed708b6d1e#diff-5ffcdb3a533058503533bac73799b825a292e9cc5fabc494e6ec10d4d26753f6L77-R76) to:
```diff
-		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to find all clusters with criteria: %v", criteria)
+		return nil, err
```

I was just wondering what the reasoning for this was?

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -68,7 +67,7 @@ func (f *FirstSchedulableWithinLimit) FindCluster(kafka *dbapi.KafkaRequest) (*a
 
 	kafkaInstanceSize, e := f.KafkaConfig.GetKafkaInstanceSize(kafka.InstanceType, kafka.SizeId)
 	if e != nil {
-		return nil, errors.NewWithCause(errors.ErrorInstancePlanNotSupported, e, "failed to find cluster with criteria '%v'", criteria)
+		return nil, e
```
ziccardi, line 70: Why this change?

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -121,9 +120,9 @@ func searchClusterObjInArray(clusters []*api.Cluster, clusterId string) *api.Clu
 }
 
 // findClusterKafkaInstanceCount searches DB for the number of Kafka instance associated with each OSD Clusters
-func (f *FirstSchedulableWithinLimit) findClusterKafkaInstanceCount(clusterIDs []string) (map[string]int, *errors.ServiceError) {
+func (f *FirstSchedulableWithinLimit) findClusterKafkaInstanceCount(clusterIDs []string) (map[string]int, error) {
 	if instanceLst, err := f.ClusterService.FindKafkaInstanceCount(clusterIDs); err != nil {
-		return nil, errors.NewWithCause(err.Code, err, "failed to find kafka instance count for clusters '%v'", clusterIDs)
+		return nil, err
```
ziccardi, line 125: Why this change?

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -150,29 +149,31 @@ func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cl
 	}
 
 	clusters, findAllClusterErr := f.ClusterService.FindAllClusters(criteria)
-	if findAllClusterErr != nil || len(clusters) == 0 {
+	if findAllClusterErr != nil {
```
ziccardi, line 152: Why did you remove this again?

JameelB: I had to remove this due to a previous commit when I changed the return type of `FindAllClusters` but the cluster placement strategy still returns a `ServiceError` so i needed to cast it, which caused tests to fails. 

I'm supposed to re-add this, good spot!

internal/kafka/internal/services/cluster_placement_strategy.go
JameelB: These changes were due to the change of returned error type in the cluster placement strategy. All of the internal calls now returns an `error` type so we no longer need to create a `ServiceError` out of them. 

We can wrap these errors with the original message, but note that most of these will be database failures.

internal/kafka/internal/services/cluster_placement_strategy.go
JameelB: This is to ensure that each commit to refactor each method is isolated. I didn't want to change the return type of a method that I wasn't refactoring. In a later commit, I then refactored the method that calls it to use `error` type instead of a `serviceError` so I no longer had to call `errors.NewWithCause`. 

Isolating each change to a single commit ensures that if we decide to remove one change, we can drop that commit without affecting others that much.

VanillaSpoon: Thanks for the explanation, Very nice tip :)

internal/kafka/internal/services/cluster_placement_strategy.go
machi1990: A follow up on @ziccardi question.

> All of the internal calls now returns an error type so we no longer need to create a ServiceError out of them.

the `*errors.ServiceError` implements the `error` interface, I think it'd have been fine to only leave the returned service error.

 > but note that most of these will be database failures
 
 Just to confirm that the database failures are not plainly exposed to the end users, correct? to avoid leaking out db credentials or something ;-)

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -121,9 +121,9 @@ func searchClusterObjInArray(clusters []*api.Cluster, clusterId string) *api.Clu
 }
 
 // findClusterKafkaInstanceCount searches DB for the number of Kafka instance associated with each OSD Clusters
-func (f *FirstSchedulableWithinLimit) findClusterKafkaInstanceCount(clusterIDs []string) (map[string]int, *errors.ServiceError) {
+func (f *FirstSchedulableWithinLimit) findClusterKafkaInstanceCount(clusterIDs []string) (map[string]int, error) {
 	if instanceLst, err := f.ClusterService.FindKafkaInstanceCount(clusterIDs); err != nil {
-		return nil, errors.NewWithCause(err.Code, err, "failed to find kafka instance count for clusters '%v'", clusterIDs)
+		return nil, err
```
JameelB, line 126: no need to wrap this one, it should be wrapped by the caller since this is the only place where an error is returned in this method.

internal/kafka/internal/services/clusters.go
```diff
@@ -285,7 +284,7 @@ func (c clusterService) FindCluster(criteria FindClusterCriteria) (*api.Cluster,
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			return nil, nil
 		}
-		return nil, apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to find cluster with criteria")
+		return nil, err
```
JameelB, line 287: can return the database error as is since it's the only place that returns an error in this method. The caller of this method can wrap this error instead.

internal/kafka/internal/services/clusters.go
```diff
@@ -415,7 +414,7 @@ func (c clusterService) FindKafkaInstanceCount(clusterIDs []string) ([]ResKafkaI
 	for _, k := range kafkas {
 		kafkaInstanceSize, e := c.kafkaConfig.GetKafkaInstanceSize(k.InstanceType, k.SizeId)
 		if e != nil {
-			return nil, apiErrors.NewWithCause(apiErrors.ErrorInstancePlanNotSupported, e, "failed to query kafkas")
+			return nil, e
```
JameelB, line 417: Didn't see the need to wrap this error as this method returns an already descriptive error message. It returns one of the following error message:
- `Unable to find kafka instance type for '<instance-type>'`
- `Kafka instance size id: '<size-id>' not found for '<region>' instance type`.

The caller of this method can wrap the error returned by this instead.

internal/kafka/internal/services/cluster_placement_strategy.go
JameelB: > the *errors.ServiceError implements the error interface, I think it'd have been fine to only leave the returned service error and only change the method signature.

I assumed that we shouldn't mix that two even if we can, maybe I misunderstood that though :) 

> Just to confirm that the database failures are not plainly exposed to the end users, correct? to avoid leaking out db credentials or something ;-)

I think we were wrapping them originally in the Service Errors but only logged them, they're not exposed to users. I believe we return a general error message in the handlers.

internal/kafka/internal/services/clusters.go
```diff
@@ -462,7 +461,7 @@ func (c clusterService) FindAllClusters(criteria FindClusterCriteria) ([]*api.Cl
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			return nil, nil
 		}
-		return nil, apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to find all clusters with criteria: %v", clusterDetails)
+		return nil, err
```
JameelB, line 464: same as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1201#discussion_r927803547

internal/kafka/internal/services/cluster_placement_strategy.go
machi1990: > I assumed that we shouldn't mix that two even if we can, maybe I misunderstood that though :)

I was thinking only the method signature may be impacted and we as well as return the service error which is just an implementation of the `error` interface, without any issue. I think we can safely return it and only change the error type from the method signature.

PS: no need to revert the changes here due to explanations you've given above.

machi1990: > I think we were wrapping them originally in the Service Errors but only logged them, they're not exposed to users. I believe we return a general error message in the handlers.

Thanks.

machi1990 (approved):
Thanks for the followup @JameelB 

I'll let others finish up their reviews and approve as well.

internal/kafka/internal/services/cluster_placement_strategy.go
ziccardi: > I assumed that we shouldn't mix that two even if we can, maybe I misunderstood that though :)

What is important is that the returned value is an `error`, which is just an interface.
The returned concrete object can be anything you want as long as it implements `error`.

ziccardi (approved):
lgtm. Thanks, @JameelB !

PR #1202, opened 2022-07-22
feat(prewarming): make it possible for manual clusters to make use of pre-warming

machi1990:
## Description

Create a dedicated pre-warming configuration that's generic enough and can be used irrespective of whether dynamic scaling is turned on or not. 
This allows for manual clusters to make use of the pre-warming feature as well. 

The configuration looks like
```yaml
---
developer:
  reserved_streaming_units: m # if 0 then no pre-warming
  base_streaming_unit_size: x1 
standard:
  reserved_streaming_units: n # if 0 then no pre-warming
  base_streaming_unit_size: x1
```
where `m` and `n` can be integers. 
The `base_streaming_unit_size` allows to configure the size and it can be `x1`, `x2` etc depending on which size is available.
During startup, a validation is perfomed to make sure that the size and instance types are available, otherwise we throw an error.
We explicitly allow for the no configurability of pre-warming as it is an optional feature i.e if the file is not configured or does not exist, we do not throw an error during startup.

**NOTE**: The default behavior is that no pre-warming will be applied. The default value in service template is an empty object i.e {} and in development config, the instance types are there with `0` for reserved streaming units field.

Closes https://issues.redhat.com/browse/MGDSTRM-9235

## Verification Steps

1. Modified tests should pass
2. Against a real cluster verify that pre-warming managed kafkas are created irrespective of the scaling mode.
For this, make sure that the pre-warming configuration have been setup for each instance type.

- [ ] Code Review completed


templates/service-template.yml
```diff
@@ -165,7 +165,12 @@ parameters:
 - name: DYNAMIC_SCALING_CONFIG
   displayName: Dynamic Scaling configuration
   description: "YAML content containing a map of the dynamic scaling configuration for each instance type"
-  value: "{developer: {reserved_streaming_units: 1, compute_nodes_config: {max_compute_nodes: 3}}, standard: {reserved_streaming_units: 1, compute_nodes_config: {max_compute_nodes: 9}}}"
+  value: "{developer: {compute_nodes_config: {max_compute_nodes: 3}}, standard: {compute_nodes_config: {max_compute_nodes: 9}}}"
+
+- name: NODE_PREWARMING_CONFIG
+  displayName: Node prewarming configuration
+  description: "YAML content containing a map of the node prewarming configuration for each instance type"
+  value: "{}"
```
machi1990, line 173: This equates to pre-warming being disabled completely for each instance type.

It can also be disabled on per instance type level by setting reserved_streaming_units to `0`

machi1990:
I'll need to test this against https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/775 using kas-installer.

machi1990:
Added the do not merge label due to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1202#issuecomment-1192731545 

Will perform the test on Monday.

machi1990:
Tested this with https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/775 and everything works as expected

templates/service-template.yml
machi1990: cc @pb82 as mentioned in the internal call

internal/kafka/internal/config/node_prewarming_config.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	if instanceTypeConfig.BaseStreamingUnitSize == "" {
+		instanceTypeConfig.BaseStreamingUnitSize = defaultBaseStreamingUnitSize
+	}
+
+	return instanceTypeConfig, true
+}
+
+func (c *NodePrewarmingConfig) InstanceTypeConfigs() map[string]InstanceTypeNodePrewarmingConfig {
```
JameelB, line 40: The name seems a bit vague to me on what this method is for/doing. Maybe add `New` to it since it's creating a new config? Something like `NewInstanceTypeConfig()`?

JameelB (commented):
Looks good to me!

JameelB (approved):

internal/kafka/internal/config/node_prewarming_config.go
machi1990: Thanks. I'll revisit this :-)

machi1990: This method is not used anywhere. I've removed it

JameelB: thanks for the follow up :)

VanillaSpoon (approved):
Looks good to me, thanks @machi1990

internal/kafka/internal/config/node_prewarming_config.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if err != nil {
+		return errors.Wrapf(err, "Error validating node prewarming configuration for instance type %s", instanceType)
+	}
+
+	if c.ReservedStreamingUnits == 0 {
+		logger.Logger.Warningf("no capacity reservation will be applied for instance type: %s.", instanceType)
+	} else if c.ReservedStreamingUnits < 0 {
+		return fmt.Errorf("reserved_streaming_units cannot be a negative number for instance type: %s", instanceType)
+	}
```
ziccardi, line 86: Maybe a switch Instead of `if... else if` would be more readable:
```suggestion
switch {
   case c.ReservedStreamingUnits == 0: 
      logger.Logger.Warningf("no capacity reservation will be applied for instance type: %s.", instanceType)
   case c.ReservedStreamingUnits < 0:
      return fmt.Errorf("reserved_streaming_units cannot be a negative number for instance type: %s", instanceType)
}
```

WDYT?

ziccardi (approved):
Just an optional comment.
LGTM, Thanks @machi1990 !

internal/kafka/internal/config/node_prewarming_config.go
machi1990: Thanks. I've committed the suggestion via UI, will rebase and force push again.

PR #1205, opened 2022-07-25
feat: Add first draft of scaleup reconciler logic

miguelsorianod:
## Description

Part of https://issues.redhat.com/browse/MGDSTRM-9133.

This PR provides a rough implementation of the data plane scale up detection logic. It is part of the broader dynamic scaling functionality

It includes:
* Detection of when a scale up of a new data plane cluster in a provider region for a given instance type is needed
* Logic that triggers registration of a new data plane cluster

The purpose and focus of this PR is to review the algorithm around the scale up.
The PR does not include documentation, tests or verifications. It also does not focus on quality of the code, organization nor efficiency. Those will come later after confirming that the implemented approach is correct. Those will be submitted in another PR.

- [ ] Code Review completed


internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -57,14 +61,18 @@ func (m *DynamicScaleUpManager) Stop() {
 }
 
 func (m *DynamicScaleUpManager) Reconcile() []error {
```
miguelsorianod, line 63: If we want we can add a return at the beginning of this method so we can merge safely and iterate on that

miguelsorianod: I modified the condition at the beginning to always be true in order to do that

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -124,3 +144,147 @@ func (m *DynamicScaleUpManager) reconcileClustersForRegions() []error {
...
+				kafkaStreamingUnitCountPerCluster.Status != api.ClusterCleanup.String() {
+				maxStreamingUnitsInRegion = maxStreamingUnitsInRegion + int(kafkaStreamingUnitCountPerCluster.MaxUnits)
+			}
+		}
+	}
+
+	freeStreamingUnitsInRegion := maxStreamingUnitsInRegion - consumedStreamingUnitsInRegion
+	capacitySlackInRegion := instanceTypeConfig.MinAvailableCapacitySlackStreamingUnits
+	notEnoughCapacitySlackInRegion := false
+	if freeStreamingUnitsInRegion < capacitySlackInRegion { // TODO is it <= or < ?
```
machi1990, line 249: > // TODO is it <= or < ?

I think `<` because we want to be inclusive. i.e if free capacity is `4` and slack is `4`, there is no need to trigger a scale up in my opinion.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -124,3 +144,147 @@ func (m *DynamicScaleUpManager) reconcileClustersForRegions() []error {
...
+		notEnoughCapacitySlackInRegion = true
+	}
+
+	streamingUnitsLimitInRegion := instanceTypeConfig.Limit
+	regionLimitReached := false
+	if streamingUnitsLimitInRegion != nil && consumedStreamingUnitsInRegion >= *streamingUnitsLimitInRegion {
```
machi1990, line 255: if region limits is `nil`, 

do we want to compare `freeStreamingUnitsInRegion > 0` and used that to set `regionLimitReached` to `true`?

That could be redundant with`atLeastOneClusterHasCapacityForLargestInstanceType` but it really ensures that we've not consumed all current available max limits in the region for cases where we do not have region limit set.

miguelsorianod: If region limits is nil then regionLimitReached should never be true

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: If it is `<` then if the capacity slack is not set then it would never perform data plane scale up, asfreeStreamingUnitsInRegion would never be less than 0 . Do we want to define it as that?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -135,17 +136,21 @@ func (k *KafkaManager) setClusterStatusCapacityMetrics() error {
 
 	// always publish used capacity irrespective of scaling mode
 	for _, usedStreamingUnitCount := range usedStreamingUnitsCountByRegion {
-		metrics.UpdateClusterStatusCapacityUsedCount(usedStreamingUnitCount.CloudProvider, usedStreamingUnitCount.Region, usedStreamingUnitCount.InstanceType, usedStreamingUnitCount.ClusterId, float64(usedStreamingUnitCount.Count))
+		if usedStreamingUnitCount.Status != string(api.ClusterAccepted) {
```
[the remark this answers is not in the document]
miguelsorianod: :+1: . No need to change this as it will be refactored in another PR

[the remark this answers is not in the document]
miguelsorianod: :+1: updated

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990: Makes sense. 

> do we want to compare freeStreamingUnitsInRegion > 0 and used that to set regionLimitReached to true?

> That could be redundant withatLeastOneClusterHasCapacityForLargestInstanceType but it really ensures that we've not consumed all current available max limits in the region for cases where we do not have region limit set.


what I was saying above is 100% percent covered by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/2b7ba102694ac844d1d5421fc02636c43f38a554/internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go#L234

miguelsorianod: :+1: . Marking this as resolved then.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990: That will depend on the other conditions right?

Let me try to enumerate the cases:
- region limits have not been reached and no scale up action is ongoing.
- capacity slack is `0` and freeCapacityInRegion is greater than `0`. Under this `notEnoughCapacitySlackInRegion` will be `false`
- let's assume that `atLeastOneClusterHasCapacityForLargestInstanceType` is `false` (because in the total free capacity in the region is not enough to accomodate the largest kafka), under this condition we need to perform scale up.

miguelsorianod: It shouldn't depend on any other conditions. We should deal with that condition independently.

The question/condition is: "is there enough capacity slack or not?" independently on the other conditions

miguelsorianod: As discussed directly, having `<` makes sense. The behavior would be:
* If the configured capacity slack is 0 or unset that means that no matter the value of free capacity the outcome will be that there is always capacity slack (notEnoughCapacitySlackInRegion will always be false). This means that in that case the only scenario where data plane scale up will be performed is the scenario where the largest kafka size don't fit in any single cluster in the region
* If the configured capacity slack is > 0 then notEnoughCapacitySlackInRegion will be set to true when the free capacity is less than the configured slack

machi1990: Thanks @miguelsorianod for the summary. +1

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -57,14 +62,29 @@ func (m *DynamicScaleUpManager) Stop() {
...
+	// TODO remove this method call and the method itself the new dynamic scale up
+	// logic is ready
+	err := m.reconcileClustersForRegions()
+	if err != nil {
+		errs = append(errs, err...)
+	}
+
+	// TODO remove the "if false" condition once the new dynamic scale up
+	// logic is ready
+	if false {
```
machi1990, line 82: As discussed via chat with other members of the dynamic scaling work, @miguelsorianod  will work on a separate PR to:
- Remove the TODOs 
- Enable scale up with the new logic
- Update unit tests
- Update integration test to take into account the new logic

/cc @VanillaSpoon @JameelB

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -124,3 +144,147 @@ func (m *DynamicScaleUpManager) reconcileClustersForRegions() []error {
...
+					}
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+// shouldScaleUpNewClusterForInstanceType indicates whether a new data plane
```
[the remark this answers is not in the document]
JameelB: This looks good to me too

JameelB (approved):

machi1990 (approved):
:ring:

PR #1207, opened 2022-07-25
refactor: dynamic data plane scale up reconciler logic

miguelsorianod:
## Description
Based on work done in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1205, this PR refactors the introduced logic.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/config/kafka_supported_instance_types_test.go
+++ b/internal/kafka/internal/config/kafka_supported_instance_types_test.go
@@ -868,6 +868,67 @@ func TestKafkaInstanceType_HasAnInstanceSizeWithLifespan(t *testing.T) {
 
 }
 
+func TestKafkaInstanceType_GetBiggestCapacityConsumedSize(t *testing.T) {
+	tests := []struct {
+		name              string
+		kafkaInstanceType KafkaInstanceType
+		want              *KafkaInstanceSize
+	}{
+		{
+			name: "The kafka instance size with the biggest capacity consumed attribute is returned",
+			kafkaInstanceType: KafkaInstanceType{
+				Id: "t1",
+				Sizes: []KafkaInstanceSize{
+					KafkaInstanceSize{Id: "s1", CapacityConsumed: 2},
+					KafkaInstanceSize{Id: "s2", CapacityConsumed: 7},
+					KafkaInstanceSize{Id: "s3", CapacityConsumed: 5},
+				},
+			},
+			want: &KafkaInstanceSize{Id: "s2", CapacityConsumed: 7},
+		},
+		{
+			name: "When there are multiple kafka instance sizes with the highest capacity consumed the first one is returned",
+			kafkaInstanceType: KafkaInstanceType{
+				Id: "t1",
+				Sizes: []KafkaInstanceSize{
+					KafkaInstanceSize{Id: "s1", CapacityConsumed: 2},
+					KafkaInstanceSize{Id: "s2", CapacityConsumed: 7},
+					KafkaInstanceSize{Id: "s3", CapacityConsumed: 7},
+					KafkaInstanceSize{Id: "s4", CapacityConsumed: 6},
+				},
+			},
+			want: &KafkaInstanceSize{Id: "s2", CapacityConsumed: 7},
+		},
+		{
+			name: "When the sizes list of the type is empty nil is returned",
+			kafkaInstanceType: KafkaInstanceType{
+				Id:    "t1",
+				Sizes: []KafkaInstanceSize{},
+			},
+			want: nil,
+		},
+		{
+			name: "When the sizes list of the type is nil nil is returned",
+			kafkaInstanceType: KafkaInstanceType{
+				Id:    "t1",
+				Sizes: nil,
+			},
+			want: nil,
+		},
+	}
+
+	for _, testcase := range tests {
+		tt := testcase
+
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			res := tt.kafkaInstanceType.GetBiggestCapacityConsumedSize()
+			g.Expect(res).To(gomega.Equal(tt.want))
+		})
+	}
+
+}
+
 func buildTestStandardKafkaInstanceSize() KafkaInstanceSize {
 	return KafkaInstanceSize{
 		Id:                          "x1",
--- a/internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
+++ b/internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
@@ -33,7 +33,6 @@ func NewDynamicScaleUpManager(
 	dataplaneClusterConfig *config.DataplaneClusterConfig,
 	clusterProvidersConfig *config.ProviderConfig,
 	kafkaConfig *config.KafkaConfig,
-
 	clusterService services.ClusterService,
 ) *DynamicScaleUpManager {
 
@@ -87,10 +86,6 @@ func (m *DynamicScaleUpManager) Reconcile() []error {
 	}
 
 	return errs
-}
-
-func (m *DynamicScaleUpManager) logReconcileEventEnd() {
-	glog.Infoln("dynamic scale up reconcile event finished")
 }
 
 // reconcileClustersForRegions creates an OSD cluster for each supported cloud provider and region where no cluster exists.
@@ -155,12 +150,25 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 		for _, region := range provider.Regions {
 			for supportedInstanceTypeName := range region.SupportedInstanceTypes {
 				supportedInstanceTypeConfig := region.SupportedInstanceTypes[supportedInstanceTypeName]
-				shouldScaleUp, err := m.shouldScaleUpNewClusterForInstanceType(provider.Name, region.Name, supportedInstanceTypeName, &supportedInstanceTypeConfig, kafkaStreamingUnitCountPerClusterList)
+				var dynamicScaleUpProcessor dynamicScaleUpProcessor = &standardDynamicScaleUpProcessor{
+					locator: supportedInstanceTypeLocator{
+						provider:         provider.Name,
+						region:           region.Name,
+						instanceTypeName: supportedInstanceTypeName,
+					},
+					instanceTypeConfig:                    &supportedInstanceTypeConfig,
+					kafkaStreamingUnitCountPerClusterList: kafkaStreamingUnitCountPerClusterList,
+					supportedKafkaInstanceTypesConfig:     &m.KafkaConfig.SupportedInstanceTypes.Configuration,
+					clusterService:                        m.ClusterService,
+					dryRun:                                false,
+				}
+
+				shouldScaleUp, err := dynamicScaleUpProcessor.ShouldScaleUp()
 				if err != nil {
 					return err
 				}
 				if shouldScaleUp {
-					err := m.registerNewClusterForInstanceType(provider.Name, region.Name, supportedInstanceTypeName)
+					err := dynamicScaleUpProcessor.ScaleUp()
 					if err != nil {
 						return err
 					}
@@ -172,7 +180,82 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 	return nil
 }
 
-// shouldScaleUpNewClusterForInstanceType indicates whether a new data plane
+func (m *DynamicScaleUpManager) logReconcileEventEnd() {
+	glog.Infoln("dynamic scale up reconcile event finished")
+}
+
+// supportedInstanceTypeLocator is a data structure
+// that contains all the information to help locate a
+// supported instance type in a region's cluster
+type supportedInstanceTypeLocator struct {
+	provider         string
+	region           string
+	instanceTypeName string
+}
+
+func (l *supportedInstanceTypeLocator) Equal(other supportedInstanceTypeLocator) bool {
+	return l.provider == other.provider &&
+		l.region == other.region &&
+		l.instanceTypeName == other.instanceTypeName
+}
+
+// dynamicScaleUpExecutor is able to perform dynamic ScaleUp execution actions
+type dynamicScaleUpExecutor interface {
+	ScaleUp() error
+}
+
+// dynamicScaleUpEvaluator is able to perform dynamic ScaleUp evaluation actions
+type dynamicScaleUpEvaluator interface {
+	ShouldScaleUp() (bool, error)
+}
+
+// dynamicScaleUpProcessor is able to process dynamic ScaleUp reconcile events
+type dynamicScaleUpProcessor interface {
+	dynamicScaleUpExecutor
+	dynamicScaleUpEvaluator
+}
+
+// noopDynamicScaleUpProcessor is a dynamicScaleUpProcessor where
+// the scale up is a noop and it always returns that no scale up action
+// is needed
+type noopDynamicScaleUpProcessor struct {
+}
+
+var _ dynamicScaleUpEvaluator = &noopDynamicScaleUpProcessor{}
+var _ dynamicScaleUpExecutor = &noopDynamicScaleUpProcessor{}
+var _ dynamicScaleUpProcessor = &noopDynamicScaleUpProcessor{}
+
+func (p *noopDynamicScaleUpProcessor) ScaleUp() error {
+	return nil
+}
+
+func (p *noopDynamicScaleUpProcessor) ShouldScaleUp() (bool, error) {
+	return false, nil
+}
+
+// standardDynamicScaleUpProcessor is the default dynamicScaleUpProcessor
+// used when dynamic scaling is enabled.
+// It assumes the provided kafkaStreamingUnitCountPerClusterList does not
+// contain any element with a Status attribute with value 'failed'
+type standardDynamicScaleUpProcessor struct {
+	locator            supportedInstanceTypeLocator
+	instanceTypeConfig *config.InstanceTypeConfig
+	// kafkaStreamingUnitCountPerClusterList must not contain any element
+	// with a Status attribute with value 'failed'
+	kafkaStreamingUnitCountPerClusterList services.KafkaStreamingUnitCountPerClusterList
+	supportedKafkaInstanceTypesConfig     *config.SupportedKafkaInstanceTypesConfig
+	clusterService                        services.ClusterService
+
+	// dryRun controls whether the ScaleUp method performs real actions.
+	// Useful when you don't want to trigger a real scale up.
+	dryRun bool
+}
+
+var _ dynamicScaleUpEvaluator = &standardDynamicScaleUpProcessor{}
+var _ dynamicScaleUpExecutor = &standardDynamicScaleUpProcessor{}
+var _ dynamicScaleUpProcessor = &standardDynamicScaleUpProcessor{}
+
+// ShouldScaleUp indicates whether a new data plane
 // cluster should be created for a given instance type in the given provider
 // and region.
 // It returns true if all the following conditions happen:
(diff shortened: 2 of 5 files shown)


miguelsorianod:
This PR does not include documentation, testing or removal of the obsolete `reconcileClustersForRegions` method. It will be added later once we review that the refactor looks good.

The visibility of some of the attributes of the newly added types might change in the near future when testing is introduced.

I recommend looking at the whole file and not only the diff to get the full view of the new code.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -155,12 +150,24 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
+				shouldScaleUp, err := dynamicScaleUpProcessor.ShouldScaleUp()
 				if err != nil {
 					return err
 				}
 				if shouldScaleUp {
-					err := m.registerNewClusterForInstanceType(provider.Name, region.Name, supportedInstanceTypeName)
+					err := dynamicScaleUpProcessor.ScaleUp()
 					if err != nil {
 						return err
 					}
```
machi1990, line 173: Should we accumalate the list of errors and then send out an `ErrorList` contained the error for each region?
The advantage is that we won't be blocking up scale up evaluation for others regions if we encounter an error in the first regions

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -155,12 +150,24 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 		for _, region := range provider.Regions {
 			for supportedInstanceTypeName := range region.SupportedInstanceTypes {
 				supportedInstanceTypeConfig := region.SupportedInstanceTypes[supportedInstanceTypeName]
-				shouldScaleUp, err := m.shouldScaleUpNewClusterForInstanceType(provider.Name, region.Name, supportedInstanceTypeName, &supportedInstanceTypeConfig, kafkaStreamingUnitCountPerClusterList)
+				var dynamicScaleUpProcessor dynamicScaleUpProcessor = &standardDynamicScaleUpProcessor{
```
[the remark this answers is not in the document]
miguelsorianod: correct

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -155,12 +150,24 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
+					locator: supportedInstanceTypeLocator{
+						provider:         provider.Name,
+						region:           region.Name,
+						instanceTypeName: supportedInstanceTypeName,
+					},
+					instanceTypeConfig:                    &supportedInstanceTypeConfig,
+					kafkaStreamingUnitCountPerClusterList: kafkaStreamingUnitCountPerClusterList,
+					supportedKafkaInstanceTypesConfig:     &m.KafkaConfig.SupportedInstanceTypes.Configuration,
+					dryRun:                                false,
```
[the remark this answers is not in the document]
machi1990: That super useful. e.g in a certain region we could make `dry` run as true and that won't trigger any scale up while observing the scale up evaluation logic, correct? 
It has potential to allow us to control how this feature behaves in finer details.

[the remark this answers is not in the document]
miguelsorianod: :+1:, as long we expose that configurability somehow

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990: Thanks. 

We could go with the current way of doing things i.e return a list of `error` i.e `[]error` and we could have a followup on doing it properly with the `ErrorList` type

miguelsorianod: Even with the `ErrorList` I think it needs some more thinking because the issue is the same about the callers

miguelsorianod: I will go with the []error approach for now and think about it

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -172,6 +179,81 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
+
+func (l *supportedInstanceTypeLocator) Equal(other supportedInstanceTypeLocator) bool {
+	return l.provider == other.provider &&
+		l.region == other.region &&
+		l.instanceTypeName == other.instanceTypeName
+}
+
+// dynamicScaleUpExecutor is able to perform dynamic ScaleUp execution actions
+type dynamicScaleUpExecutor interface {
+	ScaleUp() error
```
JameelB, line 203: Is there a reason why these methods are exported?

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +277,204 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
+		SupportedInstanceType: p.locator.instanceTypeName,
 		MultiAZ:               newClusterMultiAZ,
 		Status:                api.ClusterAccepted,
 		ProviderType:          api.ClusterProviderOCM,
 	}
 
-	err := m.ClusterService.RegisterClusterJob(clusterRequest)
+	if p.dryRun {
+		return nil
+	}
```
JameelB, line 338: If dryRun is enabled, we don't need to set `newClusterMultiAZ` and `clusterRequest` right? Can we move this to the top of this method?

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990: here is my thinking thoughts:
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5c2eb2ca43caef5d1e19711bcd8388e58bdc217e/internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go#L92 already returns a list of error.
- we could change the https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5c2eb2ca43caef5d1e19711bcd8388e58bdc217e/internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go#L143 to return the list of `error`, the same for the methods called by this.

We do not need to check for the error type here and once the `.Reconcile()` is called, if errors occurred then they are logged automatically

Later on, on a possibly bing bang refactor, all the reconcilers will be modified to start using `ErrorList` or the CompoundError approach.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +277,204 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
+// instanceTypeConsumptionSummaryCalculator calculates a consumption summary
+// for the provided supportedInstanceTypeLocator based on the
+// provided KafkaStreamingUnitCountPerClusterList
+type instanceTypeConsumptionSummaryCalculator struct {
+	locator                               supportedInstanceTypeLocator
+	kafkaStreamingUnitCountPerClusterList services.KafkaStreamingUnitCountPerClusterList
+	supportedKafkaInstanceTypesConfig     *config.SupportedKafkaInstanceTypesConfig
+}
+
+// Calculate returns a instanceTypeConsumptionSummary containing a consumption
```
machi1990, line 404: I love these comments:

Once we've the implementation in place we could copy-paste comments like these onto the https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/docs/architecture/data-plane-osd-cluster-dynamic-scaling.md

JameelB: +:100:  really makes it clear on what each component is for and makes it easier to follow the different conditions in the methods :)

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: I wouldn't like to return an slice of errors in the `processDynamicScaleUpReconcileEvent` method.
I will make use of the `ErrorList` approach.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: No reason in particular. I still not have a clear idea on what we will export.
I am guessing integration testing will influence this.
My preference in general is to have attributes/methods unexported if possible.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: We can move this to the top of the method yes :+1:

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
JameelB: +1 no issue for me with these being exportable for now until we know what we need to export. Thanks for the explanation! :)

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: Moved

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: I updated the PR.
Now processDynamicScaleUpReconcileEvent does not directly return when one of the region's provider returns an error. Instead, the error is appended to a list of errors and when all of the region's providers have been processed it is returned.
The signature of the method returns an `error` intentionally (as opposed to a []error or an ErrorList).
The ErrorList type has been updated to allow adding errors to it.

miguelsorianod:
Unit tests for the following types have been added:
* noopDynamicScaleUpProcessor
* instanceTypeConsumptionSummaryCalculator
* supportedInstanceTypeLocator

miguelsorianod:
Unit tests added for KafkaInstanceType.GetBiggestCapacityConsumedSize

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990: Thank you Miguel

miguelsorianod:
Unit tests for `standardDynamicScaleUpProcessor` have been added.
There are a couple of method for the `DynamicScaleUpManager` type that have not been covered: Reconcile and processDynamicScaleUpReconcileEvent, but I think those should be covered once we add integration tests.

miguelsorianod:
Let me know if there are some scenarios in the unit tests that have not been covered that would be interesting to keep track of

internal/kafka/internal/config/kafka_supported_instance_types_test.go
```diff
@@ -868,6 +868,67 @@ func TestKafkaInstanceType_HasAnInstanceSizeWithLifespan(t *testing.T) {
...
+	}{
+		{
+			name: "The kafka instance size with the biggest capacity consumed attribute is returned",
+			kafkaInstanceType: KafkaInstanceType{
+				Id: "t1",
+				Sizes: []KafkaInstanceSize{
+					KafkaInstanceSize{Id: "s1", CapacityConsumed: 2},
+					KafkaInstanceSize{Id: "s2", CapacityConsumed: 7},
+					KafkaInstanceSize{Id: "s3", CapacityConsumed: 5},
+				},
```
ziccardi, line 885: `KafkaInstanceSize` inside array initialization is redundant
```suggestion
				Sizes: []KafkaInstanceSize{
					{Id: "s1", CapacityConsumed: 2},
					{Id: "s2", CapacityConsumed: 7},
					{Id: "s3", CapacityConsumed: 5},
				},
```

internal/kafka/internal/config/kafka_supported_instance_types_test.go
```diff
@@ -868,6 +868,67 @@ func TestKafkaInstanceType_HasAnInstanceSizeWithLifespan(t *testing.T) {
...
+		},
+		{
+			name: "When there are multiple kafka instance sizes with the highest capacity consumed the first one is returned",
+			kafkaInstanceType: KafkaInstanceType{
+				Id: "t1",
+				Sizes: []KafkaInstanceSize{
+					KafkaInstanceSize{Id: "s1", CapacityConsumed: 2},
+					KafkaInstanceSize{Id: "s2", CapacityConsumed: 7},
+					KafkaInstanceSize{Id: "s3", CapacityConsumed: 7},
+					KafkaInstanceSize{Id: "s4", CapacityConsumed: 6},
```
ziccardi, line 897: See previous comment

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +282,206 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
-			consumedStreamingUnitsInRegion = consumedStreamingUnitsInRegion + int(kafkaStreamingUnitCountPerCluster.Count)
-			if kafkaStreamingUnitCountPerCluster.Status != api.ClusterDeprovisioning.String() &&
-				kafkaStreamingUnitCountPerCluster.Status != api.ClusterCleanup.String() {
-				maxStreamingUnitsInRegion = maxStreamingUnitsInRegion + int(kafkaStreamingUnitCountPerCluster.MaxUnits)
-			}
-		}
+	regionLimitReached := p.regionLimitReached(instanceTypeConsumptionInRegionSummary)
+	if regionLimitReached {
+		return false, nil
 	}
```
ziccardi, line 305:
```suggestion
	if p.regionLimitReached(instanceTypeConsumptionInRegionSummary) {
		return false, nil
	}
```

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +282,206 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
 
-	freeStreamingUnitsInRegion := maxStreamingUnitsInRegion - consumedStreamingUnitsInRegion
-	capacitySlackInRegion := instanceTypeConfig.MinAvailableCapacitySlackStreamingUnits
-	notEnoughCapacitySlackInRegion := false
-	if freeStreamingUnitsInRegion < capacitySlackInRegion { // TODO is it <= or < ?
-		notEnoughCapacitySlackInRegion = true
+	ongoingScaleUpActionInRegion := p.ongoingScaleUpAction(instanceTypeConsumptionInRegionSummary)
+	if ongoingScaleUpActionInRegion {
+		return false, nil
 	}
```
ziccardi, line 310:
```suggestion
	if p.ongoingScaleUpAction(instanceTypeConsumptionInRegionSummary) {
		return false, nil
	}
```

internal/kafka/internal/config/kafka_supported_instance_types_test.go
miguelsorianod: I am aware of that but it helps me identify the types even if they are in the slice above

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: Style preference. In this case I preferred to have a shorter condition

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
JameelB: Thanks!

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: same as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1207#discussion_r930813654

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +282,206 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
 
-	streamingUnitsLimitInRegion := instanceTypeConfig.Limit
-	regionLimitReached := false
-	if streamingUnitsLimitInRegion != nil && consumedStreamingUnitsInRegion >= *streamingUnitsLimitInRegion {
-		regionLimitReached = true
+	freeCapacityForBiggestInstanceSizeInRegion := p.freeCapacityForBiggestInstanceSize(instanceTypeConsumptionInRegionSummary)
+	if !freeCapacityForBiggestInstanceSizeInRegion {
+		return true, nil
 	}
```
ziccardi, line 315:
```suggestion
	if !p.freeCapacityForBiggestInstanceSize(instanceTypeConsumptionInRegionSummary) {
		return true, nil
	}
```

miguelsorianod: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1207#discussion_r930813906

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +282,206 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
 
-	if !regionLimitReached && !scaleUpActionIsOngoing {
-		if !atLeastOneClusterHasCapacityForLargestInstanceType || (atLeastOneClusterHasCapacityForLargestInstanceType && notEnoughCapacitySlackInRegion) {
-			return true, nil
-		}
+	enoughCapacitySlackInRegion := p.enoughCapacitySlackInRegion(instanceTypeConsumptionInRegionSummary)
+	if !enoughCapacitySlackInRegion {
+		return true, nil
 	}
```
ziccardi, line 320:
```suggestion
	if !p.enoughCapacitySlackInRegion(instanceTypeConsumptionInRegionSummary) {
		return true, nil
	}
```

internal/kafka/internal/config/kafka_supported_instance_types_test.go
miguelsorianod: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1207#discussion_r930813132

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1207#discussion_r930815196

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +282,206 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
 
 	return false, nil
 }
 
 // registerNewClusterForInstanceType triggers a new data plane cluster
 // registration for a given instance type in a provider region.
-func (m *DynamicScaleUpManager) registerNewClusterForInstanceType(provider string, region string, instanceTypeName string) error {
+func (p *standardDynamicScaleUpProcessor) ScaleUp() error {
```
ziccardi, line 327: The comment should be updated

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -195,96 +282,206 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
...
 		ProviderType:          api.ClusterProviderOCM,
 	}
 
-	err := m.ClusterService.RegisterClusterJob(clusterRequest)
+	err := p.clusterService.RegisterClusterJob(clusterRequest)
 	if err != nil {
 		return err
 	}
 
 	return nil
```
ziccardi, line 350: return p.clusterService.RegisterClusterJob(clusterRequest)

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: :+1: fixed

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: Style preference

miguelsorianod:
Integration tests and logging will be added in another PR

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
ziccardi: Not really style preference: with the current code you are declaring a variable: the value returned by the function call gets copied into the variable and the memory is released only when the variable is dismissed, while with the proposed approach there is no copy and memory can be released immediately.

I leave however the decision to you.

miguelsorianod: :+1: I prefer to keep it as it is. I am not worried about memory consumption of 4 booleans. In practical terms I consider it a style preference

machi1990: I often have the same style as @miguelsorianod.

I tend to prefer it for code readability reasons: you understand what the boolean (or even other variable is for) other than trying to evaluating or guess it from the the expression in the if condition.

ziccardi: Depends on the case. I don't think `regionLimitReached` variable gives more info than the call to `p.regionLimitedReached`, so in this case, I don't agree. But I don't want to start another long discussion, it is not important. I just suggested what I thought was an improvement, albeit small.

miguelsorianod: It is correct for that case. But if you look at the overall `shouldScaleUp` method it helps readability that it is consistent with the way other conditions are checked where the variable gives more information

miguelsorianod: +1 on it depending on the case, and not really important as in practical terms it depends on style preference

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr_test.go
```diff
@@ -11,6 +11,852 @@ import (
...
+				},
+			},
+			args: args{
+				summary: instanceTypeConsumptionSummary{
+					consumedStreamingUnits: 1000,
+				},
+			},
+			want: false,
+		},
+	}
```
JameelB, line 501: could also add a test here if limit was set to 0? In that case even if consumedStreamingUnits is `0`, region limit should always be reached.

miguelsorianod: That case should be kind of covered with the `When the supported instance type consumed streaming units has reached the region limit true is returned` as 0 is not any special number in that regard.
But even in that case it does no harm, and could cover potential changes of  behavior of the logic in the future so I will add it.

miguelsorianod: added

JameelB: thanks! :)

JameelB (approved):

machi1990 (approved):
Thanks @miguelsorianod this is great work.
Excited for the next phase of this.

PR #1208, opened 2022-07-26
fix(cluster_mgr): harden setKafkaPerClusterCountMetrics to not err when ClusterID is empty

machi1990:
## Description

Harden setKafkaPerClusterCountMetrics to not err when ClusterID is empty.

This avoids sending out an error for kafkas counts that have not been assigned to an OSD cluster
yet. Follows up on
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1167#issuecomment-1190536398 and
address the first error shown there

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/workers/clusters_mgr.go
+++ b/internal/kafka/internal/workers/clusters_mgr.go
@@ -1221,16 +1221,26 @@ func (c *ClusterManager) setClusterStatusCountMetrics() error {
 }
 
 func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
-	if counters, err := c.ClusterService.FindKafkaInstanceCount([]string{}); err != nil {
+	counters, err := c.ClusterService.FindKafkaInstanceCount([]string{})
+	if err != nil {
 		return err
-	} else {
-		for _, counter := range counters {
-			clusterExternalID, err := c.ClusterService.GetExternalID(counter.Clusterid)
-			if err != nil {
-				return err
-			}
-			metrics.UpdateKafkaPerClusterCountMetric(counter.Clusterid, clusterExternalID, counter.Count)
-		}
 	}
+
+	for _, counter := range counters {
+		// Ignore counters that do not have the cluster id set as they'll err when retrieving the cluster external id with not found
+		// For this to occur, either the counter included:
+		// 1. rejected kafkas but not re-assigned
+		// 2. or accepted kafkas but have not assigned in an OSD cluster
+		if counter.Clusterid == "" {
+			continue
+		}
+
+		clusterExternalID, err := c.ClusterService.GetExternalID(counter.Clusterid)
+		if err != nil {
+			return err
+		}
+		metrics.UpdateKafkaPerClusterCountMetric(counter.Clusterid, clusterExternalID, counter.Count)
+	}
+
 	return nil
 }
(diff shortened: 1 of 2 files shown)


machi1990:
@JameelB can you have a look? This follows the following comment https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1167#issuecomment-1190536398

The situation that's avoided here can pretty much arise in the context of dynamic scaling as:
- Kafkas are re-assigned if they get rejected
- New Kafkas are accepted if region limits not reached but no cluster found at the moment of Kafka creation

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -1225,6 +1225,11 @@ func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
 		return err
 	} else {
 		for _, counter := range counters {
+			// ignore counts that do not have the cluster id set as they'll err when retrieving the cluster external id with not found
```
machi1990, line 1228: either the kafkas count includes:
- rejected kafkas but not re-assigned
- or accepted kafkas but have not assigned in an OSD cluster

machi1990:
We could also argue if the `setKafkaPerClusterCountMetrics` is needed now that we have a capacity used metric for each cluster (see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1167#issuecomment-1190836289).

The underlying metric produced by this is used in some dashaboards, I didn't want to break them. If we deem that the metrics are becoming too redundant, we could remove this metric at a later stage.

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -1225,6 +1225,14 @@ func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
 		return err
 	} else {
```
JameelB, line 1226: Wdyt of moving the `err != nil` check? This way we won't need to surround the rest of the code with an `else`.

JameelB (commented):
Change looks good to me, we shouldn't include Kafkas that hasn't been assigned to a data plane cluster in the metric. 

> We could also argue if the `setKafkaPerClusterCountMetrics` is needed now that we have a capacity used metric for each cluster (see [#1167 (comment)](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1167#issuecomment-1190836289)).
> 
> The underlying metric produced by this is used in some dashaboards, I didn't want to break them. If we deem that the metrics are becoming too redundant, we could remove this metric at a later stage.

+1 on revisiting, both metrics result in a streaming unit count so they're the exact same.

internal/kafka/internal/workers/clusters_mgr.go
machi1990: I've updated the logic.
Can you have another look?

> +1 on revisiting, both metrics result in a streaming unit count so they're the exact same.

I'll create a JIRA for this.

machi1990:
> We could also argue if the `setKafkaPerClusterCountMetrics` is needed now that we have a capacity used metric for each cluster (see [#1167 (comment)](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1167#issuecomment-1190836289)).
> 
> The underlying metric produced by this is used in some dashaboards, I didn't want to break them. If we deem that the metrics are becoming too redundant, we could remove this metric at a later stage.

Created https://issues.redhat.com/browse/MGDSTRM-9283

internal/kafka/internal/workers/clusters_mgr.go
machi1990: of course, I  updated the change and forgot to push. Everything has been pushed now.

JameelB (approved):
Thanks for this @machi1990! :rocket:

machi1990:
Thank you for the review @JameelB

PR #1222, opened 2022-08-01
feat: GetAvailableSizesInRegion only checks region limts when dynamic scaling enabled

miguelsorianod:
## Description
Related to https://issues.redhat.com/browse/MGDSTRM-9309

The `GetAvailableSizesInRegion` method, used in the list cloud provider regions endpoint (/{id}/regions) has been updated so only region limit checks are performed when dynamic scaling is enabled. The behavior when manual scaling is disabled has been kept the same.

The reason of this change is due to in dynamic scaling mode as long as the region limit is not hit there should evantually be capacity available. So we consider that size should be available as long as the region limit is not hit, even if at that moment the kafka size cannot fit in any already existing cluster.

## Verification Steps
unit and integration tests should still pass.
Calling the cloud provider endpoint enabling and disabling dynamic scaling should behave differently when there is no capacity in any existing cluster but the region limit has not been hit.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -220,6 +220,7 @@ func (k *kafkaService) GetAvailableSizesInRegion(criteria *FindClusterCriteria) 
 		return nil, err
 	}
 
+	indexOfBiggestKafkaSizeAllowed := -1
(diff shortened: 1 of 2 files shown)


machi1990 (approved):
lgtm

Thanks @miguelsorianod

PR #1224, opened 2022-08-01
feat: accept a kafka even if there is no ready cluster but region limits not reached during dynamic scaling

machi1990:
## Description

The option that's chosen here is the one that:

- Does not change how manual mode works
- Add an IF statement to check for scaling mode, if dynamic scaling accept the kafka anyway if region limits not reached
- Add a reconciliation steps that checks for Kafkas in accepted that do not have ClusterID set and assigns the ClusterID. Do this irrespective of the scaling mode and do this within some reasonable timeout of an hour (enough for a new cluster to be provisioned and terraformed).
 
Closes https://issues.redhat.com/browse/MGDSTRM-8914

## Verification Steps

Added unit tests should pass

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr.go
+++ b/internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr.go
@@ -8,6 +8,8 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
 	"github.com/google/uuid"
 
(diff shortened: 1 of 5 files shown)


internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr.go
```diff
@@ -78,35 +78,86 @@ func (k *AcceptedKafkaManager) Reconcile() []error {
...
+		cluster = foundCluster
 	}
-	kafka.DesiredStrimziVersion = latestReadyStrimziVersion.Version
 
+	latestReadyStrimziVersion, err := cluster.GetLatestAvailableAndReadyStrimziVersion()
+	if err != nil {
+		logger.Logger.Error(err)
+	}
+
+	noLatestReadyStrimziVersionFound := err != nil || latestReadyStrimziVersion == nil
```
miguelsorianod, line 106: If an error happens I think we can just return the error instead of calling `markTheAssignedKafkaAsFailedOrAllowRetryStrimziVersionPickingReconciliation`. The same happens when trying to get the ibp version or the kafka version.

machi1990: Thanks. That makes sense, I've updated the code. I think for an error to occur here, it's mostly due to json parsing issues which we may not encounter at this stage as we only store valid json in the database.

miguelsorianod (approved):
looks good to me :+1:

machi1990:
Thanks @miguelsorianod

machi1990:
I'll run the test suite against a real env before mergin.

machi1990:
> I'll run the test suite against a real env before merging.

Done with success

PR #1226, opened 2022-08-03
fix: add dynamic flag max-allowed-developer-instances

ajaypratap003:
## Description
Add dynamic flag max-allowed-developer-instances and set default value one.
https://issues.redhat.com/browse/MGDSTRM-8159

## Verification Steps
set max-allowed-developer-instances flag value and will bale to create kafka trial instances based on flag value. 
If you will try to create kafka trial instances more than flag value , it should return error.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/config/kafka.go
+++ b/internal/kafka/internal/config/kafka.go
@@ -56,7 +56,7 @@ func (c *KafkaConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.BrowserUrl, "browser-url", c.BrowserUrl, "Browser url to kafka admin UI")
 	fs.BoolVar(&c.EnableKafkaOwnerConfig, "enable-kafka-owner-config", c.EnableKafkaOwnerConfig, "Enable configuration for setting kafka owners")
 	fs.StringVar(&c.KafkaOwnerListFile, "kafka-owner-list-file", c.KafkaOwnerListFile, "File containing list of kafka owners")
-	fs.IntVar(&c.Quota.MaxAllowedTrialInstance, "max-allowed-trial-instances", c.Quota.MaxAllowedTrialInstance, "Max allowed trial instances")
+	fs.IntVar(&c.Quota.MaxAllowedDeveloperInstances, "max-allowed-developer-instances", c.Quota.MaxAllowedDeveloperInstances, "As a user, one can create up to N defined max developer instances if they do not have quota to create standard instances")
 }
 
 func (c *KafkaConfig) ReadFiles() error {
--- a/internal/kafka/internal/config/kafka.go
+++ b/internal/kafka/internal/config/kafka.go
@@ -56,6 +56,7 @@ func (c *KafkaConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.BrowserUrl, "browser-url", c.BrowserUrl, "Browser url to kafka admin UI")
 	fs.BoolVar(&c.EnableKafkaOwnerConfig, "enable-kafka-owner-config", c.EnableKafkaOwnerConfig, "Enable configuration for setting kafka owners")
 	fs.StringVar(&c.KafkaOwnerListFile, "kafka-owner-list-file", c.KafkaOwnerListFile, "File containing list of kafka owners")
+	fs.IntVar(&c.Quota.MaxAllowedTrialInstance, "max-allowed-trial-instances", c.Quota.MaxAllowedTrialInstance, "Max allowed trial instances")
 }
 
 func (c *KafkaConfig) ReadFiles() error {
--- a/internal/kafka/internal/config/kafka_quota.go
+++ b/internal/kafka/internal/config/kafka_quota.go
@@ -3,15 +3,15 @@ package config
 import "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 
 type KafkaQuotaConfig struct {
-	Type                    string
-	AllowDeveloperInstance  bool
-	MaxAllowedTrialInstance int
+	Type                         string
+	AllowDeveloperInstance       bool
+	MaxAllowedDeveloperInstances int
 }
 
 func NewKafkaQuotaConfig() *KafkaQuotaConfig {
 	return &KafkaQuotaConfig{
-		Type:                    api.QuotaManagementListQuotaType.String(),
-		AllowDeveloperInstance:  true,
-		MaxAllowedTrialInstance: 1,
+		Type:                         api.QuotaManagementListQuotaType.String(),
+		AllowDeveloperInstance:       true,
+		MaxAllowedDeveloperInstances: 1,
 	}
 }
--- a/internal/kafka/internal/config/kafka_quota.go
+++ b/internal/kafka/internal/config/kafka_quota.go
@@ -3,13 +3,15 @@ package config
 import "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 
 type KafkaQuotaConfig struct {
-	Type                   string
-	AllowDeveloperInstance bool
+	Type                    string
+	AllowDeveloperInstance  bool
+	MaxAllowedTrialInstance int
 }
 
 func NewKafkaQuotaConfig() *KafkaQuotaConfig {
 	return &KafkaQuotaConfig{
-		Type:                   api.QuotaManagementListQuotaType.String(),
-		AllowDeveloperInstance: true,
+		Type:                    api.QuotaManagementListQuotaType.String(),
+		AllowDeveloperInstance:  true,
+		MaxAllowedTrialInstance: 1,
 	}
 }
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -309,8 +309,10 @@ func (k *kafkaService) reserveQuota(kafkaRequest *dbapi.KafkaRequest) (subscript
 			return "", errors.NewWithCause(errors.ErrorGeneral, err, "failed to count kafka %s instances", instType.DisplayName)
 		}
 
-		if count > 0 {
-			return "", errors.TooManyKafkaInstancesReached("only one %s instance is allowed", instType.DisplayName)
+		maxAllowedTrialInstances := k.kafkaConfig.Quota.MaxAllowedTrialInstance
+
+		if count > int64(maxAllowedTrialInstances) {
+			return "", errors.TooManyKafkaInstancesReached(fmt.Sprintf("only %d %s instance is allowed", maxAllowedTrialInstances, instType.DisplayName))
 		}
 	}
 
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -297,7 +297,7 @@ func (k *kafkaService) reserveQuota(kafkaRequest *dbapi.KafkaRequest) (subscript
 			return "", errors.NewWithCause(errors.ErrorForbidden, err, "kafka %s instances are not allowed", instType.DisplayName)
 		}
 
-		// Only one DEVELOPER instance is admitted. Let's check if the user already owns one
+		//N DEVELOPER instance is admitted. Let's check if the user already owns N instances
 		dbConn := k.connectionFactory.New()
 		var count int64
 		if err := dbConn.Model(&dbapi.KafkaRequest{}).
@@ -309,10 +309,10 @@ func (k *kafkaService) reserveQuota(kafkaRequest *dbapi.KafkaRequest) (subscript
 			return "", errors.NewWithCause(errors.ErrorGeneral, err, "failed to count kafka %s instances", instType.DisplayName)
 		}
 
-		maxAllowedTrialInstances := k.kafkaConfig.Quota.MaxAllowedTrialInstance
+		maxAllowedDeveloperInstances := k.kafkaConfig.Quota.MaxAllowedDeveloperInstances
 
-		if count > int64(maxAllowedTrialInstances) {
-			return "", errors.TooManyKafkaInstancesReached(fmt.Sprintf("only %d %s instance is allowed", maxAllowedTrialInstances, instType.DisplayName))
+		if count >= int64(maxAllowedDeveloperInstances) {
+			return "", errors.TooManyKafkaInstancesReached(fmt.Sprintf("only %d %s instance is allowed", maxAllowedDeveloperInstances, instType.DisplayName))
 		}
 	}
 
--- a/internal/kafka/internal/services/kafka_test.go
+++ b/internal/kafka/internal/services/kafka_test.go
@@ -1222,7 +1222,6 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
 				mocket.Catcher.NewMock().WithQuery(`SELECT count(1) FROM "kafka_requests" WHERE instance_type = $1 AND owner = $2 AND (organisation_id = $3) AND "kafka_requests"."deleted_at" IS NULL`).
 					WithArgs(types.DEVELOPER.String(), testUser, "org-id").
 					WithReply(totalCountResponse)
-				mocket.Catcher.NewMock().WithQueryException().WithExecException()
 
 				mocket.Catcher.NewMock().
 					WithQuery(`SELECT * FROM "kafka_requests" WHERE region = $1 AND cloud_provider = $2 AND instance_type = $3 AND "kafka_requests"."deleted_at" IS NULL`).
@@ -1288,7 +1287,6 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
 				mocket.Catcher.NewMock().WithQuery(`SELECT count(1) FROM "kafka_requests" WHERE instance_type = $1 AND owner = $2 AND (organisation_id = $3) AND "kafka_requests"."deleted_at" IS NULL`).
 					WithArgs(types.DEVELOPER.String(), testUser, "org-id").
 					WithReply(totalCountResponse)
-				mocket.Catcher.NewMock().WithQueryException().WithExecException()
 			},
 			error: errorCheck{
 				wantErr:  true,
--- a/internal/kafka/internal/services/kafka_test.go
+++ b/internal/kafka/internal/services/kafka_test.go
@@ -1289,19 +1289,6 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
 					WithArgs(types.DEVELOPER.String(), testUser, "org-id").
 					WithReply(totalCountResponse)
 				mocket.Catcher.NewMock().WithQueryException().WithExecException()
-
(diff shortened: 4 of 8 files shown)


internal/kafka/internal/services/kafka.go
```diff
@@ -309,8 +309,10 @@ func (k *kafkaService) reserveQuota(kafkaRequest *dbapi.KafkaRequest) (subscript
 			return "", errors.NewWithCause(errors.ErrorGeneral, err, "failed to count kafka %s instances", instType.DisplayName)
 		}
 
-		if count > 0 {
-			return "", errors.TooManyKafkaInstancesReached("only one %s instance is allowed", instType.DisplayName)
+		maxAllowedTrialInstances := k.kafkaConfig.Quota.MaxAllowedTrialInstance
+
+		if count > int64(maxAllowedTrialInstances) {
```
ziccardi, line 314: I believe this should be `>=` to avoid allowing `maxAllowedTrialInstances + 1` instances.

internal/kafka/internal/config/kafka.go
```diff
@@ -56,6 +56,7 @@ func (c *KafkaConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.BrowserUrl, "browser-url", c.BrowserUrl, "Browser url to kafka admin UI")
 	fs.BoolVar(&c.EnableKafkaOwnerConfig, "enable-kafka-owner-config", c.EnableKafkaOwnerConfig, "Enable configuration for setting kafka owners")
 	fs.StringVar(&c.KafkaOwnerListFile, "kafka-owner-list-file", c.KafkaOwnerListFile, "File containing list of kafka owners")
+	fs.IntVar(&c.Quota.MaxAllowedTrialInstance, "max-allowed-trial-instances", c.Quota.MaxAllowedTrialInstance, "Max allowed trial instances")
```
miguelsorianod, line 59: If this ends up being merged, this should be MaxAllowedTrialInstances (plural)

templates/service-template.yml
```diff
@@ -568,6 +568,11 @@ parameters:
   displayName: Kas-fleetshard operator subscription starting config
   description: Kas-fleetshard operator subscription starting config
 
+- name: MAX_ALLOWED_TRIAL_INSTANCES
+  displayName: Max allowed trial instances config 
+  description: Max allowed trial instances config
```
miguelsorianod, line 573: The descriptions should be more specific, are we talking about max trial instances per kas fleet manager? per account? per organization?
Also I guess the concept of `trial` does not apply in kfm code, it should be eval instead? What do you think @machi1990 ?

internal/kafka/internal/config/kafka.go
miguelsorianod: Also see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1226/files#r936580215. Depending on the outcome of that discussion the attribute should be renamed too

internal/kafka/internal/services/kafka.go
miguelsorianod: Is that correct?

The count attribute is the the number of already existing kafka developer instances.
For example, if there is 1 already existing kafka instance the count would be 1.
If you set maxAllowedTrialInstances to 1 then the condition would be
1 >= 1 which would be true, which means that creation would be allowed, which is not what we want.

ziccardi: if count is 1, 1>=1 would be true, so it would enter the `if` and return an error

internal/kafka/internal/config/kafka_quota.go
```diff
@@ -3,13 +3,15 @@ package config
...
+	MaxAllowedTrialInstance int
 }
 
 func NewKafkaQuotaConfig() *KafkaQuotaConfig {
 	return &KafkaQuotaConfig{
-		Type:                   api.QuotaManagementListQuotaType.String(),
-		AllowDeveloperInstance: true,
+		Type:                    api.QuotaManagementListQuotaType.String(),
+		AllowDeveloperInstance:  true,
+		MaxAllowedTrialInstance: 1,
```
miguelsorianod, line 15: I wonder if now `AllowDeveloperInstance` and `MaxAllowedTrialInstance` are now redundant?
Wouldn't a `MaxAllowedTrialInstance` equaling 0 the same as AllowDeveloperInstance = false?

What do you think @ziccardi @machi1990 ?

templates/service-template.yml
ziccardi: I agree and I believe that both `MAX_ALLOWED_TRIAL_INSTANCES` and `MaxAllowedTrialInstance` should be renamed to `MAX_ALLOWED_DEVELOPER_INSTANCES` and `MaxAllowedDeveloperInstance`

internal/kafka/internal/config/kafka_quota.go
ziccardi: Good point. I think merging this PR will obsolete `AllowDeveloperInstance`

templates/service-template.yml
machi1990: +1. I think the flags should be specific to `developer` instances, `trial` instance is a different thingy which may come at a later stage.

internal/kafka/internal/config/kafka_quota.go
machi1990: +1.

internal/kafka/internal/services/kafka.go
machi1990: +1. It'll be good to have/add a unit test for this as well.

machi1990, line 314: let's also update https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ce15e67a61a882a968291405074eb736053c6178/internal/kafka/internal/services/kafka.go#L300

ajaypratap003: Yes, you are right. It should be `>=`. I'll make this change.

templates/service-template.yml
ajaypratap003: As I'm exploring KFM. Don't know much knowledge about KFM.
Can someone explain to me, would this flag will be per kfm or per  account or per organization  ?

machi1990: The flag is per user.

as a user, one can create up to `N` defined max instances if they do not have quota to create standard instances.

internal/kafka/internal/services/kafka.go
ajaypratap003: > +1. It'll be good to have/add a unit test for this as well.

We have unit test for it. But it's failing. Can you suggest me what is wrong in test ? This is my first test in kfm and golang :(

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ce15e67a61a882a968291405074eb736053c6178/internal/kafka/internal/services/kafka_test.go#L1180

internal/kafka/internal/config/kafka_quota.go
ajaypratap003: should I remove AllowDeveloperInstance flag in this PR ?

internal/kafka/internal/services/kafka.go
machi1990: You'll need to mock the `count` SQL query in the test.
Something like
```go
mocket.Catcher.NewMock().WithQuery(`SELECT count(1) FROM "kafka_requests" WHERE instance_type = $1 AND owner = $2 AND (organisation_id = $3) AND "kafka_requests"."deleted_at" IS NULL`).
					WithArgs(types.DEVELOPER.String(), testUser, "org-id").
					WithReply([]map[string]interface{}{{"count": 1}})
```
added before https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ce15e67a61a882a968291405074eb736053c6178/internal/kafka/internal/services/kafka_test.go#L1229 will do.

machi1990: I'll also suggest to:
change the value of https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ce15e67a61a882a968291405074eb736053c6178/internal/kafka/internal/services/kafka_test.go#L1190 to 2 and the returned mock count (see above comment) db query response to 2. 

Also, I'll suggest to add another test case that pass the max developer instance check.
i.e `MaxAllowedTrialInstance` is say maybe 1 and the count in the database is `0`. With this setup, the test should pass and able to create the developer kafka.

machi1990, line 314:
```suggestion
		if count >= int64(maxAllowedTrialInstances) {
```

This has to be `>=`

internal/kafka/internal/services/kafka_test.go
```diff
@@ -1176,6 +1176,66 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
...
+					WithArgs("us-east-1", "aws", "developer").
+					WithReply(converters.ConvertKafkaRequest(buildKafkaRequest(func(kafkaRequest *dbapi.KafkaRequest) {
+						kafkaRequest.ID = ""
+						kafkaRequest.InstanceType = types.DEVELOPER.String()
+						kafkaRequest.SizeId = "x1"
+						kafkaRequest.Owner = testUser
+						kafkaRequest.OrganisationId = "org-id"
+					})))
+				mocket.Catcher.NewMock().WithQuery(`INSERT INTO "kafka_requests"`)
+				mocket.Catcher.NewMock().WithQuery(`INSERT INTO "kafka_requests"`)
```
machi1990, line 1230:
```suggestion
```
This mock response isn't needed for this test case as an insert shouldn't be attempted.

it'll be needed for other test case as suggested in   https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1226#discussion_r937441694

internal/kafka/internal/config/kafka_quota.go
machi1990: No harm in keeping it.

I'd suggest we contain this PR in just making the max developer instance configurable and open a follow up JIRA to remove `AllowDeveloperInstance` flag.

machi1990 (changes requested):
Thanks for the PR @ajaypratap003 

I've added a few suggestions

internal/kafka/internal/config/kafka_quota.go
miguelsorianod: I agree, but in this PR itself the `MaxAllowedTrialInstance` attribute should be renamed to `MaxAllowedDeveloperInstances`, and its corresponding CLI flags too

ajaypratap003:
Thank you so much guys for providing your expert feedback and helping me to make this PR better. 
I've addressed all feedback. Please let me know in case I'm missing anything.

@machi1990 I tried to use your suggestion to mock SQL query in test but it doesn't work. I'm not able to mock count in SQL query. Please suggest me what is wrong in unit test

machi1990:
> Thank you so much guys for providing your expert feedback and helping me to make this PR better. I've addressed all feedback. Please let me know in case I'm missing anything.
> 
> @machi1990 I tried to use your suggestion to mock SQL query in test but it doesn't work. I'm not able to mock count in SQL query. Please suggest me what is wrong in unit test

the test setups is still missing a few mocks configuration. Due to that you are getting a Db exception. 

 
I’ll put comments below each setup to say what’s missing and we could sync up tomorrow to talk about them in detail.

internal/kafka/internal/services/kafka_test.go
```diff
@@ -1176,6 +1176,124 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
 				httpCode: http.StatusForbidden,
 			},
 		},
+		{
+			name: "should register kafka job successful when max-allowed-developer-instances is one",
```
machi1990, line 1180:
```suggestion
			name: "should register kafka job successful when developer instances count for the user is less than max-allowed-developer-instances",
```

internal/kafka/internal/services/kafka_test.go
```diff
@@ -1176,6 +1176,124 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
...
+				mocket.Catcher.NewMock().WithQuery(`INSERT INTO "kafka_requests"`)
+				mocket.Catcher.NewMock().WithQuery(``)
+				mocket.Catcher.NewMock().WithQueryException().WithExecException()
+			},
+			error: errorCheck{
+				wantErr: false,
+			},
+		},
+		{
+			name: "should register kafka job unsuccessful when max-allowed-developer-instances reached limit more than one",
```
machi1990, line 1245:
```suggestion
			name: "should register kafka job unsuccessful when developer kafka instances count for the user has reached the max-allowed-developer-instances limit",
```

internal/kafka/internal/services/kafka_test.go
```diff
@@ -1176,6 +1176,124 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
...
+					kafkaRequest.OrganisationId = "org-id"
+				}),
+			},
+			setupFn: func() {
+				mocket.Catcher.Reset()
+				totalCountResponse := []map[string]interface{}{{"count": 2}}
+
+				mocket.Catcher.NewMock().WithQuery(`SELECT count(1) FROM "kafka_requests" WHERE instance_type = $1 AND owner = $2 AND (organisation_id = $3) AND "kafka_requests"."deleted_at" IS NULL`).
+					WithArgs(types.DEVELOPER.String(), testUser, "org-id").
+					WithReply(totalCountResponse)
```
machi1990, line 1289:
```suggestion
				mocket.Catcher.Reset()
				mocket.Catcher.NewMock().
					WithQuery(`SELECT * FROM "kafka_requests" WHERE region = $1 AND cloud_provider = $2 AND instance_type = $3 AND "kafka_requests"."deleted_at" IS NULL`).
					WithArgs("us-east-1", "aws", "developer").
					WithReply(converters.ConvertKafkaRequest(buildKafkaRequest(func(kafkaRequest *dbapi.KafkaRequest) {
						kafkaRequest.ID = ""
						kafkaRequest.InstanceType = types.DEVELOPER.String()
						kafkaRequest.SizeId = "x1"
						kafkaRequest.Owner = testUser
						kafkaRequest.OrganisationId = "org-id"
					})))

				totalCountResponse := []map[string]interface{}{{"count": 2}}
				mocket.Catcher.NewMock().WithQuery(`SELECT count(1) FROM "kafka_requests" WHERE instance_type = $1 AND owner = $2 AND (organisation_id = $3) AND "kafka_requests"."deleted_at" IS NULL`).
					WithArgs(types.DEVELOPER.String(), testUser, "org-id").
					WithReply(totalCountResponse)
				mocket.Catcher.NewMock().WithQueryException().WithExecException()
```

internal/kafka/internal/services/kafka_test.go
```diff
@@ -1176,6 +1176,124 @@ func Test_kafkaService_RegisterKafkaJob(t *testing.T) {
...
+					WithReply(converters.ConvertKafkaRequest(buildKafkaRequest(func(kafkaRequest *dbapi.KafkaRequest) {
+						kafkaRequest.ID = ""
+						kafkaRequest.InstanceType = types.DEVELOPER.String()
+						kafkaRequest.SizeId = "x1"
+						kafkaRequest.Owner = testUser
+						kafkaRequest.OrganisationId = "org-id"
+					})))
+				mocket.Catcher.NewMock().WithQuery(`INSERT INTO "kafka_requests"`)
+				mocket.Catcher.NewMock().WithQuery(``)
+				mocket.Catcher.NewMock().WithQueryException().WithExecException()
```
machi1990, line 1238:
```suggestion
				mocket.Catcher.Reset()
				totalCountResponse := []map[string]interface{}{{"count": 0}}

				mocket.Catcher.NewMock().WithQuery(`SELECT count(1) FROM "kafka_requests" WHERE instance_type = $1 AND owner = $2 AND (organisation_id = $3) AND "kafka_requests"."deleted_at" IS NULL`).
					WithArgs(types.DEVELOPER.String(), testUser, "org-id").
					WithReply(totalCountResponse)

				mocket.Catcher.NewMock().
					WithQuery(`SELECT * FROM "kafka_requests" WHERE region = $1 AND cloud_provider = $2 AND instance_type = $3 AND "kafka_requests"."deleted_at" IS NULL`).
					WithArgs("us-east-1", "aws", "developer").
					WithReply(converters.ConvertKafkaRequest(buildKafkaRequest(func(kafkaRequest *dbapi.KafkaRequest) {
						kafkaRequest.ID = ""
						kafkaRequest.InstanceType = types.DEVELOPER.String()
						kafkaRequest.SizeId = "x1"
						kafkaRequest.Owner = testUser
						kafkaRequest.OrganisationId = "org-id"
					})))
				mocket.Catcher.NewMock().WithQuery(`INSERT INTO "kafka_requests"`)
				mocket.Catcher.NewMock().WithQueryException().WithExecException()
```

machi1990 (changes requested):
Left a few suggestions to improve and harden the test setup.

Once those are addressed, the change looks good to me. Thanks for the great job @ajaypratap003 

I'll let @miguelsorianod and @ziccardi have another look as well

internal/kafka/internal/services/kafka_test.go
machi1990: This removes as it is not needed 				
```go
mocket.Catcher.NewMock().WithQuery(``)
 ```

File: internal/kafka/internal/services/kafka_test.go (hunk 8 above)
Comment (machi1990) [reply internal/kafka/internal/services/kafka_test.go:1289]: This adds a mock for kafkas db query and adds the `mocket.Catcher.NewMock().WithQueryException().WithExecException()` to return an error for everything else that won't match

File: internal/kafka/internal/services/kafka_test.go (hunk 9 above)
Comment (akoserwal) [reply internal/kafka/internal/services/kafka_test.go:1238]: ```totalCountResponse := []map[string]interface{}{{"count": 0}}``` this will be totalCountResponse := []map[string]interface{}{{"count": **3**}}. As the test trying to evaluate the condition where the count is greater than the max allowed?

**Review by machi1990 — APPROVED** [review]

**Review by miguelsorianod — APPROVED** [review]
looks good to me :+1:

File: internal/kafka/internal/services/kafka_test.go (hunk 9 above)
Comment (akoserwal) [reply internal/kafka/internal/services/kafka_test.go]: make sense

File: templates/service-template.yml [hunk 10]
```diff
@@ -1082,6 +1087,7 @@ objects:
             - --redhat-sso-client-id-file=/secrets/service/redhatsso-service.clientId
             - --redhat-sso-client-secret-file=/secrets/service/redhatsso-service.clientSecret
             - --redhat-sso-base-url=${REDHAT_SSO_BASE_URL}
+            - --max-allowed-developer-instances=${MAX_ALLOWED_DEVELOPER_INSTANCES}
```
akoserwal, line 1090: Add the `MAS_ALLOWED_DEVELOPER_INSTANCES` in the make file

akoserwal:
A couple of follow-up task
- Adding the flag in app-interface
- Checking the impact on kas-installer

PR #1227, opened 2022-08-04
fix: fix pieces of code currently omitted by the linter

JasonPowr:
## Description
This PR is to Investigate and fix pieces of code currently omitted by the linter, i.e lines that have `//nolint`
Jira: https://issues.redhat.com/browse/MGDSTRM-900

- [ ] Code Review completed

### DIFF
--- a/pkg/db/context.go
+++ b/pkg/db/context.go
@@ -5,14 +5,9 @@ import (
 	"database/sql"
 	"fmt"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/constants"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
-)
-
-type contextKey int
-
-const (
-	transactionKey contextKey = iota
 )
 
 // NewContext returns a new context with transaction stored in it.
@@ -23,10 +18,8 @@ func (c *ConnectionFactory) NewContext(ctx context.Context) (context.Context, er
 		return ctx, err
 	}
 
-	// adding txid explicitly to context with a simple string key and int value
-	// due to a cyclical import cycle between pkg/db and pkg/logging
-	ctx = context.WithValue(ctx, "txid", tx.txid) //nolint
-	ctx = context.WithValue(ctx, transactionKey, tx)
+	ctx = context.WithValue(ctx, constants.TransactionIDkey, tx.txid)
+	ctx = context.WithValue(ctx, constants.TransactionKey, tx)
 
 	return ctx, nil
 }
@@ -38,7 +31,7 @@ func (c *ConnectionFactory) TxContext() (ctx context.Context, err error) {
 
 // Resolve resolves the current transaction according to the rollback flag.
 func Resolve(ctx context.Context) error {
-	tx, ok := ctx.Value(transactionKey).(*txFactory)
+	tx, ok := ctx.Value(constants.TransactionKey).(*txFactory)
 	if !ok {
 		return fmt.Errorf("Could not retrieve transaction from context")
 	}
@@ -67,7 +60,7 @@ func Resolve(ctx context.Context) error {
 }
 
 func Begin(ctx context.Context) error {
-	tx, ok := ctx.Value(transactionKey).(*txFactory)
+	tx, ok := ctx.Value(constants.TransactionKey).(*txFactory)
 	if !ok {
 		return fmt.Errorf("Could not retrieve transaction from context")
 	}
@@ -80,7 +73,7 @@ func Begin(ctx context.Context) error {
 }
 
 func AddPostCommitAction(ctx context.Context, f func()) error {
-	tx, ok := ctx.Value(transactionKey).(*txFactory)
+	tx, ok := ctx.Value(constants.TransactionKey).(*txFactory)
 	if !ok {
 		return fmt.Errorf("Could not retrieve transaction from context")
 	}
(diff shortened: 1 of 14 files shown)


machi1990:
Thanks @JasonPowr fantastic work. 

> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/common/kafka_pollers.go#L67 I’ve looked into this a bit and I believe I am closing the resps body correctly but I am still running into issues with the linter stating that it must be closed.

I think this is by design and it's up to the caller to Close the response. @ziccardi is this correct?

> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/auth/helper.go#L158 ParseRSAPrivateKeyFromPEMWithPassword() Is Deprecated however the linter is returning this error when it runs:
jwt.ParseRSAPrivateKeyFromPEMWithPassword is deprecated: This function is deprecated and should not be used anymore. It uses the deprecated x509.DecryptPEMBlock function, which was deprecated since RFC 1423 is regarded insecure by design. Unfortunately, there is no alternative in the Go standard library for now. See https://github.com/golang/go/issues/8860
So I'm not sure if there is another way around this one.

This one we need to wait for a drop in replacement implementation to be available or use external libs. 
Since it is only used for unit testing, I think we can continue to ignore the deprecated method for linting until we've a replacement.

> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L28 Wasn't sure how to fix this one.

For this one, the issue is the string "txid" which is used as key. We need to define a type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L12, the declare the "txid" as of that type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L15  and use it in place.

ziccardi:
> Thanks @JasonPowr fantastic work.
> 
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/test/common/kafka_pollers.go#L67 I’ve looked into this a bit and I believe I am closing the resps body correctly but I am still running into issues with the linter stating that it must be closed.
> 
> I think this is by design and it's up to the caller to Close the response. @ziccardi is this correct?
> 

Correct.

pawelpaszki (approved):
lgtm

JasonPowr:
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L28 Wasn't sure how to fix this one.
> 
> For this one, the issue is the string "txid" which is used as key. We need to define a type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L12, the declare the "txid" as of that type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L15 and use it in place.

So I tried to fix this like you said but it still does not seem to work. It fixes the linter Issue all right but seems to break everything else.

machi1990:
> > > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L28 Wasn't sure how to fix this one.
> > 
> > 
> > For this one, the issue is the string "txid" which is used as key. We need to define a type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L12, the declare the "txid" as of that type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L15 and use it in place.
> 
> So I tried to fix this like you said but it still does not seem to work. It fixes the linter Issue all right but seems to break everything else. @machi1990

@JasonPowr I think we need to change all the other references of `txid`. There is also a cyclic dependency on the `logger.go` so we'll need break this cyclic dependency. I can push a follow PR to this or if you are okay I can push a separate commit.

JasonPowr:
> > > > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L28 Wasn't sure how to fix this one.
> > > 
> > > 
> > > For this one, the issue is the string "txid" which is used as key. We need to define a type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L12, the declare the "txid" as of that type just like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/context.go#L15 and use it in place.
> > 
> > 
> > So I tried to fix this like you said but it still does not seem to work. It fixes the linter Issue all right but seems to break everything else. @machi1990
> 
> @JasonPowr I think we need to change all the other references of `txid`. There is also a cyclic dependency on the `logger.go` so we'll need break this cyclic dependency. I can push a follow PR to this or if you are okay I can push a separate commit.

Sure yeah work away thanks for the help :)

pkg/db/context.go
```diff
@@ -23,10 +18,8 @@ func (c *ConnectionFactory) NewContext(ctx context.Context) (context.Context, er
 		return ctx, err
 	}
 
-	// adding txid explicitly to context with a simple string key and int value
-	// due to a cyclical import cycle between pkg/db and pkg/logging
-	ctx = context.WithValue(ctx, "txid", tx.txid) //nolint
```
machi1990, line 28: This was fixed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1227/commits/de3c8de28fba0d5031cfc2e1cf8d701bf3df4ccd which introduced the https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/de3c8de28fba0d5031cfc2e1cf8d701bf3df4ccd/pkg/constants/db_context_keys.go to also fix the cyclic dependency issue

PR #1233, opened 2022-08-08
refactor(clusters_providers): add supported cluster providers types in the provider factory

machi1990:
## Description

This avoids the need to enlist supported clusters providers from the database all the time. The list
of provider types is known at compilation time, we can just return it wherever it is needed. 
1. The advantage of this approach is that it saves us a few query to the database and the roundtrip.
2. Another advantage is that we do not need to have at least a cluter from the given cluster provider
to have a functional cloud providers endpoints

The later point was discovered in various testing scenarios during local development

/cc @miguelsorianod 

## Verification Steps

Unit tests and integration tests should continue to pass

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/cloud_providers.go
+++ b/internal/kafka/internal/services/cloud_providers.go
@@ -9,7 +9,6 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/clusters/types"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/patrickmn/go-cache"
 )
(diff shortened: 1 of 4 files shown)


internal/kafka/internal/services/cloud_providers.go
```diff
@@ -203,21 +190,6 @@ func convertToCloudProviderRegions(cachedCloudProviderRegions interface{}) ([]ap
 	return nil, nil
 }
 
-func (p cloudProvidersService) getAvailableClusterProviderTypes() ([]Cluster, *errors.ServiceError) {
```
machi1990, line 206: this database call is not needed anymore

miguelsorianod:
This PR alters the behavior:

Before this PR if there were no clusters created of a given provider type then that method would not return that provider, even if it was set in the supported providers configuration file.

After this PR if there are no clusters created of a given provider type then that method will still return that provider if it is set in the supported providers configuration file.

Do we know if the previous behavior was intentional or accidental for some reason? UI side? others that were thought at the time?

If there is no need to keep the previous behavior for any reason I would be ok with the changes

machi1990:
> This PR alters the previous behavior:

Correct. 
 
> Before this PR if there were no clusters created of a given provider type then that method would not return that provider, even if it was set in the supported providers configuration file. The provider would also not be returned if there was one or more clusters created for a given provider type but all of them were in one of the `ClusterDeletionStatuses` (`ClusterCleanup` and `ClusterDeprovisioning` at the moment of writing this).
> 
> After this PR if there are no clusters created of a given provider type then that method will still return that provider if it is set in the supported providers configuration file.
> 
> Do we know if the previous behavior was intentional or accidental for some reason? 

The previous behavior was intentional and accidental at the same time. 
- It was intentional to avoid having to list cloud providers for standalone clusters if they are not configured i.e not present in the database. This is not an issue anymore as the standalone provider fetches cloud providers/regions from the database. 
- It is accidental since it ended up introducing a requirement of having a cluster of a given provider type to be able to list cloud providers using the corresponding cluster provider - this further meant tight coupling between having a cluster and being able to list cloud providers which are independent concepts. 

> UI side? others that were thought at the time?

It was added that way when standalone provider was added (I think I did add it).
The change was transparent to the UI as we do not have any standalone clusters in Stage/Prod.  

This might have some impact with the way the standalone mode works for kas-installer. I'll double check if at all kas-installer consumers enlist cloud providers, if they do, then this change introduce a requirement of having ocm credentials set.
If that's fine for kas-installer then a new version of it will be created and users of it be informed of the new requirement.  

> If there is no need to keep the previous behavior for any reason I would be ok with the changes

Thanks. I'll raise it to the team for discussion as well - to get their thoughts.

machi1990:
> This might have some impact with the way the standalone mode works for kas-installer

This might cause issue with kas-installer in standalone mode as OCM credentials are not given and we rely on the OCM mock flag https://github.com/bf2fc6cc711aee1a0c2a/kas-installer/blob/main/kas-fleet-manager/deploy-kas-fleet-manager.sh#L167
The underlying mock service does not extensively cover all the interaction we've with OCM and merging this PR might potentially cause `panic` errors.

I'll close this for now in favor of the current state.

PR #1235, opened 2022-08-09
feat: Expose GCP configuration

miguelsorianod:
## Description

Expose configurability of GCP credentials as a secret file.
Related to https://issues.redhat.com/browse/MGDSTRM-9132.

Information about GCP API credentials can be found in: https://cloud.google.com/iam/docs/creating-managing-service-account-keys

* For local development with a local process a file named `gcp.api-credentials` is expected in the secrets directory
* For development in an OSD cluster it is created as the `gcp.api-credentials` in the K8s secret named 'kas-fleet-manager' from the secrets template
  * Its value can be configured through the `GCP_API_CREDENTIALS` template parameter
    * The `GCP_API_CREDENTIALS` template parameter must be a base64 encoded string of the GCP API credentials configuration file contents (similarly to the image-pull.dockerconfigjson secret)
* The default location of the credentials file can be modified by using the newly added flag '--gcp-api-credentials-file' which by default it is set to 'secrets/gcp.api-credentials'

Additional remarks:
* Validation of the credentials file contents is only performed when `gcp` is listed as a supported cloud provider in the providers configuration file
* The credentials file is not mandatory when `gcp` is not a supported cloud provider
* If the credential file exists but it is empty or only contains blank it is treated in the same way as if it did not exist
* To perform validation at Go level a new library https://github.com/go-playground/validator has been used. I think this might be useful to perform validations at a general level

This change shouldn't cause any issue when merging it as `gcp` is not a supported provider at the moment in stage nor in production. To enable it in those environments it will need to be added as a supported provider and the corresponding secrets configured accordingly with the secrets mechanism used in stage and production.

## Verification Steps

* Verify the different scenarios and combinations between having the GCP credentails file and GCP as a supported provider.
* Verify that KAS Fleet Manager runs successfully both as a local process and in an OSD dev cluster, with and without GCP credentials and with and without gcp as a supported cloud provider.

Credentials can be obtained from Vault. Ask them to me for directions on how to get them if needed.

- [ ] Code Review completed


Makefile
```diff
@@ -741,7 +742,7 @@ deploy/secrets:
 		-p KAFKA_TLS_CERT="$(shell ([ -s './secrets/kafka-tls.crt' ] && [ -z '${KAFKA_TLS_CERT}' ]) && cat ./secrets/kafka-tls.crt || echo '${KAFKA_TLS_CERT}')" \
 		-p KAFKA_TLS_KEY="$(shell ([ -s './secrets/kafka-tls.key' ] && [ -z '${KAFKA_TLS_KEY}' ]) && cat ./secrets/kafka-tls.key || echo '${KAFKA_TLS_KEY}')" \
 		-p OBSERVABILITY_CONFIG_ACCESS_TOKEN="$(shell ([ -s './secrets/observability-config-access.token' ] && [ -z '${OBSERVABILITY_CONFIG_ACCESS_TOKEN}' ]) && cat ./secrets/observability-config-access.token || echo '${OBSERVABILITY_CONFIG_ACCESS_TOKEN}')" \
-		-p IMAGE_PULL_DOCKER_CONFIG="$(shell ([ -s './secrets/image-pull.dockerconfigjson' ] && [ -z '${IMAGE_PULL_DOCKER_CONFIG}' ]) && cat ./secrets/image-pull.dockerconfigjson | base64 || echo '${IMAGE_PULL_DOCKER_CONFIG}')" \
+		-p IMAGE_PULL_DOCKER_CONFIG="$(shell ([ -s './secrets/image-pull.dockerconfigjson' ] && [ -z '${IMAGE_PULL_DOCKER_CONFIG}' ]) && cat ./secrets/image-pull.dockerconfigjson | base64 -w 0 || echo '${IMAGE_PULL_DOCKER_CONFIG}')" \
```
miguelsorianod, line 745: I added a new commit to fix an issue where the image pull docker config was not correctly being passed making the makefile target crash when the config was in a file

internal/kafka/internal/config/gcp.go
```diff
@@ -0,0 +1,80 @@
...
+
+	cloudProviderNameGCP = "gcp"
+)
+
+type GCPConfig struct {
+	GCPCredentials         GCPCredentials
+	gcpCredentialsFilePath string
+}
+
+type GCPCredentials struct {
```
miguelsorianod, line 24: For now we make all attributes of the credentials required. We might relax this in the future if we find out some of them are not needed

pb82: :+1:

pb82: any other validations that would make sense? E.g. correct format of `ClientEmail` or `TokenURI`?

Makefile
machi1990: :+1:

machi1990 (approved):
lgtm

internal/kafka/internal/config/gcp.go
miguelsorianod: I don't think so. If those are incorrect the provisioning will fail. Let's let OCM/GCP (the source of truth) deal with it.

templates/service-template.yml
```diff
@@ -967,6 +967,7 @@ objects:
             - --aws-secret-access-key-file=/secrets/service/aws.secretaccesskey
             - --aws-route53-access-key-file=/secrets/service/aws.route53accesskey
             - --aws-route53-secret-access-key-file=/secrets/service/aws.route53secretaccesskey
+            - --gcp-api-credentials-file=/secrets/service/gcp.api-credentials
```
machi1990, line 970: when adding the credentials to vault for stage/prod. The key has to be `gcp.api-credentials`

pb82 (approved):

PR #1237, opened 2022-08-09
feat: expose GCP compute instance type configuration

miguelsorianod:
## Description

Allow configurability of compute instance type for the data plane clusters that are created in GCP.
Related to https://issues.redhat.com/browse/MGDSTRM-9362.

This PR is about introducing the configurability.
There will be other PRs about making use of it.

Merging this PR should be safe because although the flag is introduced it is not being used at the moment.
This will also require template parameter changes in stage and production before enabling GCP functionality.

## Verification Steps

Verify that kas fleet manager runs
Verify that the new added flag appears as part of kas fleet manager binary cli flags. Verify that if you change the value the data plane cluster config changes.
Verify that fleet manager is able to run from template

- [ ] Code Review completed


miguelsorianod:
As a consequence of this changes, I would like to rename the existing `cluster-compute-machine-type` CLI flag to `aws-cluster-compute-machine-type` as the name has become very generic.

An approach for rolling that would be:
1. Introduce a new flag `aws-cluster-compute-machine-type` and its corresponding template parameter
2. Deploy the changes in stage and production
3. Add the the template parameter value for stage and production
4. Change the code to use the new flag value
5. Deploy the changes in stage and production
6. Remove the old template parameter value for stage and production
7. Remove the old `cluster-compute-machine-type` cli flag from kfm code
8. Deploy the changes in stage and production

internal/kafka/internal/config/dataplane_cluster_config.go
```diff
@@ -273,7 +280,8 @@ func (c *DataplaneClusterConfig) IsReadyDataPlaneClustersReconcileEnabled() bool
 
 func (c *DataplaneClusterConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.OpenshiftVersion, "cluster-openshift-version", c.OpenshiftVersion, "The version of openshift installed on the cluster. An empty string indicates that the latest stable version should be used")
-	fs.StringVar(&c.ComputeMachineType, "cluster-compute-machine-type", c.ComputeMachineType, "The compute machine type")
+	fs.StringVar(&c.ComputeMachineType, "cluster-compute-machine-type", c.ComputeMachineType, "The instance type of the AWS compute instances for Data Planes created in AWS")
```
machi1990, line 283:
```suggestion
	fs.StringVar(&c.ComputeMachineType, "aws-cluster-compute-machine-type", c.ComputeMachineType, "The instance type of the AWS compute instances for Data Planes created in AWS")
```

templates/service-template.yml
```diff
@@ -417,10 +417,15 @@ parameters:
   value: ""
 
 - name: CLUSTER_COMPUTE_MACHINE_TYPE
```
machi1990, line 419:
```suggestion
- name: AWS_CLUSTER_COMPUTE_MACHINE_TYPE
```

templates/service-template.yml
```diff
@@ -1054,6 +1059,7 @@ objects:
             - --max-allowed-instances=${MAX_ALLOWED_INSTANCES}
             - --cluster-openshift-version=${CLUSTER_OPENSHIFT_VERSION}
             - --cluster-compute-machine-type=${CLUSTER_COMPUTE_MACHINE_TYPE}
```
machi1990, line 1061:
```suggestion
            - --aws-cluster-compute-machine-type=${AWS_CLUSTER_COMPUTE_MACHINE_TYPE}
```

machi1990 (commented):
> This PR is about introducing the configurability.
> There will be other PRs about making use of it.

Sounds good. Just mentioning, when we start using this field, we'll also have to update to consider the cloud provider 

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d3c7da600c4756cd83a37199080151a657059ec3/internal/kafka/internal/workers/clusters_mgr.go#L1103

Regarding https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1237#issuecomment-1209487780

I am +1 on the renaming of the CLI flag to prefix it with `aws-`
I left some suggestions on updating the change right away as the old flag is not used at the moment and we can safely rename the parameter and the CLI flag. 
We could as well update AppInterface along this change tomorrow. 

If you are okay with the suggestion, we'll need to update the environments files which also contain some defaults e.g https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d3c7da600c4756cd83a37199080151a657059ec3/internal/kafka/internal/environments/development.go#L25

miguelsorianod:
> > This PR is about introducing the configurability.
> > There will be other PRs about making use of it.
> 
> Sounds good. Just mentioning, when we start using this field, we'll also have to update to consider the cloud provider
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d3c7da600c4756cd83a37199080151a657059ec3/internal/kafka/internal/workers/clusters_mgr.go#L1103
> 
> Regarding [#1237 (comment)](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1237#issuecomment-1209487780)
> 
> I am +1 on the renaming of the CLI flag to prefix it with `aws-` I left some suggestions on updating the change right away as the old flag is not used at the moment and we can safely rename the parameter and the CLI flag. We could as well update AppInterface along this change tomorrow.
> 
> If you are okay with the suggestion, we'll need to update the environments files which also contain some defaults e.g
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d3c7da600c4756cd83a37199080151a657059ec3/internal/kafka/internal/environments/development.go#L25

:+1:. I will perform the renaming of the variable in another PR. Let's focus this PR on the gcp related change.

miguelsorianod:
For the new `--gcp-cluster-compute-machine-type` flag in this PR I am not going to set any value in the {development,integration,production,stage,testing}.go files. I don't see any value on adding them with the same value as the hardcoded one.
I've observed that in the case of cluster-compute-machine-type there is an exception where in the `integration` environment the instance type is a different one than the default. However, I don't see why we would need a different instance type in that environment, at least for GCP so I prefer to keep it simple and use the same.

machi1990:
> For the new `--gcp-cluster-compute-machine-type` flag in this PR I am not going to set any value in the {development,integration,production,stage,testing}.go files. I don't see any value on adding them with the same value as the hardcoded default one.
> I've observed that in the case of `--cluster-compute-machine-type` there is an exception where in the `integration` environment the instance type is a different one than the hardcoded default one. However, I don't see why we would need a different instance type in that environment, at least for GCP, so I prefer to keep it simple and use the same.

+1

templates/service-template.yml
miguelsorianod: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1237#issuecomment-1209607169

templates/service-template.yml
miguelsorianod: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1237#issuecomment-1209607169

internal/kafka/internal/config/dataplane_cluster_config.go
miguelsorianod: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1237#issuecomment-1209607169

machi1990 (approved):
lgtm

PR #1238, opened 2022-08-09
feat: add GetQuotaCost method to the cluster provider

JameelB:
## Description
Add GetQuotaCost() to the cluster provider interface which will be responsible for getting all quota information that belongs to the authenticated user's organisation. This will include information on the max allowed and quotas consumed by the provider's resources.

This is added to the providers interface as we need to get this information and expose them as metrics in the cluster manager. It is best practice to avoid using the ocm client directly in the workers to avoid coupling worker code to any specific cluster provider. 

The provider to be used will be determined by the credentials provided by the user. If the ocm credentials were specified, use ocm provider. Otherwise, use the standalone provider (this will be added in a follow on PR).

## Verification Steps
1. Ensure all tests passing

- [ ] Code Review completed


machi1990 (approved):
The code change lgtm

JameelB:
Thanks for the review @machi1990, I'll add some tests now. Will ping you again for another round of review :)

internal/kafka/internal/clusters/standalone_provider.go
```diff
@@ -558,3 +558,8 @@ func (s *StandaloneProvider) CreateMachinePool(request *types.MachinePoolRequest
 	// TODO implement
 	return nil, nil
 }
+
+// noop method, it will always return an empty list as a standalone provider does not have any resource quotas
+func (s *StandaloneProvider) GetQuotaCost() ([]types.QuotaCost, error) {
+	return nil, nil
```
machi1990, line 564: Can we return an empty slice instead of nil?

machi1990 (approved):
lgtm

left one suggestion about returning an empty slice instead of `nil`.

internal/kafka/internal/clusters/standalone_provider.go
JameelB: I think it's fine to return nil here? i think it's preferred in golang (see [here](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)). I believe they behave the same way, the built in functions for slices are still valid.

JameelB: We could do this if it makes it clearer?

```suggestion
        var quotaCostList []types.QuotaCost
	return quotaCostList, nil
```

miguelsorianod: The description of the method indicates that an empty list is always returned. The description should match the behavior, whatever is the decided outcome.
Isn't the name of the method also a little bit confusing? It is named `GetQuotaCost`, singular, but a list of QuotaCost is returned. Shouldn't it be named `GetQuotaCosts` then?

machi1990: > I think it's fine to return nil here? i think it's preferred in golang (see [here](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)). I believe they behave the same way, the built in functions for slices are still valid.

Thanks.

JameelB: `GetQuotaCosts` sounds better, thanks @miguelsorianod. I'll update the name to that.

JameelB: updated it there, please have another look @machi1990 @miguelsorianod

machi1990 (approved):
lgtm

Thanks @JameelB

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -441,3 +441,30 @@ func (o *OCMProvider) CreateMachinePool(request *types.MachinePoolRequest) (*typ
...
 }
+
+// GetQuotaCosts returns a list of ocm resource quota cost information for the authenticated user
+func (o *OCMProvider) GetQuotaCosts() ([]types.QuotaCost, error) {
+	var quotaCostList []types.QuotaCost
+	account, err := o.ocmClient.GetCurrentAccount()
+	if err != nil {
+		return quotaCostList, err
+	}
+	orgID := account.Organization().ID()
```
miguelsorianod, line 452: As a note:
OCM types usually provide two types of getters for the attributes of an object:
* A getter with the name of the attribute
* A getter Get<AttributeName>

For example, for organization of an AMS account you have the methods:
* Organization()
* GetOrganization()

The second type of getter is useful because it allows you to check whether the attribute has been set or not, and also allows differentiating between the zero value of the type and whether it's been set or not for the cases where the zero value is an accepted value.

internal/kafka/internal/clusters/standalone_provider.go
```diff
@@ -558,3 +558,9 @@ func (s *StandaloneProvider) CreateMachinePool(request *types.MachinePoolRequest
 	// TODO implement
 	return nil, nil
 }
+
+// noop method, it will always return an empty list as a standalone provider does not have any resource quotas
```
miguelsorianod, line 562: The description should be updated to match what's being returned

internal/kafka/internal/clusters/ocm_provider.go
machi1990: Good to know. I was always wondering what's the difference between the two.

JameelB: I changed it to use GetId() instead so we wont have to check the value of the id :+1:

internal/kafka/internal/clusters/standalone_provider.go
JameelB: updated it there :)

miguelsorianod (approved):

PR #1241, opened 2022-08-10
feat: add gcp example

pb82:
## Description

Adds GCP to the supported cloud providers example. Validation is already present and no further changes were required.

## Verification Steps

1. Add an entry to `dataplane-clusters-configuration.yaml` like
```
  - name: anyname
    cluster_id: dummyid
    cloud_provider: gcp
    region: us-east1
    multi_az: true
    schedulable: true
    kafka_instance_limit: 2
    status: "ready"
    provider_type: "ocm"
    supported_instance_type: "developer,standard"
```
2. Create a file named `gcp.api-credentials` under secrets:
```
{
  "type": "service_account",
  "project_id": "dummy",
  "private_key_id": "dummy",
  "private_key": "dummy",
  "client_email": "dummy@gmail.com",
  "client_id": "dummy",
  "auth_uri": "dummy",
  "token_uri": "dummy",
  "auth_provider_x509_cert_url": "dummy",
  "client_x509_cert_url": "dummy"
}
```
3. Run the fleet manager and query the [/api/kafkas_mgmt/v1/cloud_providers/{id}/regions](http://localhost:8082/?urls.primaryName=Public%20API#/default/getCloudProviderRegions) endpoint with `id` set to `gcp`.
4. Make sure a result is returned for the `us-east1` region.

### DIFF
--- a/cmd/kas-fleet-manager/main_test.go
+++ b/cmd/kas-fleet-manager/main_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka"
@@ -19,7 +20,9 @@ func TestInjections(t *testing.T) {
 	)
 	g.Expect(err).To(gomega.BeNil())
 	err = env.CreateServices()
-	g.Expect(err).To(gomega.BeNil())
+	if err != nil && !strings.Contains(err.Error(), "error validating GCP API credentials") {
+		t.Error(err)
+	}
 
 	var bootList []environments.BootService
 	env.MustResolve(&bootList)
--- a/config/provider-configuration.yaml
+++ b/config/provider-configuration.yaml
@@ -44,3 +44,11 @@ supported_providers:
         supported_instance_type:
           standard: {}
           developer: {}
+  - name: gcp
+    default: false
+    regions:
+      - name: us-east1
+        default: true
+        supported_instance_type:
+          standard: {}
+          developer: {}
--- a/internal/kafka/test/helper.go
+++ b/internal/kafka/test/helper.go
@@ -60,12 +60,32 @@ func NewKafkaHelper(t *testing.T, server *httptest.Server) (*test.Helper, *publi
 
 func NewKafkaHelperWithHooks(t *testing.T, server *httptest.Server, configurationHook interface{}) (*test.Helper, *public.APIClient, func()) {
 	h, teardown := test.NewHelperWithHooks(t, server, configurationHook, kafka.ConfigProviders(), di.ProvideValue(environments.BeforeCreateServicesHook{
-		Func: func(dataplaneClusterConfig *config.DataplaneClusterConfig, kafkaConfig *config.KafkaConfig, observabilityConfiguration *observatorium.ObservabilityConfiguration, kasFleetshardConfig *config.KasFleetshardConfig) {
+		Func: func(dataplaneClusterConfig *config.DataplaneClusterConfig, kafkaConfig *config.KafkaConfig, observabilityConfiguration *observatorium.ObservabilityConfiguration, kasFleetshardConfig *config.KasFleetshardConfig, providerConfig *config.ProviderConfig) {
 			kafkaConfig.KafkaLifespan.EnableDeletionOfExpiredKafka = true
 			observabilityConfiguration.EnableMock = true
 			dataplaneClusterConfig.DataPlaneClusterScalingType = config.NoScaling // disable scaling by default as it will be activated in specific tests
 			dataplaneClusterConfig.RawKubernetesConfig = nil                      // disable applying resources for standalone clusters
 			dataplaneClusterConfig.EnableDynamicScaleUpManagerScaleUpTrigger = false
+
+			// enable only aws for integration tests
+			providerConfig.ProvidersConfig = config.ProviderConfiguration{
+				SupportedProviders: config.ProviderList{
+					{
+						Name:    "aws",
+						Default: true,
+						Regions: []config.Region{
+							{
+								Name:    "us-east-1",
+								Default: true,
+								SupportedInstanceTypes: map[string]config.InstanceTypeConfig{
+									"standard":  {},
+									"developer": {},
+								},
+							},
+						},
+					},
+				},
+			}
 		},
 	}))
 	if err := h.Env.ServiceContainer.Resolve(&TestServices); err != nil {


miguelsorianod:
By enabling this this would try to create a GCP cluster automatically. I think we should update our integration tests suite so the only supported provider is AWS at the moment to avoid triggering an automatic gcp scale up.

cc @machi1990

machi1990:
> By enabling this this would try to create a GCP cluster automatically when dynamic scaling is enabled. I think we should update our integration tests suite so the only supported provider there is AWS at the moment to avoid triggering an automatic GCP scale up.
> 
> cc @machi1990

The test already limits the providers to only `aws`: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/98a582ee9344d235adf852377e7a4151a3ff443e/internal/kafka/test/integration/cluster_mgr_test.go#L37-L56

machi1990:
I think only adding the `config` without the secret is causing issues in the CI and even local integrations tests. 

Maybe we need to add a make target to setup the secret e.g `make gcp-credentials/setup` to dummy values which we can call locally and also in the CI.

miguelsorianod:
> > By enabling this this would try to create a GCP cluster automatically when dynamic scaling is enabled. I think we should update our integration tests suite so the only supported provider there is AWS at the moment to avoid triggering an automatic GCP scale up.
> > cc @machi1990
> 
> The test already limits the providers to only `aws`:
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/98a582ee9344d235adf852377e7a4151a3ff443e/internal/kafka/test/integration/cluster_mgr_test.go#L37-L56

Even though that's correct, there is still the risk that someone else would enable autoscale and not specify the cloud providers in a new test. That means the default file would be read and thus a cluster in GCP would be autoscaled too.

I think we should update the PR so integration tests only have AWS as the provider by default. Then if we want a test that involves GCP the test would specify the providers it desires.

machi1990:
> I think we should update the PR so integration tests only have AWS as the provider by default. Then if we want a test that involves GCP the test would specify the providers it desires.

+1 on this. I also think this will remove the need for the suggestion in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1241#issuecomment-1210535566

internal/kafka/providers.go
```diff
@@ -22,6 +22,20 @@ import (
 	"github.com/goava/di"
 )
 
+func DataPlaneConfigProviders() di.Option {
+	// do not inject gcp config when running integration test to avoid accidental cluster creation
+	if environments2.GetEnvironmentStrFromEnv() == environments2.IntegrationEnv {
```
miguelsorianod, line 27: This is not correct.
The issue does not only happen when running integration environment but also when running integration tests with the development environment set. This is common when developing locally and wanting to run the integration tests against real OCM, and also on our nightly tests.

I also think that the enablement of this should probably be done at the test suite level and not in the DI framework side, as in some tests we might want to have the GCP configuration provider set

pb82: Makes sense. I'm thinking to remove the gcp provider in `NewKafkaHelperWithHooks` which is used in the integration tests.
The Injection test is a special case though and I'll have to find a different fix for that.

internal/kafka/test/helper.go
```diff
@@ -60,12 +60,26 @@ func NewKafkaHelper(t *testing.T, server *httptest.Server) (*test.Helper, *publi
...
 	h, teardown := test.NewHelperWithHooks(t, server, configurationHook, kafka.ConfigProviders(), di.ProvideValue(environments.BeforeCreateServicesHook{
-		Func: func(dataplaneClusterConfig *config.DataplaneClusterConfig, kafkaConfig *config.KafkaConfig, observabilityConfiguration *observatorium.ObservabilityConfiguration, kasFleetshardConfig *config.KasFleetshardConfig) {
+		Func: func(dataplaneClusterConfig *config.DataplaneClusterConfig, kafkaConfig *config.KafkaConfig, observabilityConfiguration *observatorium.ObservabilityConfiguration, kasFleetshardConfig *config.KasFleetshardConfig, providerConfig *config.ProviderConfig) {
 			kafkaConfig.KafkaLifespan.EnableDeletionOfExpiredKafka = true
 			observabilityConfiguration.EnableMock = true
 			dataplaneClusterConfig.DataPlaneClusterScalingType = config.NoScaling // disable scaling by default as it will be activated in specific tests
 			dataplaneClusterConfig.RawKubernetesConfig = nil                      // disable applying resources for standalone clusters
 			dataplaneClusterConfig.EnableDynamicScaleUpManagerScaleUpTrigger = false
+
+			// remove the gcp provider to avoid integration tests creating instances
```
miguelsorianod, line 70: Maybe instead of removing the gcp provider we could set completely the providers value totally ignoring the originally read value from the file? In that case we would have full control of the providers we want by default, totally ignoring the provider configuration file for our integration tests.
cc @machi1990 to have another view around this.

Additionally, did you verify this indeeds changes the original value and not a copy?

cmd/kas-fleet-manager/main_test.go
```diff
@@ -19,7 +20,9 @@ func TestInjections(t *testing.T) {
 	)
 	g.Expect(err).To(gomega.BeNil())
 	err = env.CreateServices()
-	g.Expect(err).To(gomega.BeNil())
+	if err != nil && !strings.Contains(err.Error(), "error validating GCP API credentials") {
```
miguelsorianod, line 23: Maybe we could play with `ProvideValue` instead to provide a hardcoded value for the providers in this test. Could you explore that potential approach?

config/provider-configuration.yaml
```diff
@@ -44,3 +44,11 @@ supported_providers:
         supported_instance_type:
           standard: {}
           developer: {}
+  - name: gcp
```
miguelsorianod, line 47: I am thinking about what would be our preferred defaults regarding the provider configuration and dynamic scaling functionality including both AWS and GCP for our development environment.

With the proposed configuration (and also with already pre-existing configuration with AWS) when dynamic scaling is enabled 4 new clusters would be created: one for each supported instance type in each supported cloud provider. This could easily create undesired clusters when people are developing without realizing about it.

Maybe we should by default limit the scaling of clusters in the development configuration?
Two options I can think of if we were to decide that:
1. Set a limit of 0 for all instance types for all providers in the provider configuration
2. Set the `EnableDynamicScaleUpManagerScaleUpTrigger` and its corresponding scaledown option to false by default in development environment.

I prefer the first option as it would be more explicit and the.
In this case, if someone wants to really trigger scale ups would need to modify the limit for the desired instance type in the configuration file.

cc @machi1990 @pb82

machi1990: adding @JameelB @JasonPowr  @VanillaSpoon to the discussion as well

machi1990: I am thinking maybe we could keep the scaling mode to `manual` for development as the default.
And have section in the documentation explaining how to switch to `auto` mode: this will contain a  `NOTE` on the implication of having `auto` mode etc

internal/kafka/test/helper.go
machi1990: > Maybe instead of removing the gcp provider we could set completely the providers value totally ignoring the originally read value from the file? In that case we would have full control of the providers we want by default, totally ignoring the provider configuration file for our integration tests.

+1. That way we won't have to ask ourselves the same questions when we support other cloud providers

config/provider-configuration.yaml
pb82: @machi1990 do you mean when `OCM_ENV` is set to development? We can do that, but people could still set the env to a different value when running locally.

machi1990: > @machi1990 do you mean when `OCM_ENV` is set to development? 

Yes. The scaling type is already set to `manual` as the default. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/dc3eb0e33955fa8329b44dd58bdbc6bfe4599ef9/internal/kafka/internal/environments/development.go#L29

> We can do that, but people could still set the env to a different value when running locally.

For other environments, if scaling type is not specified manual will be the default scaling type. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/dc3eb0e33955fa8329b44dd58bdbc6bfe4599ef9/internal/kafka/internal/config/dataplane_cluster_config.go#L90

Essentially, we are saying that by default we won't turn autoscaling on. When someone turns it on, they understand the impact (scaling up/down, taints addition etc)

pb82: Ok, that sounds reasonable to me. Do we need documentation for turning on dynamic scaling for development and what the implications are?

machi1990: We already have docs here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/docs/architecture/data-plane-osd-cluster-dynamic-scaling.md

Which is already linked from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/docs/data-plane-osd-cluster-options.md#configuring-osd-cluster-creation-and-autoscaling 

We could add a big `NOTE` there that someone reads the architectural proposal to understand the implication of turning dynamic scaling on.

cmd/kas-fleet-manager/main_test.go
pb82: I've experimented with the dependency injection and `ProvideValue` a bit, but I don't think it is useful in this case. `ProvideValue` can be used to provide an additional instance of the GCP config with pre-populated values that pass validation.

However, this only provides an additional instance, validation will still fail on the original GCP config that was provided in `ConfigProviders`. Using tags to exclude the GCP config only from this test doesn't seem to be practical because `goava` doesn't allow you to have 'not' conditions on tags. So you can't express something like `always provide this, except if a tag has a certain value`. At least I didn't find a way to do that.

So conclusion: checking the error string seems to be the most practical option.

internal/kafka/test/helper.go
```diff
@@ -60,12 +60,18 @@ func NewKafkaHelper(t *testing.T, server *httptest.Server) (*test.Helper, *publi
...
+		Func: func(dataplaneClusterConfig *config.DataplaneClusterConfig, kafkaConfig *config.KafkaConfig, observabilityConfiguration *observatorium.ObservabilityConfiguration, kasFleetshardConfig *config.KasFleetshardConfig, providerConfig *config.ProviderConfig) {
 			kafkaConfig.KafkaLifespan.EnableDeletionOfExpiredKafka = true
 			observabilityConfiguration.EnableMock = true
 			dataplaneClusterConfig.DataPlaneClusterScalingType = config.NoScaling // disable scaling by default as it will be activated in specific tests
 			dataplaneClusterConfig.RawKubernetesConfig = nil                      // disable applying resources for standalone clusters
 			dataplaneClusterConfig.EnableDynamicScaleUpManagerScaleUpTrigger = false
+
+			// enable only aws for integration tests
+			awsProvider, _ := providerConfig.ProvidersConfig.SupportedProviders.GetByName("aws")
+			providerConfig.ProvidersConfig = config.ProviderConfiguration{
```
miguelsorianod, line 72: Did we verify if this change is propagated to the tests effectively when they run??

cmd/kas-fleet-manager/main_test.go
miguelsorianod: :+1: I see. it's unfortunate that the DI framework does not provide that.
The test is quite fragile with this change (if the error string value changes the test will break, etc...) but for now it will work to workaround the DI framework / mechanism we have on top of it.

internal/kafka/test/helper.go
pb82: Yes, modification of the providers config here is passed to the tests. I've verified it by providing an empty supported providers list which causes the tests to fail with:

`Unable to initialize testing environment: expected 1 default provider in provider list, got 0`

miguelsorianod: :+1:

internal/kafka/test/helper.go
miguelsorianod: I see this has been updated to include aws but it is still reading the value from the configuration file instead of hardcoding the values. The best is probably to completely hardcode the provider information

pb82: This is done now, the aws provider is hardcoded.

miguelsorianod (approved):
:+1:

PR #1251, opened 2022-08-16
feat: add cluster_provider_resource_quota_max_allowed metric

JameelB:
## Description
Adds and exposes `kas_fleet_manager_cluster_provider_resource_quota_max_allowed` metric in the KAS Fleet Manager /metric endpoint. This provides information on the maximum allowed number of resources that can be consumed by the provided cluster provider user (i.e. ocm user). 

We only need to expose metrics for `ocm` cluster provider. OCM quota metrics will be set when the user provides ocm client credentials to the kfm service, either ocm clientid+secret or the token.

Standalone providers doesn't have any resource quota limits so no quota metrics will be exposed. 

Example metric: 
```
kas_fleet_manager_cluster_provider_resource_quota_max_allowed{quota_id="<quota-id>", provider="ocm"} 100
```

## Verification Steps
- [ ] Ensure metric is exposed correctly
    - Run kas fleet manager with ocm credentials specified
    - Curl http://localhost:8080/metrics
    - Ensure the quota metric as defined above are included in the response with correct values
- [ ] All CI checks passing

- [ ] Code Review completed
- [ ] ~~All PR comments are resolved either by addressing them or creating follow up tasks~~

### DIFF
--- a/internal/kafka/internal/clusters/ocm_provider.go
+++ b/internal/kafka/internal/clusters/ocm_provider.go
@@ -26,6 +26,14 @@ const (
 	ocmMultiAZClusterNodeScalingMultiple = 3
 )
 
+var (
+	// AMS quota filters
+	AMSQuotaAddonResourceNamePrefix        = "addon"
+	AMSQuotaOSDProductName          string = "OSD"
+	AMSQuotaComputeNodeResourceType string = "compute.node"
+	AMSQuotaClusterResourceType     string = "cluster"
+)
+
 type OCMProvider struct {
 	ocmClient      ocm.Client
 	clusterBuilder ClusterBuilder
@@ -442,9 +450,11 @@ func (o *OCMProvider) CreateMachinePool(request *types.MachinePoolRequest) (*typ
 	return request, err
 }
 
-// GetQuotaCosts returns a list of ocm resource quota cost information for the authenticated user
-// Returns a nil slice when no resource quota is assigned to the user
-func (o *OCMProvider) GetQuotaCosts() ([]types.QuotaCost, error) {
+// GetClusterResourceQuotaCosts returns a list of quota cost information related to ocm resources used for the provisioning and
+// terraforming of data plane clusters for the authenticated user.
+//
+// Returns a nil slice when no ocm resource quota is assigned to the user
+func (o *OCMProvider) GetClusterResourceQuotaCosts() ([]types.QuotaCost, error) {
 	var quotaCostList []types.QuotaCost
 
 	account, err := o.ocmClient.GetCurrentAccount()
@@ -456,12 +466,36 @@ func (o *OCMProvider) GetQuotaCosts() ([]types.QuotaCost, error) {
 		return quotaCostList, errors.New("failed to get quota cost: organisation id for the current authenticated user can't be found")
 	}
 
-	ocmQuotaCostList, err := o.ocmClient.GetQuotaCosts(orgID, false, false)
+	strimziOperatorAddonName := fmt.Sprintf("%s-%s", AMSQuotaAddonResourceNamePrefix, o.ocmConfig.StrimziOperatorAddonID)
+	fleetshardOperatorAddonName := fmt.Sprintf("%s-%s", AMSQuotaAddonResourceNamePrefix, o.ocmConfig.KasFleetshardAddonID)
+
+	filters := []ocm.QuotaCostRelatedResourceFilter{
+		// Filter for Strimzi operator add-on
+		{
+			ResourceName: &strimziOperatorAddonName,
+		},
+		// Filter for Fleetshard operator add-on
+		{
+			ResourceName: &fleetshardOperatorAddonName,
+		},
+		// Filter for osd compute nodes
+		{
+			ResourceType: &AMSQuotaComputeNodeResourceType,
+			Product:      &AMSQuotaOSDProductName,
+		},
+		// Filter for osd cluster
+		{
+			ResourceType: &AMSQuotaClusterResourceType,
+			Product:      &AMSQuotaOSDProductName,
+		},
+	}
+
+	ocmQuotaCostList, err := o.ocmClient.GetQuotaCosts(orgID, true, false, filters...)
 	if err != nil {
 		return quotaCostList, err
 	}
 
-	for _, qc := range ocmQuotaCostList.Slice() {
+	for _, qc := range ocmQuotaCostList {
 		quotaCostList = append(quotaCostList, types.QuotaCost{
 			ID:         qc.QuotaID(),
 			MaxAllowed: qc.Allowed(),
--- a/internal/kafka/test/integration/cluster_metrics_test.go
+++ b/internal/kafka/test/integration/cluster_metrics_test.go
@@ -1,10 +1,12 @@
 package integration
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/clusters"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/kafkas/types"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test/common"
@@ -12,10 +14,13 @@ import (
 	clusterMocks "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test/mocks/clusters"
 	kafkaMocks "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test/mocks/kafkas"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/ocm"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/metrics"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/workers"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/test/mocks"
 	"github.com/onsi/gomega"
+
+	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 )
 
 func TestClusterCapacityUsedMetric(t *testing.T) {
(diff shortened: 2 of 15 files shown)


internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -189,19 +189,20 @@ func (c *ClusterManager) Reconcile() []error {
 }
 
 func (c *ClusterManager) processMetrics() []error {
+	var errs []error
 	if err := c.setClusterStatusCountMetrics(); err != nil {
-		return []error{errors.Wrapf(err, "failed to set cluster status count metrics")}
+		errs = append(errs, errors.Wrapf(err, "failed to set cluster status count metrics"))
```
machi1990, line 194: :+1:

internal/kafka/test/integration/cluster_metrics_test.go
```diff
@@ -77,3 +79,31 @@ func TestClusterCapacityUsedMetric(t *testing.T) {
...
+
+	h, _, teardown := test.NewKafkaHelperWithHooks(t, ocmServer, func(reconcilerConfig *workers.ReconcilerConfig) {
+		// set the interval to 1 second to have sufficient time for the metric to be propagate
+		// so that metrics checks does not timeout after 10s
+		reconcilerConfig.ReconcilerRepeatInterval = 1 * time.Second
+	})
+	defer teardown()
+
+	// skip the test if not running on ocm mock mode. We cannot test using an actual ocm env as the values for the quota costs
+	// cannot be pre-determined for the assertions as they may change overtime.
```
machi1990, line 99: makes sense!

internal/kafka/internal/workers/clusters_mgr_test.go
```diff
@@ -337,13 +335,13 @@ func TestClusterManager_processMetrics(t *testing.T) {
...
-					ProviderFactory:        tt.fields.ProviderFactory,
+					OCMConfig:              tt.fields.ocmConfig,
+					ProviderFactory:        tt.fields.providerFactory,
 				},
 			}
 			// processMetrics accumulates all the errors encountered during metrics processing in an array.
-			// If that array is non empty then an error should be expected. Otherwise, no errors should be expected.
-			g.Expect(len(c.processMetrics()) > 0).To(gomega.Equal(tt.wantErr))
+			gotErrors := c.processMetrics()
+			g.Expect(len(gotErrors)).To(gomega.Equal(len(tt.wantErrs)), "Errors received: %v", gotErrors)
```
ziccardi, line 344: Can we use `Equal` so that we will check that the received errors will be precisely the expected errors?

JameelB: Can't use equal on the error list itself unfortunately because the error stack will be different. What we could have here though is check the error message and ensure they're what we expect. wdyt?

JameelB: Added this in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1251/commits/59c21cb7be738860e05428f734666a969fcddc0f, let me know what you think

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -190,18 +190,20 @@ func (c *ClusterManager) Reconcile() []error {
...
 	}
 
 	if err := c.setKafkaPerClusterCountMetrics(); err != nil {
-		return []error{errors.Wrapf(err, "failed to set kafka per cluster count metrics")}
+		errs = append(errs, errors.Wrapf(err, "failed to set kafka per cluster count metrics"))
 	}
 
 	if err := c.setClusterProviderResourceQuotaMetrics(); err != nil {
```
miguelsorianod, line 202: Question around this:

This method iterates on all QuotaCost of the organization that the AMS user account making the request belongs to.

If you take a look for example to the AMS stage organization we are using the results include non-rhosak QuotaCosts too like for example:
```
    {
      "allowed": 33024,
      "consumed": 24,
      "href": "/api/accounts_mgmt/v1/organizations/1gxCqJ5UorERrM42S8fRz8lT3ML/quota_cost",
      "kind": "QuotaCost",
      "organization_id": "1gxCqJ5UorERrM42S8fRz8lT3ML",
      "quota_id": "compute.node|cpu|byoc|osd",
      "version": "1ee76705-de4f-4c60-a39c-1c914719dd6c"
    },
```

or

```
    {
      "allowed": 12,
      "consumed": 0,
      "href": "/api/accounts_mgmt/v1/organizations/1gxCqJ5UorERrM42S8fRz8lT3ML/quota_cost",
      "kind": "QuotaCost",
      "organization_id": "1gxCqJ5UorERrM42S8fRz8lT3ML",
      "quota_id": "cluster|rhinfra|rhosr|any",
      "version": "1ee76705-de4f-4c60-a39c-1c914719dd6c"
    },
```

No filtering of any kind is being performed. It seems to me that is not what we really want. Is that intentional?

JameelB: it's intentional to avoid hardcoding values for filtering these quotas. It was discussed here: https://issues.redhat.com/browse/MGDSTRM-9403

miguelsorianod: A while ago I implemented the `GetQuotaCostsForProduct` method where you can filter by the product name and resource name and you don't need to filter specifying the entire quota IDs.
The logic is based on filtering based on the RelatedResources in AMS.
Maybe we could reuse that to get the information for the products and resource names we are interested on?

JameelB: We could. We may also need to include `resource_type` in the filters as the compute nodes returns ~300 related resources.  We'll need to hardcode the following in the ocm provider: product name ('OSD'/'ANY'), resource_name (addon names) and resource_type ('add-on', 'compute.node', 'cluster'). One other thing to note is we'll need to update these in case the resource information changes/a new quota has been added, which doesn't really happen often.

Does the above sound good to you @miguelsorianod @machi1990 @pb82?

miguelsorianod: We would need to check if it is possible to filter at ReservedResource level on AMS API server side.

I am not too worried about updating when changes/new quota has been added. It shouldn't happen often: we don't add new addons or products frequently

JameelB: Not sure if there is a way. I had a look at the api docs for that [endpoint](https://api.openshift.com/?urls.primaryName=Accounts%20management%20service#/default/get_api_accounts_mgmt_v1_organizations__orgId__quota_cost), there is a search parameter but I don't think it's possible to use it on the `related_resources` column since that's an array. We may have to filter the results on our side.

JameelB: I've updated this PR with the proposed changes in this thread, see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1251/commits/7294f6f31ecbd2d1c0b719f4b0956efc1f596c45. Please take a look when you get a chance.

machi1990: the new commit lgtm.

I'll let @miguelsorianod have a look as well

machi1990 (approved):
lgtm, thanks @JameelB 

I'll let @miguelsorianod have a look as well

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -26,6 +26,14 @@ const (
 	ocmMultiAZClusterNodeScalingMultiple = 3
 )
 
+var (
+	// OCM quota filters
+	QuotaAddonResourceNamePrefix        = "addon"
```
miguelsorianod, line 31: Can we namespace the names to include `OCM` or `AMS` to make it more clear that are quota concepts tied to that? ,a s opposed to the general quota concept in kas fleet manager or quota management list

pkg/client/ocm/client.go
```diff
@@ -496,31 +514,43 @@ func (c client) GetQuotaCosts(organizationID string, fetchRelatedResources, fetc
 		return nil, err
 	}
 	quotaCostList := quotaCostResp.Items()
-	return quotaCostList, nil
+
+	if !fetchRelatedResources || len(relatedResourceFilters) == 0 {
+		return quotaCostList.Slice(), nil
+	}
+
+	var quotaCosts []*amsv1.QuotaCost
```
miguelsorianod, line 522: Have we verified the behavior is the same when there are no results with this change? is a nil slice returned? an empty one?

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -442,7 +450,7 @@ func (o *OCMProvider) CreateMachinePool(request *types.MachinePoolRequest) (*typ
 	return request, err
 }
 
-// GetQuotaCosts returns a list of ocm resource quota cost information for the authenticated user
+// GetQuotaCosts returns a list of data plane related ocm resource quota cost information for the authenticated user
 // Returns a nil slice when no resource quota is assigned to the user
 func (o *OCMProvider) GetQuotaCosts() ([]types.QuotaCost, error) {
```
miguelsorianod, line 455: Is there some reason we don't include rhosak related quotacosts in here?
In any case we should document the behavior of this as the method name is very generic but we return very specific results

pkg/client/ocm/client.go
```diff
@@ -487,7 +487,25 @@ func (c *client) CreateMachinePool(clusterID string, machinePool *clustersmgmtv1
 	return createdMachinePool, nil
 }
 
-func (c client) GetQuotaCosts(organizationID string, fetchRelatedResources, fetchCloudAccounts bool) (*amsv1.QuotaCostList, error) {
+// QuotaCostRelatedResourceFilter represents the properties of the related resource, associated
+// to each quota cost, that can be used to filter the result of the get quota costs request
+type QuotaCostRelatedResourceFilter struct {
+	ResourceName *string
```
miguelsorianod, line 493: Can we document the behavior about one of the attributes of the filter being nil meaning it is not evaluated?

pkg/client/ocm/client.go
```diff
@@ -496,31 +514,43 @@ func (c client) GetQuotaCosts(organizationID string, fetchRelatedResources, fetc
 		return nil, err
 	}
 	quotaCostList := quotaCostResp.Items()
-	return quotaCostList, nil
+
+	if !fetchRelatedResources || len(relatedResourceFilters) == 0 {
```
miguelsorianod, line 518: I think the behavior of this method is incorrect now:

In the case where fetchRelatedResources is `true` and relatedResourceFilters is empty(or nil) we are now returning a nil slice. That would be incorrect. The result should be all the QuotaCosts.

Can we also cover that in a test to ensure we don't break it in the future?

miguelsorianod: edit: I was not right on my comment. In that case this if would be `true` and it would return the quotas as they are

pkg/client/ocm/client.go
JameelB: A nil slice is returned. yes it's the same, i tested it manually though but i'll add a test for this.

internal/kafka/internal/clusters/ocm_provider.go
JameelB: The rhosak related quota costs are for creating kafkas and we don't need to care about that since those are only used for testing purposes.

Would changing the method name here be better? Maybe change it to `GetClusterResourceQuotaCosts`? Then I can add a comment in here that states this will only return quota related to cluster resources used when provisioning and terraforming a data plane cluster. wdyt?

pkg/client/ocm/client.go
JameelB: I'll add a test for this.

internal/kafka/internal/clusters/ocm_provider.go
miguelsorianod: Changing the name seems a good idea.

internal/kafka/test/integration/ocm_client_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"testing"
+
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/ocm"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/test/mocks"
+	"github.com/onsi/gomega"
+	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+)
+
+func TestOCMClientGetQuotaCosts(t *testing.T) {
```
JameelB, line 13: @miguelsorianod had to do an integration test to test the get quota cost logic because I couldn't mock the ocm client itself.

machi1990:
@miguelsorianod can you have another look at the PR as well? Thanks

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -456,12 +466,44 @@ func (o *OCMProvider) GetQuotaCosts() ([]types.QuotaCost, error) {
...
+		// Filter for osd cluster
+		{
+			ResourceType: &AMSQuotaClusterResourceType,
+			Product:      &AMSQuotaOSDProductName,
+		},
+	}
+
+	// Only include cluster logging operator filter if its specified
+	// If not specified, this won't install on the data plane cluster so we don't need to retrieve the quota cost information
+	if o.ocmConfig.ClusterLoggingOperatorAddonID != "" {
```
miguelsorianod, line 495: As a side comment there is current work to move cluster logging addon installation to be a SyncSet instead of an addon so maybe not worth tracking this?

pkg/metrics/metrics.go
```diff
@@ -719,9 +745,10 @@ func init() {
 	prometheus.MustRegister(clusterStatusCountMetric)
 	prometheus.MustRegister(kafkaPerClusterCountMetric)
 	prometheus.MustRegister(clusterStatusCapacityMaxMetric)
-	prometheus.MustRegister(KasFleetManagerClusterProviderResourceQuotaConsumedMetric)
 	prometheus.MustRegister(clusterStatusCapacityUsedMetric)
 	prometheus.MustRegister(clusterStatusCapacityAvailableMetric)
+	prometheus.MustRegister(ClusterProviderResourceQuotaConsumedMetric)
+	prometheus.MustRegister(clusterProviderResourceQuotaMaxAllowedMetric)
```
miguelsorianod, line 751: Is there some reason one of the variables is exported and the other one is unexported?

JameelB: ah good catch, didn't notice that. No reason but it should be unexported to keep consistent with the rest of the metrics defined here.

internal/kafka/internal/clusters/ocm_provider.go
JameelB: Thanks for pointing that out. +1 for me, i don't think it's worth tracking this. I'll remove it.

ziccardi (approved):
Great job!

miguelsorianod:
I see there are currently 14 commits for this PR. Are those relevant or should we squash some?

pkg/client/ocm/client_test.go
```diff
@@ -0,0 +1,116 @@
...
+				},
+			},
+			want: true,
+		},
+	}
+	for _, tt := range tests {
+		tc := tt
+		t.Run(tc.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			qcf := &QuotaCostRelatedResourceFilter{
```
miguelsorianod, line 103: Is there some reason we build the QuotaCostRelatedResourceFilter here instead of directly defining it in the `fields` struct?

JameelB: That's because that's the type that the method belongs to which we're testing in this test. Mostly we construct the object in this way. The properties of this are usually what we include in the fields rather than constructing the whole object within it. It's how vscode auto-generated this test. 

If it's preferred to define this in the field let me know and i'll update it.

miguelsorianod (approved):
Looks good to me :+1:

PR #1258, opened 2022-08-18
feat: add cluster_provider_resource_quota_consumed_ metric

JasonPowr:
## Description
Adds kas_fleet_manager_quota_consumed to the KAS Fleet Manager metrics. This represents the quota currently used by the organisation of the ocm account used by KAS Fleet Manager.

Jira: [Here](https://issues.redhat.com/browse/MGDSTRM-9404)

## Verification Steps
1.Run kas fleet manager with ocm credentials specified
2.Curl http://localhost:8080/metrics
3.Ensure the quota metric as defined above are included in the response with correct values

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/workers/clusters_mgr.go
+++ b/internal/kafka/internal/workers/clusters_mgr.go
@@ -197,6 +197,9 @@ func (c *ClusterManager) processMetrics() []error {
 		return []error{errors.Wrapf(err, "failed to set kafka per cluster count metrics")}
 	}
 
+	if err := c.setKasFleetManagerClusterProviderResourceQuotaConsumedMetric(); err != nil {
+		return []error{errors.Wrapf(err, "failed to set Kas Fleet Manager Cluster Provider Resource Quota Consumed metric")}
+	}
 	return []error{}
 }
 
@@ -1241,6 +1244,25 @@ func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
 			return err
 		}
 		metrics.UpdateKafkaPerClusterCountMetric(counter.Clusterid, clusterExternalID, counter.Count)
+	}
+
+	return nil
+}
+
+func (c *ClusterManager) setKasFleetManagerClusterProviderResourceQuotaConsumedMetric() error {
+	ocmCredentialsSpecified := (c.OCMConfig.ClientID != "" && c.OCMConfig.ClientSecret != "") || c.OCMConfig.SelfToken != ""
+	if ocmCredentialsSpecified {
+		provider, err := c.ProviderFactory.GetProvider(api.ClusterProviderOCM)
+		if err != nil {
+			return err
+		}
+		quota, err := provider.GetQuotaCosts()
+		if err != nil {
+			return err
+		}
+		for _, q := range quota {
+			metrics.UpdateKasFleetManagerClusterProviderResourceQuotaConsumed(q.ID, api.ClusterProviderOCM.String(), q.Consumed)
+		}
 	}
 
 	return nil
--- a/internal/kafka/internal/workers/clusters_mgr.go
+++ b/internal/kafka/internal/workers/clusters_mgr.go
@@ -197,7 +197,7 @@ func (c *ClusterManager) processMetrics() []error {
 		return []error{errors.Wrapf(err, "failed to set kafka per cluster count metrics")}
 	}
 
-	if err := c.setKasFleetManagerClusterProviderResourceQuotaConsumedMetric(); err != nil {
+	if err := c.setClusterProviderResourceQuotaMetrics(); err != nil {
 		return []error{errors.Wrapf(err, "failed to set Kas Fleet Manager Cluster Provider Resource Quota Consumed metric")}
 	}
 	return []error{}
@@ -1249,21 +1249,21 @@ func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
 	return nil
 }
 
-func (c *ClusterManager) setKasFleetManagerClusterProviderResourceQuotaConsumedMetric() error {
-	ocmCredentialsSpecified := (c.OCMConfig.ClientID != "" && c.OCMConfig.ClientSecret != "") || c.OCMConfig.SelfToken != ""
-	if ocmCredentialsSpecified {
-		provider, err := c.ProviderFactory.GetProvider(api.ClusterProviderOCM)
-		if err != nil {
-			return err
-		}
-		quota, err := provider.GetQuotaCosts()
-		if err != nil {
-			return err
-		}
-		for _, q := range quota {
-			metrics.UpdateKasFleetManagerClusterProviderResourceQuotaConsumed(q.ID, api.ClusterProviderOCM.String(), q.Consumed)
-		}
+func (c *ClusterManager) setClusterProviderResourceQuotaMetrics() error {
+	ocmCredentialsSpecified := c.OCMConfig.HasCredentials()
+	if !ocmCredentialsSpecified {
+		return nil
 	}
-
+	provider, err := c.ProviderFactory.GetProvider(api.ClusterProviderOCM)
+	if err != nil {
+		return err
+	}
+	quota, err := provider.GetQuotaCosts()
+	if err != nil {
+		return err
+	}
+	for _, q := range quota {
+		metrics.UpdateKasFleetManagerClusterProviderResourceQuotaConsumed(q.ID, api.ClusterProviderOCM.String(), q.Consumed)
+	}
 	return nil
 }
--- a/internal/kafka/test/integration/cloudprovider_test.go
+++ b/internal/kafka/test/integration/cloudprovider_test.go
@@ -369,9 +369,9 @@ func TestListCloudProviderRegions(t *testing.T) {
 	defer teardown()
 	var clusterService services.ClusterService
 	h.Env.MustResolve(&clusterService)
-	cluster := dataPlaneClusterConfig.ClusterConfig.GetManualClusters()
+	clusters := dataPlaneClusterConfig.ClusterConfig.GetManualClusters()
 
-	for _, c := range cluster {
+	for _, c := range clusters {
 		_, err := common.WaitForClusterStatus(h.DBFactory(), &clusterService, c.ClusterId, api.ClusterReady)
 		g.Expect(err).NotTo(gomega.HaveOccurred())
 	}
--- a/internal/kafka/test/integration/cloudprovider_test.go
+++ b/internal/kafka/test/integration/cloudprovider_test.go
@@ -7,7 +7,9 @@ import (
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test/common"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/test/mocks"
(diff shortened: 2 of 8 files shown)


JasonPowr:
@JameelB  :)

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -1245,3 +1248,22 @@ func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
 
 	return nil
 }
+
+func (c *ClusterManager) setKasFleetManagerClusterProviderResourceQuotaConsumedMetric() error {
+	ocmCredentialsSpecified := (c.OCMConfig.ClientID != "" && c.OCMConfig.ClientSecret != "") || c.OCMConfig.SelfToken != ""
```
machi1990, line 1253: This is also here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1251 maybe it is good to have a `.CheckIfCredentialsAreProvided()` method in the `OCMConfig` struct that can be used here and in the linked PR

/cc @JameelB @JasonPowr

internal/kafka/test/integration/cloudprovider_test.go
```diff
@@ -360,9 +363,18 @@ func TestListCloudProviderRegions(t *testing.T) {
...
 	})
 	defer teardown()
+	var clusterService services.ClusterService
+	h.Env.MustResolve(&clusterService)
+	cluster := dataPlaneClusterConfig.ClusterConfig.GetManualClusters()
+
+	for _, c := range cluster {
+		_, err := common.WaitForClusterStatus(h.DBFactory(), &clusterService, c.ClusterId, api.ClusterReady)
+		g.Expect(err).NotTo(gomega.HaveOccurred())
+	}
```
machi1990, line 377: Why is this test modified?

internal/kafka/test/integration/kafkas_test.go
```diff
@@ -774,6 +774,11 @@ func TestKafkaCreate_TooManyKafkas(t *testing.T) {
 	h, client, teardown := test.NewKafkaHelperWithHooks(t, ocmServer, configHook)
 	defer teardown()
 
+	var clusterService services.ClusterService
+	h.Env.MustResolve(&clusterService)
+	_, err := common.WaitForClusterStatus(h.DBFactory(), &clusterService, "test01", api.ClusterReady)
+	g.Expect(err).NotTo(gomega.HaveOccurred())
```
machi1990, line 780: Why is this modified?

pkg/metrics/metrics.go
```diff
@@ -256,6 +275,16 @@ var clusterStatusCapacityMaxMetric = prometheus.NewGaugeVec(
 	clusterStatusCapacityLabels,
 )
 
+// create a new gaugeVec for the quota an organisation is currently using
+var KasFleetManagerClusterProviderResourceQuotaConsumedMetric = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Subsystem: KasFleetManager,
+		Name:      KasFleetManagerClusterProviderResourceQuotaConsumed,
+		Help:      "quota an organisation is currently using",
```
machi1990, line 283:
```suggestion
		Help:      "cluster resource quota that are current consumed",
```

internal/kafka/internal/workers/clusters_mgr.go
JameelB: having a method for it sounds good. We can use it in other places too if needed in the future.

JameelB:
Majority of the changes here will be the same for https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1251. Looking to get this in first then I can rebase my changes once merged.

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -1245,3 +1248,22 @@ func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
 
 	return nil
 }
+
+func (c *ClusterManager) setKasFleetManagerClusterProviderResourceQuotaConsumedMetric() error {
```
JameelB, line 1252: Can we change the name to this? 
```suggestion
func (c *ClusterManager) setClusterProviderResourceQuotaMetrics() error {
```

I don't think we need separate methods betwen max allowed and consumed since both requires the same logic. The max allowed metric updated can be added in here as well.

internal/kafka/internal/workers/clusters_mgr.go
miguelsorianod: Shouldn't we check instead whether the cluster provider is OCM first at all?

miguelsorianod: Additionally, why is this metric limited to OCM?

internal/kafka/test/integration/cloudprovider_test.go
JasonPowr: Hey @machi1990  After making the changes for this PR These two integration tests started to fail. I believe there was some sort of timing issue that was causing them to fail and @JameelB  helped me fix them.

internal/kafka/internal/workers/clusters_mgr.go
machi1990: > Shouldn't we check instead whether the cluster provider is OCM instead?

These are not cluster specific metrics but cluster provider e.g OCM, wide metrics.  

> Additionally, why is this metric limited to clusters provisioned through OCM?

For other only provider that we've i.e `standalone` the count will be zero everywhere (`consumed` and `max`). I think once we've other cluster providers we could add them.

internal/kafka/test/integration/cloudprovider_test.go
machi1990: Thanks

internal/kafka/internal/workers/clusters_mgr.go
```diff
@@ -1245,3 +1248,22 @@ func (c *ClusterManager) setKafkaPerClusterCountMetrics() error {
 
 	return nil
 }
+
+func (c *ClusterManager) setClusterProviderResourceQuotaMetrics() error {
+	ocmCredentialsSpecified := c.OCMConfig.CheckIfCredentialsAreProvided()
+	if ocmCredentialsSpecified {
```
machi1990, line 1254: we could do an early return here by inversing the condition.
```suggestion
	if !ocmCredentialsSpecified {
```

what do you think? @JameelB @JasonPowr

pkg/client/ocm/config.go
```diff
@@ -63,6 +63,13 @@ func (c *OCMConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.ClusterLoggingOperatorAddonID, "cluster-logging-operator-addon-id", "", "The name of the cluster logging operator addon. An empty string indicates that the operator should not be installed")
 }
 
+func (c *OCMConfig) CheckIfCredentialsAreProvided() bool {
+	if (c.ClientID != "" && c.ClientSecret != "") || c.SelfToken != "" {
+		return true
+	}
+	return false
```
machi1990, line 70:
```suggestion
	return c.ClientID != "" && c.ClientSecret != "") || c.SelfToken != ""
```

pkg/client/ocm/config.go
```diff
@@ -63,6 +63,13 @@ func (c *OCMConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&c.ClusterLoggingOperatorAddonID, "cluster-logging-operator-addon-id", "", "The name of the cluster logging operator addon. An empty string indicates that the operator should not be installed")
 }
 
+func (c *OCMConfig) CheckIfCredentialsAreProvided() bool {
```
machi1990, line 66: we can add a unit test for this?

machi1990 (approved):
the code change lgtm

I've left some minor suggestions, once these are addressed.  
I'll leave it out to @JameelB @miguelsorianod to see if their reviews have been addressed and merge

internal/kafka/test/integration/cloudprovider_test.go
JameelB: It's because this change adds additional time to the processMetrics which is the first step in the cluster manager reconcile. This results in the dummy clusters to not be recorded in the database in time for the assertions in this test. 

Having to wait for the cluster to be ready and recorded in the database makes sense here either way since the cloud provider regions endpoint relies on a cluster being there, otherwise it the cloud provider regions endpoint returns an empty list.

internal/kafka/internal/workers/clusters_mgr.go
JameelB: Yeah that makes sense. 

Just note that this will need to be modified if we ever add new providers that we want to expose this metric for since we wouldn't want to return straight away in that case. This is probably far in the future though :)

internal/kafka/test/integration/cloudprovider_test.go
```diff
@@ -360,9 +363,18 @@ func TestListCloudProviderRegions(t *testing.T) {
...
 		})
+		dataPlaneClusterConfig = dc
 
 	})
 	defer teardown()
+	var clusterService services.ClusterService
+	h.Env.MustResolve(&clusterService)
+	cluster := dataPlaneClusterConfig.ClusterConfig.GetManualClusters()
+
+	for _, c := range cluster {
```
JameelB, line 374: let's make it clear that this is a slice
```suggestion
	clusters := dataPlaneClusterConfig.ClusterConfig.GetManualClusters()

	for _, c := range clusters {
```

pkg/metrics/metrics.go
```diff
@@ -73,6 +73,11 @@ const (
 	// ClusterStatusCapacityAvailable - metric name for the number of available instances
 	ClusterStatusCapacityAvailable = "cluster_status_capacity_available"
 
+	// KasFleetManagerClusterProviderResourceQuotaConsumed - metric name for how much quota an organisation is currently using
+	KasFleetManagerClusterProviderResourceQuotaConsumed = "kas_fleet_manager_cluster_provider_resource_quota_consumed"
```
JameelB, line 77: No need to add KasFleetManager as a prefix. This is taken care of by `Subsystem: KasFleetManager,` in the metric definition. 

```suggestion
	// KasFleetManagerClusterProviderResourceQuotaConsumed - metric name for how much quota an organisation is currently using
	ClusterProviderResourceQuotaConsumed = "cluster_provider_resource_quota_consumed"
```

pkg/metrics/metrics.go
```diff
@@ -73,6 +73,11 @@ const (
...
+
+	LabelQuotaId    = "quota_id"
+	LabelProvider   = "provider"
```
JameelB, line 80: Let's make this more specific to cluster_provider to differentiate it with other provider types like cloud provider.

pkg/metrics/metrics.go
```diff
@@ -158,6 +163,11 @@ var clusterStatusCapacityLabels = []string{
 	LabelCloudProvider,
 }
 
+var KasFleetManagerClusterProviderResourceQuotaConsumedLabels = []string{
```
JameelB, line 166: Remove KasFleetManager prefix to be consistent with other variables. 
```suggestion
var ClusterProviderResourceQuotaConsumedLabels = []string{
```

pkg/metrics/metrics.go
```diff
@@ -73,6 +73,11 @@ const (
 	// ClusterStatusCapacityAvailable - metric name for the number of available instances
 	ClusterStatusCapacityAvailable = "cluster_status_capacity_available"
 
+	// KasFleetManagerClusterProviderResourceQuotaConsumed - metric name for how much quota an organisation is currently using
```
JameelB, line 76:
```suggestion
	// KasFleetManagerClusterProviderResourceQuotaConsumed - metric name for how much quota, given to a user by a cluster provider, is currently used.
```

quota by organisation is an ocm detail and other providers may not implement it this way.

pkg/metrics/metrics.go
```diff
@@ -214,6 +224,15 @@ func UpdateClusterStatusCapacityMaxCount(provider string, region, instanceType,
 	clusterStatusCapacityMaxMetric.With(labels).Set(count)
 }
 
+// UpdateKasFleetManagerClusterProviderResourceQuotaConsumed - sets quota an organisation is currently using
+func UpdateKasFleetManagerClusterProviderResourceQuotaConsumed(quotaId, provider string, count int) {
```
JameelB, line 228: Move this closer to the metric definition it's updating. Best to keep them together to find related uses of the metric easier.

JameelB: update the description to replace `organisation` by `user`.

[optional] change `provider` -> `clusterProvider` to make it clear what provider this method is looking for.

pkg/client/ocm/config.go
```diff
@@ -63,6 +63,10 @@ func (c *OCMConfig) AddFlags(fs *pflag.FlagSet) {
...
```
ziccardi, line 66: Function that just returns a boolean, are usually used to answer to a specific question with a yes or no. IMO this would be better called something like `HasCredentials`, 'CredentialsExists`, `AreCredentialsProvided` or something like that.

JameelB (approved):

PR #1265, opened 2022-08-26
feat(connectors): fixes MGDCTRS-1230, added support for API to list connector type labels/categories and counts

dhirajsb:
## Description
Added API to list connector type labels/categories and counts, including count of featured connector types implemented using a new column `featured_rank`.

## Verification Steps
Included integration tests for new API to list connector type labels. 

- [ ] Code Review completed


openapi/connector_mgmt.yaml
```diff
@@ -1315,6 +1353,10 @@ components:
               type: array
               items:
                 type: string
+            featured_rank:
```
lburgazzoli, line 1356: It is not clear to be what is the purpose of this filed, how it is consumed and set.
Can you clarify ?

dhirajsb: Yeah, this field was kinda born out of tech discussions with UI and UX teams, so it has barely been captured in MGDCTRS-1230. The idea is to allow us to decide an order for all the _featured_ connector types. So, a value of `0` means not a featured connector, and if it's not `0` it's used to order featured types. So the UI will add `orderBy=featured_rank desc` param when listing connector types. 

Although, the implementation doesn't really care what the non-zero values are, we could come up with some convention like values would be `10,100,1000,` etc.

@gashcrumb did I explain that correctly?

lburgazzoli (approved):

PR #1266, opened 2022-08-29
refactor(cluster-reconciler): create a cleanup and deprovisioning clusters reconcilers

machi1990:
## Description

Create a cleanup and deprovisioning clusters reconcilers
First part of https://issues.redhat.com/browse/MGDSTRM-9137.
I'll followup with this by:
- creating a dynamic scale down reconciler as well
- connecting the scale down logic to deprovisioning and cleanup logic

## Verification Steps

Unit test and integration tests(where deprovisioning and cleanup logic is executed) should continue to pass

- [ ] Code Review completed


internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -119,6 +119,10 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 		}
 	}
 
+	if len(errList) > 0 {
```
miguelsorianod, line 122: There is no need to add this condition here. I think we can just return `errList`

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
```diff
@@ -0,0 +1,145 @@
...
+type CleanupClusterExternalResourcesManager struct {
+	id           string
+	workerType   string
+	isRunning    bool
+	imStop       chan struct{} //a chan used only for cancellation.
+	syncTeardown sync.WaitGroup
+	CleanupClusterExternalResourcesManagerOptions
+}
+
+type CleanupClusterExternalResourcesManagerOptions struct {
```
miguelsorianod, line 35: I personally prefer explicitly passing the dependencies individually in the constructor to be aware of them instead of using `di.Inject` in a struct (as it's done in the different kafka managers) , but that's a personal style preference.

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
```diff
@@ -0,0 +1,145 @@
...
+
+	for _, cluster := range cleanupClusters {
+		glog.V(10).Infof("cleanup cluster ClusterID = %s", cluster.ClusterID)
+		metrics.UpdateClusterStatusSinceCreatedMetric(cluster, api.ClusterCleanup)
+		if err := c.reconcileCleanupCluster(cluster); err != nil {
+			errList.AddErrors(errors.Wrapf(err, "failed to reconcile cleanup cluster %s", cluster.ID))
+		}
+	}
+
+	if len(errList) > 0 {
```
miguelsorianod, line 120: No need to conditionally return errList

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_cluster_mgr.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	deprovisioningWorkerType = "cluster_deprovisioning"
+)
+
+// DeprovisioningClusterManager represents a cluster manager that periodically reconciles data plane clusters in deprovisioning state.
+type DeprovisioningClusterManager struct {
+	id           string
```
miguelsorianod, line 27: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1266/files#r957100730

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990: This is needed because otherwise the lint ended up complaining that a non nil value is never returned.

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/pkg/errors"
+)
+
+const (
+	cleanupClusterExternalResourcesWorkerType = "cleanup_cluster_external_resources"
+)
+
+// CleanupClusterExternalResourcesManager represents a worker that periodically reconciles data plane clusters in cleanup state.
+type CleanupClusterExternalResourcesManager struct {
```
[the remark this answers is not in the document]
machi1990: > We can embed worker.Baseworker. 

Yes, I am aware of that. The reason I didn't choose to go `worker.Baseworker` is that the `worker.Baseworker.Stop()` methods ends up resetting the kafka specific metrics which I something I didn't want.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990:
```
internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go:94:5: SA4023: this comparison is always true (staticcheck)
	if err != nil {
	   ^
internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go:93:9: SA4023(related information): the lhs of the comparison is the 1st return value of this function call (staticcheck)
	err := c.processCleanupClusters()
	       ^
internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go:101:50: SA4023(related information): (*github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/workers/cluster_mgrs.CleanupClusterExternalResourcesManager).processCleanupClusters never returns a nil interface value (staticcheck)
func (c *CleanupClusterExternalResourcesManager) processCleanupClusters() error {
```

it was complaining with something like this..

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
machi1990: I can change it.  The reason I used it here was mostly because the two added workers were copy-pasted and since the number of arguments for the new worker was quite big as well I thought we can continue to keep them in the arg containter

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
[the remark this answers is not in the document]
machi1990: Actually, not resetting the kafka specific metrics is not necessarily a bad thing and also it could be solved differently. 
Let me re-work this part

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_cluster_mgr.go
machi1990: adjusted in the latest commit

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
machi1990: adjusted

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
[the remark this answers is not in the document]
machi1990: changed, thanks for the review

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: :+1:

miguelsorianod: I think the issue here is that
	var errList fleeterrors.ErrorList

is `nil`,

but when converted to an error it becomes an interface that is not nil, but it contains a nil value

internal/kafka/internal/workers/cluster_mgrs/cleanup_cluster_external_resources_mgr.go
```diff
@@ -0,0 +1,113 @@
...
```
miguelsorianod, line 23: I am thinking about the name of `CleanupClusterExternalResourcesManager` will a cluster in `cleanup` state be only limited to cleaning up of external cluster resources or could it do other cleanup actions? the name is quite specific and it matches the current behavior but will it match potential future behavior? what if for example as part of cleanup it needs to send some notification, event...?

machi1990: Thinking about it the name also diverges a bit from the pattern `state_mgr.go`. In that regard, I am thinking of naming the file to just `cleanup_cluster_mgr.go` and the type to `CleanupClusterManager`, what do you think?

miguelsorianod: The proposed name in plural would be good.
The reason for plural is to be consistent with the kafka managers

machi1990: Perfect, thanks. Changing now

machi1990: Change pushed

JameelB (approved):
lgtm, i'll leave the final approval to @miguelsorianod

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+func (m *CleanupClustersManager) Reconcile() []error {
+	glog.Infoln("reconciling clusters")
+
+	var errList fleeterrors.ErrorList
+	err := m.processCleanupClusters()
+	if err != nil {
+		errList.AddErrors(err)
+	}
+
+	return errList.ToErrorSlice()
```
ziccardi, line 66: Why do we convert it to a slice? I believe the function should return the `error` interface, and it will be up to the caller to cast it to `ErrorList` if he needs the embedded error list.

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+		glog.V(10).Infof("cleanup cluster ClusterID = %s", cluster.ClusterID)
+		metrics.UpdateClusterStatusSinceCreatedMetric(cluster, api.ClusterCleanup)
+		if err := m.reconcileCleanupCluster(cluster); err != nil {
+			errList.AddErrors(errors.Wrapf(err, "failed to reconcile cleanup cluster %s", cluster.ID))
+		}
+	}
+
+	if len(errList) > 0 {
+		return errList
+	}
```
ziccardi, line 90: This is something that probably will be done often. I believe it would be helpful to add an `IsEmpty` method to ErrorList to make this more readable. WDYT?

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return nil
+}
+
+func (m *CleanupClustersManager) reconcileCleanupCluster(cluster api.Cluster) error {
+	glog.Infof("Removing Dataplane cluster %s IDP client", cluster.ClusterID)
+	keycloakDeregistrationErr := m.osdIdpKeycloakService.DeRegisterClientInSSO(cluster.ID)
+	if keycloakDeregistrationErr != nil {
+		return errors.Wrapf(keycloakDeregistrationErr, "Failed to removed Dataplance cluster %s IDP client", cluster.ClusterID)
```
ziccardi, line 99: Errors should start with lowercase letter
```suggestion
		return errors.Wrapf(keycloakDeregistrationErr, "failed to removed Dataplance cluster %s IDP client", cluster.ClusterID)
```

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	glog.Infof("Removing Dataplane cluster %s fleetshard service account", cluster.ClusterID)
+	serviceAcountRemovalErr := m.kasFleetshardOperatorAddon.RemoveServiceAccount(cluster)
+	if serviceAcountRemovalErr != nil {
+		return errors.Wrapf(serviceAcountRemovalErr, "Failed to removed Dataplance cluster %s fleetshard service account", cluster.ClusterID)
```
ziccardi, line 104:
```suggestion
		return errors.Wrapf(serviceAcountRemovalErr, "failed to removed Dataplance cluster %s fleetshard service account", cluster.ClusterID)
```

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	glog.Infof("Soft deleting the Dataplane cluster %s from the database", cluster.ClusterID)
+	deleteError := m.clusterService.DeleteByClusterID(cluster.ClusterID)
+	if deleteError != nil {
+		return errors.Wrapf(deleteError, "Failed to soft delete Dataplance cluster %s from the database", cluster.ClusterID)
```
ziccardi, line 110:
```suggestion
		return errors.Wrapf(deleteError, "failed to soft delete Dataplance cluster %s from the database", cluster.ClusterID)
```

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (m *DeprovisioningClustersManager) Reconcile() []error {
+	glog.Infoln("reconciling clusters")
+
+	var errList fleeterrors.ErrorList
+	err := m.processDeprovisioningClusters()
+	if err != nil {
+		errList.AddErrors(err)
+	}
```
ziccardi, line 59: We always do this check. Maybe we could make `AddErrors` nil safe and make it no-op when the received error is `nil` so that we don't need to make this test anymore. WDYT?

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
miguelsorianod: The reason we convert it to an empty slice is to avoid refactoring the caller code in this PR, as `Reconcile` is an interface method and changing the signature would mean refactoring all the workers in this PR

machi1990: The `.Reconciler` signatures needs a slice hence the conversion to a slice.
Changing the signature of that method requires a whole lot of other modification which are not part of this PR.

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
machi1990: Sounds good. I'll add it.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
ziccardi: I don't think that is the correct interpretation. The issue here, IMO, is that ErrList is not a pointer, so it can't be `nil`.
Internally it is declared as a slice, so ErrorList, in the beginning, is an empty slice.

You can see try it yourself here:
https://go.dev/play/p/94SP5EqWyE_z

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
machi1990: I do not know, no opinion either way. What if (hypothetically) someone wants to append `nil` in the list? 
I'll keep it like this for now. At the same time,  I'll tag @miguelsorianod in the discussion as well. I am happy for the evaluation to be done of whether we want to do the above suggestion and we could change it all over in a separate PR

ziccardi: It is a list of errors, so I believe appending a `nil` would make no sense.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: `ErrList` is a pointer and it can be nil.
See https://go.dev/play/p/Gbs3mrR7_tJ

ziccardi: Never said it can't be nil, just it would be a long discussion to explain it here. I could make a blog for that :)
Have you looked at my code? That is **exactly** as the code in this PR and it is an empty array.

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
machi1990: > It is a list of errors, so I believe appending a nil would make no sense.

Seems like we want to make it `The list of non nil errors` because  `nil` is also an `error` ;-)

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
miguelsorianod: I looked at your code yes, and if I understand the behavior properly what I said still applies.

The confusion you are having is that Println doesn't show the difference between empty and nil in the case of slices. It prints both cases as `[]`.

Add the following code at the end of your example:
```
	fmt.Println(reflect.ValueOf(err).IsNil())
```

and you will be able to see that

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
machi1990: Added

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
ziccardi: You are totally right. Sorry for the confusion. Probably my holiday was not enough 😄

machi1990: :-) Thanks for the discussion @ziccardi @miguelsorianod 

Just adding in my2cents: I didn't dig further the reason why but I believe the debugger was showing the similar behavior described in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1266#discussion_r957177361

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+const (
+	cleanupClustersWorkerType = "cleanup_clusters"
+)
+
+// CleanupClustersManager represents a worker that periodically reconciles data plane clusters in cleanup state.
+type CleanupClustersManager struct {
+	workers.BaseWorker
+	clusterService             services.ClusterService
+	kasFleetshardOperatorAddon services.KasFleetshardOperatorAddon
+	osdIdpKeycloakService      sso.OsdKeycloakService
```
miguelsorianod, line 27: let's call this `osdIDPKeycloakService`: acronyms in Go should be written in uppercase and not mixed case (as long as it doesn't affect exportability), see https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// NewCleanupClustersManager creates a new worker that reconciles data plane clusters in cleanup state.
+func NewCleanupClustersManager(reconciler workers.Reconciler,
+	clusterService services.ClusterService,
+	kasFleetshardOperatorAddon services.KasFleetshardOperatorAddon,
+	osdIdpKeycloakService sso.OsdKeycloakService) *CleanupClustersManager {
```
miguelsorianod, line 34: Same as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1266/files#r957409647

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -0,0 +1,113 @@
...
+		return errList
+	}
+
+	glog.Infof("cleanup clusters count = %d", len(cleanupClusters))
+
+	for _, cluster := range cleanupClusters {
+		glog.V(10).Infof("cleanup cluster ClusterID = %s", cluster.ClusterID)
+		metrics.UpdateClusterStatusSinceCreatedMetric(cluster, api.ClusterCleanup)
+		if err := m.reconcileCleanupCluster(cluster); err != nil {
+			errList.AddErrors(errors.Wrapf(err, "failed to reconcile cleanup cluster %s", cluster.ID))
```
miguelsorianod, line 84: Can we use `cluster.ClusterID` instead, and make explicit in the message that it is a ClusterID? Asking this because on stage/prod environments we don't usually have access to the DB to get the KFM internal IDs

machi1990: done

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if errList.IsEmpty() {
+		return nil
+	}
+
+	return errList
+}
+
+func (m *DeprovisioningClustersManager) reconcileDeprovisioningCluster(cluster *api.Cluster) error {
+	if m.dataplaneClusterConfig.IsDataPlaneAutoScalingEnabled() {
```
miguelsorianod, line 93: I guess part of this logic will be moved out of the deprovisioning reconciler and moved to the future scale down manager right? and even then we will still have a deprovisioning clusters manager

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/onsi/gomega"
+
+	apiErrors "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+)
+
+var deprovisionCluster = api.Cluster{
+	Status: api.ClusterDeprovisioning,
+}
+
+func TestCleanupClustersManager_processCleanupClusters(t *testing.T) {
```
miguelsorianod, line 18: Not to change in this PR if not wanted but I think we should start thinking on a test method naming scheme:

Sometimes I see test names written as:
Test_<TypeBeingTested>_<method_or_behavior_BeingTested>

and sometimes I see:
Test<TypeBeingTested>_<method_or_behavior_BeingTested>

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
machi1990: The plan was to keep the logic here as well:
- the future scale down manager will perform a similar logic
- and the deprovisioning manager will perform that same logic to be sure that it can safely delete the cluster.

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr_test.go
machi1990: +1

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/onsi/gomega"
+	"github.com/pkg/errors"
+
+	apiErrors "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+)
+
+var autoScalingDataPlaneConfig = &config.DataplaneClusterConfig{
```
miguelsorianod, line 16: can we set it as a variable inside the test method if it is only used in one test? If it is used in more than one test and we keep it as a package level variable can we rename this to be more specific? for example adding `test` at the beginning of the name. The reason in the later case is due to all code in the package share that same namespace so from non-test code we would see `autoScalingDataPlane` config variable which I think is too generic as it is a package level namespace where non test code coexists

machi1990: This is only used for testing. It makes sense to make them test specific. Doing that also avoid the need to modify one test ending up messing other tests because the variable is global

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
machi1990: It is a sort of double verification that I think we can keep.
- the one in scale down manager reconciler will be used to trigger deprovisioning and marking the cluster as `deprovisioning`
- the one in here will be used to assure that we can safely delete the cluster without causing issues.

miguelsorianod: It might be worth  thinking a little bit about where to perform the sibling logic.
In the case of scale up mgr for example the logic around empty clusters was deleted from any other reconciler and dynamic scale up mgr takes care of that part

miguelsorianod (approved):
:+1: great work

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
machi1990: The plan I've for #1267 is to have a scale down evaluator which performs all the scale down evaluation logic, this will tell us whether a given cluster can be scaled down or no: via the `.ShouldScaleDown()` method of the processor.

Then that logic will be called in the scale down manager by the scale down executor.

And over here, it'll be called in this reconciler as well. 

> It might be worth thinking a little bit about where to perform the sibling logic in next PRs.

So it won't be the sibling logic but it'll be the whole `can we safely scale down` logic being performed here at this part as well. 
The need for that over here is mainly as a safe guard / being extra cautious. 
We could talk more about this on that follow up PR

PR #1267, opened 2022-08-29
Add scale down manager

machi1990:
## Description

This PR adds the dynamic scale down manager.
It's implementation is heavily influenced by the DynamicScaleUpManager design. 

Additionally, the dynamic scale down evaluator re-uses the dynamic scale up evaluation when evaluating for the below conditions: 

```
      * There is at least a cluster in the provider's region that has enough capacity to allocate
        the biggest instance size of the instance types supported by the empty cluster
      * There will be free capacity (in streaming units) for each instance type supported by the cluster in
        the provider's region that is smaller or equal than the defined slack capacity (also in streaming units) of the given instance type. 
        Free capacity is defined as max(total) capacity - consumed capacity.
        For the calculation of the max capacity:
        * Clusters in `deprovisioning` and `cleanup` state are excluded, as
          clusters into those states don't accept kafka instances anymore.
        * Clusters that are still not `ready` to accept kafka instance are also excluded from the capacity calculation
```

The motivation was that if a cluster is to be removed, then it shouldn't provoke an immediate scale up. So instead of duplicating that evaluation logic, re-using the dynamic scale up evaluator made total sense. 
 
~~The idea behind opening this one as a draft is to get feedback first on the proposed `.ShouldScaleDown()` pseudo code logic. Once enough feedback is given, I'll go ahead with coding the change and then promote the PR as ready for review. 
With that being said, any early feedback on something else will not be addressed up until the PR is ready for review.~~

~~Please only concentrate on the last commit: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/commit/6393ac9928e1bdd4b4fa3574354d7bf932996252 the first one is taken from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1266 I'll rebase after it is merged~~

## Verification Steps

1 Added unit test should pass
2 Updated integration test should pass in both environments. 

- ~~[] Documentation added for the feature~~
- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/config/dataplane_cluster_config.go
+++ b/internal/kafka/internal/config/dataplane_cluster_config.go
@@ -40,22 +40,23 @@ type DataplaneClusterConfig struct {
 	// 'manual' to use OSD Cluster configuration file,
 	// 'auto' to use dynamic scaling
 	// 'none' to disabled scaling all together, useful in testing
-	DataPlaneClusterScalingType                 string
-	DataPlaneClusterConfigFile                  string
-	ReadOnlyUserList                            userv1.OptionalNames
-	ReadOnlyUserListFile                        string
-	KafkaSREUsers                               userv1.OptionalNames
-	KafkaSREUsersFile                           string
-	ClusterConfig                               *ClusterConfig
-	EnableReadyDataPlaneClustersReconcile       bool
-	EnableKafkaSreIdentityProviderConfiguration bool
-	EnableDynamicScaleUpManagerScaleUpTrigger   bool
-	Kubeconfig                                  string
-	RawKubernetesConfig                         *clientcmdapi.Config
-	StrimziOperatorOLMConfig                    OperatorInstallationConfig
-	KasFleetshardOperatorOLMConfig              OperatorInstallationConfig
-	DynamicScalingConfig                        DynamicScalingConfig
-	NodePrewarmingConfig                        NodePrewarmingConfig
+	DataPlaneClusterScalingType                   string
+	DataPlaneClusterConfigFile                    string
+	ReadOnlyUserList                              userv1.OptionalNames
+	ReadOnlyUserListFile                          string
+	KafkaSREUsers                                 userv1.OptionalNames
+	KafkaSREUsersFile                             string
+	ClusterConfig                                 *ClusterConfig
+	EnableReadyDataPlaneClustersReconcile         bool
+	EnableKafkaSreIdentityProviderConfiguration   bool
+	EnableDynamicScaleUpManagerScaleUpTrigger     bool
+	EnableDynamicScaleDownManagerScaleDownTrigger bool
+	Kubeconfig                                    string
+	RawKubernetesConfig                           *clientcmdapi.Config
+	StrimziOperatorOLMConfig                      OperatorInstallationConfig
+	KasFleetshardOperatorOLMConfig                OperatorInstallationConfig
+	DynamicScalingConfig                          DynamicScalingConfig
+	NodePrewarmingConfig                          NodePrewarmingConfig
 }
 
 type OperatorInstallationConfig struct {
@@ -101,20 +102,21 @@ func getDefaultKubeconfig() string {
 
 func NewDataplaneClusterConfig() *DataplaneClusterConfig {
 	return &DataplaneClusterConfig{
-		OpenshiftVersion:                            "",
-		AWSComputeMachineType:                       defaultAWSComputeMachineType,
-		GCPComputeMachineType:                       defaultGCPComputeMachineType,
-		ImagePullDockerConfigContent:                "",
-		ImagePullDockerConfigFile:                   "secrets/image-pull.dockerconfigjson",
-		DataPlaneClusterConfigFile:                  "config/dataplane-cluster-configuration.yaml",
-		ReadOnlyUserListFile:                        "config/read-only-user-list.yaml",
-		KafkaSREUsersFile:                           "config/kafka-sre-user-list.yaml",
-		DataPlaneClusterScalingType:                 ManualScaling,
-		ClusterConfig:                               &ClusterConfig{},
-		EnableReadyDataPlaneClustersReconcile:       true,
-		EnableKafkaSreIdentityProviderConfiguration: true,
-		EnableDynamicScaleUpManagerScaleUpTrigger:   true,
-		Kubeconfig:                                  getDefaultKubeconfig(),
+		OpenshiftVersion:                              "",
+		AWSComputeMachineType:                         defaultAWSComputeMachineType,
+		GCPComputeMachineType:                         defaultGCPComputeMachineType,
+		ImagePullDockerConfigContent:                  "",
+		ImagePullDockerConfigFile:                     "secrets/image-pull.dockerconfigjson",
+		DataPlaneClusterConfigFile:                    "config/dataplane-cluster-configuration.yaml",
+		ReadOnlyUserListFile:                          "config/read-only-user-list.yaml",
+		KafkaSREUsersFile:                             "config/kafka-sre-user-list.yaml",
+		DataPlaneClusterScalingType:                   ManualScaling,
+		ClusterConfig:                                 &ClusterConfig{},
+		EnableReadyDataPlaneClustersReconcile:         true,
+		EnableKafkaSreIdentityProviderConfiguration:   true,
+		EnableDynamicScaleUpManagerScaleUpTrigger:     true,
+		EnableDynamicScaleDownManagerScaleDownTrigger: true,
+		Kubeconfig: getDefaultKubeconfig(),
 		StrimziOperatorOLMConfig: OperatorInstallationConfig{
 			IndexImage:             defaultStrimziOperatorIndexImage,
 			Namespace:              constants.StrimziOperatorNamespace,
--- a/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
+++ b/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
@@ -84,52 +84,22 @@ func (m *DynamicScaleDownManager) processDynamicScaleDownReconcileEvent() error 
 		return errList
 	}
 
-	type processed struct {
-		indexesOfStreamingUnitForSameClusterID []int
-		processed                              bool
-	}
-
-	var alreadyProcessedClusters map[string]processed = map[string]processed{}
-
-	for i, kafkaStreamingUnitCountPerCluster := range kafkaStreamingUnitCountPerClusterList {
-		clusterID := kafkaStreamingUnitCountPerCluster.ClusterId
-		existing, ok := alreadyProcessedClusters[clusterID]
-
-		if !ok {
-			alreadyProcessedClusters[clusterID] = processed{
-				// consider non ready cluster as already processed for scale down evaluation and thus they'll be skipped from scale down consideration
-				processed:                              kafkaStreamingUnitCountPerCluster.Status != api.ClusterReady.String(),
-				indexesOfStreamingUnitForSameClusterID: []int{i},
-			}
-		} else {
-			alreadyProcessedClusters[clusterID] = processed{
-				indexesOfStreamingUnitForSameClusterID: append(existing.indexesOfStreamingUnitForSameClusterID, i),
-				processed:                              existing.processed,
-			}
-		}
-	}
+	processedClusters := m.createAMapOfProcessedClusters(kafkaStreamingUnitCountPerClusterList)
 
 	for _, suCount := range kafkaStreamingUnitCountPerClusterList {
 		clusterID := suCount.ClusterId
-		existing, alreadyProcessed := alreadyProcessedClusters[clusterID]
-		if alreadyProcessed && existing.processed { // skip already processed
+		existing := processedClusters[clusterID]
+		if existing.processed { // skip already processed
 			glog.V(10).Infof("cluster with cluster_id %s is already processed", suCount.ClusterId)
 			continue
 		}
 
-		alreadyProcessedClusters[clusterID] = processed{
+		processedClusters[clusterID] = processed{
 			indexesOfStreamingUnitForSameClusterID: existing.indexesOfStreamingUnitForSameClusterID,
 			processed:                              true,
 		}
 
-		var regionsSupportedInstanceType config.InstanceTypeMap
-		provider, ok := m.clusterProvidersConfig.ProvidersConfig.SupportedProviders.GetByName(suCount.CloudProvider)
-		if ok {
-			region, regionFound := provider.Regions.GetByName(suCount.Region)
-			if regionFound {
-				regionsSupportedInstanceType = region.SupportedInstanceTypes
-			}
-		}
+		regionsSupportedInstanceType := m.findRegionInstanceTypeConfiguration(suCount)
 
 		var dynamicScaleDownProcessor dynamicScaleDownProcessor = &standardDynamicScaleDownProcessor{
 			kafkaStreamingUnitCountPerClusterList:  kafkaStreamingUnitCountPerClusterList,
@@ -162,6 +132,68 @@ func (m *DynamicScaleDownManager) processDynamicScaleDownReconcileEvent() error 
 	}
 
 	return errList
+}
+
+// processed is a struct that holds indexes with same cluster is and whether the cluster
+// representing the streaming unit index has been processed or not
+type processed struct {
+	indexesOfStreamingUnitForSameClusterID []int
+	processed                              bool
+}
+
+// createAMapOfProcessedClusters creates a map struct that indicates if a cluster_id has been processed for all occurrences of cluster_id.
+// The struct returned is a mmap[string]processed.
+// For example:
+//
+//		{
+//		 "some-cluster-id": {
+//		   processed: false,
+//	       indexesOfStreamingUnitForSameClusterID: [0, 2],
+//		},
+//	     "some-other-cluster-id": {
+//			processed: true,
+//		    indexesOfStreamingUnitForSameClusterID: [1],
+//		},
+//
+// }
+// Once the struct is built, a cluster is considered processed if it is not in a "ready" state.
+// NOTE: For all the clusters in "ready" state, the processed marker is updated once scale down evaluation has been perfomed for at least one occurrence of a cluster.
+// This is done in the caller of this method.
+func (m *DynamicScaleDownManager) createAMapOfProcessedClusters(kafkaStreamingUnitCountPerClusterList services.KafkaStreamingUnitCountPerClusterList) map[string]processed {
+	var processedClusters map[string]processed = map[string]processed{}
+
+	for i, kafkaStreamingUnitCountPerCluster := range kafkaStreamingUnitCountPerClusterList {
+		clusterID := kafkaStreamingUnitCountPerCluster.ClusterId
+		existing, ok := processedClusters[clusterID]
+
+		if !ok {
+			processedClusters[clusterID] = processed{
+				// consider non ready cluster as already processed for scale down evaluation and thus they'll be skipped from scale down consideration
+				processed:                              kafkaStreamingUnitCountPerCluster.Status != api.ClusterReady.String(),
+				indexesOfStreamingUnitForSameClusterID: []int{i},
+			}
+		} else {
+			processedClusters[clusterID] = processed{
+				indexesOfStreamingUnitForSameClusterID: append(existing.indexesOfStreamingUnitForSameClusterID, i),
+				processed:                              existing.processed,
+			}
+		}
+	}
+
+	return processedClusters
+}
+
+// findRegionInstanceTypeConfiguration finds the instance type configuration for a region represented in the given streaming unit
+func (m *DynamicScaleDownManager) findRegionInstanceTypeConfiguration(suCount services.KafkaStreamingUnitCountPerCluster) config.InstanceTypeMap {
+	var regionsSupportedInstanceType config.InstanceTypeMap
+	provider, ok := m.clusterProvidersConfig.ProvidersConfig.SupportedProviders.GetByName(suCount.CloudProvider)
+	if ok {
+		region, regionFound := provider.Regions.GetByName(suCount.Region)
+		if regionFound {
+			regionsSupportedInstanceType = region.SupportedInstanceTypes
+		}
+	}
+	return regionsSupportedInstanceType
 }
 
 // dynamicScaleDownExecutor is able to perform dynamic ScaleDown execution actions
@@ -220,12 +252,8 @@ var _ dynamicScaleDownProcessor = &standardDynamicScaleDownProcessor{}
 // 3. Clusters that are still not ready to accept kafka instance are also excluded from the capacity calculation
 func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
 	// First let's check if the cluster is empty
-	for _, i := range p.indexesOfStreamingUnitForSameClusterID {
-		clusterIsNotEmpty := p.kafkaStreamingUnitCountPerClusterList[i].Count > 0
-		if clusterIsNotEmpty {
-			glog.Infof("cluster with cluster_id %s is not empty. It is not going to be removed", p.clusterID)
-			return false, nil
-		}
+	if p.isClusterNotEmpty() {
+		return false, nil
 	}
 
 	// let's check if the cluster can be safely removed without causing a scale up event
@@ -234,6 +262,34 @@ func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
 		return true, nil
 	}
 
+	newkafkaStreamingUnitCountPerClusterList := p.createNewStreamingUnitPerClusterListAfterRemovalOfCandidateCluster()
+	scaleUpNeededAfterRemoval, err := p.isScaleUpNeededAfterCandidateClusterRemoval(newkafkaStreamingUnitCountPerClusterList)
+	if err != nil {
+		return false, err
+	}
+
+	// to safely perform scale down, there shouldn't a need of scale up immediately afterwards
+	return !scaleUpNeededAfterRemoval, nil
+}
+
+// isClusterNotEmpty checks whether the cluster is not empty.
+// The method iterates through all occurrences of cluster_id
+func (p *standardDynamicScaleDownProcessor) isClusterNotEmpty() bool {
+	for _, i := range p.indexesOfStreamingUnitForSameClusterID {
+		clusterIsNotEmpty := p.kafkaStreamingUnitCountPerClusterList[i].Count > 0
+		if clusterIsNotEmpty {
+			glog.Infof("cluster with cluster_id %s is not empty. It is not going to be removed", p.clusterID)
+			return true
+		}
+	}
+
+	return false
+}
+
+// createNewStreamingUnitPerClusterListAfterRemovalOfCandidateCluster creates a new kafka streaming unit list with the following charcteristics:
+// 1. The candidate deletion cluster is removed from the original list
+// 2. Ignore clusters in terraforming states for scale up evaluation as they are not ready yet. We only want to delete the cluster if we've a sibling ready cluster
+func (p *standardDynamicScaleDownProcessor) createNewStreamingUnitPerClusterListAfterRemovalOfCandidateCluster() services.KafkaStreamingUnitCountPerClusterList {
 	newkafkaStreamingUnitCountPerClusterList := services.KafkaStreamingUnitCountPerClusterList{}
 
 	clusterStatesTowardReadyState := []string{
@@ -242,6 +298,7 @@ func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
 	}
 
 	for _, suCount := range p.kafkaStreamingUnitCountPerClusterList {
+
 		// Ignore clusters in terraforming states for scale up evaluation as they are not ready yet.
 		// We only want to delete the cluster if we've a sibling ready cluster
 		if arrays.Contains(clusterStatesTowardReadyState, suCount.Status) {
@@ -258,7 +315,12 @@ func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
 
 		newkafkaStreamingUnitCountPerClusterList = append(newkafkaStreamingUnitCountPerClusterList, suCount)
 	}
+	return newkafkaStreamingUnitCountPerClusterList
+}
 
+// isScaleUpNeededAfterCandidateClusterRemoval evaluates wheteher scale up is needed given the newkafkaStreamingUnitCountPerClusterList which does not contain the
+// candidate cluster to delete occurrences and clusters in terraforming states.
+func (p *standardDynamicScaleDownProcessor) isScaleUpNeededAfterCandidateClusterRemoval(newkafkaStreamingUnitCountPerClusterList services.KafkaStreamingUnitCountPerClusterList) (bool, error) {
 	for _, i := range p.indexesOfStreamingUnitForSameClusterID {
 		suCount := p.kafkaStreamingUnitCountPerClusterList[i]
 
@@ -286,16 +348,16 @@ func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
 		shouldScaleUp, err := dynamicScaleUpProcessor.ShouldScaleUp()
 		if err != nil {
 			glog.Infof("scale up evaluation results returned an error. Not removing cluster with cluster_id %s:", suCount.ClusterId)
-			return false, err
+			return shouldScaleUp, err
 		}
 
 		if shouldScaleUp {
 			glog.Infof("scale up will be needed if the cluster with cluster_id %s is removed. The decision is not to remove it", suCount.ClusterId)
-			return false, nil
+			return shouldScaleUp, nil
 		}
 	}
 
-	return true, nil
+	return false, nil
 }
 
 // ScaleDown marks the cluster as deprovisioning.
--- a/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
+++ b/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
@@ -1,0 +1,322 @@
+package cluster_mgrs
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
+	fleeterrors "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+	"github.com/golang/glog"
+
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/workers"
+	"github.com/google/uuid"
+)
+
+const (
+	dynamicScaleDownWorkerType = "dynamic_scale_down"
+)
+
+type DynamicScaleDownManager struct {
+	workers.BaseWorker
+
+	dataplaneClusterConfig *config.DataplaneClusterConfig
+	clusterProvidersConfig *config.ProviderConfig
+	kafkaConfig            *config.KafkaConfig
+	clusterService         services.ClusterService
+}
+
+var _ workers.Worker = &DynamicScaleDownManager{}
+
+func NewDynamicScaleDownManager(
+	reconciler workers.Reconciler,
+	dataplaneClusterConfig *config.DataplaneClusterConfig,
+	clusterProvidersConfig *config.ProviderConfig,
+	kafkaConfig *config.KafkaConfig,
+	clusterService services.ClusterService,
+) *DynamicScaleDownManager {
+
+	return &DynamicScaleDownManager{
+		BaseWorker: workers.BaseWorker{
+			Id:         uuid.New().String(),
+			WorkerType: dynamicScaleDownWorkerType,
+			Reconciler: reconciler,
+		},
+
+		dataplaneClusterConfig: dataplaneClusterConfig,
+		clusterProvidersConfig: clusterProvidersConfig,
+		kafkaConfig:            kafkaConfig,
+		clusterService:         clusterService,
+	}
+}
+
+func (m *DynamicScaleDownManager) Start() {
+	m.StartWorker(m)
+}
+
+func (m *DynamicScaleDownManager) Stop() {
+	m.StopWorker(m)
+}
+
+func (m *DynamicScaleDownManager) Reconcile() []error {
+	var errList fleeterrors.ErrorList
+	if !m.dataplaneClusterConfig.IsDataPlaneAutoScalingEnabled() {
+		glog.Infoln("dynamic scaling is disabled. Dynamic scale down reconcile event skipped")
+		return nil
+	}
+
+	glog.Infoln("running dynamic scale down reconcile event")
+
+	err := m.processDynamicScaleDownReconcileEvent()
+	if err != nil {
+		errList.AddErrors(err)
+	}
+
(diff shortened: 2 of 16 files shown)


internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,241 @@
...
+// It returns true if all the following conditions happen:
+// 1. If specified the cluster is empty i.e it does not contain any streaming unit
+// 2. If the cluster can be removed without triggering a scale down action
+// Otherwise false is returned.
+// Note: This method assumes kafkaStreamingUnitCountPerClusterList does not
+//       contain elements with the Status attribute with the 'failed' value.
+//       Thus, if the type is constructed with the assumptions being true, it
+//       can be considered as the 'failed' state Clusters are not included in
+//       the calculations.
+func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
```
machi1990, line 197: @JameelB @miguelsorianod looking for your feedback on this logic

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		if shouldScaleDown {
+			glog.Infof("data plane scale down need detected for cluster '%s'", clusterID)
+			err := dynamicScaleDownProcessor.ScaleDown()
+			if err != nil {
+				errList.AddErrors(err)
+				continue
+			}
+		}
+
```
machi1990, line 115: @miguelsorianod this is heavily inspired by the scale up reconciler design.
Let me know what you think

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,241 @@
...
+	dataplaneClusterConfig *config.DataplaneClusterConfig
+	clusterProvidersConfig *config.ProviderConfig
+	kafkaConfig            *config.KafkaConfig
+	clusterService         services.ClusterService
+}
+
+var _ workers.Worker = &DynamicScaleDownManager{}
+
+func NewDynamicScaleDownManager(
```
machi1990, line 29: Test will be added later, for now we want to nail the scale down detection logic

internal/kafka/internal/config/dataplane_cluster_config.go
```diff
@@ -114,7 +115,8 @@ func NewDataplaneClusterConfig() *DataplaneClusterConfig {
 		EnableReadyDataPlaneClustersReconcile:       true,
 		EnableKafkaSreIdentityProviderConfiguration: true,
 		EnableDynamicScaleUpManagerScaleUpTrigger:   true,
-		Kubeconfig:                                  getDefaultKubeconfig(),
+		EnableDynamicScaleUpManagerScaleDownTrigger: true,
```
miguelsorianod, line 118: I know it's still a draft but pointing it just in case: The variable should be named `EnableDynamicScaleDownManagerScaleDownTrigger` in case it ends up here

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// TODO implement a check that a cluster can be removed in the region without causing scale up
+
+	// 1. Create a list of streaming that has all occurrences of p.clusterID removed from p.kafkaStreamingUnitCountPerClusterList. Let's call this list "newkafkaStreamingUnitCountPerClusterList"
+	// shouldScaleDown := true
+	// 2. For each element in "clusterKafkaStreamingUnitCount"
+	//      retrieve the instance type and check whether we need scale up to happen in the region
+	//         use the "dynamicScaleUpProcessor" for this, passing in "newkafkaStreamingUnitCountPerClusterList" as the value for the "dynamicScaleUpProcessor.kafkaStreamingUnitCountPerClusterList" field
+	//      if shouldScaleUp
+	//          shouldScaleDown = false
```
machi1990, line 218: the for-loop for step (2) ends here

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,241 @@
...
+// It returns true if all the following conditions happen:
+// 1. If specified the cluster is empty i.e it does not contain any streaming unit
+// 2. If the cluster can be removed without triggering a scale down action
+// Otherwise false is returned.
+// Note: This method assumes kafkaStreamingUnitCountPerClusterList does not
+//       contain elements with the Status attribute with the 'failed' value.
+//       Thus, if the type is constructed with the assumptions being true, it
+//       can be considered as the 'failed' state Clusters are not included in
+//       the calculations.
+func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
```
miguelsorianod, line 197: As mentioned in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1266#discussion_r957451954 let's think about the sibling clusters logic: with the new dynamic scaling behavior is that logic still needed at all or are we already dealing with it somehow? maybe there's a better way to ensure we don't end up with 0 cluters in the region that does not require changing the state of already existing clusters to ready? like, not setting the cluster to deprovision at all if we detect that would end up happening. Alternatively, if we end up determining that we still need that same sibling logic, where should be more appropriate to put that?

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,241 @@
...
+	// TODO implement checking if the cluster is empty
+
+	// 1. Evaluate if the cluster is empty
+	//     From p.kafkaStreamingUnitCountPerClusterList find the streaming unit count for p.clusterID.
+	//     store this result in a sub-list "clusterKafkaStreamingUnitCount"
+	//     clusterIsEmpty := true
+	//     For each of the element in the sub-list "clusterKafkaStreamingUnitCount"
+	//         if element.Count > 0
+	//				clusterIsEmpty = false
```
machi1990, line 206: the for-loop for step (1) ends here

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: > like, not setting the cluster to deprovision at all if we detect that would end up happening

This is already taken care of by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1267/files#diff-7750701a0ae3440f5aacd7bdbe12d43b25c1c2f01eaa827dd85f10a10aa39c95R130

i.e after the evaluation of all the scaling down logic has been completed. We are not going to mark the scale for deprovisioning if either of the checks returned false i.e 
if one of the following conditions are true, then the cluster shouldn't be deleted: 
- the cluster is non empty 
- or we do not have a sibling cluster in the region after the cluster is deleted, 
- we'll be eating the slack capacity after the cluster is deleted 
- or we'll be unable to provision the largest streaming unit in the region after the cluster is deleted.  

> maybe there's a better way to ensure we don't end up with 0 cluters in the region that does not require changing the state back of already existing clusters to ready?

Because of the checks we've, I am not so much worried about the sibling clusters cases and we can remove the logic we discussed in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1266#discussion_r957451954

The case where I was thinking that bringing back the state of the cluster will be useful is when:
- we saw that we can delete a cluster, 
- but before we mark it as empty, a Kafka is assigned onto that cluster: an event we do not know about so we end up marking the cluster for deletion anyway.  

However, thinking afresh about this, an SQL query might be enough when marking the cluster for deletion. 
```
set status=deprovisioning where given_cluster_id not in (select cluster_id from kafka_requests).
```
I'll give it more thought today but I feel like that should be enough and we can delete the other logic effectively.

machi1990: discussed internally:

The decision was to keep the check that the cluster is empty, if not empty then bring back the state of the cluster to `ready`, otherwise we go ahead and delete the cluster from OCM. 
I've updated the code in that direction.

machi1990:
I've promoted the PR as ready for review. The pseudo code logic from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/commit/6393ac9928e1bdd4b4fa3574354d7bf932996252 was the one that's ended up being implemented.
It was rather straight forward and I'll be happy to take any feedback along that direction with this ready for review PR.

internal/kafka/internal/config/dataplane_cluster_config.go
machi1990: updated

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: I've added unit test and updated the integration test

miguelsorianod:
I don't see any reference to capacity slack in the scaledown mgr code. Where is that managed?

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
```diff
@@ -27,6 +27,10 @@ type DynamicScaleUpManager struct {
 }
 
 var _ workers.Worker = &DynamicScaleUpManager{}
+var clusterStatesTowardReadyState = []string{
```
miguelsorianod, line 30: This has been added as a global variable but not used in dynamic scale up mgr.

I also wonder if it would be a good idea to just duplicate this to not affect both dynamic scaleup/scaledown mgrs if the set of states changes

machi1990:
> I don't see any reference to capacity slack in the scaledown mgr code. Where is that managed?

That's managed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1267/files#diff-7750701a0ae3440f5aacd7bdbe12d43b25c1c2f01eaa827dd85f10a10aa39c95R281

internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go
machi1990: an IDE refactoring not completely finished. 
I should have remove the local reference.

I am fine with duplicating it either way as I think the value won't change

machi1990: done

miguelsorianod:
> > I don't see any reference to capacity slack in the scaledown mgr code. Where is that managed?
> 
> That's managed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1267/files#diff-7750701a0ae3440f5aacd7bdbe12d43b25c1c2f01eaa827dd85f10a10aa39c95R281

Could you expand on that? the link points me to the following line content:
` 			clusterService:                        p.clusterService,`

machi1990:
> > > I don't see any reference to capacity slack in the scaledown mgr code. Where is that managed?
> > 
> > 
> > That's managed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1267/files#diff-7750701a0ae3440f5aacd7bdbe12d43b25c1c2f01eaa827dd85f10a10aa39c95R281
> 
> Could you expand on that? the link points me to the following line content: ` clusterService: p.clusterService,`

ah sorry, the link got outdated.  

```
There is at least a cluster in the provider's region that has enough capacity to allocate
        the biggest instance size of the instance types supported by the empty cluster
      * There will be free capacity (in streaming units) for each instance type supported by the cluster in
        the provider's region that is smaller or equal than the defined slack capacity (also in streaming units) of the given instance type. 
        Free capacity is defined as max(total) capacity - consumed capacity.
```
All the above considerations are taken care by:
  
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/521b681eb155df6243ceb2dff273a16b52b11280/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go#L286-L293

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,321 @@
...
+		if alreadyProcessed && existing.processed { // skip already processed
+			continue
+		}
+
+		alreadyProcessedClusters[clusterID] = processed{
+			indexesOfStreamingUnitForSameClusterID: existing.indexesOfStreamingUnitForSameClusterID,
+			processed:                              true,
+		}
+
+		if suCount.Status != api.ClusterReady.String() { // do not consider non ready cluster for scale down evaluation
```
miguelsorianod, line 122: Maybe we can apply cluster filtering above on kafkaStreamingUnitCountPerClusterList ?

machi1990: At this line https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/4627eb783d268e54ca1da2102f22b33ac367f9e5/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go#L100 
with something like
```
processed: kafkaStreamingUnitCountPerCluster.Status != api.ClusterReady.String()
```
?

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,321 @@
...
+		var regionsSupportedInstanceType config.InstanceTypeMap
+		provider, ok := m.clusterProvidersConfig.ProvidersConfig.SupportedProviders.GetByName(suCount.CloudProvider)
+		if ok {
+			region, regionFound := provider.Regions.GetByName(suCount.Region)
+			if regionFound {
+				regionsSupportedInstanceType = region.SupportedInstanceTypes
+			}
+		}
+
+		dynamicScaleDownProcessor := &standardDynamicScaleDownProcessor{
```
miguelsorianod, line 135: Here using `dynamicScaleDownProcessor` type instead of `standardDynamicScaleDownProcessor` might be a good idea to do less changes in case dependency injection is needed in the near future (you can see scale up mgr code for an example)

machi1990: Sounds good. I'll add the type.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,321 @@
...
+			indexesOfStreamingUnitForSameClusterID: existing.indexesOfStreamingUnitForSameClusterID,
+			processed:                              true,
+		}
+
+		if suCount.Status != api.ClusterReady.String() { // do not consider non ready cluster for scale down evaluation
+			continue
+		}
+
+		var regionsSupportedInstanceType config.InstanceTypeMap
+		provider, ok := m.clusterProvidersConfig.ProvidersConfig.SupportedProviders.GetByName(suCount.CloudProvider)
```
miguelsorianod, line 127: I would like to think a little bit about this:

Seems unlikely a cluster will be in a non supported cloudprovider and region, but if that's the case how should we proceed? I think ignoring it at this level would be a good idea. Automatically removing it might be too aggressive/surprising behavior

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,321 @@
...
+			region:           suCount.Region,
+			instanceTypeName: suCount.InstanceType,
+		}
+
+		instanceTypeConfig, ok := p.regionsSupportedInstanceType[suCount.InstanceType]
+		if !ok {
+			continue
+		}
+
+		dynamicScaleUpProcessor := &standardDynamicScaleUpProcessor{
```
miguelsorianod, line 276: I think passing the scale up processor as a dependency of this type as an interface might be a good idea as it would let us unit test with custom dynamic scale up implementations

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: I think it is fine as long as as the cluster is empty, it can be removed. 
There are no region considerations that we can take on the cluster since the region is not supported anymore.
If the region is added back, then we should be fine as dynamic scale up will be bring up the needed cluster.

The approach taken here favors cost saving us leaving the cluster around while it can be deleted is not desirable cost wise, so as long as the cluster (in a region that's does not exists ) does not contain kafka workload we can remove it.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: For the standard scale down processor, I wanted to use the standard scale up processor i.e the same scale up logic.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: done

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
miguelsorianod: I think the comment was not understood.
What I am suggesting is that a `dynamicScaleUpProcessor` (an interface) is passed to the standardDynamicScaleDownProcessor` instead of instantiating it inside with a concrete type

miguelsorianod: I think this might not be trivial without refactoring as the instantiation depends on new data calculated in the method

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
miguelsorianod: No. I mean just remove the entries from the original list

miguelsorianod: Maybe it's not feasible if we depend on having complete information

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: Yes, that's what I understood as well and I do not think that (customization) is needed. The standard scale down processor needs the standard scale up processor concrete type (basically it needs to evaluate if scale up can happen based on the logic in the standard scale up processor and nothing else). 

As such, the unit test for the `standardDynamicScaleDownProcessor` have been written with this "ensuring no scale up will be triggered" logic in mind.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,318 @@
...
```
miguelsorianod, line 273: I am thinking if using the dynamic scale up processor is reliable for this purpose the way it is implemented in this PR now:
One of the conditions to scale down is that the number of allocatable kafka instances in the region would still be higher than the defined capacity slacks for **all** supported instance types if the cluster were to be deleted.
However the dynamic scale up processor only evaluates it for an individual instance type.
What happens if the cluster being evaluated supports multiple instance types?

miguelsorianod: Also are we redundantly calculating it? The list is at kafka size level and not at kafka instance type level

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: I ended up considering them as already processed. 

We could filter them, as well since over here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6c071fb6699f56cbc3dc379abdaf1c9a9fa15df8/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go#L246 we are already filtering some of them (the status in terraforming states). That could be a potential future perf improvement

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: That's a great question. 

We evaluate scale up for all the supported instance types of the candidate cluster.
The logic in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6c071fb6699f56cbc3dc379abdaf1c9a9fa15df8/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go#L259-L291 handles the multiple instance type cases

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,318 @@
...
+
+	newkafkaStreamingUnitCountPerClusterList := services.KafkaStreamingUnitCountPerClusterList{}
+
+	clusterStatesTowardReadyState := []string{
+		api.ClusterProvisioning.String(), api.ClusterProvisioned.String(),
+		api.ClusterAccepted.String(), api.ClusterWaitingForKasFleetShardOperator.String(),
+	}
+
+	for _, suCount := range p.kafkaStreamingUnitCountPerClusterList {
+		// Ignore clusters in terraforming states for scale up evaluation as they are not ready yet.
```
miguelsorianod, line 244: Is ignoring this correct?

The ScaleUp processor uses that information to check if there is an ongoing scale up action in progress so we should keep them.

miguelsorianod: They are also used to perform capacity calculations

machi1990: Before deleting a cluster, we want to be sure that there is a sibling cluster in a ready state i.e capable of accepting a kafka. If we end up considering the clusters that are not ready yet, we might delete a cluster and find ourselves having to wait until the other clusters that are not ready yet to be ready to start provisioning Kafkas on them.

The approach taken here is to be conservative and only evaluate if scale up needs to be performed against sibling cluster that are "ready". 

To summarise:

If we've cluster `A` that's ready and empty and cluster `B` that's in `accepted`.
We do not want to delete cluster `A` up until we are sure that we've one sibling cluster that's ready i.e until cluster `B` becomes ready.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,318 @@
+package cluster_mgrs
```
miguelsorianod, line 1: For when the PR is refined, can we add debugging logging similarly to scale up mgr?

machi1990: +1, i'll add more log lines on this. 

I was thinking of metrics too for the two reconcilers (I am open to suggestions on them).

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
miguelsorianod: Can't we just check that before?
I am finding this quite confusing/fragile as it has lots of implicit behavior around scale up manager and it assumes clusters toward ready state are there to evaluate some of the conditions.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
miguelsorianod: Not only multiple instance types, but instance types **and** instance sizes of that type no?

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
miguelsorianod: I am not sure that's possible as dynamic scale up mgr needs the global information about all clusters for a given cloudprovider-cloudproviderregion-instancetype

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: Yes. That's already handled in that loop.

miguelsorianod: But what I mean is that we are redundantly calculating the shouldScaleUp no? because shouldScaleUp is at instance type level. It considers all sizes of the instance type

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: > Can't we just check that before?

Can you expand more on this? You want these clusters filtered out of the list completely at the start of scale down evaluation?

> scale up manager assumes clusters toward ready state are there to evaluate some of the conditions

the scale up evaluator is doing it is job correctly for scale. 

But for dynamic scaling down, we do not want to rely only on the "mare" presence of a sibling cluster for us to delete a candidate cluster. The sibling cluster has to be in `ready` state hence this filter: this ensures that no any false signaling that "scale up won't be triggered" will be thrown by the scale up evaluation after the candidate cluster is deleted.

miguelsorianod: > Can you expand more on this? You want these clusters filtered out of the list completely at the start of scale down evaluation?

Check if there is another cluster in Ready state in that region. Then we don't need to call the scale up mgr at all. It is clear we won't want to delete it

miguelsorianod: My concern is that dynamic scale up mgr uses clusters toward ready state to calculate different conditions and here we are excluding them so those capacities will be ignored which they shouldn't

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: Thanks for the clarification. 

> But what I mean is that we are redundantly calculating the shouldScaleUp no? because shouldScaleUp is at instance type level. It considers all sizes of the instance type.

No. Because, here we are dealing only with instance type. The largest size consideration (which I initially wrongly thought was your question) is handled in the `.ShouldScaleUp()` call. 
So at the end of the loop, for each instance type, we'll call `.ShouldScaleUp()` once.

> For example if you have instance type standard with 6 different sizes you only need to call shouldScaleUp once.  However the current logic calls it 6 times

For each instance type we'll call `.ShouldScaleUp()` once. In the `kafkaStreamingUnitCountPerClusterList` list, an instance type is only represented once. All the different sizes are agregated in `.Count` field of each element in that list.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: > Check if there is another cluster in Ready state in that region. That way we know we won't want to delete it if it's the only cluster

Ah right. This is really something that already handled nicely in the `.ShouldScaleUp()`.  Hence the re-use of that behavior to ensure that when we consider that only ready clusters in the region, then no scale up will need to be performed in that region if a candidate cluster is removed.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: added the debug lines.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
miguelsorianod: > No. Because, here we are dealing only with instance type.

Is that correct? I understand that if we have an instance type 'standard' with 2 sizes then wouldn't `kafkaStreamingUnitCountPerClusterList` have 2 entries and thus shouldScaleUp be called for each one?

machi1990: For the same `cluster_id` then there will be one entry and the sizes as considered/treated in the `.Count` - https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8aa0a7614ce607c101f0ef85533fb8f9684c7b13/internal/kafka/internal/services/clusters.go#L799

miguelsorianod: I see. I missed that part. It seems correct then :+1:

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: All the information are provided, except for the clusters in terraforming phase which are filtered out intentionally.
That's because, we don't want https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8aa0a7614ce607c101f0ef85533fb8f9684c7b13/internal/kafka/internal/workers/cluster_mgrs/dynamic_scaleup_mgr.go#L254-L257 to report `no need to scale up after the candidate cluster is removed because we've an ongoing scale up action.` Because if that happens, it might create a situation where we do not have a ready cluster in a region because we have just wrongly deleted the only left

miguelsorianod: The information of those clusters is not only taken into account for that condition but also for other ones

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,324 @@
...
+			kafkaStreamingUnitCountPerClusterList: newkafkaStreamingUnitCountPerClusterList,
+			supportedKafkaInstanceTypesConfig:     p.supportedKafkaInstanceTypesConfig,
+			clusterService:                        p.clusterService,
+			dryRun:                                true,
+		}
+
+		glog.Infof("evaluating whether deleting the cluster %s would trigger scale up for locator '%+v'", p.clusterID, currLocator)
+		shouldScaleUp, err := dynamicScaleUpProcessor.ShouldScaleUp()
+		if err != nil {
+			glog.Infof("scale up evaluation results returned an error. Not removing cluster %s:", suCount.ClusterId)
```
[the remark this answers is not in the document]
machi1990: the pattern applied everywhere of course..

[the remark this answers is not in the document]
miguelsorianod: Yes

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: The other part where they are used is in capacity calculations? if so, I think `.Count` and `.MaxUnits` will be `0` for these clusters until they are `ready: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8aa0a7614ce607c101f0ef85533fb8f9684c7b13/internal/kafka/internal/services/data_plane_cluster.go#L89-L101. 
So even though the information are considered, they do not add/remove much.

miguelsorianod: :thinking: that's correct I think.
Can we expand the documentation around all of this? I think it is not trivial to understand what it is being done and why.
For example the current doc says that it would scale up if:
`// 2. If the cluster can be removed without triggering a scale up action`

But the triggering of a scale up action is really modified as the provided list of clusters is altered to kind of alter what is being returned by shouldScaleUp.

machi1990: That's true. 

This doc https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/docs/architecture/data-plane-osd-cluster-dynamic-scaling.md#osd-cluster-deletion-evaluation is up to date. I think only the method `ShouldScaleDown()` description  is to be updated.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,324 @@
...
+	// Useful when you don't want to trigger a real scale down.
+	dryRun bool
+}
+
+var _ dynamicScaleDownEvaluator = &standardDynamicScaleDownProcessor{}
+var _ dynamicScaleDownExecutor = &standardDynamicScaleDownProcessor{}
+var _ dynamicScaleDownProcessor = &standardDynamicScaleDownProcessor{}
+
+// ShouldScaleDown indicates whether a new data plane
+// cluster should be created for a given instance type in the given provider
```
miguelsorianod, line 212: This is incorrect, I guess due to partial copy of the ShouldScaleUp doc. It is talking about creating a new cluster instead of deleting it, and at a different scope level too

machi1990: Thanks for this. 
This was a heavy copy and paste #ashamed

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,325 @@
...
+	if errList.IsEmpty() {
+		return nil
+	}
+
+	return errList
+}
+
+func (m *DynamicScaleDownManager) logReconcileEventEnd() {
+	glog.Infoln("dynamic scale down reconcile event finished")
+}
```
ziccardi, line 168: Why create a method that calls `Infoln` instead of using only `Infoln`?
I saw we did the same in `dynamic scale up mgr`: I believe I remember it was meant to be extended with some extra info, but that never happened.
My suggestion is to add these methods only when we have some logic for them.

machi1990: Thanks for the suggestion @ziccardi I've re-adjusted this part.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: discussed on another comment.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -0,0 +1,322 @@
...
+// contain elements with the Status attribute with the 'failed' value.
+// Thus, if the type is constructed with the assumptions being true, it
+// can be considered as the 'failed' state Clusters are not included in
+// the calculations.
+// 2. Clusters in deprovisioning and cleanup state are excluded, as clusters into those states don't accept kafka instances anymore.
+// 3. Clusters that are still not ready to accept kafka instance are also excluded from the capacity calculation
+func (p *standardDynamicScaleDownProcessor) ShouldScaleDown() (bool, error) {
+	// First let's check if the cluster is empty
+	for _, i := range p.indexesOfStreamingUnitForSameClusterID {
+		clusterIsNotEmpty := p.kafkaStreamingUnitCountPerClusterList[i].Count > 0
```
akoserwal, line 224: Is there any situation where **kafkaStreamingUnitCountPerClusterList[i]** can be empty?

akoserwal:
@machi1990 great work

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: You mean `.Count` to be `0`?

Yes, if the cluster does not have any Kafka at all.

akoserwal: sorry, I updated my comment, `nil`?

machi1990: Thanks.

No, it'll never be `nil`.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
miguelsorianod: Meaning to be extended was not the only reason. Aside from that, which might be useful:
1. Makes clear the intention of that logic. There are several places where creating a method that only includes one line is reasonable. See the ScaleUp manager for some cases where some conditions where added as 1 line methods to make intent more clear
1. Unconditionally log that line. By having removed that the behavior has been altered: Now when there's an error in the intermediate logic between where the defer originally was and the end of the method we won't perform logging. Whereas before it was unconditional. defer could be called with the glog infolfn call directly but again having it encapsulated as mentioned in the previous point seems reasonable to me.

Again I don't see any issue with having that even if there is no more than one line.

machi1990: > Now when there's an error in the intermediate logic between where the defer originally was and the end of the method we won't perform logging. 

I am trying to process this in context of https://github.com/machi1990/kas-fleet-manager/blob/9399c93d0c24692a61438c884ea8745af6446cee/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go#L67-L75

Given that all the errors are accumulated and returned at the end, what errors do you think can cause the log from being processed?

machi1990: Thanks for confirming this. I'll change it there as well.

miguelsorianod (approved):
Looks good to me :+1: .
I would like to see some of the logic "blocks" encapsulated into their own methods to improve readability but we can do that in a follow-up PR if we want to change it.

machi1990:
> I would like to see some of the logic "blocks" encapsulated into their own methods to improve readability but we can do that in a follow-up PR if we want to change it.

Makes sense @miguelsorianod 
I found room to squeeze some of the changes in this PR. 
I'll push a separate commit for easy of review.

machi1990:
> > I would like to see some of the logic "blocks" encapsulated into their own methods to improve readability but we can do that in a follow-up PR if we want to change it.
> 
> Makes sense @miguelsorianod I found room to squeeze some of the changes in this PR. I'll push a separate commit for easy of review.

@miguelsorianod I've pushed the change here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1267/commits/ea911557de6caaf4353bc89871ca522b5217e339, let me know what you think.

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -164,6 +134,51 @@ func (m *DynamicScaleDownManager) processDynamicScaleDownReconcileEvent() error
...
 }
 
+// processed is a struct that holds indexes with same cluster is and whether the cluster
+// representing the streaming unit index has been processed or not
+type processed struct {
+	indexesOfStreamingUnitForSameClusterID []int
+	processed                              bool
+}
+
+// createAMapOfProcessedClusters creates a map struct that indicates if a cluster_id has been processed for all occurrences of cluster_id
```
miguelsorianod, line 144: Can we document what do we consider a processed cluster, and what are we specifically returning here?

miguelsorianod:
Looks good to me :+1:

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
machi1990: updated.

machi1990:
Thanks I've pushed the latest change. I'll merge this PR once the test passes. We can iterate on this on follow up work.

PR #1279, opened 2022-09-01
Increase db max connections limit

JameelB:
## Description
Increase the max_connections limit for both local and CI Postgres container to resolve issues with integration tests reaching the default (100) limit.

Value for the max_connections ranges from 1..262143. Setting the value to 200 for now as our integration tests only require around 105.

Note that this is only a temporary fix until the underlying issue has been resolved (see [MGDSTRM-9553](https://issues.redhat.com/browse/MGDSTRM-9553))

## Verification Steps
Run the following commands to verify local db set up.
1. `make db/setup`
2. `make db/login`
3. `show max_connections;`
    - Value should be 200

For CI changes, see the Lint & Test run of this PR. 
- `Show inital max_connections value` step should show the default value
- `Show max_connections after new configuration has been applied` step should show the new value (200). 

- [x] Code Review completed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -73,6 +73,7 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+          --name kas-fleet-manager-db
 
       keycloak:
         image: quay.io/keycloak/keycloak:17.0.1-legacy
@@ -103,6 +104,10 @@ jobs:
           filters: |
             openapi:
               - 'openapi/*.yaml'
+      - name: Increase Postgres max_connections limit
+        run: docker exec -i kas-fleet-manager-db /bin/bash docker-entrypoint.sh psql -U kas_fleet_manager -d serviceapitests -c 'alter system set max_connections=200;'
+      - name: Restart Postgres container
+        run: docker restart --time 0 kas-fleet-manager-db
       - name: Setup Keycloak realm config
         run: make sso/config
       - name: Cache go module
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:11
+        image: postgres:13
         env:
           POSTGRES_PASSWORD: foobar-bizz-buzz
           POSTGRES_USER: kas_fleet_manager
--- a/db_setup_docker.sql
+++ b/db_setup_docker.sql
@@ -1,3 +1,4 @@
 CREATE USER postgres with password 'postgres';
 GRANT ALL PRIVILEGES ON DATABASE serviceapitests TO postgres;
 GRANT ALL PRIVILEGES ON DATABASE serviceapitests TO kas_fleet_manager;
+ALTER SYSTEM SET max_connections=200;
--- a/scripts/local_db_setup.sh
+++ b/scripts/local_db_setup.sh
@@ -18,4 +18,5 @@ docker run \
   -e POSTGRES_USER=$(cat secrets/db.user) \
   -e POSTGRES_DB=$(cat secrets/db.name) \
   -p $(cat secrets/db.port):5432 \
-  -d postgres:13
+  -d postgres:13 \
+  postgres -N 200


JameelB:
Commit **e2e71a395b032a27fa562ec3a206604196ea5204** is only used for verification purposes. Will remove this commit once this PR has been approved.

machi1990:
I assume setting the `dbConfig.MaxOpenConnection` (used in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/connection.go#L53) to `0` didn't have effect?

scripts/local_db_setup.sh
```diff
@@ -18,4 +18,5 @@ docker run \
   -e POSTGRES_USER=$(cat secrets/db.user) \
   -e POSTGRES_DB=$(cat secrets/db.name) \
   -p $(cat secrets/db.port):5432 \
-  -d postgres:13
\ No newline at end of file
+  -d postgres:13 \
+  postgres -N 200
```
machi1990, line 22: can we also do this for:
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e2e71a395b032a27fa562ec3a206604196ea5204/docker/Dockerfile_template#L86 and 
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e2e71a395b032a27fa562ec3a206604196ea5204/docker/Dockerfile_template_mocked#L77 which are used for the midstream PR checks run

pb82:
@JameelB are there two different settings? One on the database itself and one for how many connections Gorm can open?

miguelsorianod:
`dbConfig.MaxOpenConnection` controls the maximum number of open connections in a connection pool that Gorm maintains

whereas the one being changed here is at postgreSQL level

machi1990:
> `dbConfig.MaxOpenConnection` controls the maximum number of open connections in a connection pool that Gorm maintains
> 
> whereas the one being changed here is at postgreSQL level
> 
> Basically dbConfig.MaxOpenconnection should be <= the setting at PostgreSQL level, as if it is bigger it will simply hit the max connections at postgreSQL level before filling the Gorm connection pool connections

Thanks.

miguelsorianod:
> `dbConfig.MaxOpenConnection` controls the maximum number of open connections in a connection pool that Gorm maintains
> 
> whereas the one being changed here is at postgreSQL level
> 
> Basically dbConfig.MaxOpenconnection should be <= the setting at PostgreSQL level, as if it is bigger it will simply hit the max connections at postgreSQL level before filling the Gorm connection pool connections

@JameelB, based on this, does that mean that we are creating multiple gorm connection pools? If we are limiting gorm connection pool to 50 but we are observing ~100 DBs I understand it would mean that

JameelB:
@miguelsorianod yes, it seems like that is happening. The MaxOpenConnection setting is set to 50 at the moment but when I look at the connections open in the db, I can see ~100.

miguelsorianod:
> @miguelsorianod yes, it seems like that is happening. The MaxOpenConnection setting is set to 50 at the moment but when I look at the connections open in the db, I can see ~100.

One of the possible causes then in that case might be that in integration tests we might be calling `gorm.Open` multiple times which is what I think creates a new gorm connection pool

pb82:
`gorm.Open` is called in in `NewMockConnectionFactory` (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/connection.go#L66) which in turn is called in many test cases.

miguelsorianod:
`gorm.Open is called in in NewMockConnectionFactory (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/connection.go#L66) which in turn is called in many test cases.

> `gorm.Open` is called in in `NewMockConnectionFactory` (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/connection.go#L66) which in turn is called in many test cases.

My understanding is that the issue happens with integration tests, and in there `NewMockConnectionFactory` is not used. For unit tests there is no real database used.

JameelB:
> `gorm.Open` is called in in `NewMockConnectionFactory` (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/connection.go#L66) which in turn is called in many test cases.

I think this is only used for unit tests which had no issues, unit tests don't make connections to the db. The leftover connections are all from integration tests.

machi1990:
> @miguelsorianod yes, it seems like that is happening. The MaxOpenConnection setting is set to 50 at the moment but when I look at the connections in the db, I can see ~100 (using `select count(*) from pg_stat_activity;`)
> 
> Edit:
> 
> > I assume setting the dbConfig.MaxOpenConnection (used in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pkg/db/connection.go#L53) to 0 didn't have effect?
> 
> @machi1990 This has no affect disappointed

Thanks for looking at this. I think then a followup to this PR will be to configure other connections pools parameters especially idle max time.

scripts/local_db_setup.sh
JameelB: Updated in 619c98528a5cf4ce9ef3958904de1692dcb90f43

.github/workflows/ci.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:11
+        image: postgres:13
```
JameelB, line 63: Any objections to this? This is the version we're currently using for our local db setup: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/scripts/local_db_setup.sh#L21

machi1990: +1, let's go with the change.

machi1990 (approved):
lgtm, thanks @JameelB for this.

.github/workflows/ci.yaml
miguelsorianod: +1

.github/workflows/ci.yaml
```diff
@@ -103,6 +104,14 @@ jobs:
           filters: |
             openapi:
               - 'openapi/*.yaml'
+      - name: Show initial max_connections value
```
miguelsorianod, line 107: We should cleanup this before merging the PR right?

miguelsorianod (approved):
Looks good to me assuming the verification commit will be removed before the merge and that the Dockerfile changes have been tested

.github/workflows/ci.yaml
JameelB: yep i'll remove this.

db_setup_docker.sql
```diff
@@ -1,3 +1,4 @@
 CREATE USER postgres with password 'postgres';
 GRANT ALL PRIVILEGES ON DATABASE serviceapitests TO postgres;
 GRANT ALL PRIVILEGES ON DATABASE serviceapitests TO kas_fleet_manager;
+ALTER SYSTEM SET max_connections=200;
```
machi1990, line 4: nice.

machi1990, line 4:
```suggestion
ALTER SYSTEM SET max_connections=200;

```
adding in the new line

scripts/local_db_setup.sh
machi1990, line 22:
```suggestion
  postgres -N 200
  
```

JameelB: @machi1990  `CMD ["postgres", "-N", "200"]` didn't work when i tried it locally :disappointed: It started the database but didn't apply the max connection configuration. 

Instead, I added a sql command to update this in the db setup sql script which is used in both Dockerfile_template and Dockerfile_template_mocked. 

See cd6deb1d40a7ff9148fd55f35e635dd1dd577030

machi1990: Thanks @JameelB 

I am happy with the alternative solution. It's even better than the duplication in both docker template files

JameelB: Max connections config of db in pr check container: 
image

machi1990 (approved):
Thanks for the great work @JameelB

db_setup_docker.sql
miguelsorianod: Could you explai why has this been added?

machi1990: > Could you explain why has this been added?

Adding a newline at the end of the file or the addition of the SQL statement?

miguelsorianod: the addition of the sql statement

machi1990: Ah right. This comment https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1279#discussion_r961580190 from @JameelB  explains it.

miguelsorianod: How was the test made? I tried it locally without this change and this is what I observe:
```
msoriano@localhost:~/go/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager (main)$ git diff
diff --git a/db_setup_docker.sql b/db_setup_docker.sql
index c3fd21ff..f1d4c905 100644
--- a/db_setup_docker.sql
+++ b/db_setup_docker.sql
@@ -1,4 +1,3 @@
 CREATE USER postgres with password 'postgres';
 GRANT ALL PRIVILEGES ON DATABASE serviceapitests TO postgres;
 GRANT ALL PRIVILEGES ON DATABASE serviceapitests TO kas_fleet_manager;
-ALTER SYSTEM SET max_connections=200;

msoriano@localhost:~/go/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager (main)$ make db/setup
./scripts/local_db_setup.sh
bb89ade25fd27f66652dd58daf73f2b47707f1ad37327883250e07682ff30bfc
54a43ca2e401424c052fa9f92b7559e9a69844d60983d9238474033f75123571

msoriano@localhost:~/go/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager (main)$ make db/migrate
go run ./cmd/kas-fleet-manager migrate
I0902 16:13:50.213081  341164 environment.go:58] Environment variable "OCM_ENV" not specified, using default "development"
I0902 16:13:50.214918  341164 cmd.go:19] Migration starting
I0902 16:13:50.947170  341164 cmd.go:22] Database has 62 migrations applied

msoriano@localhost:~/go/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager (main)$ docker exec -it  kas-fleet-manager-db sh
# psql serviceapitests kas_fleet_manager
psql (13.7 (Debian 13.7-1.pgdg110+1))
Type "help" for help.

serviceapitests=# show max_connections;
 max_connections 
-----------------
 200
(1 row)
```

cc @JameelB

machi1990: For the local development/test database - the maximum connection is controlled by https://github.com/JameelB/kas-fleet-manager/blob/f42a96e031bc0d42f06890b0f4f9452ea49789aa/scripts/local_db_setup.sh#L22

The SQL statement is what controls the database used in the midstream PR checks, and nightly run. The file is used as an an init-db file in the dockerfiles used to setup the environment to run the integration tests: 
1. https://github.com/JameelB/kas-fleet-manager/blob/f42a96e031bc0d42f06890b0f4f9452ea49789aa/docker/Dockerfile_template_mocked#L44
2. https://github.com/JameelB/kas-fleet-manager/blob/f42a96e031bc0d42f06890b0f4f9452ea49789aa/docker/Dockerfile_template#L53

JameelB: This is for the [pr_check.sh](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/pr_check.sh) script used by the midstream prs. The make target uses a different script, [local_db_setup.sh](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/scripts/local_db_setup.sh)

You'll need to run `OCM_ENV=integration ./pr_check.sh` with mas sso stage credentials specified in the env vars. It should bring up a docker container. The screenshot I took from the comment above was from the docker container.

machi1990: To test the midstream CI dockerfiles:
change 
- https://github.com/JameelB/kas-fleet-manager/blob/f42a96e031bc0d42f06890b0f4f9452ea49789aa/docker/Dockerfile_template#L1 and use `postgres` instead of `quay.io/app-sre/postgres`
- and make sure you've the mas-sso environment variables set 
- then run the `pr_check.sh` file

Once the container is running, you can run `psql` against it to observe the max_connections set using this added SQL statement

miguelsorianod: Thanks I will try it :+1:

miguelsorianod: An additional/side comment:

In the dockerfile `quay.io/app-sre/postgres` is used. I understand that means that it uses the `latest` tag. I also understand that that image is a mirror of https://hub.docker.com/_/postgres/. The `latest` tag is currently using PostgreSQL 14.
You can verify this by running:
```sh
msoriano@localhost:~/go/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager (main)$ docker pull docker.io/postgres
Using default tag: latest
latest: Pulling from library/postgres
...

msoriano@localhost:~/go/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager (main)$ docker run -it --rm postgres --version
postgres (PostgreSQL) 14.5 (Debian 14.5-1.pgdg110+1)
```

We should update the Dockerfiles to use postgresql 13 there too to be consistent. I am assuming 13 has been chosen due to the db we are using for stage and prod is postgresql 13 (let me know if that's not accurate)

machi1990: > In the dockerfile quay.io/app-sre/postgres is used. I understand that means that it uses the latest tag. I also understand that that image is a mirror of https://hub.docker.com/_/postgres/. The latest tag is currently using PostgreSQL 14.

Yes. This is correct. We should probably use a specific tag and not a floating one. Given that the quay repo mirrors the postgres upstream one, we should be good to use whichever tag that's in upstream.

> We should update the Dockerfiles to use postgresql 13 there too to be consistent. 

+1

> I am assuming 13 has been chosen due to the db we are using for stage and prod is postgresql 13 (let me know if that's not accurate)

If we want to be exact, 12.7 is the one we are using in our envs.

miguelsorianod: Regarding my previous comment I've been looking at stage/prod and if I am understanding correctly the configurations it seems we are using PostgreSQL 12.x. In that case my previous comment applies to PostgreSQL 12 but we should also change it in the other places of the repository.

cc @machi1990 do you know what version is used for stage/prod?

miguelsorianod: I found why `-N 200` didn't "work" @JameelB @machi1990:

Part of the Dockerfile build process copies the file `pr_check_docker.sh` from the project root into the `/docker-entrypoint-initdb.d/` directory of the container. Contents placed in that directory are automatically executed when the image is run.
If you pay attention to that file some of the steps are stopping the postgresql server and starting it again with some hardcoded flags, which basically ignores all provided flags.

Setting the max connections in the db_setup_docker.sql script is also a fortunate coincidence. Setting it does not change the connections until postgresql server is restarted. Due to the `pr_check_docker.sh` script stops and starts the server the change is made effective. Setting it through that way also does not make it 100% permanent. What it does is it stores the setting in `postgresql.auto.conf` which is different than the permanent `postgresql.conf` although it's also read aside from that. If someone were to perform an alter system reset or set the parameter to default that would be lost. For more detail on behavior of alter system for postgresql see: https://www.postgresql.org/docs/current/sql-altersystem.html.

JameelB: yes, the max_connection is a config that needs to be set at start/otherwise postgres needs to restart which luckily the script was already doing :) 

I think it should be fine to have it in the auto.conf, it's read last i believe and takes precendence over postgresql.conf. There also shouldn't be anything that resets the configuration of the db and restart it again during test runs :thinking:

miguelsorianod: :+1: it's not clean but it does the work so let's keep it that way.
What we need to change though is the postgresql version used in several places

PR #1282, opened 2022-09-02
ci: do not copy the bin folder when creating the docker image for CI run

machi1990:
## Description

During local testing of the CI containers files, it was observed that sometimes the pr_check.sh run fails because of existing local bin files being copied over to the CI container.
This ensure that the copied local bin files are deleted and let our make setups download the needed dependencies.
 
## Verification Steps

Verifying it requires a bit of a dance. 
@JameelB has the setup already in place due to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1279 and will perform the verification steps that the CI container run as expected.

- [ ] Code Review completed

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,47 @@
+# See https://docs.docker.com/engine/reference/builder/#dockerignore-file for more about ignoring files.
+
+
+# Ignore bundler config.
+/.bundle
+
+# Ignore all logfiles and tempfiles.
+/log/*
+/tmp/*
+!/log/.keep
+!/tmp/.keep
+
+/coverage
+
+# Ignore built binaries
+/kas-fleet-manager
+
+# Ignore generated templates
+/templates/*-template.json
+
+# Ignore editor config
+.vscode
+
+# Ignore IntelliJ config
+.idea/
+
+# cluster details (used for testing)
+/internal/kafka/test/integration/test_cluster.json
+
+# integration test's results
+/data/results/*
+!/data/results/.keep
+
+# unit test result files
+cover*.out
+
+# local bin directory
+bin
+
+# Mac DS_Store files
+**/.DS_Store
+
+# Ignore vendor
+vendor
+
+#reports dir
+reports/
--- a/docker/Dockerfile_template
+++ b/docker/Dockerfile_template
@@ -68,8 +68,6 @@ ENV CGO_ENABLED=0
 # copy over the kas-fleet-manager
 RUN mkdir -p /kas-fleet-manager/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager
 COPY . /kas-fleet-manager/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager
-# remove the copied bin folder to ensure that we always download bin files that matches the platform architecture
-RUN rm -rf /kas-fleet-manager/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/bin
 
 # create credential files
 WORKDIR /kas-fleet-manager/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager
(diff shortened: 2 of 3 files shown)


miguelsorianod:
Hi,

If we never want to include the `bin` directory when creating the image a better alternative is to use a `.dockerignore` file. See https://docs.docker.com/engine/reference/builder/#dockerignore-file.

Can we use that instead?

miguelsorianod (changes requested):
See previous comment

machi1990:
> Hi,
> 
> If we never want to include the `bin` directory when creating the mentioned image(s) a better alternative is to use a `.dockerignore` file. See https://docs.docker.com/engine/reference/builder/#dockerignore-file.
> 
> Can we use that instead?

Thanks. Yes, that's what @JameelB suggested as well. I pushed a second commit that contains the change, plus adding in some additional files that can be ignored as well.

miguelsorianod:
Thanks :+1: .

Do we want to use the `.gitignore` for all dockerfiles including the ones in the main directory? I also see it has the same content as `.gitignore`. Have we verified that we want to exclude the same as `.gitignore`.
It is also important that it looks for the file in the root of the context:
```
Before the docker CLI sends the context to the docker daemon, it looks for a file named .dockerignore in the root directory of the context.
```

So it totally depends on how docker build is called. We should make sure that the `.dockerignore` file is applied in all cases: all dockerfiles that we use, from all our environments (local, nightly tests, perftests?, ...)

machi1990:
> Thanks +1 .
> 
> Do we want to use the `.gitignore` for all dockerfiles including the ones in the main directory? I also see it has the same content as `.gitignore`. Have we verified that we want to exclude the same as `.gitignore`. 

The content are not exactly the same, the gitignore file has more contents than the newly created dockerignor file. 
All the content in the dockerignore files can be ignored without issue. 


It is also important that it looks for the file in the root of the context as per the doc:
> 
> ```
> Before the docker CLI sends the context to the docker daemon, it looks for a file named .dockerignore in the root directory of the context.
> ```
> 
> So it totally depends on how docker build is called. We should make sure that the `.dockerignore` file is applied in all cases: all dockerfiles that we use, from all our environments (local, nightly tests, perftests?, ...)

IIRC, Perftests containers are built different/ from a different repo nowdays. 
As for the nighty, MR checks, all images are build from the root project folder that's why the dockerignore is added in the root project folder.

miguelsorianod:
> > Thanks +1 .
> > Do we want to use the `.gitignore` for all dockerfiles including the ones in the main directory? I also see it has the same content as `.gitignore`. Have we verified that we want to exclude the same as `.gitignore`.
> 
> The content are not exactly the same, the gitignore file has more contents than the newly created dockerignor file. All the content in the dockerignore files can be ignored without issue.
> 
> It is also important that it looks for the file in the root of the context as per the doc:
> 
> > ```
> > Before the docker CLI sends the context to the docker daemon, it looks for a file named .dockerignore in the root directory of the context.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So it totally depends on how docker build is called. We should make sure that the `.dockerignore` file is applied in all cases: all dockerfiles that we use, from all our environments (local, nightly tests, perftests?, ...)
> 
> IIRC, Perftests containers are built different/ from a different repo nowdays. As for the nighty, MR checks, all images are build from the root project folder that's why the dockerignore is added in the root project folder.

:+1:. Just a comment about this:
> As for the nighty, MR checks, all images are build from the root project folder that's why the dockerignore is added in the root project folder.

My understanding is that it does not depend on where the images are built from, but depend on the specified context path. See https://docs.docker.com/engine/reference/commandline/build/

JameelB (approved):
I verified by running `pr_check.sh` with integration environment locally and this works. It ignored my local /bin folder and re-installed all of the required binaries.

miguelsorianod (approved):

machi1990:
Thanks @miguelsorianod and @JameelB for the review

PR #1290, opened 2022-09-06
feat: cache ams external id to org id mapping

JameelB:
## Description
The ams organisation id and external id mapping should never change. This method gets called multiple times when validating a Kafka request on POST /kafkas. We should cache the results from ams to avoid calling it multiple times. 

Related JIRA: https://issues.redhat.com/browse/MGDSTRM-8913

Investigation on reducing calls of `GetOrganisationIdFromExternalId()` was also part of this issue. No changes are made to reduce the current calls due to the following reasons:
1. It was decided previously that validators should only perform validation. It should avoid having data passed between them even if calls to retrieve certain information may be duplicated. (See discussion in [PR#988](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/988#issuecomment-1115879161))
2. Both [AssignInstanceType](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/82a4f42ff8e9280cdb30112abd8314480839fbda/internal/kafka/internal/handlers/validation.go#L64) and [ValidateBillingAccount](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/82a4f42ff8e9280cdb30112abd8314480839fbda/internal/kafka/internal/handlers/validation.go#L69) retrieve the org id. This was not changed to avoid using quota service factory within the validator. The next call to retrieve the org id will get it from the cache so it shouldn't call ams again.

## Verification Steps
1. CI checks passes

- [ ] Code Review completed

### DIFF
--- a/pkg/client/ocm/client.go
+++ b/pkg/client/ocm/client.go
@@ -3,7 +3,9 @@ package ocm
 import (
 	"fmt"
 	"net/http"
+	"time"
 
+	"github.com/patrickmn/go-cache"
 	pkgerrors "github.com/pkg/errors"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
@@ -59,6 +61,7 @@ var _ Client = &client{}
 
 type client struct {
 	connection *sdkClient.Connection
+	cache      *cache.Cache
 }
 
 type AMSClient Client
@@ -103,7 +106,10 @@ func NewOCMConnection(ocmConfig *OCMConfig, BaseUrl string) (*sdkClient.Connecti
 }
 
 func NewClient(connection *sdkClient.Connection) Client {
-	return &client{connection: connection}
+	return &client{
+		connection: connection,
+		cache:      cache.New(168*time.Hour, 1*time.Hour),
+	}
 }
 
 func (c *client) Connection() *sdkClient.Connection {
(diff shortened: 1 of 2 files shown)


pkg/client/ocm/client.go
```diff
@@ -103,7 +106,10 @@ func NewOCMConnection(ocmConfig *OCMConfig, BaseUrl string) (*sdkClient.Connecti
 }
 
 func NewClient(connection *sdkClient.Connection) Client {
-	return &client{connection: connection}
+	return &client{
+		connection: connection,
+		cache:      cache.New(15*time.Minute, 5*time.Minute),
```
JameelB, line 111: Let me know if the expiration and cleanup interval needs to be increased/decreased.

machi1990: We can increase the expiration time, to a day or an even bigger value :-D

pkg/client/ocm/client.go
```diff
@@ -129,6 +135,14 @@ func (c *client) CreateCluster(cluster *clustersmgmtv1.Cluster) (*clustersmgmtv1
 }
 
 func (c *client) GetOrganisationIdFromExternalId(externalId string) (string, error) {
+	orgId, cached := c.cache.Get(externalId)
+	if cached {
+		orgId, ok := orgId.(string)
+		if ok && orgId != "" {
+			return orgId, nil
+		}
+	}
```
machi1990, line 144: can we add a test for when a value is retrieved from cache?

pkg/client/ocm/client.go
JameelB: Since it doesn't really change we could even go for a week? :thinking: wdyt? or is that too long :joy:

machi1990: +1. No concern of going for a week.

pkg/client/ocm/client.go
JameelB: It's a bit difficult to add a unit test for this since we can't mock the ams call :disappointed: 

I can look into adding an integration test for this or if it's good enough, have a unit test that sets the cache and ensure it returns correct value without calling to ams. Otherwise, we'd need to move out the cache logic to one of the services to unit test it. wdyt?

machi1990: >  if it's good enough, have a unit test that sets the cache and ensure it returns correct value without calling to ams.

+1,  it is what I was thinking.

JameelB: added the test there :)

machi1990: Thanks

PR #1296, opened 2022-09-08
[WIP] feat: fetch all Observatorium metrics in one request

pb82:
## Description

Query all metrics from Observatorium in one request


internal/kafka/internal/environments/development.go
```diff
@@ -34,7 +34,7 @@ func NewDevelopmentEnvLoader() environments.EnvLoader {
 		"observability-red-hat-sso-token-refresher-url":   "http://localhost:8085",
 		"observability-red-hat-sso-observatorium-gateway": "https://observatorium-mst.api.stage.openshift.com",
 		"observability-red-hat-sso-tenant":                "managedkafka",
-		"observatorium-auth-type":                         "dex",
+		"observatorium-auth-type":                         "redhat",
```
pb82, line 37: Even for development we're now using a token refresher proxy talking to redhat sso, no need to default to `dex` anymore

pkg/client/observatorium/api.go
```diff
@@ -46,253 +44,192 @@ func (obs *ServiceObservatorium) GetKafkaState(name string, resourceNamespace st
 	return KafkaState, nil
 }
 
+// buildQueries takes a list of exposed metrics and a list of filters and computes the minimum number of queries
+// to run. We need to one query per label selector, but multiple metrics can share the same label selector.
```
pawelpaszki, line 48:
```suggestion
// to run. We need to run one query per label selector, but multiple metrics can share the same label selector.
```

pb82: thanks, fixed

JameelB (commented):
Looks good to me too!

internal/kafka/internal/environments/development.go
machi1990: kas-installer will base it's value off https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/2822e09e1e19355057e10e18308ac36ed627e96c/templates/service-template.yml#L245 which is still at `dex` just in case the value is not already overriden in kas-installer.

pb82: For consistency, should we then also leave this value at `dex`?

JameelB: +1 for leaving this value at dex for consistency for now.

pb82: But this is meant to be our development defaults right? We're usually not developing against KAS Installer, so `redhat` would be ok i assume?

JameelB: I think it all comes down to what we want to use for upstream :thinking: I'm assuming `redhat` auth is internal only so for external users running this project, they won't be able to use it at all.

pb82: Makes sense, I've reset it back to `dex` now

vmanley:
Looks good

internal/kafka/internal/environments/development.go
machi1990: > +1 for leaving this value at dex for consistency for now. If we think that we no longer need to use dex, then we can remove it in a separate PR

> I think it all comes down to what we want to use for upstream thinking I'm assuming redhat auth is internal only so for external users running this project, they won't be able to use it at all.

+1

JameelB (approved):
Verified locally with @pb82. All metric endpoints works as expected. Thanks for this!

PR #1305, opened 2022-09-12
fix(ocm): iterates through all the list of quota and apply filter once

machi1990:
## Description

Returning `false` in a `.Each` loop stops the iteration. This could lead to us only returning the first item from the original quota list while we are interested in applying the filter throughout all the items of the original quota list.
Instead of a return, we've to break the inner loop to only stop iterating the inner loops.

``

- [ ] Code Review completed

### DIFF
--- a/pkg/client/ocm/client.go
+++ b/pkg/client/ocm/client.go
@@ -530,6 +530,8 @@ func (qcf *QuotaCostRelatedResourceFilter) IsMatch(relatedResource *amsv1.Relate
 // Each quota cost contains information on the usage and max allowed ocm resources quota given to the specified oganization.
 //
 // relatedResourceFilters will only be applied when fetchRelatedResources is set to true.
+// When relatedResourceFilters is not specified, all the quotas are returned.
+// When relatedResourceFilters is specified, a quota is returned if one of the filters matches the related resources.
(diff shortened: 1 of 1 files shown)


pkg/client/ocm/client.go
```diff
@@ -546,15 +546,21 @@ func (c client) GetQuotaCosts(organizationID string, fetchRelatedResources, fetc
...
+	// iterates through all Quota and only return the quota whose resources matches the given filters
 	quotaCostList.Each(func(qc *amsv1.QuotaCost) bool {
 		relatedResources := qc.RelatedResources()
+		quotaMatchesFilters := false
 		for _, relatedResource := range relatedResources {
 			for _, filter := range relatedResourceFilters {
 				if filter.IsMatch(relatedResource) {
-					quotaCosts = append(quotaCosts, qc)
-					return false
+					quotaMatchesFilters = true
```
miguelsorianod, line 556: Is this correct? Shouldn't it match **all** filters?

miguelsorianod: After checking the code it seems it should add it if one of the filters match so the current code is correct.

Can we document this behavior around filters on this method documentation to avoid future confusions?

miguelsorianod: cc @JameelB to confirm

miguelsorianod (approved):

JameelB (approved):

PR #1307, opened 2022-09-12
feat: Added  generic functions to manipulate arrays

ziccardi:
## Description
This PR will add several generic functions to manipulate arrays:

- FindFirst(values, predicate) (int, K): returns the index of the value found and the value itself
- Filter(values, predicate) []K: returns a slice that contains all the values of `values` matching the predicate
- Map[K any, V any](values []K, mapper MapperFunc[K, V]) []V: Transforms a slice to another slice using the mapper function
- AnyMatch(values, predicate): returns `true` if at least one value of the slice matches the predicate
- NoneMatch(values, predicate): returns `true` if at none of the values of the slice matches the predicate
- AllMatch(values, predicate): returns `true` if all the values of the slice matches the predicate
- Reduce(values, reducer): reduces the slice to a single value applying the reducer
- Contains (values, value): returns true if the slice `values` contains the passed int value
- ForEach(values, consumer): runs the consumer on each of the value of the slice `values`

- [ ] Code Review completed


machi1990 (commented):
Though I'd have liked for us to use a library for most of the methods here, the code change in the PR lgtm. Thanks @ziccardi I'll leave the final code review to others.

Once this is merged, it might be good to port it to the fleet manager template: https://github.com/bf2fc6cc711aee1a0c2a/ffm-fleet-manager-go-template

or update the issue https://github.com/bf2fc6cc711aee1a0c2a/ffm-project/issues/27 of what has to be done

docs/utilities/arrays.md
```diff
@@ -1,67 +1,65 @@
-# Array Utilities
-This document describes how to use array utilities in `pkg/shared/utils` folder
+# Slice Utilities
```
miguelsorianod, line 1: Can we rename the file to `slices.md` too?

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,86 @@
 package arrays
 
+const NotFound = -1
+
+type PredicateFunc[K any] func(x K) bool
```
miguelsorianod, line 5: Can we document this types here?

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,86 @@
 package arrays
 
+const NotFound = -1
```
miguelsorianod, line 3: Could we use a more specific name?

Right now the usage from client side would be `arrays.NotFound`. What is that is not found? Maybe `ElementNotFound`, `SliceElementNotFound` or something similar?
Does this constant need to be exported?

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,86 @@
 package arrays
```
miguelsorianod, line 1: Maybe we can rename this package to `slices`?

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,86 @@
...
+
+type PredicateFunc[K any] func(x K) bool
+type ReducerFunc[K any, V any] func(accumulatorValue V, currentValue K) V
+type MapperFunc[K any, V any] func(x K) V
+type ConsumerFunc[K any] func(x K)
+
 // FindFirst Finds the first element into the passed in values where the predicate is true
-// WARNING: passed in values are converted to an []interface{} slice and returned value will be an interface{}
-func FindFirst(predicate func(x interface{}) bool, values ...interface{}) (int, interface{}) {
+func FindFirst[K any](values []K, predicate PredicateFunc[K]) (int, K) {
```
miguelsorianod, line 11: In Generics and Go the parametrized type is usually denoted as `T`, with the exception on when there's a key and a value (like in the case of the `Map` function). Could we update the occurrences to use `T` to be more idiomatic?

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,86 @@
...
+	res := initialValue
+	for _, val := range values {
+		val := val
+		res = reducer(res, val)
+	}
+	return res
+}
+
+// Contains - check whether the given array contains the given value
+func Contains[K comparable](values []K, s K) bool {
```
miguelsorianod, line 77: I think we should probably expand documentation around this method about how it specifically checks that, and also remark the difference with AnyMatch.
Specifically, this compares using the `==` operator. People might be wondering when to use `Contains` and when to use `AnyMatch` thinking it performs some kind of semantic comparison whereas it just uses the `==` operator.

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,86 @@
...
+		val := val
+		if predicate(val) {
+			res = append(res, val)
+		}
+	}
+	return res
+}
+
+// Map - creates a new slice where each element is the result of the mapper function to the corresponding element in the input slice
+func Map[K any, V any](values []K, mapper MapperFunc[K, V]) []V {
```
miguelsorianod, line 34: Should we add the `comparable` constraint to `K` and/or in the `MapperFunc` type?

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,86 @@
...
+		if !predicate(val) {
+			return false
+		}
+	}
+	return true
+}
+
+// Reduce - reduces an array to a single value by applying the reducer function.
+// initialValue is the starting value of the reduction
+func Reduce[K any, V any](values []K, reducer ReducerFunc[K, V], initialValue V) V {
```
miguelsorianod, line 67: Can we document here and in the doc file what are K and V in this context?

pkg/shared/utils/arrays/generic_array_utils.go
ziccardi: The idea behind exporting it was to allow the user to do something like
```go

if idx, val := arrays.Find(.....); idx == array.NotFound {
...
}
```

pkg/shared/utils/arrays/generic_array_utils.go
ziccardi: No. The mapper function receives an object and returns another object. The received object does not need to be comparable.

docs/utilities/arrays.md
ziccardi: I will rename it in another PR since that will require changing all the references.

pkg/shared/utils/arrays/generic_array_utils.go
ziccardi: I will rename it in another PR since that will require changing all the references.

docs/utilities/arrays.md
miguelsorianod: :+1:

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -1,12 +1,100 @@
...
+// U - the type of the produced value
+type MapperFunc[T any, U any] func(x T) U
+
+// ConsumerFunc - A function that perform some operation on the received object
+type ConsumerFunc[T any] func(x T)
+
 // FindFirst Finds the first element into the passed in values where the predicate is true
-// WARNING: passed in values are converted to an []interface{} slice and returned value will be an interface{}
-func FindFirst(predicate func(x interface{}) bool, values ...interface{}) (int, interface{}) {
+func FindFirst[T any](values []T, predicate PredicateFunc[T]) (int, T) {
```
miguelsorianod, line 22: Can we document what is returned?

docs/utilities/arrays.md
```diff
@@ -1,67 +1,65 @@
...
+This document describes how to use array utilities in `pkg/shared/utils` folder.
 
-## Generic Array
+## Simple API
 
 ### FindFirst
 ```go
-func FindFirst(predicate func(x interface{}) bool, values ...interface{}) (int, interface{})
+func FindFirst[T any](values []T, predicate PredicateFunc[T]) (index int, value K)
```
miguelsorianod, line 8: here I think it should be `value T`

docs/utilities/arrays.md
```diff
@@ -1,67 +1,65 @@
...
+func AllMatch[T any](values []T, predicate PredicateFunc[T]) bool
 ```
-would become:
+This function checks whether `predicate(x)` is `true` for all the element `x` of the passed-in slice.
+
+### Map
+```go
+func Map[T any, U any](values []T, mapper MapperFunc[T, U]) []U
+```
+This function transforms all the elements of a slice of `K`s to a slice of `V`s by applying the `mapper` function to each element.
```
miguelsorianod, line 47: Here names should be updated to T, U too

miguelsorianod (approved):
Looks good to me. Great work :+1:

PR #1309, opened 2022-09-15
Adds common predicates to the utility package

ziccardi:
This PR is build on top of  #1307 

## Description
Adds common predicates to the utility package:

- IsNotNilPredicate
- IsNilPredicate
- StringNotEmptyPredicate
- StringEmptyPredicate
- CompositePredicateAll: this predicate returns `true` if all its sub predicates returns true
- CompositePredicateAny: this predicate returns `true` if any of its sub predicates returns true

- [ ] Code Review completed


pkg/shared/utils/arrays/common_predicates.go
```diff
@@ -0,0 +1,57 @@
+package arrays
+
+import "reflect"
+
+// IsNotNilPredicate - returns `true` if the value is not nil
+func IsNotNilPredicate[K any](x K) bool {
```
miguelsorianod, line 6: What would be the difference if we added a `comparable` constraint to `K`?

ziccardi: You would restrict the function to comparable objects (an object that implements `==` and `!=`), but you wouldn't be able to use it anyway to check for `nil`.
Adding the `comparable` and checking for nil with `return x != nil`, we get an error at runtime:
```go
invalid operation: x != nil (mismatched types K and untyped nil)
```

miguelsorianod: :+1:

miguelsorianod: Maybe if `K` is set to nil that would work?

ziccardi: As discussed in chat: we have no warranty that K will be a pointer, so we can't assign `nil`.
We can't even use the `zero` value like this:
```go
var y K
return x == y
```
Because if K is not a pointer, the zero value won't be `nil` (for example, if K would be `string`, its zero value` would be the `empty string`)

miguelsorianod (approved):
:+1:

PR #1310, opened 2022-09-16
feat: makefile targets to setup empty secrets

pb82:
## Description

Makefile target to create empty default secret files. If the secrets already exist nothing gets overwritten.

## Verification Steps

1. run `make setup/secrets/empty`
2. Non existing secret files should get created with no content

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -692,6 +692,30 @@ deploy/db:
 	@time timeout --foreground 3m bash -c "until $(OC) get pods -n $(NAMESPACE) | grep kas-fleet-manager-db | grep -v deploy | grep -q Running; do echo 'database is not ready yet'; sleep 10; done"
 .PHONY: deploy/db
 
+# Touch all the necessary files for fleet manager to start
+.PHONY: setup/secrets/empty
+setup/secrets/empty:
+	touch secrets/ocm-service.clientId
+	touch secrets/ocm-service.clientSecret
+	touch secrets/aws.accountid
+	touch secrets/aws.accesskey
+	touch secrets/aws.secretaccesskey
+	touch secrets/aws.route53accesskey
+	touch secrets/aws.route53secretaccesskey
+	touch secrets/keycloak-service.clientId
+	touch secrets/keycloak-service.clientSecret
+	touch secrets/osd-idp-keycloak-service.clientId
+	touch secrets/osd-idp-keycloak-service.clientSecret
+	touch secrets/sentry.key
+	touch secrets/rhsso-logs.clientId
+	touch secrets/rhsso-logs.clientSecret
+	touch secrets/rhsso-metrics.clientId
+	touch secrets/rhsso-metrics.clientSecret
+	touch secrets/observability-config-access.token
+	touch secrets/kafka-tls.crt
+	touch secrets/kafka-tls.key
+	touch secrets/image-pull.dockerconfigjson
+
 # deploys the secrets required by the service to an OpenShift cluster
 deploy/secrets:
 	@$(OC) get service/kas-fleet-manager-db -n $(NAMESPACE) || (echo "Database is not deployed, please run 'make deploy/db'"; exit 1)


Makefile
```diff
@@ -692,6 +692,30 @@ deploy/db:
 	@time timeout --foreground 3m bash -c "until $(OC) get pods -n $(NAMESPACE) | grep kas-fleet-manager-db | grep -v deploy | grep -q Running; do echo 'database is not ready yet'; sleep 10; done"
 .PHONY: deploy/db
 
+# Touch all the necessary files for fleet manager to start
```
miguelsorianod, line 695: Which has been the criteria to decide which secrets to include here? The ones that are needed to not cause a crash of kfm at startup with the defaults in the `development` environment?

pb82: I took the list from the go template and renamed the dinosaur ones to kafka.

miguelsorianod: Then we should check for potentially missing secrets aside from the ones added in this PR, as the go template is not 100% synced with kfm repository

Makefile
```diff
@@ -692,6 +692,30 @@ deploy/db:
...
+.PHONY: setup/secrets/empty
+setup/secrets/empty:
+	touch secrets/ocm-service.clientId
+	touch secrets/ocm-service.clientSecret
```
JameelB, line 699: I think it would be better to specify the `ocm-service.token` file here instead since this is the one that's mostly used for dev purposes.

Makefile
```diff
@@ -692,6 +692,30 @@ deploy/db:
...
+	touch secrets/aws.accountid
+	touch secrets/aws.accesskey
+	touch secrets/aws.secretaccesskey
+	touch secrets/aws.route53accesskey
+	touch secrets/aws.route53secretaccesskey
```
JameelB, line 704: Not sure if these are needed. I think I remember this is only needed when flag `enable-kafka-cname-registration` is set to true.

Makefile
```diff
@@ -692,6 +692,30 @@ deploy/db:
...
+	touch secrets/keycloak-service.clientId
+	touch secrets/keycloak-service.clientSecret
+	touch secrets/osd-idp-keycloak-service.clientId
+	touch secrets/osd-idp-keycloak-service.clientSecret
+	touch secrets/sentry.key
```
JameelB, line 709: not required. By default, sentry is disabled.

Makefile
```diff
@@ -692,6 +692,30 @@ deploy/db:
...
+	touch secrets/rhsso-logs.clientId
+	touch secrets/rhsso-logs.clientSecret
+	touch secrets/rhsso-metrics.clientId
+	touch secrets/rhsso-metrics.clientSecret
+	touch secrets/observability-config-access.token
+	touch secrets/kafka-tls.crt
+	touch secrets/kafka-tls.key
```
JameelB, line 716: Not required, only needed when `enable-kafka-external-certificate` is set to true.

Makefile
```diff
@@ -692,6 +692,30 @@ deploy/db:
...
+	touch secrets/aws.route53secretaccesskey
+	touch secrets/keycloak-service.clientId
+	touch secrets/keycloak-service.clientSecret
+	touch secrets/osd-idp-keycloak-service.clientId
+	touch secrets/osd-idp-keycloak-service.clientSecret
+	touch secrets/sentry.key
+	touch secrets/rhsso-logs.clientId
+	touch secrets/rhsso-logs.clientSecret
+	touch secrets/rhsso-metrics.clientId
+	touch secrets/rhsso-metrics.clientSecret
```
JameelB, line 713: Not required, by default dev env uses `dex` auth for observability stack.

machi1990:
@pb82 @JameelB @miguelsorianod bumping the PR in case it has been forgotten

pb82:
thanks for the reminder @machi1990 and thanks for the review @JameelB 
I'll update the PR

machi1990:
Superseded by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1532

PR #1320, opened 2022-09-19
feat(prewarming): add reserved instances metrics

machi1990:
## Description

Adds reserved instances metrics. The details of the implementation are available in the below linked JIRA.
Closes https://issues.redhat.com/browse/MGDSTRM-9679

## Verification Steps

The added unit & integration tests should pass.
 
- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/data_plane_kafka.go
+++ b/internal/kafka/internal/services/data_plane_kafka.go
@@ -18,24 +18,39 @@ import (
 	"github.com/pkg/errors"
 )
 
-type kafkaStatus string
+type managedKafkaStatus string
+
+func (ks managedKafkaStatus) String() string {
+	return string(ks)
+}
+
+type managedKafkaDeploymentType string
+
+func (dt managedKafkaDeploymentType) String() string {
+	return string(dt)
+}
+
+type totalCountPerReservedManagedKafkaDeploymentStatus map[managedKafkaStatus]int
+type reservedManagedKafkaStatusCountPerInstanceType map[api.ClusterInstanceTypeSupport]totalCountPerReservedManagedKafkaDeploymentStatus
 
 const (
-	statusInstalling          kafkaStatus = "installing"
-	statusReady               kafkaStatus = "ready"
-	statusError               kafkaStatus = "error"
-	statusRejected            kafkaStatus = "rejected"
-	statusRejectedClusterFull kafkaStatus = "rejectedClusterFull"
-	statusDeleted             kafkaStatus = "deleted"
-	statusUnknown             kafkaStatus = "unknown"
-	strimziUpdating           string      = "StrimziUpdating"
-	kafkaUpdating             string      = "KafkaUpdating"
-	kafkaIBPUpdating          string      = "KafkaIbpUpdating"
+	statusInstalling          managedKafkaStatus         = "installing"
+	statusReady               managedKafkaStatus         = "ready"
+	statusError               managedKafkaStatus         = "error"
+	statusRejected            managedKafkaStatus         = "rejected"
+	statusRejectedClusterFull managedKafkaStatus         = "rejectedClusterFull"
+	statusDeleted             managedKafkaStatus         = "deleted"
+	statusUnknown             managedKafkaStatus         = "unknown"
+	strimziUpdating           string                     = "StrimziUpdating"
+	kafkaUpdating             string                     = "KafkaUpdating"
+	kafkaIBPUpdating          string                     = "KafkaIbpUpdating"
+	reservedDeploymentType    managedKafkaDeploymentType = "reserved"
+	realDeploymentType        managedKafkaDeploymentType = "real"
 )
 
 //go:generate moq -out data_plane_kafka_service_moq.go . DataPlaneKafkaService
 type DataPlaneKafkaService interface {
-	UpdateDataPlaneKafkaService(ctx context.Context, clusterId string, status []*dbapi.DataPlaneKafkaStatus) *serviceError.ServiceError
+	UpdateDataPlaneKafkaService(ctx context.Context, clusterID string, status []*dbapi.DataPlaneKafkaStatus) *serviceError.ServiceError
 }
 
 type dataPlaneKafkaService struct {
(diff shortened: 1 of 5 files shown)


internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -20,22 +20,34 @@ import (
 
 type kafkaStatus string
 
+func (ks kafkaStatus) String() string {
+	return string(ks)
+}
+
+type deploymentType string
```
miguelsorianod, line 27: The name seems too generic: deployment of what? can we be more specific? for example, `ManagedKafkaDeploymentType`, or `DataPlaneKafkaDeploymentType`?

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -52,67 +64,126 @@ func NewDataPlaneKafkaService(kafkaSrv KafkaService, clusterSrv ClusterService,
...
+	if err != nil {
+		log.Error(err)
+	}
+
 	return nil
 }
 
+// processRealKafkaDeployment process reserved kafka instances and updates their status in cluster
+func (d *dataPlaneKafkaService) processReservedKafkaDeployment(ks *dbapi.DataPlaneKafkaStatus, prewarmingStatusInfo map[string]map[string]int, log logger.UHCLogger, clusterID string) {
+	parsedID := strings.Split(ks.KafkaClusterId, "-")
```
miguelsorianod, line 114: Isn't KafkaClusterId different than the name?

internal/kafka/internal/services/data_plane_kafka.go
machi1990: We are in the kafka context, I' say kafka deployment.

internal/kafka/internal/services/data_plane_kafka.go
machi1990: What name?

machi1990:
> Do we need to store the information in the DB? Can't we set the metric(s) when `UpdateDataPlaneKafkaService` is called?

Yes, to ensure only one pod will emit the metrics and not N of them which can cause duplicated metrics entries.

internal/kafka/internal/services/data_plane_kafka.go
miguelsorianod: The name of the ManagedKafka.

The Kafka ID does not necessarly need to be the Kafka Name (same with the Cluster ID vs the Cluster name in the context of the Cluster type)

internal/kafka/internal/services/data_plane_kafka.go
machi1990: changed to `managedKafkaDeploymentType`

miguelsorianod: I wouldn't say kafka deployment. It's certainly ManagedKafka or DataPlaneKafka but not Kafka :thinking:

internal/kafka/internal/services/data_plane_kafka.go
machi1990: Ah right. 
For reserved MK CR, the name and the id will be the same.
I always confuse the pattern and I think this method needs some reworking as the instance type will be the third element not the second.

The pattern is `reserved-kafka-<instanceType>-<number>`

miguelsorianod:
> > Do we need to store the information in the DB? Can't we set the metric(s) when `UpdateDataPlaneKafkaService` is called?
> 
> Yes, to ensure only one pod will emit the metrics and not N of them which can cause duplicated metrics entries.

I am trying to understand this. Are we dealing incorrectly with other metrics then, as we have other metrics being set in data plane kafka too?

Also, what happens when for example there is a change of leader among the pods? And when there's a redeployment of all of them, for example when there's a new release, where all pods are renewed? all their IDs will change and thus new metrics will be created too.

This also depends on how the Prometheus K8s scraping information is configured. Is it through the Service, through Pods, through Endpoints? The behavior will be different depending on that.

internal/kafka/internal/services/data_plane_kafka.go
miguelsorianod: In that case. Shouldn't the length check be 4? :thinking:

machi1990:
> > > Do we need to store the information in the DB? Can't we set the metric(s) when `UpdateDataPlaneKafkaService` is called?
> > 
> > 
> > Yes, to ensure only one pod will emit the metrics and not N of them which can cause duplicated metrics entries.
> 
> I am trying to understand this. Are we dealing incorrectly with other metrics then, as we have other metrics being set in data plane kafka too?

The other are fine since they are mostly counts. At the end of the end, the sum will appear correctly. At least, there hasn't been an issue reported on this. 

> Also, what happens when for example there is a change of leader among the pods? And when there's a redeployment of all of them, for example when there's a new release, where all pods are renewed? all their IDs will change and thus new metrics will be created too.

We always reset the metrics when a new leader pod is elected. See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/475b1633047c882d26ed83ff4267ec21ebd41cdc/pkg/metrics/metrics.go#L828 which is called from the cluster_mgr.go 

> This also depends on how the Prometheus K8s scraping information is configured. Is it through the Service, through Pods, through Endpoints? The behavior will be different depending on that.

I do not know how Prometheus is configured (the one managed by SRE) but we've seen the duplicated metrics behaviour if a gauge metric is scraped from N pods. One such occurrence was fixed in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1149 to ensure that when a leader changes, the metrics are properly reset.

pkg/api/cluster_types.go
```diff
@@ -516,3 +520,23 @@ func (cluster *Cluster) GetSupportedInstanceTypes() []string {
 func (cluster *Cluster) GetRawSupportedInstanceTypes() string {
 	return cluster.SupportedInstanceType
 }
+
+// SetPrewarmingStatusInfo sets the prewarming status info per instance type into a json object that can be persisted in
+// the database
+func (cluster *Cluster) SetPrewarmingStatusInfo(prewarmingStatusInfo map[string]map[string]int) {
```
miguelsorianod, line 526: I am thinking that we it might be worth investing some time trying to make this more typesafe compared to previous implementations of similar data like DynamicCapacityInfo or AvailableStrimziVersions:

We are exposing that prewarmingStatusInfo as a map[string]map[string]int. It is unavoidable at JSON/db level the way we are doing it (at serialization time). However, at code level I think we can be safer and more flexible:

Make map[string]map[string]int its own type. For that we have two options:
```
type prewarmingStatusIndex map[string]map[string]int // or prewarmingStatusInfoIndex
```

or
```
type prewarmingStatusIndex struct { // or prewarmingStatusInfoIndex
  index map[string]map[string]int
}  
```
I tend to prefer the later. The difference is that with the first approach we indirectly coupling the implementation to a map, whereas with the second approach that is totally hidden and thus transparent to change. With the first approach we could directly convert it to a map and use the typical Go map operators but I think that is even counterproductive as again it ties the type to map-specific semantics.

From that, we can even go further. the second map[string]int can be its own type too, and the same approach would apply to it.

Additional useful methods and potentially safer could be then implemented on top of the new types too.

miguelsorianod: We can work on this together if needed

internal/kafka/internal/services/data_plane_kafka.go
machi1990: Yes, we could change it to 4. For the check added here, I was mainly interested in making sure that the third element (representing the instance-type) is there to avoid panic.

pkg/api/cluster_types.go
machi1990: I was leaning toward the second approach for this e.g having a 
```go
type PrewarmingInfo struct { // or PrewarmingInfo
  statusCountPerInstanceType map[string]CountPerStatus
} 

type CountPerStatus map[string]int
```
where the whole PrewarmingInfo struct is marshalled/unmarshalled. 

I had a mental blocker about adding the type now as I cant see any other field that we'll like to add in `PrewarmingInfo` struct. Can you think of any?

internal/kafka/internal/services/data_plane_kafka.go
machi1990: changed to `4` as well to ensure that the format is respected and if not we've logs of why it wasn't respected.

miguelsorianod:
> > > > Do we need to store the information in the DB? Can't we set the metric(s) when `UpdateDataPlaneKafkaService` is called?
> > > 
> > > 
> > > Yes, to ensure only one pod will emit the metrics and not N of them which can cause duplicated metrics entries.
> > 
> > 
> > I am trying to understand this. Are we dealing incorrectly with other metrics then, as we have other metrics being set in data plane kafka too?
> 
> The other are fine since they are mostly counts. At the end of the end, the sum will appear correctly. At least, there hasn't been an issue reported on this.
> 
> > Also, what happens when for example there is a change of leader among the pods? And when there's a redeployment of all of them, for example when there's a new release, where all pods are renewed? all their IDs will change and thus new metrics will be created too.
> 
> We always reset the metrics when a new leader pod is elected. See
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/475b1633047c882d26ed83ff4267ec21ebd41cdc/pkg/metrics/metrics.go#L828
> which is called from the cluster_mgr.go
> 
> > This also depends on how the Prometheus K8s scraping information is configured. Is it through the Service, through Pods, through Endpoints? The behavior will be different depending on that.
> 
> I do not know how Prometheus is configured (the one managed by SRE) but we've seen the duplicated metrics behaviour if a gauge metric is scraped from N pods. One such occurrence was fixed in #1149 to ensure that when a leader changes, the metrics are properly reset.

I think I more or less understand the problematic.

I have some thoughts about this, not really specific to this PR:

I am not sure we have been approaching this properly. Maybe what is happening is that we have it backwards:

An alternative would be adding a timestamp label to the metric, in a format it can be compared and always increases, and then from PromQL side we could use the `max` statement to get the most recent metric in case of gauges. That way the number of pods processing this does not really matter, and we would not need to tie our code business logic / implementation around the metric.

We can discuss this over other channels to evolve the idea. Not necessary for this PR as I understand it is a fundamental change on how we deal with concurrency/metrics but I think it's worth thinking a little bit deeper about it.

pkg/api/cluster_types.go
miguelsorianod: PrewarmingInfoIndex

miguelsorianod: On the other hand, `CountPerStatus` seems very generic. We should be more specific if possible

machi1990:
> We can discuss this over other channels to evolve the idea. Not necessary for this PR as I understand it is a fundamental change on how we deal with concurrency/metrics but I think it's worth thinking a little bit deeper about it.

+1 sounds good to me. 

> An alternative would be adding a timestamp label to the metric, in a format it can be compared and always increases, and then from PromQL side we could use the max statement to get the most recent metric based on that label in that case. That way the number of pods processing this does not really matter, and we would not need to tie our code business logic / implementation around the metric.

I like the idea that's being floated here. Honestly, there loads of code here more than I'd like (in the PR) just to avoid duplicated metrics.

pkg/metrics/metrics.go
```diff
@@ -735,6 +749,31 @@ func UpdateDatabaseQueryDurationMetric(status string, queryType string, elapsed
 
 // #### Metrics for Database - End ####
 
+// create a new gaugeVec for the prewarming status info count per cluster_id, instance_type and status
+var prewarmingStatusInfoCountMetric = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Subsystem: KasFleetManager,
+		Name:      PrewarmingStatusInfoCount,
+		Help:      "count of reserved instance in a given status that deployed on a given cluster_id",
```
JameelB, line 757:
```suggestion
		Help:      "count of reserved kafka instances in a given status thats deployed on the given cluster_id",
```

machi1990:
> An alternative would be adding a timestamp label to the metric, in a format it can be compared and always increases, and then from PromQL side we could use the max statement to get the most recent metric based on that label in that case. That way the number of pods processing this does not really matter, and we would not need to tie our code business logic / implementation around the metric.

@miguelsorianod @JameelB I am wondering if we could use this PR as a POC for the idea as well.
It is safe to do it since the metric are not current used anywhere in any dashboard/alert. When we start using them, we could add the the info on how the metric is supposed to be queried? With this we could remove a ton of code (DB updates etc) that we are only having to maintain just cause of this limit? Any thoughtS?

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -52,67 +64,126 @@ func NewDataPlaneKafkaService(kafkaSrv KafkaService, clusterSrv ClusterService,
...
+	_, ok := prewarmingStatusInfo[instanceType]
+	if !ok {
+		log.Error(fmt.Errorf("The reserved %q ID is not supported in the cluster with cluster_id %q", ks.KafkaClusterId, clusterID))
+		return
+	}
+
+	s := getStatus(ks)
+
+	if s != statusReady && s != statusInstalling && s != statusError {
+		// consider the rest of the statuses as 'unknown' in the prewarming context as they shouldn't be sent by the Operator
```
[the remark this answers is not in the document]
machi1990: Hmm actually, it is just the these two statuses that we are ignoring.
It does not hurt to add and initialize them to `0`. If the Fleetshard starts sending them, we'll have the count anyway. Otherwise, they'll be set to `0` always something that won't hurt the metric in any way.

JameelB:
> @miguelsorianod @JameelB I am wondering if we could use this PR as a POC for the idea as well.
It is safe to do it since the metric are not current used anywhere in any dashboard/alert. When we start using them (on dashboard or alerts), we could add the the info on how the metric is supposed to be queried? With this we could remove a ton of code (DB updates etc) that we are only having to maintain just cause of this limit? Any thoughtS?

That sounds good to me. It would also save us from adding a new column to the clusters table and removing it later. Would it be worth getting thoughts from RTS team around this? They're usually the ones that consume the metrics and configure the dashboards.

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -293,6 +293,13 @@ func (c *ClusterManager) processReadyClusters() []error {
 
 	for _, readyCluster := range readyClusters {
 		glog.V(10).Infof("ready cluster ClusterID = %s", readyCluster.ClusterID)
```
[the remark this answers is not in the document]
machi1990: Let's keep this for now. Once we've an agreement on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1320#issuecomment-1252278834 I'll look to remove the added code in this file :-)

machi1990:
> > @miguelsorianod @JameelB I am wondering if we could use this PR as a POC for the idea as well.
> > It is safe to do it since the metric are not current used anywhere in any dashboard/alert. When we start using them (on dashboard or alerts), we could add the the info on how the metric is supposed to be queried? With this we could remove a ton of code (DB updates etc) that we are only having to maintain just cause of this limit? Any thoughtS?
> 
> That sounds good to me. It would also save us from adding a new column to the clusters table and removing it later. Would it be worth getting thoughts from RTS team around this? They're usually the ones that consume the metrics and configure the dashboards.

Adding in @StevenTobin to the PR's discussion to get RTS's input.

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -399,3 +469,15 @@ func buildRoutes(routesInRequest []dbapi.DataPlaneKafkaRouteRequest, kafka *dbap
...
+// getManagedKafkaDeploymentType gets the deployment type from the data plane kafka status
+// If the id of the status begins with "reserved" then the deployment type is reserved.
+// Otherwise, it is a "real" deployment
+func getManagedKafkaDeploymentType(ks *dbapi.DataPlaneKafkaStatus) managedKafkaDeploymentType {
+	parsedID := strings.Split(ks.KafkaClusterId, "-")
+	if parsedID[0] == reservedDeploymentType.String() {
+		return reservedDeploymentType
+	}
+
+	return realDeploymentType
```
ziccardi, line 482: I think this would be more readable this way
```suggestion
	if (strings.HasPrefix(ks. KafkaClusterId,  reservedDeploymentType.String())) {
	    return reservedDeploymentType
	}

	return realDeploymentType
```

machi1990: Thanks. Updated to use the prefix check

machi1990:
I've update the PR to remove handling of duplicated metric from fleet manager as it shouldn't be the responsibility of the fleet manager. A decision was taken to keep the implementation simple and let the handling of duplicated metrics be performed when the metric is consumed (in alert or dashboards).

For example, one can use the `without` operator to ignore the `pod` & `instance` label from the metrics (these labels are coming from the pod monitor) when consuming the metric in alerts or dashboards. 

If needed, the metric could be enhanced with the addition of  `timestamp` label (see the https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1320#issuecomment-1252250768 suggestion) to indicate when was the metric produced ( useful to know the latest when interpreting metrics etc).

pkg/api/cluster_types.go
machi1990: Thanks @miguelsorianod for the review. This is no longer something that we need to pursue since this logic has been removed from the PR as of https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/commit/6e4f7374f2ff2f1d48d21d0581c5879b9e712095

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -20,22 +20,37 @@ import (
 
 type kafkaStatus string
 
+func (ks kafkaStatus) String() string {
```
miguelsorianod, line 23: I know this doesn't belong to this PR but could we rename the type `kafkaStatus` to `managedKafkaStatus` to be more specific?

machi1990: Good catch! Let's do the renaming

internal/kafka/internal/services/data_plane_kafka.go
miguelsorianod: Has this been updated? I don't see it reflected

internal/kafka/internal/services/data_plane_kafka.go
machi1990: done, throughout.

internal/kafka/internal/services/data_plane_kafka.go
machi1990: let me double check, I might have forgotten to push

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -52,67 +67,128 @@ func NewDataPlaneKafkaService(kafkaSrv KafkaService, clusterSrv ClusterService,
...
+				Count:        count,
+			})
 		}
 	}
+
 	return nil
 }
 
+// processRealKafkaDeployment process reserved kafka instances and updates their status in cluster
+func (d *dataPlaneKafkaService) processReservedKafkaDeployment(ks *dbapi.DataPlaneKafkaStatus, prewarmingStatusInfo reservedManagedKafkaStatusCountPerInstanceType, log logger.UHCLogger, clusterID string) {
```
miguelsorianod, line 124: A nice side-effect of the refactor is that no DB errors will be shown anymore when processing the reserved managed kafkas :+1:

internal/kafka/internal/services/data_plane_kafka.go
machi1990: Yes. Should be good now. Can you re-check? @miguelsorianod

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -379,6 +427,35 @@ func (d *dataPlaneKafkaService) persistKafkaRoutes(kafka *dbapi.KafkaRequest, ka
 	return nil
 }
 
+func getStatus(status *dbapi.DataPlaneKafkaStatus) managedKafkaStatus {
```
miguelsorianod, line 430: Side comment about this:

I see `getStatus` and other functions in this file like `getRoutes` are at package-level. I don't see any reason why this should be that way. The smaller the scope the better. Can we make them part of the `dataplaneKafkaService` type? We can deal with that in another PR if you prefer

internal/kafka/internal/services/data_plane_kafka.go
miguelsorianod: :+1:

internal/kafka/internal/services/data_plane_kafka.go
machi1990: No reason they couldn't be part of the dataplane kafka service. Let me look at the possibility of that.
This comment also made me realise that it should `getManagedKafkaStatus` instead of `getStatus`

machi1990: Done in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1320/commits/d58b78c0b1546b973612df26366c9cc4e6313728

Thanks for the suggestion @miguelsorianod

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -379,7 +427,36 @@ func (d *dataPlaneKafkaService) persistKafkaRoutes(kafka *dbapi.KafkaRequest, ka
...
+			}
+			if strings.EqualFold(c.Reason, "Error") {
+				return statusError
+			}
+			if strings.EqualFold(c.Reason, "Rejected") {
+				if strings.EqualFold(c.Message, "Cluster has insufficient resources") {
+					return statusRejectedClusterFull
+				}
+				return statusRejected
+			}
```
ziccardi, line 453:
```suggestion
			switch(true) {
				case strings.EqualFold(c.Reason, "Deleted"):
					return statusDeleted
				case strings.EqualFold(c.Reason, "Error"):
					return statusError
				case strings.EqualFold(c.Reason, "Rejected"):
					if strings.EqualFold(c.Message, "Cluster has insufficient resources") {
					return statusRejectedClusterFull
					  }
					return statusRejected
			}
```

machi1990: Thanks for the suggestion. I think I'll keep the `if` statement.

pkg/metrics/metrics.go
```diff
@@ -735,6 +749,40 @@ func UpdateDatabaseQueryDurationMetric(status string, queryType string, elapsed
...
 // #### Metrics for Database - End ####
 
+// create a new gaugeVec for the prewarming status info count per cluster_id, instance_type and status.
+// The metric contains the timestamp (in milliseconds) information indicating the time the metric was reported.
+// The value of the timestamp is an increasing value.
+var prewarmingStatusInfoCountMetric = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Subsystem: KasFleetManager,
+		Name:      PrewarmingStatusInfoCount,
+		Help:      "count of reserved kafka instances in a given status thats deployed on the given cluster_id. The metric contains the timestamp (in milliseconds) information indicating the time the metric was reported.",
```
miguelsorianod, line 759: `thats` can be removed

miguelsorianod: The part about timestamp doesn't apply right now no?

machi1990: Good catch! The timestamp part doesn't apply. I was in middway of adding it but then decided to keep it for later phases implementation. I'll remove it from the description.

machi1990: Done

miguelsorianod (approved):
looks good to me :+1: great work

PR #1321, opened 2022-09-20
feat: expose the fleet manager version

valeriiashapoval:
## Description
(https://issues.redhat.com/browse/MGDSTRM-9585)

## Verification Steps
1. run ` make run ` from the root
2. run ` curl -H "Authorization: Bearer $(ocm token)" http://localhost:8000/api/kafkas_mgmt/v1 | jq . `

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/test/integration/metadata_test.go
+++ b/internal/kafka/test/integration/metadata_test.go
@@ -1,0 +1,40 @@
+package integration
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/test"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/test/mocks"
+	"github.com/onsi/gomega"
+)
+
+func TestKafkaCreate_Metadata(t *testing.T) {
+	g := gomega.NewWithT(t)
+
+	// setup ocm server
+	ocmServerBuilder := mocks.NewMockConfigurableServerBuilder()
+	ocmServer := ocmServerBuilder.Build()
+	defer ocmServer.Close()
+
+	// start servers
+	h, client, teardown := test.NewKafkaHelper(t, ocmServer)
+	defer teardown()
+
+	account := h.NewRandAccount()
+	ctx := h.NewAuthenticatedContext(account, nil)
+
+	version, resp, err := client.DefaultApi.GetVersionMetadata(ctx)
+	if resp != nil {
+		resp.Body.Close()
+	}
+
+	g.Expect(err).NotTo(gomega.HaveOccurred(), "g.Unable to retrieve version metadata:  %v", err)
+	g.Expect(resp.StatusCode).To(gomega.Equal(http.StatusOK))
+	g.Expect(version.Href).To(gomega.ContainSubstring("/v1"))
+	g.Expect(version.Collections).To(gomega.HaveLen(3))
+
+	/*  Unable to test for a specific version as for now due to the debug package issue: https://github.com/golang/go/issues/33976,
+	TODO when the issue is resolved.
+	*/
+}
--- a/openapi/kas-fleet-manager.yaml
+++ b/openapi/kas-fleet-manager.yaml
@@ -1066,22 +1066,25 @@ components:
                   - $ref: "#/components/schemas/KafkaRequest"
     VersionMetadata:
       allOf:
-        - $ref: "#/components/schemas/ObjectReference"
-        - type: object
-          example:
-            kind: "APIVersion"
-            id: "v1"
-            href: "/api/kafkas_mgmt/v1"
-            collections:
-              - id: "kafkas"
-                href: "/api/kafkas_mgmt/v1/kafkas"
-                kind: "KafkaList"
-          properties:
-            collections:
-              type: array
-              items:
-                allOf:
-                  - $ref: "#/components/schemas/ObjectReference"
+      - $ref: "#/components/schemas/ObjectReference"
+      - type: object
+        example:
+          kind: "APIVersion"
+          id: "v1"
+          href: "/api/kafkas_mgmt/v1"
+          server_version: "24a263e8631d713b3104c1a70c143644ab91de6f"
+          collections:
+            - id: "kafkas"
+              href: "/api/kafkas_mgmt/v1/kafkas"
+              kind: "KafkaList"
+        properties:
+          server_version:
+            type: string
+          collections:
+            type: array
+            items:
+              allOf:
+              - $ref: "#/components/schemas/ObjectReference"
     KafkaRequestPayload:
       description: Schema for the request body sent to /kafkas POST
       required:
--- a/pkg/api/metadata.go
+++ b/pkg/api/metadata.go
@@ -22,10 +22,29 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"runtime/debug"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/getsentry/sentry-go"
 	"github.com/golang/glog"
 )
+
+// getCommitSha() returns the current version (vcs.revision attribute of BuildInfo struct) using a built-in debug package and its ReadBuildInfo() function.
+// Upon error, an empty string is returned along with a GeneralError.
+// Empty values returned when vsc.revision is empty/unavailable.
+func getCommitSha() (string, error) {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return "", errors.GeneralError("failed to retrieve commitSha")
+	}
+	for _, i := range info.Settings {
+		if i.Key == "vcs.revision" {
+			return i.Value, nil
+		}
+	}
+	return "", nil
+}
 
 // ServeHTTP sends API documentation response.
 func (m *Metadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
@@ -60,7 +79,7 @@ func (m *Metadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Send the response:
 	_, err = w.Write(data)
 	if err != nil {
-		err = fmt.Errorf("Can't send response body for request '%s'", r.URL.Path)
+		err = fmt.Errorf("can't send response body for request '%s'", r.URL.Path)
 		glog.Error(err)
 		sentry.CaptureException(err)
 		return
@@ -69,14 +88,22 @@ func (m *Metadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // ServeHTTP sends API version v1 documentation response.
 func (v *VersionMetadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Get the version:
+	version, err := getCommitSha()
+	if err != nil {
+		shared.HandleError(r, w, err.(*errors.ServiceError))
+		return
+	}
+
 	// Set the content type:
 	w.Header().Set("Content-Type", "application/json")
 
 	// Prepare the body:
 	body := VersionMetadata{
-		ID:   v.ID,
-		Kind: "APIVersion",
-		HREF: r.URL.Path,
+		ID:            v.ID,
+		Kind:          "APIVersion",
+		HREF:          r.URL.Path,
+		ServerVersion: version,
 	}
 	for _, c := range v.Collections {
 		href := c.HREF
@@ -99,7 +126,7 @@ func (v *VersionMetadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Send the response:
 	_, err = w.Write(data)
 	if err != nil {
-		glog.Errorf("Can't send response body for request '%s'", r.URL.Path)
+		glog.Errorf("can't send response body for request '%s'", r.URL.Path)
 		return
 	}
 }
--- a/internal/connector/test/integration/features/connector-api.feature
+++ b/internal/connector/test/integration/features/connector-api.feature
@@ -1582,7 +1582,8 @@ Feature: create a connector
             "collections": null,
             "href": "/api/connector_mgmt/v1",
             "id": "v1",
-            "kind": "APIVersion"
+            "kind": "APIVersion",
+            "server_version": ""
           }
         ]
       }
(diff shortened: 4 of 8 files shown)


pkg/api/metadata.go
```diff
@@ -22,11 +22,27 @@ import (
...
 
+func GetCommitSha() (string, error) {
+	var commitSha string
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return commitSha, errors.GeneralError("Failed to retrieve commitSha")
+	}
+	for _, i := range info.Settings {
+		if i.Key == "vcs.revision" {
+			commitSha = i.Value
```
machi1990, line 40:
```suggestion
			return i.Value, nil
```

pkg/api/metadata.go
```diff
@@ -22,11 +22,27 @@ import (
...
+		}
+	}
+	return commitSha, nil
```
machi1990, line 43:
```suggestion
	return "", nil
```

pkg/api/metadata.go
```diff
@@ -22,11 +22,27 @@ import (
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/getsentry/sentry-go"
 	"github.com/golang/glog"
 )
 
+func GetCommitSha() (string, error) {
+	var commitSha string
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return commitSha, errors.GeneralError("Failed to retrieve commitSha")
```
machi1990, line 36:
```suggestion
		return "", errors.GeneralError("Failed to retrieve commitSha")
```

pkg/api/metadata.go
```diff
@@ -22,11 +22,27 @@ import (
...
 	"net/http"
+	"runtime/debug"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/getsentry/sentry-go"
 	"github.com/golang/glog"
 )
 
+func GetCommitSha() (string, error) {
+	var commitSha string
```
machi1990, line 33:
```suggestion
```

pkg/api/metadata.go
```diff
@@ -69,14 +85,20 @@ func (m *Metadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // ServeHTTP sends API version v1 documentation response.
 func (v *VersionMetadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	version, ok := GetCommitSha()
+	if ok != nil {
+		version = ok.Error()
+	}
```
machi1990, line 91: Maybe we could send the error instead as a `500` http code error?

machi1990 (commented):
thanks @valeriiashapoval for this.

I've added some suggestions. Let me know what you think?

can we also add an integration test?

pkg/api/metadata.go
```diff
@@ -22,11 +22,27 @@ import (
...
 	"fmt"
 	"net/http"
+	"runtime/debug"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/getsentry/sentry-go"
 	"github.com/golang/glog"
 )
 
+func GetCommitSha() (string, error) {
```
ziccardi, line 32: Do we need to export this? If we use it only inside the package, we can rename it to `getCommitSha`.

pkg/api/metadata.go
ziccardi, line 91: Usually, `ok` is a boolean. Here it is of type `error`, so I believe we should rename it to `err`.

pb82: Should we rename the `ok` return to `err` here? That's the default name for an error returned from a function. `ok` makes it sounds like it's a boolean.

pkg/api/metadata.go
valeriiashapoval: Thanks @machi1990 ! Just committed the suggested changes.

pkg/api/metadata.go
valeriiashapoval: Thanks @pb82 @machi1990. Applied the changes!

pkg/api/metadata.go
valeriiashapoval: Thanks @ziccardi. Just made the change.

miguelsorianod:
The changes seem to have an impact on the connectors code. cc @valdar @dhirajsb

pkg/api/metadata.go
machi1990: > what do you think about returning 500 from the endpoint if fetching the version failed?

The 500 response is returned as an error in json format so it can parsed. 
Concretely, my suggestion is when an error occurs if to call
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ce9749f3a293242d72def0349a8e19d67fa65672/pkg/shared/handle_error.go#L11
and return

So the code here will look like

```go
if err != nil {
 shared.HandleError(r,w,err)	
 return
}
```

@pb82 @valeriiashapoval

machi1990 (changes requested):
Please see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1321#discussion_r976343581

pkg/api/metadata.go
```diff
@@ -69,14 +85,21 @@ func (m *Metadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
 	// Prepare the body:
 	body := VersionMetadata{
-		ID:   v.ID,
-		Kind: "APIVersion",
-		HREF: r.URL.Path,
+		ID:            v.ID,
+		Kind:          "APIVersion",
+		HREF:          r.URL.Path,
+		ServerVersion: version,
```
ziccardi, line 102: According to the `getCommitSha` implementation, `version` here could be an empty string. Is that correct?

pkg/api/metadata.go
```diff
@@ -22,11 +22,27 @@ import (
...
 	"net/http"
+	"runtime/debug"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/getsentry/sentry-go"
 	"github.com/golang/glog"
 )
 
+func getCommitSha() (string, error) {
```
ziccardi, line 33: Mind adding a comment explaining what this command does? (which commit sha it returns,  when empty string is returned)

pkg/api/metadata.go
```diff
@@ -22,11 +22,27 @@ import (
...
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return "", errors.GeneralError("Failed to retrieve commitSha")
```
ziccardi, line 36: Errors messages should start with lowercase

pkg/api/metadata.go
ziccardi: @machi1990 WDYT?

miguelsorianod: That's a good question.
It should never happen in our case :tm: but it is theoretically possible to not receive a commit sha. If the binary is built from outside the directory containing the source code then the commit information would not be there. I don't know however in that case if `ReadBuildInfo` would return an error or not. I suspect it wouldn't return an error and simply the `vcs.revision` key would not exist or be empty

miguelsorianod: +1

openapi/kas-fleet-manager.yaml
```diff
@@ -1065,23 +1065,24 @@ components:
...
-              type: array
-              items:
-                allOf:
-                  - $ref: "#/components/schemas/ObjectReference"
+      required:
+        - kind
+        - id
+        - href
+      type: object
+      properties:
```
miguelsorianod, line 1073: There is no need to duplicate the kind, id and href properties. The `VersionMetadata` object can be defined as a Composition of the `ObjectReference` object and a custom object inline that has the server_version and the collections attributes

internal/kafka/test/integration/metadata_test.go
```diff
@@ -0,0 +1,38 @@
...
+	if resp != nil {
+		resp.Body.Close()
+	}
+	g.Expect(err).NotTo(gomega.HaveOccurred(), "g.Unable to retrieve version metadata:  %v", err)
+	g.Expect(resp.StatusCode).To(gomega.Equal(http.StatusOK))
+	/*  Unable to test for a specific version as for now due to the debug package issue: https://github.com/golang/go/issues/33976,
+	TODO when the issue is resolved.
+	*/
+
+	g.Expect(version.ServerVersion).To(gomega.BeEmpty())
```
miguelsorianod, line 37: I think there is no value in testing that it is empty so I think we can omit this check and keep the comment

pkg/api/metadata.go
miguelsorianod: @valeriiashapoval can this method be documented as suggested in the previous comment?

openapi/kas-fleet-manager.yaml
```diff
@@ -1066,22 +1066,24 @@ components:
...
-                kind: "KafkaList"
-          properties:
-            collections:
-              type: array
-              items:
-                allOf:
-                  - $ref: "#/components/schemas/ObjectReference"
+      - $ref: "#/components/schemas/ObjectReference"
+      - type: object
+        example:
```
miguelsorianod, line 1071: The example should be updated to include server_version with an example value

openapi/kas-fleet-manager.yaml
valeriiashapoval: Thanks @miguelsorianod ! The changes were implemented in the last commit.

openapi/kas-fleet-manager.yaml
valeriiashapoval: Thanks a million @miguelsorianod ! Just committed the requested change.

pkg/api/metadata.go
```diff
@@ -22,11 +22,29 @@ import (
...
+	"runtime/debug"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/getsentry/sentry-go"
 	"github.com/golang/glog"
 )
 
+// getCommitSha() returns the current version (vcs.revision attribute of BuildInfo struct) using a built-in debug package and its ReadBuildInfo() function.
+// Upon error, an empty string is returned along with a GeneralError, as well as when vsc.revision is empty/unavailable.
```
miguelsorianod, line 34: // Upon error, an empty string is returned along with a GeneralError, as well as when vsc.revision is empty/unavailable.

This is not accurate.
When ReadBuildInfo returns successfully but  `vcs.revision` is empty/unavailable then no error is returned

miguelsorianod (approved):
Looks good to me. Great work :+1:

internal/kafka/test/integration/metadata_test.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	g.Expect(err).NotTo(gomega.HaveOccurred(), "g.Unable to retrieve version metadata:  %v", err)
+	g.Expect(resp.StatusCode).To(gomega.Equal(http.StatusOK))
+	g.Expect(version.Href).To(gomega.ContainSubstring("/v1"))
+	g.Expect(version.Collections).To(gomega.HaveLen(3))
+
+	/*  Unable to test for a specific version as for now due to the debug package issue: https://github.com/golang/go/issues/33976,
+	TODO when the issue is resolved.
+	*/
```
machi1990, line 39: Can we create a JIRA out of this?

openapi/kas-fleet-manager.yaml
```diff
@@ -1966,4 +1969,4 @@ components:
           for cluster id: 1g5d88q0lrcdv4g7alb7slfgnj3dhbsj%!(EXTRA *errors.Error=identifier
           is '404', code is 'CLUSTERS-MGMT-404' and operation identifier is '1g5or50viu07oealuehrkc26dgftj1ac':
           Cluster '1g5d88q0lrcdv4g7alb7slfgnj3dhbsj' not found)"
-        operation_id: "1iYTsWry6nsqb2sNmFj5bXpD7Ca"
+        operation_id: "1iYTsWry6nsqb2sNmFj5bXpD7Ca"
```
machi1990, line 1972:
```suggestion
        operation_id: "1iYTsWry6nsqb2sNmFj5bXpD7Ca"

```
nit adding newline at end of file

pkg/api/metadata.go
```diff
@@ -69,14 +88,22 @@ func (m *Metadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // ServeHTTP sends API version v1 documentation response.
 func (v *VersionMetadata) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Get the version:
+	version, err := getCommitSha()
+	if err != nil {
+		shared.HandleError(r, w, err.(*errors.ServiceError))
```
machi1990, line 94: the `err.(*errors.ServiceError)` cast assumes all errors from `getCommitSha` will be service error. It is fine for now since only the GeneralError is returned. To be on the safe side I'd have used `errors.ToServiceError(err)` here.

machi1990 (approved):
lgtm

left some minor comments for future considerations.
Let's also create a JIRA out of the TODO comment. 
Thanks for the superb work @valeriiashapoval

internal/kafka/test/integration/metadata_test.go
valeriiashapoval: Thanks @machi1990 ! Will create a Jira for it.

valeriiashapoval:
Thanks @miguelsorianod  @machi1990 ! 
The JIRA for the TODO comment: https://issues.redhat.com/browse/MGDSTRM-9842

machi1990:
Thanks @valeriiashapoval and @miguelsorianod 

When testing this change against a built KFM docker image, I see that the server_version is always empty. 
However, the `make run` command works. I am current looking if it is just my setup or there is something missing on the current PR.

machi1990:
> > Thanks @valeriiashapoval and @miguelsorianod
> > When testing this change against a built KFM docker image, I see that the server_version is always empty. However, the `make run` command works. I am current looking if it is just my setup or there is something missing on the current PR.
> 
> As potentially useful information: build VCS information is only included if the `.git` directory is available when doing the build

Yes, this is the case already: https://github.com/valeriiashapoval/kas-fleet-manager/blob/4fdc083f41258a01096f09becb52a581e7469e31/Dockerfile#L12
 
I've found the missing piece, please have a look at https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1333
@miguelsorianod @valeriiashapoval

PR #1323, opened 2022-09-22
perf(metrics): distribute metrics queries

machi1990:
## Description

The metrics queries can be processed in parallel to further improve the metrics latency time. At the moment the processing is done sequentially meaning which means that the latency will be the total time it takes for all sequential calls to observatorium to finish. Doing it in parallel means that the time it now takes will be the largest time it takes for one operation to complete

- [ ] Code Review completed

### DIFF
--- a/pkg/client/observatorium/api.go
+++ b/pkg/client/observatorium/api.go
@@ -20,6 +20,12 @@ type fetcher struct {
 	labels string
 }
 
+// metricQueryResult holds a metric result for each metric query
+type metricQueryResult struct {
+	metricResult Metric
+	metricQuery  string
+}
+
 type ServiceObservatorium struct {
 	client *Client
 }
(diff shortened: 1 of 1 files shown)


pkg/client/observatorium/api.go
```diff
@@ -203,17 +209,33 @@ func (obs *ServiceObservatorium) GetMetrics(metrics *KafkaMetrics, namespace str
...
+		if metricQueryResult.metricResult.Err != nil {
+			message := fmt.Sprintf("%q:%v", metricQueryResult.metricQuery, metricQueryResult.metricResult.Err)
+			failedMetrics = append(failedMetrics, message)
+			glog.Errorf("error running query %q: %v", metricQueryResult.metricQuery, metricQueryResult.metricResult.Err)
 			continue
 		}
 
-		*metrics = append(*metrics, result)
+		*metrics = append(*metrics, metricQueryResult.metricResult)
 	}
```
machi1990, line 239: I preferred to split the call to obseratorium vs the processing of the results.

pb82 (approved):
LGTM

JameelB (approved):

pkg/client/observatorium/api.go
```diff
@@ -203,17 +209,33 @@ func (obs *ServiceObservatorium) GetMetrics(metrics *KafkaMetrics, namespace str
...
+				metricResult: metricResult,
+				metricQuery:  query,
+			}
+		}(query)
+	}
+
+	// process the metrics result
+	var failedMetrics []string
+	for i := 1; i <= len(queries); i++ {
+		metricQueryResult := <-resultChan
```
miguelsorianod, line 230: I understand this will block until it receives a result right? We can guarantee we will wait until all metrics have been queried due to the length of the queries slice

miguelsorianod: Also the order we will receive the results is not guaranteed to be the order of the original metrics. Is that an issue?

pb82: Yes, that is correct. No metrics can be skipped, even in the case of an error, a result will be returned from the query. So there should always be as many items on the channel as there are queries.

miguelsorianod: > Yes, this was the intention. Do you think we should use the WaitGroup as well?

I don't think that's needed as the channel blocks and the length should guarantee all of them will be processed

machi1990: Thanks @pb82 & @miguelsorianod

pb82: The order won't be an issue for the e2e tests and if it were, I'd consider that a problem with the test that needs to be addressed.

miguelsorianod (approved):
:+1:

machi1990:
Thanks for the review @JameelB @miguelsorianod and @pb82 let's get this in and observe how Stage behaves as per https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1323#issuecomment-1254895848

pkg/client/observatorium/api.go
machi1990: > The order won't be an issue for the e2e tests and if it were, I'd consider that a problem with the test that needs to be addressed.

+1

PR #1332, opened 2022-09-28
fix(dynamic_scale_up): ensure that clusters in deleting state are skipped during capacity evaluation

machi1990:
## Description

ensure that clusters in deleting state are skipped during capacity evaluation
Clusters in 'deprovisiong' and 'cleanup' state can't accept kafka anymore. We shouldn't consider them as clusters that contribute to free streaming units nor as clusters that are capable of accepting the biggest instance size capcity. The clusters in these states should be skipped from any capacity evaluation consideration

## Verification Steps

Added unit test should pass

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
+++ b/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
@@ -407,6 +407,13 @@ func (i *instanceTypeConsumptionSummaryCalculator) Calculate() (instanceTypeCons
 
 		if arrays.Contains(clusterStatesTowardReadyState, kafkaStreamingUnitCountPerCluster.Status) {
 			scaleUpActionIsOngoing = true
+			continue // ignore the rest of calculation as MaxUnits and Count are 0 for clusters that are in terraforming phase
+		}
+
+		clusterIsInDeletingState := arrays.Contains(clusterStatesTowardDeletion, kafkaStreamingUnitCountPerCluster.Status)
+		// ignore cluster is deleting state as they can't accept kafka anymore
+		if clusterIsInDeletingState {
+			continue
 		}
 
 		if kafkaStreamingUnitCountPerCluster.FreeStreamingUnits() >= int32(biggestKafkaInstanceSizeCapacityConsumption) {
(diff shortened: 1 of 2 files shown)


internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
```diff
@@ -407,16 +407,21 @@ func (i *instanceTypeConsumptionSummaryCalculator) Calculate() (instanceTypeCons
...
+			continue // ignore the rest of calculation as MaxUnits and Count are 0 for clusters that are in terraforming phase
+		}
+
+		clusterIsInDeletingState := arrays.Contains(clusterStatesTowardDeletion, kafkaStreamingUnitCountPerCluster.Status)
+		// ignore cluster is deleting state as they can't accept kafka anymore
+		if clusterIsInDeletingState {
+			continue
 		}
 
 		if kafkaStreamingUnitCountPerCluster.FreeStreamingUnits() >= int32(biggestKafkaInstanceSizeCapacityConsumption) {
```
machi1990, line 419: With this condition (without the added skip above), all the clusters in deleting states were considered for this evaluation. 
If one of them had a max size big enough to accommodate the biggest streaming unit, then we were considering that we've enough space in the region for that instance size, so no scale up is required. 
While in fact, we should omit any clusters in these states from capacity evaluation consideration as the clusters that in deleting states can't accept kafka anymore: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/b27cb8ccf683b2ce71cf1d36a4c9f2b561b355ad/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go#L368

pawelpaszki (approved):

miguelsorianod:
I think the original reason on being implemented that way is that we considered the clusters in those states as consuming streaming units.

miguelsorianod:
It seems the exclusion of clusters in deleting or deprovisioning should be excluded of the biggestInstanceType capacity calculation yes.

machi1990:
> I think the original reason on being implemented that way is that we considered the clusters in those states as consuming streaming units. In that way we could scale up before waiting for the cluster to be completely deleted

Yes, this is effectively the intention and not excluding them from biggest capacity evaluation ended up having a negative effect of not causing a scale up.

> It seems the exclusion of clusters in deleting or deprovisioning should be excluded of the biggestInstanceType capacity calculation yes.

@miguelsorianod can you explain what you mean here?

miguelsorianod:
> Yes, this is effectively the intention and not excluding them from biggest capacity evaluation ended up having a negative effect of not causing a scale up.

But those are two different scenarios:
1. Evaluate whether the biggest instance type fits in any cluster
2. Consider it into consumed streaming units calculation

miguelsorianod:
> > Yes, this is effectively the intention and not excluding them from biggest capacity evaluation ended up having a negative effect of not causing a scale up.
> 
> But those are two different scenarios:
> 
>     1. Evaluate whether the biggest instance type fits in any cluster
> 
>     2. Consider it into consumed streaming units calculation

cc @machi1990

machi1990:
> But those are two different scenarios:
> 
> 1. Evaluate whether the biggest instance type fits in any cluster
> 2. Consider it into consumed streaming units calculation

Sorry, I missed this. Reply sent in chat.

PR #1336, opened 2022-10-05
feat: show build info on kfm serve start

valeriiashapoval:
## Description
**Show build-related information on kas-fleet-manager serve start** - (https://issues.redhat.com/browse/MGDSTRM-9742)

## Verification Steps
run `make run`

The output: `I1005 19:32:47.581485  896230 cmd.go:56] Current Go Version: go1.19.1. Current VSC and Operating System information: [{-compiler gc} {CGO_ENABLED 1} {CGO_CFLAGS } {CGO_CPPFLAGS } {CGO_CXXFLAGS } {CGO_LDFLAGS } {GOARCH amd64} {GOOS linux} {GOAMD64 v1} {vcs git} {vcs.revision 324a0f5bdb906853b47144148b14bd9f659c4b10} {vcs.time 2022-10-03T14:36:59Z} {vcs.modified true}]. Main package path: github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/cmd/kas-fleet-manager.`

- [ ] Code Review completed


pkg/cmd/serve/cmd.go
```diff
@@ -19,6 +32,14 @@ func NewServeCommand(env *environments.Env) *cobra.Command {
 			err := env.CreateServices()
 			if err != nil {
 				glog.Fatalf("Unable to initialize environment: %s", err.Error())
+			} else {
+				// Get the build information:
+				buildInfo, err := showBuildInfo()
+				if err != nil {
+					glog.Errorf("unable to retrieve buildinfo")
+				} else {
+					glog.Infof("Current Go Version: %s. Current VSC and Operating System information: %v. Main package path: %s.", buildInfo.GoVersion, buildInfo.Settings, buildInfo.Path)
```
miguelsorianod, line 41: I think we should be more specific about what we want. There's some information that we don't really need.
The current implementation dumps entire structs of information. I'd like to limit it to:
* Build commit SHA
* Architecture
* Operating system
* Build time
* VCS type
* Go version used

miguelsorianod:
This PR and the one where the commit sha in an endpoint was implemented duplicate some of the logic surrounding obtaining the commit sha.
Could we refactor it so both elements reuse code to obtain the commit sha?

miguelsorianod (changes requested):
See comments

pkg/cmd/serve/cmd.go
```diff
@@ -2,14 +2,27 @@ package serve
...
+	"runtime/debug"
 	"syscall"
+
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/environments"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
 )
 
+// showBuildInfo() returns the current Build Information using a built-in debug package and its ReadBuildInfo() function.
```
miguelsorianod, line 16: Parentheses as part of the comment description are not specified. Just the method/function name

miguelsorianod:
The implementation is now better :+1: but it still has room for improvements.

Avoiding package level global state variables should be preferable when possible, even when the package variables are private. Global state is hard to follow, when it is changed it impacts all the code that uses it making it hard to change, etc.
You can find information around why are global variables not good in most cases. See https://www.baeldung.com/cs/global-variables for example.
With that I don't mean that global variables are always bad. There are some cases when it is reasonable or even desirable. But for the changes that we are dealing with in here we can and should avoid it.

I see the getCommitSha code is now reused between the two packages. That's better than before :+1: but if you observe it closely the code still calls the `debug.ReadBuildInfo()`method twice which is unneeded duplicated code. We should try to reduce this.

My proposal is:
Let's create a new type that contains all the needed information. Then we can add methods to the type to retrieve all the information we want. Then the different packages that need it will be able to instantiate and use that type. That implies a new question which is, where do we create that type? given that the type is reused by multiple packages I propose creating a new package `buildinformation`in the `internal/pkg`directory that contains all the implementation related to that. Then the cmd code and the metadata code can use that package.

Let me know if it is clear.

pkg/cmd/serve/cmd.go
```diff
@@ -20,6 +41,26 @@ func NewServeCommand(env *environments.Env) *cobra.Command {
 			if err != nil {
 				glog.Fatalf("Unable to initialize environment: %s", err.Error())
 			}
+			// Get the build information:
+			buildInfo, e := getBuildInfo()
+			// Get the commitSha from the api package:
+			version, err := api.GetCommitSha()
+			if err != nil || e != nil {
+				glog.Fatalf("unable to retrieve buildinfo")
```
ziccardi, line 49: You are no passing parameter to be inserted into the log string, so there is no need to use Fatalf.

```suggestion
				glog.Fatal("unable to retrieve buildinfo")
```

valeriiashapoval:
> The implementation is now better +1 but it still has room for improvements.
> 
> Avoiding package level global state variables should be preferable when possible, even when the package variables are private. Global state is usually harder to follow, when it is changed it impacts all the code that uses it making it hard to change, etc. You can find information around why are global variables not desirable in most cases. See http://wiki.c2.com/?GlobalVariablesAreBad for example. With that I don't mean that global variables are always bad. There are some cases when it is reasonable or even desirable. But for the changes that we are dealing with in here we can and should avoid it.
> 
> I see the getCommitSha code is now reused between the two packages. That's better than before +1 but if you observe it closely the code still calls the `debug.ReadBuildInfo()`method twice which is unneeded duplicated code. We should try to reduce this.
> 
> My proposal is: Let's create a new type that contains all the needed information. Then we can add methods to the type to retrieve all the information we want. Then the different packages that need it will be able to instantiate and use that type. That implies a new question which is, where do we create that type? given that the type is reused by multiple packages I propose creating a new package `buildinformation`in the `internal`directory that contains all the implementation related to that. Then the cmd code and the metadata code can use that package. This also has the advantage that we contain the specific `runtime/debug` knowledge encapsulated and hidden in that package so other packages are not coupled to that dependency.
> 
> Before doing the change copy the current branch implementation in another branch so when the final solution is performed we can observe and analyze the differences and what advantages are we getting in comparison.
> 
> Let me know if it is clear.

Thanks @miguelsorianod ! I have committed the requested changes.

pkg/cmd/serve/cmd.go
valeriiashapoval: Thanks @ziccardi ! Added the changes in the last commit.

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
+package information
```
miguelsorianod, line 1: I'd prefer a `buildinformation` package name. The reason is that `information` is too generic. Information about what? This might open possibility to add all kinds of "information" in the package but I think we should be more specific so the package does not become an "information container"

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
...
+	vcsTime         string
+	vcsType         string
+	operatingSystem string
+	goVersion       string
+}
+
+// GetBuildInfo returns the current Build Information as a BuildInfo type (struct) using a built-in debug package and its ReadBuildInfo() function.
+// Upon error, an empty struct is returned along with a logged error.
+func GetBuildInfo() *BuildInfo {
+	b := new(BuildInfo)
```
miguelsorianod, line 21: An alternative is just doing:
b := &BuildInfo{}

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
...
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		glog.Fatal("unable to get build info")
```
miguelsorianod, line 24: This is not just logging the error but also terminating program execution.
We should just return an error instead of terminating execution. The decision on what to do when an error happens should be up to the consumers of the type

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	b.goVersion = info.GoVersion
+	for _, i := range info.Settings {
```
miguelsorianod, line 27: This is correct but could be more efficient.

`info.Settings` is a map. There is no need to iterate among all keys.
We should change the implementation to retrieve the values using the keys instead.

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
...
+		switch i.Key {
+		case "vcs.time":
```
miguelsorianod, line 29: The keys are referenced as literal "magic" strings. This is already acceptable in this case but a good practice is to put this kind of magic strings into constants. Could we change that?

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
...
+			b.operatingSystem = i.Value
+		case "vcs.revision":
+			b.commitSha = i.Value
+		}
+	}
+	return b
+}
+
+// GetCommitSha returns the current commitSha (version).
+func (b *BuildInfo) GetCommitSha() string {
```
miguelsorianod, line 45: We should document that it is possible that this might be empty depending on whether it's been built using a vcs

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
...
+	return b.commitSha
+}
+
+// GetArchitecture returns the current architecture.
+func (b *BuildInfo) GetArchitecture() string {
+	return b.architecture
+}
+
+// GetVCSTime returns the current commit time.
+func (b *BuildInfo) GetVCSTime() string {
```
miguelsorianod, line 55: We should document that it is possible that this might be empty depending on whether it's been built using a vcs

internal/information/build_info.go
```diff
@@ -0,0 +1,72 @@
...
+	return b.vcsTime
+}
+
+// GetVCSType returns the Version Control System type.
```
miguelsorianod, line 59: We should document that it is possible that this might be empty depending on whether it's been built using a vcs

pkg/cmd/serve/cmd.go
```diff
@@ -20,6 +22,8 @@ func NewServeCommand(env *environments.Env) *cobra.Command {
 			if err != nil {
 				glog.Fatalf("Unable to initialize environment: %s", err.Error())
 			}
+			info := information.GetBuildInfo()
+			glog.Infof("GoVersion: %s. Commit time: %s. Architecture: %s. Operating System: %s. VCS Type: %s. CommitSha: %s.", info.GetGoVersion(), info.GetVCSTime(), info.GetArchitecture(), info.GetOperatingSystem(), info.GetVCSType(), info.GetCommitSha())
```
miguelsorianod, line 26: I suggest changing `%s` by `%q`. `%q` adds single quotes around the value. Due to some of the returned values can be the empty string having single quotes around them is useful.

valeriiashapoval:
Thank you @miguelsorianod ! Just added the suggested changes and edited the sample output to the description.

internal/buildinformation/build_info.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+	"runtime/debug"
+
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+)
+
+const (
+	goarch      = 6
```
miguelsorianod, line 11: We shouldn't rely on the index but on the key (string) value

internal/buildinformation/build_info.go
```diff
@@ -0,0 +1,75 @@
...
+	goVersion       string
+}
+
+// GetBuildInfo returns the current Build Information as a BuildInfo type (struct) using a built-in debug package and its ReadBuildInfo() function.
+// Upon error, an empty struct is returned along with a logged error.
+func GetBuildInfo() *BuildInfo {
+	b := &BuildInfo{}
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		fmt.Print(errors.GeneralError("unable to get build info"))
```
miguelsorianod, line 33: This is not correct. We should be returning an error instead of just logging

valeriiashapoval:
Thanks @miguelsorianod . Just added the suggested changes. Could you please take a look?

internal/buildinformation/build_info.go
```diff
@@ -0,0 +1,83 @@
...
+	architecture    string
+	vcsTime         string
+	vcsType         string
+	operatingSystem string
+	goVersion       string
+}
+
+// GetBuildInfo returns the current Build Information as a BuildInfo type (struct) using a built-in debug package and its ReadBuildInfo() function.
+// Upon error, an empty struct is returned along with a logged error.
+func GetBuildInfo() (*BuildInfo, *errors.ServiceError) {
```
miguelsorianod, line 28: Instead of `*errors.ServiceError` an `error` interface should be returned

pkg/cmd/serve/cmd.go
```diff
@@ -20,6 +22,11 @@ func NewServeCommand(env *environments.Env) *cobra.Command {
 			if err != nil {
 				glog.Fatalf("Unable to initialize environment: %s", err.Error())
 			}
+			info, e := buildinformation.GetBuildInfo()
+			if e != nil {
+				glog.Fatal("unable to retrieve buildinfo")
```
miguelsorianod, line 27: Here the error contenct can be printed, similarly to what's done in the unable to initialize environment error logging

internal/buildinformation/build_info.go
```diff
@@ -0,0 +1,83 @@
...
+		case vcsrevision:
+			b.commitSha = i.Value
+		}
+	}
+	return b, nil
+}
+
+// GetCommitSha returns the current commitSha (version).
+// The returned string can be empty if binary was not built using VCS.
+func (b *BuildInfo) GetCommitSha() string {
```
miguelsorianod, line 54: `Sha` should be capitalized to `SHA`. Acronyms should be capitalized as long as don't affect exportability. The same should be done for the attribute in the struct

internal/buildinformation/build_info.go
```diff
@@ -0,0 +1,83 @@
...
+type BuildInfo struct {
+	commitSha       string
+	architecture    string
+	vcsTime         string
+	vcsType         string
+	operatingSystem string
+	goVersion       string
+}
+
+// GetBuildInfo returns the current Build Information as a BuildInfo type (struct) using a built-in debug package and its ReadBuildInfo() function.
```
miguelsorianod, line 26: No need to indicate the type being returned nor the details of it being a type nor implementation details.
An alternative documentation might be:
// GetBuildInfo returns build related information of the binary or an error

internal/buildinformation/build_info.go
```diff
@@ -0,0 +1,82 @@
...
+const (
+	goarch      string = "GOARCH"
+	goos        string = "GOOS"
+	vcstype     string = "vcs"
+	vcsrevision string = "vcs.revision"
+	vcstime     string = "vcs.time"
+)
+
+type BuildInfo struct {
+	commitSha       string
```
miguelsorianod, line 18: Sha in here should be capitalized

valeriiashapoval:
> Looks good to me. Once the tests pass this can be merged.
> 
> Good work +1

Thank you so much @miguelsorianod !

PR #1344, opened 2022-10-11
feat: add handling of suspended field in admin api patch kafka

pawelpaszki:
## Description
feat: add handling of suspended field in admin api patch kafka

## Verification Steps
CI passing and states changing as per epic brief for kafka suspension

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/api/admin/private/api/openapi.yaml
+++ b/internal/kafka/internal/api/admin/private/api/openapi.yaml
@@ -340,6 +340,7 @@ components:
         max_data_retention_size: max_data_retention_size
         kafka_version: kafka_version
         kafka_storage_size: kafka_storage_size
+        suspended: true
       properties:
         strimzi_version:
           type: string
@@ -355,6 +356,13 @@ components:
         max_data_retention_size:
           description: Maximum data storage available to this Kafka
           type: string
+        suspended:
+          description: boolean value indicating whether kafka should be suspended
+            or not depending on the value provided. Suspended kafkas have their certain
+            resources removed and become inaccessible until fully unsuspended (restored
+            to Ready state).
+          nullable: true
+          type: boolean
       type: object
     SupportedKafkaSizeBytesValueItem:
       properties:
--- a/internal/kafka/internal/handlers/admin_kafka.go
+++ b/internal/kafka/internal/handlers/admin_kafka.go
@@ -9,12 +9,14 @@ import (
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/admin/private"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/presenters"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/handlers"
 	coreServices "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services"
+	shared "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	"github.com/gorilla/mux"
 )
 
@@ -169,12 +171,32 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
 				return false
 			}
 
+			getStatusBasedOnSuspendedParam := func(susp *bool, kafka *dbapi.KafkaRequest) string {
+				if shared.IsNilPredicate(susp) {
+					return kafka.Status
+				} else {
+					if *susp {
+						if kafka.Status == constants.KafkaRequestStatusReady.String() {
+							return constants.KafkaRequestStatusSuspending.String()
+						}
+					} else {
+						if kafka.Status == constants.KafkaRequestStatusSuspended.String() || kafka.Status == constants.KafkaRequestStatusSuspending.String() {
+							return constants.KafkaRequestStatusResuming.String()
+						}
+					}
+				}
+				return kafka.Status
+			}
+
 			requestedStorageSize, _ := arrays.FirstNonEmpty(kafkaUpdateReq.MaxDataRetentionSize, kafkaUpdateReq.DeprecatedKafkaStorageSize)
 
 			updateRequired := update(&kafkaRequest.DesiredKafkaVersion, kafkaUpdateReq.KafkaVersion)
 			updateRequired = update(&kafkaRequest.DesiredStrimziVersion, kafkaUpdateReq.StrimziVersion) || updateRequired
 			updateRequired = update(&kafkaRequest.DesiredKafkaIBPVersion, kafkaUpdateReq.KafkaIbpVersion) || updateRequired
 			updateRequired = update(&kafkaRequest.KafkaStorageSize, requestedStorageSize) || updateRequired
+
+			newStatus := getStatusBasedOnSuspendedParam(kafkaUpdateReq.Suspended, kafkaRequest)
+			updateRequired = update(&kafkaRequest.Status, newStatus) || updateRequired
 
 			if updateRequired {
 				err := h.kafkaService.VerifyAndUpdateKafkaAdmin(ctx, kafkaRequest)
--- a/internal/kafka/internal/handlers/admin_kafka_test.go
+++ b/internal/kafka/internal/handlers/admin_kafka_test.go
@@ -3,12 +3,14 @@ package handlers
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/admin/private"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
@@ -18,6 +20,10 @@ import (
 	s "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services/account"
 	"github.com/onsi/gomega"
+)
+
+const (
+	kafkaByIdUrl = "/kafkas/{id}"
 )
 
 func GetHandlerParams(method string, url string, body io.Reader, t *testing.T) (*http.Request, *httptest.ResponseRecorder) {
@@ -207,7 +213,7 @@ func Test_Delete(t *testing.T) {
 		{
 			name: "should return an error if async flag is not set to true when deleting kafka",
 			args: args{
-				url: "/kafkas/{id}",
+				url: kafkaByIdUrl,
 			},
 			wantStatusCode: http.StatusBadRequest,
 		},
@@ -239,10 +245,11 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 		body []byte
 	}
 	tests := []struct {
-		name           string
-		fields         fields
-		args           args
-		wantStatusCode int
+		name            string
+		fields          fields
+		args            args
+		wantStatusCode  int
+		wantKafkaStatus constants.KafkaStatus
 	}{
 		{
 			name: "should return an error if retrieving kafka to update fails",
@@ -269,7 +276,7 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 				accountService: account.NewMockAccountService(),
 			},
 			args: args{
-				url:  "/kafkas/{id}",
+				url:  kafkaByIdUrl,
 				body: []byte(`{}`),
 			},
 			wantStatusCode: http.StatusInternalServerError,
@@ -284,7 +291,7 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 				},
 			},
 			args: args{
-				url:  "/kafkas/{id}",
+				url:  kafkaByIdUrl,
 				body: []byte(`{}`),
 			},
 			wantStatusCode: http.StatusNotFound,
@@ -302,7 +309,7 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 				},
 			},
 			args: args{
-				url:  "/kafkas/{id}",
+				url:  kafkaByIdUrl,
 				body: []byte(`{"kafka_version": "2.8.0"}`),
 			},
 			wantStatusCode: http.StatusBadRequest,
@@ -320,7 +327,7 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 				},
 			},
 			args: args{
-				url:  "/kafkas/{id}",
+				url:  kafkaByIdUrl,
 				body: []byte(`{"strimzi_version": "2.8.0"}`),
 			},
 			wantStatusCode: http.StatusBadRequest,
@@ -338,7 +345,7 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 				},
 			},
 			args: args{
-				url:  "/kafkas/{id}",
+				url:  kafkaByIdUrl,
 				body: []byte(`{"kafka_ibp_version": "2.8.0"}`),
 			},
 			wantStatusCode: http.StatusBadRequest,
@@ -383,7 +390,7 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 				accountService: account.NewMockAccountService(),
 			},
 			args: args{
-				url:  "/kafkas/{id}",
+				url:  kafkaByIdUrl,
 				body: []byte(`{"kafka_ibp_version": "2.7"}`),
 			},
 			wantStatusCode: http.StatusBadRequest,
@@ -430,7 +437,7 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
 				accountService: account.NewMockAccountService(),
 			},
 			args: args{
-				url:  "/kafkas/{id}",
+				url:  kafkaByIdUrl,
 				body: []byte(`{"kafka_ibp_version": "2.7"}`),
 			},
 			wantStatusCode: http.StatusInternalServerError,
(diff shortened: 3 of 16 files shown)


pawelpaszki:
ping @JameelB @valeriiashapoval

internal/kafka/internal/services/kafka.go
```diff
@@ -1036,6 +1037,9 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 		"bf2.org/kafkaInstanceProfileQuotaConsumed": strconv.Itoa(k.QuotaConsumed),
 		"bf2.org/kafkaInstanceProfileType":          kafkaRequest.InstanceType,
 	}
+	if strings.HasPrefix(kafkaRequest.Status, "suspend") {
+		labels["bf2.org/suspended"] = fmt.Sprintf("%t", true)
+	}
```
machi1990, line 1042: We could use https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/docs/utilities/arrays.md#contains over here: 

```suggestion 
                suspendedStatuses := []string{KafkaRequestStatusSuspending.String(), KafkaRequestStatusSuspended.String()}
		labels["bf2.org/suspended"] = fmt.Sprintf("%t", arrays.Contains(suspendedStatuses, kafkaRequest.Status))
```

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -169,13 +170,36 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
...
 			updateRequired := update(&kafkaRequest.DesiredKafkaVersion, kafkaUpdateReq.KafkaVersion)
 			updateRequired = update(&kafkaRequest.DesiredStrimziVersion, kafkaUpdateReq.StrimziVersion) || updateRequired
 			updateRequired = update(&kafkaRequest.DesiredKafkaIBPVersion, kafkaUpdateReq.KafkaIbpVersion) || updateRequired
 			updateRequired = update(&kafkaRequest.KafkaStorageSize, requestedStorageSize) || updateRequired
 
+			newStatus := getStatusBasedOnSuspendedParam(kafkaUpdateReq.Suspended, kafkaRequest)
+			if kafkaRequest.Status != newStatus {
+				kafkaRequest.Status = newStatus
+				updateRequired = true
+			}
```
machi1990, line 201:
```suggestion
			newStatus := getStatusBasedOnSuspendedParam(kafkaUpdateReq.Suspended, kafkaRequest)
			updateRequired = update(&kafkaRequest.Status, newStatus) || updateRequired
```

machi1990 (changes requested):

internal/kafka/internal/services/kafka.go
machi1990: Worth noting that the suggestion will add the label all the time: either `true`  or `false`. For existing Kafkas this will cause some pods restart but it won't cause service disruption.

pawelpaszki: makes sense to use the array, however I think the label should only be sent if suspended is true (according to the epic brief). not sure if we want to send `suspended=false`

machi1990: Sending `false` should be fine as well and it'll serve us that one extra `if`: Please see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/4b570dec3c121012e1a330fc4f2d0388c1200d67/api/src/main/java/org/bf2/operator/resources/v1alpha1/ManagedKafka.java#L203 which is part of https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/771

@MikeEdgar is my understanding above correct?

MikeEdgar: @machi1990, yes that is correct. `false` is the same as the label being absent.

internal/kafka/internal/services/kafka.go
```diff
@@ -1037,9 +1039,10 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 		"bf2.org/kafkaInstanceProfileQuotaConsumed": strconv.Itoa(k.QuotaConsumed),
 		"bf2.org/kafkaInstanceProfileType":          kafkaRequest.InstanceType,
 	}
-	if strings.HasPrefix(kafkaRequest.Status, "suspend") {
-		labels["bf2.org/suspended"] = fmt.Sprintf("%t", true)
-	}
+
+	suspendedStatuses := []string{constants.KafkaRequestStatusSuspending.String(), constants.KafkaRequestStatusSuspended.String()}
+	labels["bf2.org/suspended"] = fmt.Sprintf("%t", arrays.Contains(suspendedStatuses, kafkaRequest.Status))
```
machi1990, line 1044: Thanks @pawelpaszki 

Do you think this code https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c030410217731ebb6c3e6dce12ddcd83c1766beb/internal/kafka/internal/services/kafka.go#L1038-L1044

could be written as
```go
suspendedStatuses := []string{constants.KafkaRequestStatusSuspending.String(), constants.KafkaRequestStatusSuspended.String()}

labels := map[string]string{
		"bf2.org/kafkaInstanceProfileQuotaConsumed": strconv.Itoa(k.QuotaConsumed),
		"bf2.org/kafkaInstanceProfileType":          kafkaRequest.InstanceType,
                "bf2.org/suspended":  fmt.Sprintf("%t", arrays.Contains(suspendedStatuses, kafkaRequest.Status))
	}
```

internal/kafka/internal/services/kafka.go
```diff
@@ -1036,6 +1039,10 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 		"bf2.org/kafkaInstanceProfileQuotaConsumed": strconv.Itoa(k.QuotaConsumed),
 		"bf2.org/kafkaInstanceProfileType":          kafkaRequest.InstanceType,
 	}
+
+	suspendedStatuses := []string{constants.KafkaRequestStatusSuspending.String(), constants.KafkaRequestStatusSuspended.String()}
+	labels["bf2.org/suspended"] = fmt.Sprintf("%t", arrays.Contains(suspendedStatuses, kafkaRequest.Status))
```
machi1990, line 1044: can we add an integration test on the newly added label? There is this existing test that we could enrich https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c030410217731ebb6c3e6dce12ddcd83c1766beb/internal/kafka/test/integration/data_plane_endpoints_test.go#L223 to test for the case the label is `false` (all the other statues - existing ones) and for when the label is `true`.

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -169,13 +170,33 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
...
+				} else {
+					if *susp {
+						if kafka.Status != constants.KafkaRequestStatusSuspended.String() {
+							return constants.KafkaRequestStatusSuspending.String()
+						}
+					} else {
+						if kafka.Status == constants.KafkaRequestStatusSuspended.String() || kafka.Status == constants.KafkaRequestStatusSuspending.String() {
+							return constants.KafkaRequestStatusResuming.String()
+						}
+					}
```
machi1990, line 185: can we add a unit test for this piece of logic?

internal/kafka/internal/services/kafka.go
pawelpaszki: yes. I am also adding some more unit tests right now

internal/kafka/internal/handlers/admin_kafka.go
pawelpaszki: adding them as we speak :)

machi1990: just seen https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1344#discussion_r993107180, thanks.

internal/kafka/internal/services/kafka.go
machi1990: Thank you.

machi1990 (approved):

machi1990:
lgtm, thanks @pawelpaszki 

I'll let @JameelB and @valeriiashapoval to have the final review.

internal/kafka/internal/handlers/admin_kafka_test.go
```diff
@@ -477,10 +484,155 @@ func Test_adminKafkaHandler_Update(t *testing.T) {
...
+					},
+				},
+				accountService: account.NewMockAccountService(),
+			},
+			args: args{
+				url:  kafkaByIdUrl,
+				body: []byte(`{"suspended": false}`),
+			},
+			wantStatusCode:  http.StatusOK,
+			wantKafkaStatus: constants.KafkaRequestStatusReady,
```
JameelB, line 635: Can we also add a test case here to ensure that status is not affected if suspended was not specified in the request?

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -169,13 +170,33 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
 				return false
 			}
 
+			getStatusBasedOnSuspendedParam := func(susp *bool, kafka *dbapi.KafkaRequest) string {
```
JameelB, line 173: I see what you mean now :) :+1: :heart:

internal/kafka/test/integration/admin_kafka_test.go
```diff
@@ -1146,6 +1152,96 @@ func TestAdminKafka_Update(t *testing.T) {
...
+					Suspended: &falseB,
+				},
+			},
+			verifyResponse: func(result adminprivate.Kafka, resp *http.Response, err error) {
+				g.Expect(err).To(gomega.BeNil())
+				g.Expect(resp.StatusCode).To(gomega.Equal(http.StatusOK))
+				g.Expect(result.Id).To(gomega.Equal(suspendingKafkaID))
+				g.Expect(result.Status).To(gomega.Equal(constants.KafkaRequestStatusResuming.String()))
+			},
+		},
```
JameelB, line 1244: Similar to unit test, can we also add a test case here to ensure status does not get changed if `suspended` was omitted from the request?

pawelpaszki:
@JameelB I think I addressed your comments. can you take another look please?

MikeEdgar:
Not sure when you're planning to merge this, but if you want to keep it open I can test together with the data plane and UI to confirm that looks OK.

internal/kafka/internal/handlers/validation.go
```diff
@@ -260,13 +260,18 @@ func ValidateKafkaUpdateFields(kafkaUpdateRequest *private.KafkaUpdateRequest) h
 			stringSet(&kafkaUpdateRequest.KafkaVersion) ||
 			stringSet(&kafkaUpdateRequest.KafkaIbpVersion) ||
 			stringSet(&kafkaUpdateRequest.DeprecatedKafkaStorageSize) ||
-			stringSet(&kafkaUpdateRequest.MaxDataRetentionSize)) {
-			return errors.FieldValidationError("Failed to update Kafka Request. Expecting at least one of the following fields: strimzi_version, kafka_version, kafka_ibp_version, kafka_storage_size or max_data_retention_size to be provided")
+			stringSet(&kafkaUpdateRequest.MaxDataRetentionSize) ||
+			boolIsSet(kafkaUpdateRequest.Suspended)) {
```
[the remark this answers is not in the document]
JameelB: +1 on this, once refactored, we can remove the `boolIsSet` method and we also won't need the tests for it.

internal/kafka/internal/services/kafka.go
```diff
@@ -1032,10 +1044,13 @@ func buildManagedKafkaCR(kafkaRequest *dbapi.KafkaRequest, kafkaConfig *config.K
 	if err != nil {
 		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "unable to list kafka request")
 	}
+
 	labels := map[string]string{
 		"bf2.org/kafkaInstanceProfileQuotaConsumed": strconv.Itoa(k.QuotaConsumed),
 		"bf2.org/kafkaInstanceProfileType":          kafkaRequest.InstanceType,
+		v1.ManagedKafkaBf2SuspendedLabelKey:         fmt.Sprintf("%t", arrays.Contains(constants.GetSuspendedStatuses(), kafkaRequest.Status)),
```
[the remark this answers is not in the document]
MikeEdgar: @JameelB it's possible in the case where the control plane sets a version for Strimzi, Kafka, or Kafka IBP that is invalid in the data plane, or if the deployment is changed to reserved, the profile is changed (standard <-> developer), or the size changes (x1 <-> x2). Otherwise, the data plane overrides any other status with suspended when the label is present.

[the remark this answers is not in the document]
machi1990: I think the only valid next state from `suspending` from user point of view should be `suspended`. And the only valid state from `suspended` should be `resuming`.
ie. the CP in its state transition has to guard against these states changes `suspending -> failed`, `suspending->ready`, `suspended -> failed`, `suspended->ready`. 

and allow for `suspending->suspended` ......(allow for Kafka components version upgrades without states changes when they are needed)...  `suspended->resuming`,  ( either `resuming->failed` or `resuming->ready`) states.

or allow for `suspending->suspended` ......(allow for Kafka components version upgrades without states changes when they are needed)...  `suspended->deprovision`... kafka deletion.

pawelpaszki:
> Not sure when you're planning to merge this, but if you want to keep it open I can test together with the data plane and UI to confirm that looks OK.

@MikeEdgar - sure I'll keep it open. let me know if everything looks ok. thanks

internal/kafka/internal/services/kafka.go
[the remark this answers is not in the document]
JameelB: Thanks @MikeEdgar @machi1990. 

I was thinking the same, but we won't see if a Kafka fails to go from `suspending` -> `suspended`. I think what we're missing here are metrics + alert if a Kafka is stuck in `suspending` for quite some time. wdyt?

This can be done as part of a separate JIRA.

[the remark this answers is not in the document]
machi1990: > I was thinking the same, but we won't see if a Kafka fails to go from suspending -> suspended. I think what we're missing here are metrics + alert if a Kafka is stuck in suspending for quite some time. 

I think there will be sentry alerts (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/1007c644febe368beacedbbe0cbb5f03b01027b2/internal/kafka/internal/services/data_plane_kafka.go#L311) when a Data Plane reports that the Kafka has failed. We could add a metric on top of that.

miguelsorianod:
@machi1990 how does this affect dynamic scaling logic?

There are several places where sets of states are evaluated. I understand we are introducing new states for kafka instances

machi1990:
> @machi1990 how does this affect dynamic scaling logic?

That's a good question.
 
> There are several places where sets of states are evaluated. I understand we are introducing new states for kafka instances

The new introduced states should still be accounted for in the capacity used count: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L742

miguelsorianod:
> > @machi1990 how does this affect dynamic scaling logic?
> 
> That's a good question.
> 
> > There are several places where sets of states are evaluated. I understand we are introducing new states for kafka instances
> 
> The new introduced states should still be accounted for in the capacity used count:
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L742

What about the  set of states referenced in the dynamic scale up / down mgrs?

machi1990:
> > > @machi1990 how does this affect dynamic scaling logic?
> > 
> > 
> > That's a good question.
> > > There are several places where sets of states are evaluated. I understand we are introducing new states for kafka instances
> > 
> > 
> > The new introduced states should still be accounted for in the capacity used count:
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L742
> 
> What about the set of states referenced in the dynamic scale up / down mgrs?

Those are `clusters` Status and they are not affected by this PR. As for the Kafka Requests status, the kafka status is excluded from the returned result: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L791 and the applied filter on is only on `deleting` status: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L793.

miguelsorianod:
> > > > @machi1990 how does this affect dynamic scaling logic?
> > > 
> > > 
> > > That's a good question.
> > > > There are several places where sets of states are evaluated. I understand we are introducing new states for kafka instances
> > > 
> > > 
> > > The new introduced states should still be accounted for in the capacity used count:
> > > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L742
> > 
> > 
> > What about the set of states referenced in the dynamic scale up / down mgrs?
> 
> Those are `clusters` Status and they are not affected by this PR. As for the Kafka Requests status, the kafka status is excluded from the returned result:
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L791
> and the applied filter on is only on `deleting` status:
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L793
> .

This means that kafka resources in suspended, resuming, suspending status would be considered as consuming resources right?

machi1990:
> > > > > @machi1990 how does this affect dynamic scaling logic?
> > > > 
> > > > 
> > > > That's a good question.
> > > > > There are several places where sets of states are evaluated. I understand we are introducing new states for kafka instances
> > > > 
> > > > 
> > > > The new introduced states should still be accounted for in the capacity used count:
> > > > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L742
> > > 
> > > 
> > > What about the set of states referenced in the dynamic scale up / down mgrs?
> > 
> > 
> > Those are `clusters` Status and they are not affected by this PR. As for the Kafka Requests status, the kafka status is excluded from the returned result:
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L791
> > 
> > and the applied filter on is only on `deleting` status:
> > https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cae53dd1355fad5784b8583dbb3fb450d4029b1/internal/kafka/internal/services/clusters.go#L793
> > 
> > .
> 
> This means that kafka resources in suspended, resuming, suspending status would be considered as consuming resources in the cluster right?

Yes, exactly. In the control plane, we want to be sure that we do not overprovision Kafka on clusters that have in suspended/supsending states because we want to make sure that the Kafka in suspended/supsending/resuming states can be resumed without issue on those clusters. 

In the data plane, the kafkas in suspended state is not consuming any resource i.e all its components are scaled down which will result in the cluster autoscaler scaling down the corresponding nodes.

internal/kafka/internal/api/admin/private/api/openapi.yaml
```diff
@@ -355,6 +356,11 @@ components:
         max_data_retention_size:
           description: Maximum data storage available to this Kafka
           type: string
+        suspended:
+          description: boolean value indicating whether kafka should be suspended
```
miguelsorianod, line 360: Could we add some information about what does setting a kafka as suspended mean from the user's point of view?

pawelpaszki: @miguelsorianod - I added a bit more info there

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -169,13 +171,33 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
...
+				if shared.IsNilPredicate(susp) {
+					return kafka.Status
+				} else {
+					if *susp {
```
JameelB, line 178: This means we can update the status from `failed` to `suspending`. Do we want to do that?

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -169,13 +171,33 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
...
+						if kafka.Status != constants.KafkaRequestStatusSuspended.String() {
+							return constants.KafkaRequestStatusSuspending.String()
+						}
+					} else {
+						if kafka.Status == constants.KafkaRequestStatusSuspended.String() || kafka.Status == constants.KafkaRequestStatusSuspending.String() {
```
JameelB, line 183: We probably also want to add in `KafkaRequestStatusFailed` to this check so users can't update from `failed` -> `resuming`

JameelB: read this wrong, ignore this :)

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -169,13 +171,33 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
...
 			}
 
+			getStatusBasedOnSuspendedParam := func(susp *bool, kafka *dbapi.KafkaRequest) string {
+				if shared.IsNilPredicate(susp) {
+					return kafka.Status
+				} else {
+					if *susp {
+						if kafka.Status != constants.KafkaRequestStatusSuspended.String() {
+							return constants.KafkaRequestStatusSuspending.String()
+						}
```
MikeEdgar, line 181: Running this with my kas-fleetshard changes, the status remains as `suspending`. I think this might be missing a transition to `suspended` somewhere.

machi1990: This is taken care in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1346

MikeEdgar:
There seems to be a race condition between the Admin API setting the Kafka status and the status being reverted to ready by the data plane sync. Patching a `"suspended": true` results in the KFM status changing to `suspending`, but within a few seconds it returns to `ready` and the MK ends up with a `bf2.org/suspended=false` label. I'm checking to see if this can be accounted for in the fleetshard sync.

pawelpaszki:
> There seems to be a race condition between the Admin API setting the Kafka status and the status being reverted to ready by the data plane sync. Patching a `"suspended": true` results in the KFM status changing to `suspending`, but within a few seconds it returns to `ready` and the MK ends up with a `bf2.org/suspended=false` label. I'm checking to see if this can be accounted for in the fleetshard sync.
> 
> Update: We may be somewhat limited on the data plane side handling of this situation. This is what I think is happening
> 
> 1. Status set to `suspending` in KFM database
> 2. Data plane syncs MK list from control plane
> 3. Data plane periodic resync of MK status to control plane: status set back to ready
> 4. kas-fleetshard operator reacts to label suspended=true, suspends instance, sets Ready=False, Reason=Suspended
> 5. Data plane syncs MK list from control plane, label suspended=false due to step 3
> 6. kas-fleetshard operator reacts to label suspended=false, resumes instance
> 
> Update 2: If the above is correct, can KFM ignore the ready status from the data plane when the local state is suspending/suspended?

@JameelB - will your status transition PR handle this situation? (e.g. ignore ready status from FSO if the status is already suspended/suspending?

internal/kafka/internal/handlers/admin_kafka.go
pawelpaszki: deleting isn't one of the [updatable statuses](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1344/files#diff-08f986360fba8bdf2eb71752a89a88a34d0c4a79014ec11c306ebd6586bed919R101). thinking now - should `deprovision` also be removed from the updatable statuses? this way we wouldn't need to add extra check here

machi1990:
> > There seems to be a race condition between the Admin API setting the Kafka status and the status being reverted to ready by the data plane sync. Patching a `"suspended": true` results in the KFM status changing to `suspending`, but within a few seconds it returns to `ready` and the MK ends up with a `bf2.org/suspended=false` label. I'm checking to see if this can be accounted for in the fleetshard sync.
> > Update: We may be somewhat limited on the data plane side handling of this situation. This is what I think is happening
> > 
> > 1. Status set to `suspending` in KFM database
> > 2. Data plane syncs MK list from control plane
> > 3. Data plane periodic resync of MK status to control plane: status set back to ready
> > 4. kas-fleetshard operator reacts to label suspended=true, suspends instance, sets Ready=False, Reason=Suspended
> > 5. Data plane syncs MK list from control plane, label suspended=false due to step 3
> > 6. kas-fleetshard operator reacts to label suspended=false, resumes instance
> > 
> > Update 2: If the above is correct, can KFM ignore the ready status from the data plane when the local state is suspending/suspended?
> 
> @JameelB - will your status transition PR handle this situation? (e.g. ignore ready status from FSO if the status is already suspended/suspending?

I think this is acounted for here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/228a9c7785116a4aff8772f484bde7867ab14345/internal/kafka/internal/services/data_plane_kafka.go#L169 in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1346

internal/kafka/internal/handlers/admin_kafka.go
machi1990: > should deprovision also be removed from the updatable statuses? this way we wouldn't need to add extra check here

We need to be able to update Strimzi version of a Kafka in deprovision status. See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/787 for the motivation

pawelpaszki: @machi1990 should we set to resuming if kafka is  in `KafkaRequestStatusDeprovision` or `KafkaRequestStatusDeleting` state?

pawelpaszki: nvm. it will not happen. ignore

JameelB:
> > > There seems to be a race condition between the Admin API setting the Kafka status and the status being reverted to ready by the data plane sync. Patching a `"suspended": true` results in the KFM status changing to `suspending`, but within a few seconds it returns to `ready` and the MK ends up with a `bf2.org/suspended=false` label. I'm checking to see if this can be accounted for in the fleetshard sync.
> > > Update: We may be somewhat limited on the data plane side handling of this situation. This is what I think is happening
> > > 
> > > 1. Status set to `suspending` in KFM database
> > > 2. Data plane syncs MK list from control plane
> > > 3. Data plane periodic resync of MK status to control plane: status set back to ready
> > > 4. kas-fleetshard operator reacts to label suspended=true, suspends instance, sets Ready=False, Reason=Suspended
> > > 5. Data plane syncs MK list from control plane, label suspended=false due to step 3
> > > 6. kas-fleetshard operator reacts to label suspended=false, resumes instance
> > > 
> > > Update 2: If the above is correct, can KFM ignore the ready status from the data plane when the local state is suspending/suspended?
> > 
> > 
> > @JameelB - will your status transition PR handle this situation? (e.g. ignore ready status from FSO if the status is already suspended/suspending?
> 
> I think this is acounted for here
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/228a9c7785116a4aff8772f484bde7867ab14345/internal/kafka/internal/services/data_plane_kafka.go#L169
> 
> in #1346

As @machi1990 mentioned it should be accounted for. This PR focuses on setting suspending/resuming from admin API. #1346 bridges this change and updates from FSO. I based that PR off of this so it should have both changes. Could test using that branch @MikeEdgar?

internal/kafka/internal/handlers/admin_kafka.go
pawelpaszki: @JameelB I have added your requested change. will squash the commits after CI passes

JameelB: Thanks @pawelpaszki! Seen the changes, look good to me :+1: 

There's an open question in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1346#discussion_r995497253 which may change the logic here.

pawelpaszki: yes. saw that. will await further comments there

internal/kafka/internal/handlers/validation.go
```diff
@@ -260,8 +260,9 @@ func ValidateKafkaUpdateFields(kafkaUpdateRequest *private.KafkaUpdateRequest) h
 			stringSet(&kafkaUpdateRequest.KafkaVersion) ||
 			stringSet(&kafkaUpdateRequest.KafkaIbpVersion) ||
 			stringSet(&kafkaUpdateRequest.DeprecatedKafkaStorageSize) ||
-			stringSet(&kafkaUpdateRequest.MaxDataRetentionSize)) {
-			return errors.FieldValidationError("Failed to update Kafka Request. Expecting at least one of the following fields: strimzi_version, kafka_version, kafka_ibp_version, kafka_storage_size or max_data_retention_size to be provided")
+			stringSet(&kafkaUpdateRequest.MaxDataRetentionSize) ||
+			arrays.IsNotNilPredicate(kafkaUpdateRequest.Suspended)) {
```
JameelB, line 264: I don't think we want to add this check in here since `suspended` can be omitted from the request?

See thread in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1346#discussion_r995732444

machi1990: I think this is checking that at least one field is present in the update request body: notice the `!` in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3ab2d9f94366698babaeeebfdccc223f7caa2c6c/internal/kafka/internal/handlers/validation.go#L259

internal/kafka/test/integration/admin_kafka_test.go
```diff
@@ -1146,6 +1155,148 @@ func TestAdminKafka_Update(t *testing.T) {
...
+				ctx: func(h *coreTest.Helper) context.Context {
+					return NewAuthenticatedContextForAdminEndpoints(h, []string{testFullRole})
+				},
+				kafkaID: sampleKafkaID1,
+				kafkaUpdateRequest: adminprivate.KafkaUpdateRequest{
+					Suspended: &falseB,
+				},
+			},
+			verifyResponse: func(result adminprivate.Kafka, resp *http.Response, err error) {
+				g.Expect(err).To(gomega.BeNil())
```
ziccardi, line 1170:
```suggestion
				g.Expect(err).NotTo(gomega.HaveOccurred())
```

internal/kafka/internal/handlers/admin_kafka.go
```diff
@@ -169,13 +170,33 @@ func (h *adminKafkaHandler) Update(w http.ResponseWriter, r *http.Request) {
 				return false
 			}
 
+			getStatusBasedOnSuspendedParam := func(susp *bool, kafka *dbapi.KafkaRequest) string {
+				if !boolIsSet(susp) {
```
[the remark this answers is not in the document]
ziccardi: Those predicates were meant to be used with the array functions. If we believe they can be of common use, we should move them to another package.

internal/kafka/internal/handlers/validation.go
pawelpaszki: this statement checks if any of the parameters is provided and fails if no valid param is present in the body. maybe we should remove it completely and just return 200 in this case?

ziccardi: @JameelB No. What I wanted to ask was: "Why are we using array utils predicate functions here?"
Those predicates were meant to be used with the `Find`, `AnyMatch`, `...` functions.

If we believe they can be useful as common utilities, I believe we should move them to another package.

pawelpaszki: for isNil/ isNotNil I think it makes sense to move them to the common utils. should I move it as a part of my PR?

MikeEdgar:
> I based that PR off of this so it should have both changes. Could test using that branch

@JameelB - I'll test with that PR instead. Thanks for the heads up, I was getting myself confused between the two PRs :laughing:

MikeEdgar:
@JameelB , @pawelpaszki - testing this via #1346 now

internal/kafka/internal/handlers/validation.go
```diff
@@ -260,8 +260,9 @@ func ValidateKafkaUpdateFields(kafkaUpdateRequest *private.KafkaUpdateRequest) h
...
+			return errors.FieldValidationError("Failed to update Kafka Request. Expecting at least one of the following fields: strimzi_version, kafka_version, kafka_ibp_version, kafka_storage_size, max_data_retention_size or suspended to be provided")
```
ziccardi, line 265: Error messages should start with lowercase

internal/kafka/internal/handlers/validation_test.go
```diff
@@ -969,7 +969,7 @@ func TestValidateKafkaUpdateFields(t *testing.T) {
 					DeprecatedKafkaStorageSize: "",
 				},
 			},
-			want: errors.FieldValidationError("Failed to update Kafka Request. Expecting at least one of the following fields: strimzi_version, kafka_version, kafka_ibp_version, kafka_storage_size or max_data_retention_size to be provided"),
+			want: errors.FieldValidationError("Failed to update Kafka Request. Expecting at least one of the following fields: strimzi_version, kafka_version, kafka_ibp_version, kafka_storage_size, max_data_retention_size or suspended to be provided"),
```
ziccardi, line 972: Error messages should start with lowercase

MikeEdgar (approved):
Latest changes look good from data plane perspective

internal/kafka/internal/handlers/validation.go
pawelpaszki: @ziccardi - there are a lot of ocurrences of errors starting with uppercase across kfm. I will change this one and create a jira to tackle others

miguelsorianod: @pawelpaszki @ziccardi ,

A possible option is to add the pattern in the `forbididgo` configuration of golangci-lint. For example, adding fmt.Errorf starting with a literal string that begins with uppercase inside the first parameter of it. We could also add GeneralError with the same pattern. Same for errors.Errorf and other cases where we know we can reliably perform the check without false positives.

This won't detect all scenarios nor warn us of all of them but it would be an improvement.

miguelsorianod: A potentially better alternative, or one to be used in conjunction with my previous proposal is to use the `revive` linter with the `error-strings`  rule `https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#error-strings` although we should explore in the doc / behavior how far it hows and the chance of false positives.

Of course all of this means that it will start to detect the errors so those should be fixed together with the corresponding linters changes

PR #1346, opened 2022-10-13
feat: set suspended status reported by fleetshard

JameelB:
## Description
The Fleetshard Operator will have a new status to report called 'Suspended'. Fleetshard will report this status by setting the ManagedKafka CR 'Ready' condition to have the following values: `Status=False,Reason=Suspended`.

On top of that, there are also two new KafkaRequest status set in KAS Fleet Manager (as seen in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1344)
* `resuming`: set by a user to indicate that they want to resume their Kafka instance from a `suspended` or `suspending` state.
* `suspending`: set by a user to indicate that they want to suspend their Kafka instance. This can only be updated from a `ready` state.

All three new statuses must be accommodated for when processing real Kafka deployments when the Fleetshard calls the `/agent-clusters/{id}/kafkas/status` endpoint.

**~~NOTE: Branch is based on changes from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1344. This should only be merged once this has been merged to main. Any changes in relation to 3ab2d9f94366698babaeeebfdccc223f7caa2c6c should be addressed in #1344~~**

Related JIRA issue: [MGDSTRM-9782](https://issues.redhat.com/browse/MGDSTRM-9782)

## Verification Steps
* Unit and integration tests passing
* [Optional] test with Fleetshard operator
  * Suspend a Kafka instance using the admin endpoint (request body should have `suspended: true`)
  * Kafka should reach `suspended` state
  * Resume the `suspended` Kafka by updating it using the admin endpoint (request body should have `suspended: false`)
  * Kafka should reach `ready` state (or `failed`, if the FSO reports an error when resuming the Kafka instance) 

- [ ] Code Review completed
- [ ] ~~Required metrics/dashboards/alerts have been added (or PR created).~~ (will be done as part of [MGDSTRM-9937](https://issues.redhat.com/browse/MGDSTRM-9937))

### DIFF
--- a/internal/kafka/internal/services/data_plane_kafka.go
+++ b/internal/kafka/internal/services/data_plane_kafka.go
@@ -372,6 +372,7 @@ func (d *dataPlaneKafkaService) setKafkaClusterDeleting(kafka *dbapi.KafkaReques
 	return nil
 }
 
+// reassigns a Kafka instance to another data plane cluster. It only reassigns Kafka instances in a 'provisioning' state.
 func (d *dataPlaneKafkaService) reassignKafkaCluster(kafka *dbapi.KafkaRequest) *serviceError.ServiceError {
 	if kafka.Status == constants2.KafkaRequestStatusProvisioning.String() {
 		// If a Kafka cluster is rejected by the kas-fleetshard-operator, it should be assigned to another OSD cluster (via some scheduler service in the future).
@@ -389,6 +390,7 @@ func (d *dataPlaneKafkaService) reassignKafkaCluster(kafka *dbapi.KafkaRequest) 
 	return nil
 }
 
+// unassigns a Kafka instance from a data plane cluster. This is only done for Kafka instances in a 'provisioning' state.
 func (d *dataPlaneKafkaService) unassignKafkaFromDataplaneCluster(kafka *dbapi.KafkaRequest) *serviceError.ServiceError {
 	if kafka.Status == constants2.KafkaRequestStatusProvisioning.String() {
 		logger.Logger.Infof("kafka %s is being unassigned from cluster %s", kafka.ID, kafka.ClusterID)
@@ -420,6 +422,7 @@ func (d *dataPlaneKafkaService) checkKafkaRequestCurrentStatus(kafka *dbapi.Kafk
 	return matchStatus, nil
 }
 
+// stores routes reported by data plane to the database if not already persisted
 func (d *dataPlaneKafkaService) persistKafkaRoutes(kafka *dbapi.KafkaRequest, kafkaStatus *dbapi.DataPlaneKafkaStatus, cluster *api.Cluster) *serviceError.ServiceError {
 	if kafka.Routes != nil {
 		logger.Logger.V(10).Infof("skip persisting routes for Kafka %s as they are already stored", kafka.ID)
--- a/internal/kafka/internal/services/data_plane_kafka.go
+++ b/internal/kafka/internal/services/data_plane_kafka.go
@@ -41,6 +41,7 @@ const (
 	statusRejectedClusterFull managedKafkaStatus         = "rejectedClusterFull"
 	statusDeleted             managedKafkaStatus         = "deleted"
 	statusUnknown             managedKafkaStatus         = "unknown"
+	statusSuspended           managedKafkaStatus         = "suspended"
 	strimziUpdating           string                     = "StrimziUpdating"
 	kafkaUpdating             string                     = "KafkaUpdating"
 	kafkaIBPUpdating          string                     = "KafkaIbpUpdating"
@@ -150,14 +151,32 @@ func (d *dataPlaneKafkaService) processRealKafkaDeployment(ks *dbapi.DataPlaneKa
 		log.Warningf("clusterId for kafka cluster %s does not match clusterId. kafka clusterId = %s :: clusterId = %s", kafka.ID, kafka.ClusterID, cluster.ClusterID)
 		return
 	}
+
+	// Notes on state transitions
+	//  - 'suspending' state can only be set by an admin user from a 'ready' state via the /admin/kafkas/ endpoint.
+	//     This must only transition to 'resuming', 'suspended' or 'deprovision'.
+	//     - FSO will only send the status 'suspended' if the Kafka instance was already in a 'suspending' or 'suspended' state.
+	//     - FSO will not change the status of the ManagedKafka CR prior to transitioning to 'suspended' unless an error occurs.
+	//       e.g. If the Kafka instance was in a 'ready' state previously, FSO will keep reporting its status as 'ready' until it
+	//            finally reports 'suspended'. If an error occurs, the 'Ready' condition of the CR will have the values 'Status=False,Reason=Error'.
+	//  - 'resuming' state can only be set by an admin user from a 'suspending' or 'suspended' state via the /admin/kafkas/ endpoint.
+	//     This must only transition to 'ready', 'failed' or 'deprovision'.
+	//  - 'suspended' state must only transition to 'resuming' or 'deprovision'.
+	//  - 'deprovision' state is set by the user. This must only transition to 'deleting' or 'failed'.
+	//     - FSO will only send the status 'deleted' if the Kafka instance was already in a 'deprovisioning' state.
+	//  - 'failed' (or 'error') state may occur at any time.
+	//     - KFM must never transition a Kafka instance to 'failed' from 'suspending' and 'suspended' states.
+	//  - Routes should only be created once. They will remain uncahnged and will continue to be published in the status even if the Kafka instance was suspended.
 	var e *serviceError.ServiceError
 	switch s := d.getManagedKafkaStatus(ks); s {
 	case statusReady:
-		// Store the routes (and create them) when Kafka is ready. By the time it is ready, the routes should definitely be there.
-		e = d.persistKafkaRoutes(kafka, ks, cluster)
-		if e == nil {
-			kafka.AdminApiServerURL = ks.AdminServerURI
-			e = d.setKafkaClusterReady(kafka)
+		if kafka.Status != constants2.KafkaRequestStatusSuspending.String() && kafka.Status != constants2.KafkaRequestStatusSuspended.String() {
+			// Store the routes (and create them) when Kafka is ready. By the time it is ready, the routes should definitely be there.
+			e = d.persistKafkaRoutes(kafka, ks, cluster)
+			if e == nil {
+				kafka.AdminApiServerURL = ks.AdminServerURI
+				e = d.setKafkaClusterReady(kafka)
+			}
 		}
 	case statusInstalling:
 		// Store the routes (and create them) if they are available at this stage to lessen the length of time taken to provision the Kafka.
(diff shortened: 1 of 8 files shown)


internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -150,30 +151,57 @@ func (d *dataPlaneKafkaService) processRealKafkaDeployment(ks *dbapi.DataPlaneKa
...
 	case statusRejected:
-		e = d.reassignKafkaCluster(kafka)
+		if kafka.Status != constants2.KafkaRequestStatusSuspending.String() {
+			e = d.reassignKafkaCluster(kafka)
+		}
 	case statusRejectedClusterFull:
-		e = d.unassignKafkaFromDataplaneCluster(kafka)
+		if kafka.Status != constants2.KafkaRequestStatusSuspending.String() {
+			e = d.unassignKafkaFromDataplaneCluster(kafka)
+		}
```
JameelB, line 199: Is it possible for fleetshard to report these two statuses when a Kafka is in a `suspended` state?

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -150,30 +151,57 @@ func (d *dataPlaneKafkaService) processRealKafkaDeployment(ks *dbapi.DataPlaneKa
...
 	case statusError:
-		// when getStatus returns statusError we know that the ready
-		// condition will be there so there's no need to check for it
-		readyCondition, _ := ks.GetReadyCondition()
-		e = d.setKafkaClusterFailed(kafka, readyCondition.Message)
+		if kafka.Status != constants2.KafkaRequestStatusSuspending.String() || kafka.Status != constants2.KafkaRequestStatusSuspended.String() {
+			// when getStatus returns statusError we know that the ready
+			// condition will be there so there's no need to check for it
+			readyCondition, _ := ks.GetReadyCondition()
+			e = d.setKafkaClusterFailed(kafka, readyCondition.Message)
```
JameelB, line 188: Do we also want to set the `FailedReason` in the db when it fails during `suspending` and `suspended` state? This will be seen in the API responses, it's part of the KafkaRequest object.

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -150,30 +151,57 @@ func (d *dataPlaneKafkaService) processRealKafkaDeployment(ks *dbapi.DataPlaneKa
...
+			}
 		}
 	case statusInstalling:
-		// Store the routes (and create them) if they are available at this stage to lessen the length of time taken to provision the Kafka.
-		// The routes list will either be empty or complete.
-		e = d.persistKafkaRoutes(kafka, ks, cluster)
+		if kafka.Status != constants2.KafkaRequestStatusSuspending.String() {
+			// Store the routes (and create them) if they are available at this stage to lessen the length of time taken to provision the Kafka.
+			// The routes list will either be empty or complete.
+			e = d.persistKafkaRoutes(kafka, ks, cluster)
```
JameelB, line 181: @MikeEdgar can the routes change once a Kafka is resumed? Right now, we don't support overwriting the routes that's already persisted in the database. Do we need to accommodate for this?

MikeEdgar: The routes will remain unchanged and will continue to be published in the status.

internal/kafka/internal/services/data_plane_kafka.go
machi1990: Judging from the capacity reporting conversation, when the Kafka is in `provisioning` state, it'll first report rejected due to fullness before the cluster is full. @MikeEdgar is this something that going to be honored? 

> Should we allow Kafkas to be re-assigned to another cluster when it's in a suspending or suspended state?

No, we shouldn't support for this as we risk loosing data if the kafka state was ready before it was suspended. 
Based on what I am saying above, we shouldn't really be concerned about re-assigning the suspended/suspending kafka onto another cluster. We could add a guard against these states.

internal/kafka/internal/services/data_plane_kafka.go
machi1990: I think just logging the error is fine and we could action from there. If the error is to occur, it is not  an error due to a user action/inaction so there is nothing they could do to take the kafka from the failed state.

internal/kafka/internal/services/data_plane_kafka.go
MikeEdgar: The rejected status will take priority over the suspended status. I think that from KFM's perspective, it may appear that the cluster transitions from installing -> suspending -> rejected. In that case FSO would only notify of the rejection and those other status values would originate in the control plane.

JameelB: if we don't allow for re-assignment, how would this affect the upgrade? In Data Plane, FSO will resume suspended Kafkas in order to upgrade it (while continuing to report status reason as 'Suspended'). If there's no capacity on the cluster, would this mean the upgrade will always fail?

internal/kafka/internal/services/data_plane_kafka.go
JameelB: Thanks for confirming @MikeEdgar!

internal/kafka/internal/services/data_plane_kafka.go
JameelB: Thanks @machi1990, that makes sense :)

internal/kafka/internal/services/data_plane_kafka.go
machi1990: > I wonder to avoid this scenario, if it is the case, would it be best to not be able to suspend a Kafka instance unless it was in a ready state previously?

I think this makes sense. What do you think? @MikeEdgar

JameelB: cc @pawelpaszki

internal/kafka/internal/handlers/validation.go
```diff
@@ -260,8 +260,9 @@ func ValidateKafkaUpdateFields(kafkaUpdateRequest *private.KafkaUpdateRequest) h
 			stringSet(&kafkaUpdateRequest.KafkaVersion) ||
 			stringSet(&kafkaUpdateRequest.KafkaIbpVersion) ||
 			stringSet(&kafkaUpdateRequest.DeprecatedKafkaStorageSize) ||
-			stringSet(&kafkaUpdateRequest.MaxDataRetentionSize)) {
-			return errors.FieldValidationError("Failed to update Kafka Request. Expecting at least one of the following fields: strimzi_version, kafka_version, kafka_ibp_version, kafka_storage_size or max_data_retention_size to be provided")
+			stringSet(&kafkaUpdateRequest.MaxDataRetentionSize) ||
+			arrays.IsNotNilPredicate(kafkaUpdateRequest.Suspended)) {
```
ziccardi, line 264: Why are we using the `IsNotNilPredicate`?

internal/kafka/test/integration/admin_kafka_test.go
```diff
@@ -1146,6 +1155,148 @@ func TestAdminKafka_Update(t *testing.T) {
...
+				ctx: func(h *coreTest.Helper) context.Context {
+					return NewAuthenticatedContextForAdminEndpoints(h, []string{testFullRole})
+				},
+				kafkaID: sampleKafkaID1,
+				kafkaUpdateRequest: adminprivate.KafkaUpdateRequest{
+					Suspended: &falseB,
+				},
+			},
+			verifyResponse: func(result adminprivate.Kafka, resp *http.Response, err error) {
+				g.Expect(err).To(gomega.BeNil())
```
ziccardi, line 1170:
```suggestion
				g.Expect(err).NotTo(gomega.HaveOccurred())
```

internal/kafka/test/integration/admin_kafka_test.go
```diff
@@ -1146,6 +1155,148 @@ func TestAdminKafka_Update(t *testing.T) {
...
+				ctx: func(h *coreTest.Helper) context.Context {
+					return NewAuthenticatedContextForAdminEndpoints(h, []string{testFullRole})
+				},
+				kafkaID: deletingKafkaID,
+				kafkaUpdateRequest: adminprivate.KafkaUpdateRequest{
+					Suspended: &trueB,
+				},
+			},
+			verifyResponse: func(result adminprivate.Kafka, resp *http.Response, err error) {
+				g.Expect(err).NotTo(gomega.BeNil())
```
ziccardi, line 1242:
```suggestion
				g.Expect(err).To(HaveOccurred())
```

internal/kafka/test/integration/admin_kafka_test.go
JameelB: cc @pawelpaszki, this change is from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1344

internal/kafka/test/integration/admin_kafka_test.go
JameelB: cc @pawelpaszki, this change is from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1344

internal/kafka/internal/handlers/validation.go
JameelB: Note that this change is from #1344 

Suspended can be undefined. Now that you mention it here, we may not want to add this in this check as we don't have to specify this field. Moving conversation to other PR.

internal/kafka/internal/services/data_plane_kafka.go
JameelB: This was addressed #1344. Thanks guys!

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -452,6 +455,9 @@ func (d *dataPlaneKafkaService) getManagedKafkaStatus(status *dbapi.DataPlaneKaf
 				return statusReady
 			}
 			if strings.EqualFold(c.Status, "Unknown") {
+				if strings.EqualFold(c.Reason, "Installing") {
+					return statusResuming
```
JameelB, line 462: Added the new `resuming` state here. In this case though, we don't do anything since the Kafka should already have status `resuming`. We'll only log it as seen in L#207 above.

machi1990 (commented):
the code change for status transitions looks good to me.
On top of the unit test, might be good to add an integration test.

internal/kafka/internal/services/data_plane_kafka.go
JameelB: Looks like this is no longer the case for `Resuming`. FSO will return Ready condition with `Status=False, Reason=Installing`. Will remove this.

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -150,14 +151,30 @@ func (d *dataPlaneKafkaService) processRealKafkaDeployment(ks *dbapi.DataPlaneKa
...
 		return
 	}
+
+	// Notes on state transitions
+	//  - 'suspending' state is set by the user from a 'ready' state. This must only transition to 'resuming', 'suspended' or 'deprovision'.
+	//     - FSO will only send the status 'suspended' if the Kafka instance was already in a 'suspending' or 'suspended' state.
+	//     - FSO will not change the status of the ManagedKafka CR prior to transitioning to 'suspended' unless an error occurs.
+	//       e.g. If the Kafka instance was in a 'ready' state previously, FSO will keep reporting its status as 'ready' until it
+	//            finally reports 'suspended'. If an error occurs, the 'Ready' condition of the CR will have the values 'Status=False,Reason=Error'.
+	//  - 'resuming' state is set by the user from a 'suspending' or 'suspended' state. This must only transition to 'ready', 'failed' or 'deprovision'.
```
pawelpaszki, line 161: technically resuming can't be set by any user (only via admin endpoint). do you think we could rewrite it a little?

internal/kafka/test/integration/admin_kafka_suspension_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	_, err = common.WaitForKafkaToReachStatus(publicClientCtx, kafkatest.TestServices.DBFactory, publicClient, publicKafkaReq.Id, kafkaconstants.KafkaRequestStatusPreparing)
+	g.Expect(err).ToNot(gomega.HaveOccurred(), "error waiting for kafka to reach status %q", kafkaconstants.KafkaRequestStatusPreparing)
+
+	// suspend Kafka instance
+	// set 'suspending: true' on a non-ready Kafka instance: should not update the status
+	suspendKafkaRequestPayload := private.KafkaUpdateRequest{
+		Suspended: &[]bool{true}[0],
+	}
+	privateKafkaReq, resp, err := adminClient.DefaultApi.UpdateKafkaById(adminClientCtx, publicKafkaReq.Id, suspendKafkaRequestPayload)
```
JameelB, line 93: I wonder if it makes sense to return an error if a user tries to suspend a Kafka from a non-ready state? Right now we accept it but do nothing. It may be a weird user experience as the request is successful but there would be no changes to the Kafka instance. 

wdyt @pawelpaszki @machi1990 @MikeEdgar?

machi1990: > It may be a weird user experience as the request is successful but there would be no changes to the Kafka instance.

+1, makes sense to return a 400 error.

JameelB: Thanks, I can probably add those changes in this PR. Unless it's preferable to have a separate PR for it?

machi1990: I think this PR is fine.

internal/kafka/internal/services/data_plane_kafka.go
JameelB: good point, will update this :)

internal/kafka/test/integration/admin_kafka_suspension_test.go
MikeEdgar: +1 to reject the request

machi1990 (approved):
lgtm

Thanks for adding the various test cases!

I'll leave the final approval to @MikeEdgar, @valeriiashapoval  and @pawelpaszki

pawelpaszki (approved):

internal/kafka/internal/services/data_plane_kafka.go
MikeEdgar: @JameelB - after some additional testing, I've found that the routes are being set to an empty array in the status while the MK is suspended. The reason is that kas-fleetshard operator uses the Kafka broker pods to determine the AZ for each route, and they are not available when suspended, of course. After resuming the cluster, the routes will again be populated. Please let me know if this could be a problem on the control plane side.

JameelB: Thanks for the update @MikeEdgar. I'll have a look to check if this causes a problem on our side.

PR #1357, opened 2022-10-21
feat: set route53 regional endpoints per cloud provider

miguelsorianod:
## Description
When kafka instances are created from cloud providers different than AWS, AWS Route 53 is also used to create DNS entries when kafka external names are enabled. Before this PR the AWS client was instantiated using a region value obtained from the kafka request being managed. The issue with that approach is that kafka instances allocated in cloud providers different than AWS, for example GCP, have different region names, which don't exist in AWS.

This changes the logic of the methods that make use of Route53 to make sure a valid AWS region is specified, no matter the cloud provider of the kafka instance being managed.

We define a AWS Route53 region for each cloud provider, independently on the cloud provider's region where the kafka is allocated. This is acceptable due to most Route53 features are global and not region-specific. The region in those case is used merely to access to a regional AWS API endpoint. We are currently not using regional features of Route53 so we can use the implemented approach.
AWS and GCP route53 regions have been defined to use `us-east-1`.

The option to do a mapping between cloud providers with AWS regions like for example mapping GCP regions to AWS regions  was evaluated and deem it as not feasible nor maintainable. The reason for it are
* We would need to maintain them over time which is very costly
* Applying the concept itself is not possible: Not all regions exist in all cloud providers: For example, GCP has a region `europe-north1-a` (Hamina, Finland, Europe) whereas AWS does not have any region named similarly nor one located in Finland so the mapping is not possible.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -14,6 +14,7 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
 	constants2 "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/cloudproviders"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/kafkas/types"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
@@ -939,12 +940,17 @@ func (k *kafkaService) ChangeKafkaCNAMErecords(kafkaRequest *dbapi.KafkaRequest,
 
 	domainRecordBatch := buildKafkaClusterCNAMESRecordBatch(routes, string(action))
 
-	// Create AWS client with the region of this Kafka Cluster
 	awsConfig := aws.Config{
 		AccessKeyID:     k.awsConfig.Route53AccessKey,
 		SecretAccessKey: k.awsConfig.Route53SecretAccessKey,
 	}
-	awsClient, err := k.awsClientFactory.NewClient(awsConfig, kafkaRequest.Region)
+
+	route53Region, err := k.getRoute53RegionFromKafkaRequest(kafkaRequest)
+	if err != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "error getting route 53 region from kafka request")
+	}
+
+	awsClient, err := k.awsClientFactory.NewClient(awsConfig, route53Region)
 	if err != nil {
 		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "Unable to create aws client")
 	}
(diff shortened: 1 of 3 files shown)


internal/kafka/internal/services/kafka.go
```diff
@@ -944,7 +945,13 @@ func (k *kafkaService) ChangeKafkaCNAMErecords(kafkaRequest *dbapi.KafkaRequest,
...
 		SecretAccessKey: k.awsConfig.Route53SecretAccessKey,
 	}
-	awsClient, err := k.awsClientFactory.NewClient(awsConfig, kafkaRequest.Region)
+
+	route53Region, err := k.getRoute53RegionFromKafkaRequest(kafkaRequest)
+	if err != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "error getting route 53 region from kafka request")
+	}
+
+	awsClient, err := k.awsClientFactory.NewClient(awsConfig, route53Region)
```
miguelsorianod, line 954: I was thinking that it might be a good idea to change the type of the aws client and instead of being `type Client` rename it to `type Route53Client` as the methods in there are only related to Route53. If in the future we need to interact with another AWS API for example EC2 then we would create another interface `type EC2Client` and so on. That would separate the interfaces by AWS service, similarly to what AWS Go SDK does

pb82: I think it is useful, we will likely want to offer Google Cloud DNS at some point in the future, at least for GCP clusters.

pb82 (approved):
LGTM

machi1990 (approved):

davidffrench (approved):

internal/kafka/internal/services/kafka.go
```diff
@@ -944,7 +945,13 @@ func (k *kafkaService) ChangeKafkaCNAMErecords(kafkaRequest *dbapi.KafkaRequest,
 		AccessKeyID:     k.awsConfig.Route53AccessKey,
 		SecretAccessKey: k.awsConfig.Route53SecretAccessKey,
 	}
-	awsClient, err := k.awsClientFactory.NewClient(awsConfig, kafkaRequest.Region)
+
+	route53Region, err := k.getRoute53RegionFromKafkaRequest(kafkaRequest)
```
machi1990, line 949: Can we remove this comment https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/656bde30a3a4d827387b25513638b38ceb1705d7/internal/kafka/internal/services/kafka.go#L943 as well?

miguelsorianod: Removed

internal/kafka/internal/services/kafka.go
```diff
@@ -1261,3 +1273,25 @@ func (k *kafkaService) AssignBootstrapServerHost(kafkaRequest *dbapi.KafkaReques
...
+// There are some parts of the Route53 functionality that are regional.
+// For what we perform which is create hosted zones and entries in them
+// that is a global functionality.
+// See: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/disaster-recovery-resiliency.html
+// If at some point we end up needing Route53 regional functionalities this
+// mechanism should be reevaluated
+func (k *kafkaService) getRoute53RegionFromKafkaRequest(kafkaRequest *dbapi.KafkaRequest) (string, error) {
+	switch kafkaRequest.CloudProvider {
+	case cloudproviders.AWS.String():
+		return aws.DefaultAWSRoute53Region, nil
```
machi1990, line 1291: Quick question: Will this affect the deletion of existing Kafkas (whose region is different from `us-east-1`) i.e there won't be any issues with deleting an already created CNAMEs on other (different from the default) AWS regions? 

Based on the comment on this method, I think we should be good but thought I would ask to get confirmation.

@miguelsorianod

internal/kafka/internal/services/kafka.go
machi1990: Thanks @miguelsorianod

internal/kafka/internal/services/kafka.go
miguelsorianod: We should be good

PR #1355, opened 2022-10-20
Reduce number of calls made from integration test to the SSO service

machi1990:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-7819. 
This reduce calls to sso from 23 to only 1 call to provision fleetshard service account. The
improvement should be of the same margin for the provisioning of a ClientID used to setup cluster
IDP configuration.

*Note:* The PR is only concerned about offering an improvement on fleetshard service account creation and cluster IDP clientID creation.

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has been created for changes required on the client side

PR #1359, opened 2022-10-24
Put in place config and code change to allow the midstream pipelines to run against RH SSO as the SSO provider

machi1990:
## Description

Closes https://issues.redhat.com/browse/MGDSTRM-7819

This allow to run the midstream CI using RH SSO as the provider without leaking service accounts and with an even reduced load to the SSO service thanks to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1355.

See the linked documentation in https://issues.redhat.com/browse/MGDSTRM-7819 for further background on this work. 

## Verification Steps

Run the `pr_chech.sh` file providing it with the needed credentials:
To successfully run that script, the following has to be done:
1. Change https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/dd5b40ab62cad808f26b6fcd39aeea15e5579751/docker/Dockerfile_template_mocked#L1 and https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/dd5b40ab62cad808f26b6fcd39aeea15e5579751/docker/Dockerfile_template#L1 to `FROM postgres:13`
2. Make sure you've the needed credentials. See the pr_check.sh for the needed environment variables. The env var in  https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/dd5b40ab62cad808f26b6fcd39aeea15e5579751/pr_check.sh#L58-L61 are not so mandatory for this verification as they are only needed to publish the test output to the test portal, so dummy values can be used.
3. Run the script with first, `OCM_ENV` set to `integration` and then `OCM_ENV` set to `development` so that we are sure that each mode works as expected.

- [ ] Code Review completed


pr_check_docker.sh
```diff
@@ -23,7 +23,7 @@ PGDATA=/var/lib/postgresql/data /usr/lib/postgresql/*/bin/pg_ctl -w stop
 PGDATA=/var/lib/postgresql/data /usr/lib/postgresql/*/bin/pg_ctl start -o "-c listen_addresses='*' -p 5432"
 
 # check the code. Then run the unit and integration tests and cleanup cluster (if running against real OCM)
-make -k lint verify test test/integration test/report-portal/push test/cluster/cleanup 
+make -k lint verify test test/integration/kafka test/report-portal/push test/cluster/cleanup
```
miguelsorianod, line 26: That means we don't run connectors test anymore as part of the CI / release process. Is that intentional?

scripts/mas_sso_config.sh
```diff
@@ -69,8 +69,6 @@ echo "$svcUserId"
 FINAL=$(curl -sk --data-raw '[{"id": '"$manageUser"',"name": "manage-users"},{"id": '"$manageRealm"',"name": "manage-realm"},{"id": '"$manageClients"',"name": "manage-clients"}]' --header "Content-Type: application/json" --header "Authorization: Bearer $TOKEN" "$KEYCLOAK_URL"/auth/admin/realms/rhoas/users/"$svcUserId"/role-mappings/clients/"$realmMgmtClientId")
 echo "$FINAL"
 
-
```
miguelsorianod, line 72: This script is called mas_sso_config. There are other references in other parts. Is MAS SSO still used for something? should we be renaming elements?

docker/Dockerfile_template
```diff
@@ -20,10 +20,8 @@ ENV ROUTE53_ACCESS_KEY <aws_route53_access_key>
 ENV ROUTE53_SECRET_ACCESS_KEY <aws_route53_secret_access_key>
 ENV OCM_OFFLINE_TOKEN <ocm_offline_token>
 ENV OBSERVATORIUM_CONFIG_ACCESS_TOKEN <observatorium_config_access_token>
-ENV MAS_SSO_CLIENT_ID <mas_sso_client_id>
-ENV MAS_SSO_CLIENT_SECRET <mas_sso_client_secret>
-ENV OSD_IDP_MAS_SSO_CLIENT_ID <osd_idp_mas_sso_client_id>
```
miguelsorianod, line 25: What is then used for IDP SSO?

internal/kafka/internal/environments/development.go
```diff
@@ -18,8 +18,10 @@ func NewDevelopmentEnvLoader() environments.EnvLoader {
 		"enable-deny-list":                                "true",
 		"enable-access-list":                              "false",
 		"enable-instance-limit-control":                   "false",
-		"mas-sso-base-url":                                "https://identity.api.stage.openshift.com",
+		"mas-sso-base-url":                                "http://127.0.0.1:8180",
```
miguelsorianod, line 21: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1359/files#r1004224020

pr_check_docker.sh
machi1990: This change is intentional. 

Connectors test are still run as part of the Github CI. 
The only common parts we've with them is that we share the repository (this repo) and the code base so running the connectors integration test in Github CI makes sense. 
The reason for this change is that, the commonality between KFM and COS-Fleet-manager stops at code level, the connector team have a different release process (which is based on https://github.com/bf2fc6cc711aee1a0c2a/cos-fleet-manager ) and as far as I am aware, they are not part of the different downstream pipelines dedicated for Managed Kafka. 

/cc @lburgazzoli @dhirajsb to chime in if this should be reverted.

scripts/mas_sso_config.sh
miguelsorianod: Is there some difference between mas sso and native keycloak?

docker/Dockerfile_template
machi1990: mas-sso is still used for IDP.
But since the midstream test are running against rh_sso, we cannot configure the IDP with it hence this removal

scripts/mas_sso_config.sh
machi1990: Yes, I think the real mas sso image has some extra capabilities that the native keycloak container doesnt have e.g enablement of  searchAttributes - to allow filtering service accounts (keycloak clients) by client's attributes. 
There could be other which I am missing @akoserwal can you think of others?

I believe we are using the native keycloak here and we do not have the mas-sso capabilities.

docker/Dockerfile_template
miguelsorianod: That means there's no way to access the cluster using our users in case we need it right?

machi1990: > That means there's no way to access the cluster using our users in case we need it right?

I think this has always been the case even with the current setup, we do not have all the right access to the clusters since we are not part of the default user list: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/35cf0417534cc5bb5b390f6d97586cb66f811549/config/kafka-sre-user-list.yaml and https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/35cf0417534cc5bb5b390f6d97586cb66f811549/config/read-only-user-list.yaml

And I've been always using kubeconfig to access the nightly cluster. 
With the change in this PR https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/35cf0417534cc5bb5b390f6d97586cb66f811549/internal/kafka/test/helper.go#L72, IDP configuration will be disabled completely and kubeconfig access option will be the default.

pawelpaszki:
@machi1990 - I have verified your changes against both environments and both runs were successful

machi1990:
> @machi1990 - I have verified your changes against both environments and both runs were successful

Thank you @pawelpaszki 

Let's wait for @akoserwal review and a reply from connectors team regarding https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1359#discussion_r1004222827 before merging.

pr_check_docker.sh
dhirajsb: While it's true that connectors don't share any build output from kfm downstream, the integration tests can only be run as part of the kfm test process downstream. Since c-f-m pulls k-f-m as a dependency, there's no way to run these integration tests as part of c-f-m downstream CI process. 
Is there an issue in the connector integration tests that's preventing them to be run downstream?

machi1990: @dhirajsb thanks for the comments. I've tagged you in a documentation that outlines the intention behind this: 

In short summary, we are to stop using mas_sso rhoas realm in these pipelines and start using rh sso stage env. Is that possible with the current connectors test?

machi1990: > ..the integration tests can only be run as part of the kfm test process downstream. Since c-f-m pulls k-f-m as a dependency, there's no way to run these integration tests as part of c-f-m downstream CI process.

Would you happen to have a documentation about how this works downstream for the connector team?
"Ideally" and "in my imagination", the way kfm and cfm releases processes should be decouple from one another. 
i.e any failure to kfm code shouldn't impact cfm releases and vice versa. 

One way to run the connectors integration test downstream would still be when the kfm version bumps occurs. The cfm pipeline will then clone the "kfm/common code repo" and run the connectors integration test before doing any releases.

machi1990: Following a discussion in private, I've reverted this change. 

Thanks for calling it out @miguelsorianod @dhirajsb

machi1990: Just to note, I'll still like to progress the discussion around decoupling the the KFM & CFM release process. I've opened a separate discussion topic on that.

dhirajsb (changes requested):
Can't we switch the sso provider type in environments to `redhat_sso`?

machi1990:
> Can't we switch the sso provider type in environments to `redhat_sso`?

No, the default ones are what we use in Github CI. 
For midstream pipelines, the switch is done here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d0a3a1bf148422379aeb4ead556c656c561769ab/pr_check.sh#L76-L79

scripts/mas_sso_config.sh
machi1990: > Is MAS SSO still used for something? should we be renaming elements?

@miguelsorianod I've actually remembered that some time ago I created this https://issues.redhat.com/browse/MGDSTRM-3890 to rename the sso related flags (to be fair, everything) so that they are generic enough. I think the work could be progressed there?

dhirajsb:
> 
> No, the default ones are what we use in Github CI and local development. And we do not want to use rh sso in local development by default nor use it in Github CI. For midstream pipelines, the switch is done here:
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/d0a3a1bf148422379aeb4ead556c656c561769ab/pr_check.sh#L76-L79

I was wondering about just switching to using the to `redhat_sso` provider type not the rh sso server per se. Connector tests I ran locally against local keycloak were actually able to use `redhat_sso` client.

dhirajsb (approved):
Since midstream tests using `redhat_sso` provider, it's not critical that upstream is still using `mas_sso`. 

lgtm

PR #1363, opened 2022-10-26
chore: added Suspended, Suspending and Resuming statuses to the request status count metric

ziccardi:

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
+++ b/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
@@ -26,6 +26,9 @@ var kafkaMetricsStatuses = []constants2.KafkaStatus{
 	constants2.KafkaRequestStatusDeprovision,
 	constants2.KafkaRequestStatusDeleting,
 	constants2.KafkaRequestStatusFailed,
+	constants2.KafkaRequestStatusSuspended,
+	constants2.KafkaRequestStatusSuspending,
+	constants2.KafkaRequestStatusResuming,
 }
 
 // KafkaManager represents a kafka manager that periodically reconciles kafka requests
--- a/internal/kafka/test/integration/admin_kafka_suspension_test.go
+++ b/internal/kafka/test/integration/admin_kafka_suspension_test.go
@@ -26,6 +26,12 @@ import (
 )
 
 func TestAdminKafka_KafkaSuspension(t *testing.T) {
+
+	type metricValue struct {
+		metric kafkaconstants.KafkaStatus
+		value  string
+	}
+
 	g := gomega.NewWithT(t)
 
 	ocmServer := mocks.NewMockConfigurableServerBuilder().Build()
@@ -50,7 +56,6 @@ func TestAdminKafka_KafkaSuspension(t *testing.T) {
 		d.DataPlaneClusterScalingType = config.ManualScaling
 		d.ClusterConfig = config.NewClusterConfig(clusterList)
 	})
-
 	defer teardown()
 
 	// run test only on mock mode - fleetshard sync will always be mocked so there's no point running against real env.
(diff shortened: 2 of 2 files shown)


internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -26,6 +26,9 @@ var kafkaMetricsStatuses = []constants2.KafkaStatus{
 	constants2.KafkaRequestStatusDeprovision,
 	constants2.KafkaRequestStatusDeleting,
 	constants2.KafkaRequestStatusFailed,
+	constants2.KafkaRequestStatusSuspended,
+	constants2.KafkaRequestStatusSuspending,
+	constants2.KafkaRequestStatusResuming,
```
machi1990, line 31: Might be good to add an integration test for this

ziccardi: Added check into existing integration test

internal/kafka/test/integration/admin_kafka_suspension_test.go
```diff
@@ -109,18 +110,26 @@ func TestAdminKafka_KafkaSuspension(t *testing.T) {
 	}
 	g.Expect(err).ToNot(gomega.HaveOccurred(), "failed to suspend Kafka instance")
 	g.Expect(privateKafkaReq.Status).To(gomega.Equal(kafkaconstants.KafkaRequestStatusSuspending.String()))
+	err = common.WaitForMetricToBePresent(h, t, metrics.KafkaRequestsStatusCount, kafkaconstants.KafkaRequestStatusSuspending.String())
```
machi1990, line 113: can we also check for the metric value? 

The kafka_requests_count metric always produced even if we do not have a Kafka in a given state i.e it'll report `0` if a state doesn't have a Kafka.

What will be good is to check for a particular value: for example
if we change the status to updating the we should expect the value to `1` for that state and `0` for all the other states if we only have one Kafka in the database.
```suggestion
	err = common.WaitForMetricToBePresent(h, t, metrics.KafkaRequestsStatusCount, kafkaconstants.KafkaRequestStatusSuspending.String(), "1")
	err = common.WaitForMetricToBePresent(h, t, metrics.KafkaRequestsStatusCount, kafkaconstants.KafkaRequestStatusSuspended.String(), "0")
	err = common.WaitForMetricToBePresent(h, t, metrics.KafkaRequestsStatusCount, kafkaconstants.KafkaRequestStatusResuming.String(), "0")
```

Same for all the other added checks.

internal/kafka/test/integration/admin_kafka_suspension_test.go
```diff
@@ -109,18 +110,26 @@ func TestAdminKafka_KafkaSuspension(t *testing.T) {
...
+
 	privateKafkaReq, resp, err = adminClient.DefaultApi.GetKafkaById(adminClientCtx, privateKafkaReq.Id)
 	if resp != nil {
 		resp.Body.Close()
 	}
 	g.Expect(err).ToNot(gomega.HaveOccurred(), "failed to get Kafka instance from admin get endpoint")
 	g.Expect(privateKafkaReq.Status).To(gomega.Equal(kafkaconstants.KafkaRequestStatusSuspended.String()))
 
+	err = common.WaitForMetricToBePresent(h, t, metrics.KafkaRequestsStatusCount, kafkaconstants.KafkaRequestStatusResuming.String())
+	g.Expect(err).ToNot(gomega.HaveOccurred())
```
machi1990, line 131: This check has to go in line https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e736b24f36667f564497721800dec67cef5a2a92/internal/kafka/test/integration/admin_kafka_suspension_test.go#L152 after the Kafka has been set to `resuming` via the admin call: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e736b24f36667f564497721800dec67cef5a2a92/internal/kafka/test/integration/admin_kafka_suspension_test.go#L146-L151

At the moment, the test is passing since the metric is always reported for all the statuses but the metric value is `0`.
We want to make sure that we assert against the reported metric value as well as state on the other comment

machi1990 (changes requested):

machi1990:
Thanks for adding the integration test @ziccardi 

I've left two comments, the PR should be good to merge once they are addressed.

machi1990 (approved):

PR #1364, opened 2022-10-26
feat: expose metric with kafka status set timestamp

pawelpaszki:

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/data_plane_kafka.go
+++ b/internal/kafka/internal/services/data_plane_kafka.go
@@ -228,6 +228,7 @@ func (d *dataPlaneKafkaService) setKafkaClusterReady(kafka *dbapi.KafkaRequest) 
 	}
 	// only send metrics data if the current kafka request is in "provisioning" status as this is the only case we want to report
 	shouldSendMetric, err := d.checkKafkaRequestCurrentStatus(kafka, constants2.KafkaRequestStatusProvisioning)
+
 	if err != nil {
 		return err
 	}
@@ -243,6 +244,7 @@ func (d *dataPlaneKafkaService) setKafkaClusterReady(kafka *dbapi.KafkaRequest) 
 		metrics.IncreaseKafkaSuccessOperationsCountMetric(constants2.KafkaOperationCreate)
 		metrics.IncreaseKafkaTotalOperationsCountMetric(constants2.KafkaOperationCreate)
 	}
+
 	return nil
 }
 
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -79,6 +79,8 @@ type KafkaService interface {
 	// The Kafka Request in the database will be updated with a deleted_at timestamp.
 	Delete(*dbapi.KafkaRequest) *errors.ServiceError
 	List(ctx context.Context, listArgs *services.ListArguments) (dbapi.KafkaList, *api.PagingMeta, *errors.ServiceError)
+	// Lists all kafkas. As this returns all Kafka requests without need for authentication, this should only be used for internal purposes
+	ListAll() (dbapi.KafkaList, *errors.ServiceError)
 	GetManagedKafkaByClusterID(clusterID string) ([]managedkafka.ManagedKafka, *errors.ServiceError)
 	// GenerateReservedManagedKafkasByClusterID returns a list of reserved managed
 	// kafkas for a given clusterID. The number of generated reserved managed
@@ -423,6 +425,7 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
 	}
 
 	metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusAccepted, kafkaRequest.ID, kafkaRequest.ClusterID, time.Since(kafkaRequest.CreatedAt))
+
 	return nil
 }
 
@@ -471,7 +474,6 @@ func (k *kafkaService) PrepareKafkaRequest(kafkaRequest *dbapi.KafkaRequest) *er
 	if err := k.Update(updatedKafkaRequest); err != nil {
 		return errors.NewWithCause(errors.ErrorGeneral, err, "failed to update kafka request")
 	}
-
 	return nil
 }
 
@@ -720,6 +722,17 @@ func (k *kafkaService) Delete(kafkaRequest *dbapi.KafkaRequest) *errors.ServiceE
 	metrics.IncreaseKafkaSuccessOperationsCountMetric(constants2.KafkaOperationDelete)
 
 	return nil
+}
+
+// Lists all kafkas. As this returns all Kafka requests without need for authentication, this should only be used for internal purposes
+func (k *kafkaService) ListAll() (dbapi.KafkaList, *errors.ServiceError) {
+	var kafkaRequestList dbapi.KafkaList
+	dbConn := k.connectionFactory.New()
+	if err := dbConn.Find(&kafkaRequestList).Error; err != nil {
+		return kafkaRequestList, errors.NewWithCause(errors.ErrorGeneral, err, "unable to list kafka requests")
+	}
+
+	return kafkaRequestList, nil
 }
 
 // List returns all Kafka requests belonging to a user.
--- a/internal/kafka/internal/services/kafka_test.go
+++ b/internal/kafka/internal/services/kafka_test.go
@@ -1991,6 +1991,108 @@ func Test_kafkaService_List(t *testing.T) {
 	}
 }
 
+func Test_kafkaService_ListAll(t *testing.T) {
+	type fields struct {
+		connectionFactory *db.ConnectionFactory
+	}
+
+	tests := []struct {
+		name    string
+		fields  fields
+		want    dbapi.KafkaList
+		wantErr bool
+		setupFn func(dbapi.KafkaList)
+	}{
+		{
+			name: "success list all kafkas",
+			fields: fields{
+				connectionFactory: db.NewMockConnectionFactory(nil),
+			},
+			want: dbapi.KafkaList{
+				&dbapi.KafkaRequest{
+					Region:        testKafkaRequestRegion,
+					ClusterID:     testClusterID,
+					CloudProvider: testKafkaRequestProvider,
+					MultiAZ:       false,
+					Name:          "dummy-cluster-name",
+					Status:        "accepted",
+					Owner:         testUser,
+					Meta: api.Meta{
+						CreatedAt: time.Now(),
+						UpdatedAt: time.Now(),
+						DeletedAt: gorm.DeletedAt{Valid: true},
+					},
+				},
+				&dbapi.KafkaRequest{
+					Region:        testKafkaRequestRegion,
+					ClusterID:     testClusterID,
+					CloudProvider: testKafkaRequestProvider,
+					MultiAZ:       false,
+					Name:          "dummy-cluster-name2",
+					Status:        "accepted",
+					Owner:         testUser,
+					Meta: api.Meta{
+						CreatedAt: time.Now(),
+						UpdatedAt: time.Now(),
+						DeletedAt: gorm.DeletedAt{Valid: true},
+					},
+				},
+			},
+			wantErr: false,
+			setupFn: func(kafkaList dbapi.KafkaList) {
+				mocket.Catcher.Reset()
+
+				query := fmt.Sprintf(`SELECT * FROM "%s"`, kafkaRequestTableName)
+				response := converters.ConvertKafkaRequestList(kafkaList)
+				mocket.Catcher.NewMock().WithQuery(query).WithReply(response)
+				mocket.Catcher.NewMock().WithExecException().WithQueryException()
+			},
+		},
+		{
+			name: "fail: database returns an error",
+			fields: fields{
+				connectionFactory: db.NewMockConnectionFactory(nil),
+			},
+			want:    dbapi.KafkaList{},
+			wantErr: true,
+			setupFn: func(kafkaList dbapi.KafkaList) {
+				mocket.Catcher.Reset().NewMock().WithQuery("SELECT").WithQueryException()
+			},
+		},
+	}
+
+	for _, testcase := range tests {
+		tt := testcase
+
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			tt.setupFn(tt.want)
+			k := &kafkaService{
+				connectionFactory: tt.fields.connectionFactory,
+				kafkaConfig:       config.NewKafkaConfig(),
+				awsConfig:         config.NewAWSConfig(),
+			}
+
+			result, err := k.ListAll()
+
+			// check errors
+			if (err != nil) != tt.wantErr {
+				t.Errorf("kafkaService.List() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			// compare wanted vs actual results
+			if len(result) != len(tt.want) {
+				t.Errorf("kafka.Service.List(): total number of results: got = %d want = %d", len(result), len(tt.want))
+			}
+
+			for i, got := range result {
+				g.Expect(got).To(gomega.Equal(tt.want[i]))
+			}
+		})
+	}
+}
+
 func Test_kafkaService_ListByStatus(t *testing.T) {
 	type fields struct {
 		connectionFactory *db.ConnectionFactory
--- a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
+++ b/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/acl"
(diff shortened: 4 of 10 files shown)


internal/kafka/internal/services/kafka.go
```diff
@@ -905,6 +908,8 @@ func (k *kafkaService) VerifyAndUpdateKafkaAdmin(ctx context.Context, kafkaReque
 		return errors.NewWithCause(errors.ErrorGeneral, err, "Failed to update kafka")
 	}
 
+	metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusProvisioning, kafkaRequest.ID, kafkaRequest.ClusterID)
```
machi1990, line 911: Why `provisioning` here: should it be `kafkaRequest.Status`?

internal/kafka/internal/workers/kafka_mgrs/preparing_kafkas_mgr.go
```diff
@@ -60,6 +60,7 @@ func (k *PreparingKafkaManager) Reconcile() []error {
 	for _, kafka := range preparingKafkas {
 		glog.V(10).Infof("preparing kafka id = %s", kafka.ID)
 		metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusPreparing, kafka.ID, kafka.ClusterID, time.Since(kafka.CreatedAt))
+		metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusPreparing, kafka.ID, kafka.ClusterID)
```
machi1990, line 63: I think if a reconciliation occurs for a given kafka X, the metric will be updated:
I am wondering if a better strategy might be to set this status when the Kafka transition state: e.g when it moves from `accepted` to `preparing`: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ee6c732df5c8a67e06a5c2c74e26705e85d2c253/internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr.go#L121 we fire this metric: 		metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusPreparing, kafka.ID, kafka.ClusterID)

internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr.go
```diff
@@ -68,6 +68,7 @@ func (k *AcceptedKafkaManager) Reconcile() []error {
 	for _, kafka := range acceptedKafkas {
 		glog.V(10).Infof("accepted kafka id = %s", kafka.ID)
 		metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusAccepted, kafka.ID, kafka.ClusterID, time.Since(kafka.CreatedAt))
+		metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusAccepted, kafka.ID, kafka.ClusterID)
```
machi1990, line 71: For accepted status, this has to be set only in `KafkaService#RegisterKafkaJob()`: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ee6c732df5c8a67e06a5c2c74e26705e85d2c253/internal/kafka/internal/services/kafka.go#L426

internal/kafka/internal/workers/kafka_mgrs/provisioning_kafkas_mgr.go
```diff
@@ -69,6 +69,7 @@ func (k *ProvisioningKafkaManager) Reconcile() []error {
 			}
 		}
 		metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusProvisioning, kafka.ID, kafka.ClusterID, time.Since(kafka.CreatedAt))
+		metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusProvisioning, kafka.ID, kafka.ClusterID)
```
machi1990, line 72: For provisioning status, the metric has to be done only here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ee6c732df5c8a67e06a5c2c74e26705e85d2c253/internal/kafka/internal/services/kafka.go#L476

pawelpaszki:
@machi1990 - thanks for your comments. I think I addressed them all

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -239,6 +240,7 @@ func (d *dataPlaneKafkaService) setKafkaClusterReady(kafka *dbapi.KafkaRequest)
 
 	if shouldSendMetric {
 		metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusReady, kafka.ID, kafka.ClusterID, time.Since(kafka.CreatedAt))
+		metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusReady, kafka.ID, kafka.ClusterID)
```
pawelpaszki, line 243: @machi1990 @JameelB @pb82 - should I actually move this outside of the `shouldSendMetric` check and only send it, if previous status was resuming?

machi1990: At the moment, with this we won't update the Kafka metric if the previous status is `resuming` and we'll consider that Kafka in resuming state. 

I think the `metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusReady, kafka.ID, kafka.ClusterID)` has to be fired when the previous status is either: `provisioning` or `resuming`.

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -580,7 +582,7 @@ func TestKafkaCreate_ValidatePlanParam(t *testing.T) {
 	g.Expect(err).ToNot(gomega.HaveOccurred())
 	instanceTypeSizeConfig, err := instanceTypeConfig.GetKafkaInstanceSizeByID("x1")
 	g.Expect(err).ToNot(gomega.HaveOccurred())
-	g.Expect(*kafka.ExpiresAt).To(gomega.Equal(kafka.CreatedAt.Add(time.Duration(*instanceTypeSizeConfig.LifespanSeconds) * time.Second)))
+	g.Expect(kafka.ExpiresAt.Equal(kafka.CreatedAt.Add(time.Duration(*instanceTypeSizeConfig.LifespanSeconds) * time.Second))).To(gomega.BeTrue())
```
pawelpaszki, line 585: @JameelB @machi1990 - the changed line started to fail for me locally since this morning with the error:

```
kas-fleetshard-sync.go:215: Starting mock fleetshard sync
I1028 10:57:32.761929   57616 kafkas_routes_cname_mgr.go:48] kafkas need routes created count = 0
    kafka_create_test.go:586: 
        Expected
            <time.Time>: 2022-10-30T09:57:32.810862Z
        to be equivalent to
            <time.Time>: 2022-10-30T09:57:32.810862Z
```

see [this](https://stackoverflow.com/questions/69350229/go-assert-equal-some-interface-with-time-time-object-inside) for why it might happen.

internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -405,6 +416,7 @@ func (d *dataPlaneKafkaService) unassignKafkaFromDataplaneCluster(kafka *dbapi.K
 		}
 
 		metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusProvisioning, kafka.ID, kafka.ClusterID, time.Since(kafka.CreatedAt))
+		metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusProvisioning, kafka.ID, kafka.ClusterID)
```
JameelB, line 419: Same comment as above. We'll lose track of the time this Kafka instance spent in this state before this data plane cluster unassignment.

internal/kafka/internal/services/kafka.go
```diff
@@ -905,6 +908,8 @@ func (k *kafkaService) VerifyAndUpdateKafkaAdmin(ctx context.Context, kafkaReque
 		return errors.NewWithCause(errors.ErrorGeneral, err, "Failed to update kafka")
 	}
 
+	metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaStatus(kafkaRequest.Status), kafkaRequest.ID, kafkaRequest.ClusterID)
```
JameelB, line 911: Is there a way for us to check the current status of the Kafka instance here? We probably only want to update this metric if the status is updated to a different state.

internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr.go
```diff
@@ -133,6 +134,7 @@ func (k *AcceptedKafkaManager) markTheUnassignedKafkaAsFailedOrAllowRetryCluster
 		return errors.Wrapf(err2, "failed to update failed kafka %s", kafka.ID)
 	}
 	metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusFailed, kafka.ID, kafka.ClusterID, durationSinceCreation)
+	metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusFailed, kafka.ID, kafka.ClusterID)
```
JameelB, line 137: let's only update this if the status is not already `Failed`. This is called on every reconcile so it will get updated every few seconds.

internal/kafka/internal/workers/kafka_mgrs/accepted_kafkas_mgr.go
```diff
@@ -152,6 +154,7 @@ func (k *AcceptedKafkaManager) markTheAssignedKafkaAsFailedOrAllowRetryStrimziVe
 		return errors.Wrapf(err, "failed to update failed kafka %s", kafka.ID)
 	}
 	metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusFailed, kafka.ID, kafka.ClusterID, durationSinceCreation)
+	metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusFailed, kafka.ID, kafka.ClusterID)
```
JameelB, line 157: Same as above, let's only update this if the status is not already `Failed`. This is called on every reconcile so it will get updated every few seconds.

internal/kafka/internal/workers/kafka_mgrs/preparing_kafkas_mgr.go
```diff
@@ -92,6 +92,7 @@ func (k *PreparingKafkaManager) handleKafkaRequestCreationError(kafkaRequest *db
 				return errors.Wrapf(updateErr, "Failed to update kafka %s in failed state. Kafka failed reason %s", kafkaRequest.ID, kafkaRequest.FailedReason)
 			}
 			metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusFailed, kafkaRequest.ID, kafkaRequest.ClusterID, time.Since(kafkaRequest.CreatedAt))
+			metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusFailed, kafkaRequest.ID, kafkaRequest.ClusterID)
```
JameelB, line 95: Same as above, let's only update this if the status is not already `Failed`. This is called on every reconcile so it will get updated every few seconds.

internal/kafka/internal/workers/kafka_mgrs/preparing_kafkas_mgr.go
```diff
@@ -103,6 +104,7 @@ func (k *PreparingKafkaManager) handleKafkaRequestCreationError(kafkaRequest *db
 			return errors.Wrapf(err, "Failed to update kafka %s in failed state", kafkaRequest.ID)
 		}
 		metrics.UpdateKafkaRequestsStatusSinceCreatedMetric(constants2.KafkaRequestStatusFailed, kafkaRequest.ID, kafkaRequest.ClusterID, time.Since(kafkaRequest.CreatedAt))
+		metrics.UpdateKafkaRequestsStatusSinceLastSetMetric(constants2.KafkaRequestStatusFailed, kafkaRequest.ID, kafkaRequest.ClusterID)
```
JameelB, line 107: Same as above, let's only update this if the status is not already `Failed`. This is called on every reconcile so it will get updated every few seconds.

pkg/metrics/metrics.go
```diff
@@ -474,6 +482,26 @@ func UpdateKafkaRequestsStatusSinceCreatedMetric(status constants2.KafkaStatus,
 	kafkaStatusSinceCreatedMetric.With(labels).Set(elapsed.Seconds())
 }
 
+// create a new GaugeVec for kafkas status duration
+var kafkaStatusSinceLastSetMetric = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Subsystem: KasFleetManager,
+		Name:      KafkaRequestsStatusSinceLastSet,
+		Help:      "metrics to track the status of a Kafka instance and the timestamp of the last update",
```
JameelB, line 490: @pawelpaszki how do you see the query for getting the duration of how long a Kafka is in its current state using this metric?

For getting the current status, the only way I can think of is to get the metric with the value of the latest timestamp, maybe there's a way we can do that with promql :thinking:. Otherwise, is there some other metric we can use together with this one to do that?

pawelpaszki: I was under the impression that we could compare timestamp to current time, but not sure if its even possible

internal/kafka/internal/services/kafka.go
pawelpaszki: I could add additional boolean param to indicate that status will be updated. would it make sense? unfortunately the metric can't be exposed from within admin kafka handler, because we can't be sure that the update will be successful

pawelpaszki: I realized that I can actually do it in the handler. will push shortly with all of the updates

pkg/metrics/metrics.go
```diff
@@ -34,6 +34,7 @@ const (
 	// KafkaRequestsStatus - kafka requests status metric
 	KafkaRequestsStatusSinceCreated = "kafka_requests_status_since_created_in_seconds"
 	KafkaRequestsStatusCount        = "kafka_requests_status_count"
+	KafkaRequestsStatusSinceLastSet = "kafka_requests_status_since_last_set"
```
JameelB, line 37: I wonder if the name still makes sense because we no longer aim to return the timestamp as a value for this metric. This now gives us information of the current status of a Kafka instance. Maybe something like `kafka_request_current_status_info`?

pkg/metrics/metrics.go
JameelB, line 490: Let's update this description too.

internal/kafka/internal/services/kafka.go
```diff
@@ -722,6 +723,17 @@ func (k *kafkaService) Delete(kafkaRequest *dbapi.KafkaRequest) *errors.ServiceE
 	return nil
 }
 
+// List all kafkas (used by the kafkas worker)
+func (k *kafkaService) ListAll() (dbapi.KafkaList, *errors.ServiceError) {
+	var kafkaRequestList dbapi.KafkaList
+	dbConn := k.connectionFactory.New()
+	if err := dbConn.Find(&kafkaRequestList).Error; err != nil {
+		return kafkaRequestList, errors.NewWithCause(errors.ErrorGeneral, err, "Unable to list kafka requests")
```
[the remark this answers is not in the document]
pawelpaszki: keep making this mistake :D

pb82:
LGTM

pb82 (approved):

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
```diff
@@ -84,6 +91,9 @@ func TestKafkaManager_Reconcile(t *testing.T) {
 					DeprovisionExpiredKafkasFunc: func() *errors.ServiceError {
 						return errors.GeneralError("failed to deprovision expired kafkas")
 					},
+					ListAllFunc: func() (dbapi.KafkaList, *errors.ServiceError) {
+						return dbapi.KafkaList{}, nil
+					},
 				},
 				clusterService: &services.ClusterServiceMock{
 					FindStreamingUnitCountByClusterAndInstanceTypeFunc: func() (services.KafkaStreamingUnitCountPerClusterList, error) {
```
JameelB, line 99: Might be worth adding another test case here to ensure that if the newly added ListAll call fails, Reconcile() should return that error.

pkg/metrics/metrics.go
```diff
@@ -116,14 +117,21 @@ var JobsMetricsLabels = []string{
...
-// kafkaStatusSinceCreatedMetricLabels  is the slice of labels to add to
+// kafkaStatusSinceCreatedMetricLabels is the slice of labels to add to
 var kafkaStatusSinceCreatedMetricLabels = []string{
 	LabelStatus,
 	LabelID,
 	LabelClusterID,
 }
 
-// kafkaStatusCountMetricLabels  is the slice of labels to add to
+// kafkaStatusSinceLastSetMetricLabels is the slice of labels to add to
```
JameelB, line 127: Let's change this var names to reflect the metric it's used for as well.

pkg/metrics/metrics.go
```diff
@@ -474,6 +482,26 @@ func UpdateKafkaRequestsStatusSinceCreatedMetric(status constants2.KafkaStatus,
 	kafkaStatusSinceCreatedMetric.With(labels).Set(elapsed.Seconds())
 }
 
+// create a new GaugeVec for kafkas status duration
```
JameelB, line 485:
```suggestion
// create a new GaugeVec for tracking current kafka status
```

internal/kafka/internal/services/kafka_test.go
```diff
@@ -1991,6 +1991,108 @@ func Test_kafkaService_List(t *testing.T) {
...
+				connectionFactory: tt.fields.connectionFactory,
+				kafkaConfig:       config.NewKafkaConfig(),
+				awsConfig:         config.NewAWSConfig(),
+			}
+
+			result, err := k.ListAll()
+
+			// check errors
+			if (err != nil) != tt.wantErr {
+				t.Errorf("kafkaService.List() error = %v, wantErr %v", err, tt.wantErr)
```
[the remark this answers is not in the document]
pawelpaszki: will fix these in the subsequent PR

PR #1291, opened 2022-09-07
[RHOC] Implement patch connectors, admin api

lgarciaaco:
## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has been created for changes required on the client side

PR #1365, opened 2022-10-27
Patch connectors

lgarciaaco:


lgarciaaco:
Follow up on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1291

internal/connector/internal/services/authz/authz_service.go
```diff
@@ -148,6 +148,10 @@ func (u *ValidationUser) AuthorizedNamespaceUser(errorCode errors.ServiceErrorCo
 		if u.err != nil {
 			err = u.err
 		} else {
+			if u.IsAdmin() {
```
dhirajsb, line 151: This is wrong. User's tenancy has to be verified first, even for admins. In other words, admin user MUST belong to the namespace org.

dhirajsb (changes requested):

internal/connector/internal/services/authz/authz_service.go
lgarciaaco: Fixed with [353540f](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1365/commits/353540f5ca760f41e21353e687834f1912982304)

valdar: @dhirajsb per my understanding this has been removed from the PR, can you please have a look and confirm all is good authentication side?

openapi/connector_mgmt-private-admin.yaml
```diff
@@ -765,6 +765,65 @@ paths:
...
+          application/merge-patch+json:
+            schema:
+              type: object
+          application/json-patch+json:
+            schema:
+              description: A JSON Patch, RFC 6902 - https://tools.ietf.org/html/rfc6902
+              type: object
+          application/json:
+            schema:
+              $ref: "connector_mgmt.yaml#/components/schemas/ConnectorRequest"
```
dhirajsb, line 789: The admin API follows the pattern that if there is a request/response type that's specific to the admin API. It uses a type defined in this file with the naming convention like `*AdminView`. 
Since the admin patch API is only allowed to patch the state, the admin API should use a type ConnectorAdminRequest, which only has properties that are allowed to be changed by the admin API. Otherwise the API makes it seem like even the connector configuration can be patched.

internal/connector/test/integration/features/connector-agent-api.feature
```diff
@@ -1193,6 +1193,41 @@ Feature: connector agent API
...
+            "desired_state": "ready"
+        }
+        """
+    Then the response code should be 202
+
+    When I GET path "/v1/admin/kafka_connectors/${connector_id}"
+    Then the response code should be 200
+    And the ".desired_state" selection from the response should match "ready"
+
+    When I PATCH path "/v1/admin/kafka_connectors/${connector_id}" with json body:
```
dhirajsb, line 1223: Instead of a generic non-existent field like `some_stuff`, the test should use an input json that would be valid with the public API with connector properties and kafka connection settings to test that only the `desired_state` can be patched. 
The test should also check the error response for correct error code and message.

dhirajsb (changes requested):
Sorry for the late review. 

The PR needs a little more polish to use a `ConnectorAdminRequest` type with only modifiable property `desired_state`, and a more fleshed out negative test.

internal/connector/test/integration/features/connector-agent-api.feature
lgarciaaco: I don't agree this test is needed. Having `ConnectorAdminRequest` with only `desired_state` makes this scenario impossible. 

Currently this test aims to check for a error response when the request is malformed, which can be a valid scenario.

openapi/connector_mgmt-private-admin.yaml
lgarciaaco: Fixed with https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1365/commits/59704582f0cc293bba388ecfd4a4a760ae46114d

heiko-braun:
@dhirajsb Mind another review? Luis did work on the changes requested.

internal/connector/test/integration/features/connector-agent-api.feature
dhirajsb: Agreed, with the change in request type, we don't have to worry about unmodefiable fields.

dhirajsb (approved):
Great job, looks perfect. Appreciate the contribution @lgarciaaco

PR #1368, opened 2022-11-01
feat: Added support for KFM Billing Models to QUOTA-LIST

ziccardi:
## Description
Added support for KFM billing model to quota-list so that different billing model are supported for the same instance type (standard, eval, marketplace, ...)

## Verification Steps

1) Run KFM locally with QUOTA-LIST and the specified configuration file. Try to create kafkas and check that it gives an error after reaching the quota.
2) Try to create a kafka with a user not listed in the QUOTA-LIST (set any_user to false) and ensure it allows you to create only one developer instance.
3) Run KFM locally with  AMS and check that it still works as before

**Example configuration file**
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/bde4796f5700b99671e532ac008d4c8be039c1fd/config/quota-management-list-configuration.yaml

- [ ] Code Review completed

### DIFF
--- a/config/quota-management-list-configuration.yaml
+++ b/config/quota-management-list-configuration.yaml
@@ -5,25 +5,62 @@
 # In the future it may support service accounts outside RH.
 # The structure of registered service accounts is:
 #       - username: is the account of the user. The username must be unique
-#       - max_allowed_instances: is the maximum number of instances this user can create.
+#       - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
+#         This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
 #         Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+#       - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
+#         - instance_type_id: the id of the instance type for which we are configuring the quota
+#         - billing_models: the list of billing models that we are going to configure. This is defined as follows:
+#           - id: the id of the billing model we are going to configure
+#           - allowed: the number of instances of this type the user can create
+#           - expiration: after how many days the instance will expire (it will be suspended). Defaults to 0 (unlimited)
+#           - grace_period: after how many days after being suspended, the instance will be deleted. Defaults to 0 (unlimited)
 registered_service_accounts:
   - username: testuser1@example.com
     max_allowed_instances: 3
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
   - username: testuser2@example.com
     max_allowed_instances: 2
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
   - username: testuser3@example.com
     max_allowed_instances: 2
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
   - username: testuser4@example.com
     max_allowed_instances: 2
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
   - username: testuser5@example.com
     max_allowed_instances: 2
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
 
 # A list of registered users per organisation (see structure below). The list current contains known organisations - add yours if it is missing.
 # If a user is not in this or in the `registered_service_accounts` list, only DEVELOPER kafka instances will be allowed.
 # - "id": is the organisation id
 # - "any_user": "any_user": Controls whether to allow all users to create standard kafka instances with this organisation if "registered_users" list is empty.
-# - max_allowed_instances: is the maximum number of instances this orgnisation. Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+# - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
+#   This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
+#   Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+# - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
+#   - instance_type_id: the id of the instance type for which we are configuring the quota
+#   - billing_models: the list of billing models that we are going to configure. This is defined as follows:
+#     - id: the id of the billing model we are going to configure
+#     - allowed: the number of instances of this type the user can create
+#     - expiration: after how many days the instance will expire (it will be suspended). Defaults to 0 (unlimited)
+#     - grace_period: after how many days after being suspended, the instance will be deleted. Defaults to 0 (unlimited)
 # - "registered_users": A list of registered users for this organisation. If empty, no one is registered unless "any_user" is set to true.
 #      - username: is the account of the user. The username must be unique within the organisation and across organisations.
 registered_users_per_organisation:
@@ -32,23 +69,48 @@ registered_users_per_organisation:
     any_user: true
     max_allowed_instances: 5
     registered_users: []
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
+            allowed: 5
     # The App SRE team
   - id: 12147054
     any_user: true
     max_allowed_instances: 1
     registered_users: []
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
+            allowed: 1
     # The Kafka Instances team
   - id: 13639843
     any_user: true
     max_allowed_instances: 1
     registered_users: []
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
+            allowed: 1
     # The MK Security team
   - id: 13785172
     any_user: true
     max_allowed_instances: 1
     registered_users: []
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
+            allowed: 1
     # The Running the Service team
   - id: 13645369
     any_user: true
     max_allowed_instances: 3
     registered_users: []
+    granted_quota:
+      - instance_type_id: standard
+        billing_models:
+          - id: standard
+            allowed: 3
--- a/config/quota-management-list-configuration.yaml
+++ b/config/quota-management-list-configuration.yaml
@@ -6,45 +6,52 @@
 # The structure of registered service accounts is:
 #       - username: is the account of the user. The username must be unique
 #       - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
-#         This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
+#         This value is used if `max_allowed_instances` field (inside `billing_model`) is 0 or omitted.
 #         Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
-#       - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
-#         - instance_type_id: the id of the instance type for which we are configuring the quota
-#         - billing_models: the list of billing models that we are going to configure. This is defined as follows:
-#           - id: the id of the billing model we are going to configure
-#           - allowed: the number of instances of this type the user can create
-#           - expiration: after how many days the instance will expire (it will be suspended). Defaults to 0 (unlimited)
-#           - grace_period: after how many days after being suspended, the instance will be deleted. Defaults to 0 (unlimited)
+#       - granted_quota: the configuration of the quota granted to the user.
+#         If granted_quota is empty or is not defined, it will default to an array containing quota for `standard` instance.
+#         The configuration is defined as follows:
+#         - instance_type_id: the id of the instance type for which we are configuring the quota. This must match the IDs used in the KFM
+#           `supported_instance_types` configuration.
+#         - kafka_billing_models: the list of billing models that we are going to configure.
+#           If `kafka_billing_models` is empty or is not defined, the default value will be used (a list containing only the "standard" billing model)
+#           The set of accepted billing models, depends on the KafkaBillingModel configuration.
+#           The content of this field is defined as follows:
+#           - id: the id of the billing model we are going to configure. The set of valid IDs here is the set of the KafkaBillingModel IDs
+#             available in the 'supported_billing_models' set per instance type in the 'supported_instance_types' configuration.
+#           - max_allowed_instances: the number of instances of this type the user can create
+#           - expiration_days: after how many days the instance will expire (it will be suspended). Defaults to 0 (unlimited)
+#           - grace_period_days: when in suspended status, defines after how many days the instance will be deleted. Defaults to 0 (unlimited)
 registered_service_accounts:
   - username: testuser1@example.com
     max_allowed_instances: 3
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
   - username: testuser2@example.com
     max_allowed_instances: 2
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
   - username: testuser3@example.com
     max_allowed_instances: 2
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
   - username: testuser4@example.com
     max_allowed_instances: 2
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
   - username: testuser5@example.com
     max_allowed_instances: 2
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
 
 # A list of registered users per organisation (see structure below). The list current contains known organisations - add yours if it is missing.
@@ -52,17 +59,19 @@ registered_service_accounts:
 # - "id": is the organisation id
 # - "any_user": "any_user": Controls whether to allow all users to create standard kafka instances with this organisation if "registered_users" list is empty.
 # - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
-#   This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
+#   This value is used if `max_allowed_instances` field (inside `billing_model`) is 0 or omitted.
 #   Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
 # - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
 #   - instance_type_id: the id of the instance type for which we are configuring the quota
-#   - billing_models: the list of billing models that we are going to configure. This is defined as follows:
-#     - id: the id of the billing model we are going to configure
-#     - allowed: the number of instances of this type the user can create
-#     - expiration: after how many days the instance will expire (it will be suspended). Defaults to 0 (unlimited)
-#     - grace_period: after how many days after being suspended, the instance will be deleted. Defaults to 0 (unlimited)
-# - "registered_users": A list of registered users for this organisation. If empty, no one is registered unless "any_user" is set to true.
-#      - username: is the account of the user. The username must be unique within the organisation and across organisations.
+#   - kafka_billing_models: the list of billing models that we are going to configure.
+#     If `kafka_billing_models` is empty or is not defined, the default value will be used (a list containing only the "standard" billing model)
+#     The set of accepted billing models, depends on the KafkaBillingModel configuration.
+#     The content of this field is defined as follows:
+#     - id: the id of the billing model we are going to configure. The set of valid IDs here is the set of the KafkaBillingModel IDs
+#       available in the 'supported_billing_models' set per instance type in the 'supported_instance_types' configuration.
+#     - max_allowed_instances: the number of instances of this type the user can create
+#     - expiration_days: after how many days the instance will expire (it will be suspended). Defaults to 0 (unlimited)
+#     - grace_period_days: when in suspended status, defines after how many days the instance will be deleted. Defaults to 0 (unlimited)
 registered_users_per_organisation:
     # The Control Plane team
   - id: 13640203
@@ -71,9 +80,9 @@ registered_users_per_organisation:
     registered_users: []
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
-            allowed: 5
+            max_allowed_instances: 5
     # The App SRE team
   - id: 12147054
     any_user: true
@@ -81,9 +90,9 @@ registered_users_per_organisation:
     registered_users: []
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
-            allowed: 1
+            max_allowed_instances: 1
     # The Kafka Instances team
   - id: 13639843
     any_user: true
@@ -91,9 +100,9 @@ registered_users_per_organisation:
     registered_users: []
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
-            allowed: 1
+            max_allowed_instances: 1
     # The MK Security team
   - id: 13785172
     any_user: true
@@ -101,9 +110,9 @@ registered_users_per_organisation:
     registered_users: []
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
-            allowed: 1
+            max_allowed_instances: 1
     # The Running the Service team
   - id: 13645369
     any_user: true
@@ -111,6 +120,6 @@ registered_users_per_organisation:
     registered_users: []
     granted_quota:
       - instance_type_id: standard
-        billing_models:
+        kafka_billing_models:
           - id: standard
-            allowed: 3
+            max_allowed_instances: 3
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -286,7 +286,7 @@ func (k *kafkaService) AssignInstanceType(owner string, organisationId string) (
 		return "", errors.NewWithCause(errors.ErrorGeneral, factoryErr, "unable to check quota")
 	}
 
-	hasRhosakQuota, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(owner, organisationId, types.STANDARD)
+	hasRhosakQuota, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(owner, organisationId, types.STANDARD, "")
 	if err != nil {
 		return "", err
 	}
@@ -333,7 +333,7 @@ func (k *kafkaService) reserveQuota(kafkaRequest *dbapi.KafkaRequest) (subscript
 	if factoryErr != nil {
 		return "", errors.NewWithCause(errors.ErrorGeneral, factoryErr, "unable to check quota")
 	}
-	subscriptionId, err = quotaService.ReserveQuota(kafkaRequest, types.KafkaInstanceType(kafkaRequest.InstanceType))
+	subscriptionId, err = quotaService.ReserveQuota(kafkaRequest)
 	return subscriptionId, err
 }
 
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -291,11 +291,11 @@ func (k *kafkaService) AssignInstanceType(owner string, organisationId string) (
 			continue
 		}
 		for _, bm := range instanceType.SupportedBillingModels {
-			hasRhosakQuota, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(owner, organisationId, types.KafkaInstanceType(instanceType.Id), bm)
+			hasQuota, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(owner, organisationId, types.KafkaInstanceType(instanceType.Id), bm)
 			if err != nil {
 				return "", err
 			}
-			if hasRhosakQuota {
+			if hasQuota {
 				return types.KafkaInstanceType(instanceType.Id), nil
 			}
 		}
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -286,12 +286,19 @@ func (k *kafkaService) AssignInstanceType(owner string, organisationId string) (
 		return "", errors.NewWithCause(errors.ErrorGeneral, factoryErr, "unable to check quota")
 	}
 
-	hasRhosakQuota, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(owner, organisationId, types.STANDARD, "")
-	if err != nil {
-		return "", err
-	}
-	if hasRhosakQuota {
-		return types.STANDARD, nil
+	for _, instanceType := range k.kafkaConfig.SupportedInstanceTypes.Configuration.SupportedKafkaInstanceTypes {
+		if instanceType.Id == types.DEVELOPER.String() {
+			continue
+		}
+		for _, bm := range instanceType.SupportedBillingModels {
+			hasRhosakQuota, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(owner, organisationId, types.KafkaInstanceType(instanceType.Id), bm)
+			if err != nil {
+				return "", err
+			}
+			if hasRhosakQuota {
+				return types.KafkaInstanceType(instanceType.Id), nil
+			}
+		}
 	}
 
 	return types.DEVELOPER, nil
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -2,6 +2,7 @@ package quota
 
 import (
 	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/cloudproviders"
@@ -17,6 +18,8 @@ type amsQuotaService struct {
 	amsClient   ocm.AMSClient
 	kafkaConfig *config.KafkaConfig
 }
+
+var _ services.QuotaService = &amsQuotaService{}
 
 const (
 	CloudProviderAWS   = "aws"
@@ -167,7 +170,7 @@ func (q amsQuotaService) ValidateBillingAccount(organisationId string, instanceT
 	return nil
 }
 
-func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType) (bool, *errors.ServiceError) {
+func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType, billingModelName string) (bool, *errors.ServiceError) {
 	orgId, err := q.amsClient.GetOrganisationIdFromExternalId(externalId)
 	if err != nil {
 		return false, errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("error checking quota: failed to get organization with external id %v", externalId))
@@ -338,7 +341,8 @@ func (q amsQuotaService) getBillingModel(kafka *dbapi.KafkaRequest, instanceType
 	return "", errors.InsufficientQuotaError("no matching marketplace quota found for product %s", instanceType.GetQuotaType().GetProduct())
 }
 
-func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (string, *errors.ServiceError) {
+func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *errors.ServiceError) {
+	instanceType := types.KafkaInstanceType(kafka.InstanceType)
 	kafkaId := kafka.ID
 
 	rr := q.newBaseQuotaReservedResourceBuilder(kafka)
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -170,7 +170,9 @@ func (q amsQuotaService) ValidateBillingAccount(organisationId string, instanceT
 	return nil
 }
 
-func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType, billingModelName string) (bool, *errors.ServiceError) {
+// TODO: added the `billing model` parameter to the QuotaService interface when adding KafkaBillingModel support to the QUOTA-LIST
+// The parameter is currently unused for AMSQuota management: will be used in a subsequent PR when KafkaBillingModel support will be added.
+func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType, kafkaBillingModel config.KafkaBillingModel) (bool, *errors.ServiceError) {
 	orgId, err := q.amsClient.GetOrganisationIdFromExternalId(externalId)
 	if err != nil {
 		return false, errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("error checking quota: failed to get organization with external id %v", externalId))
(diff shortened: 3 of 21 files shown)


internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -13,11 +14,20 @@ import (
 	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 )
 
+type AmsBillingModel interface {
```
miguelsorianod, line 17: I find this name confusing.

AmsBillingModel is an AMS specific attribute. Which is independent of the AMS product and resources. At the same time this interface has GetAMSBillingModels, adding to the confusion.
I think we should think of an alternative name for this if possible

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -167,7 +177,15 @@ func (q amsQuotaService) ValidateBillingAccount(organisationId string, instanceT
 	return nil
 }
 
-func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType) (bool, *errors.ServiceError) {
+func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType, billingModel services.KafkaBillingModel) (bool, *errors.ServiceError) {
+	//
```
miguelsorianod, line 181: This is commented. Is this intentional? If so, we can just add it in another PR or at least add some context. Also it seems that the `AmsBillingModel` interface is only used for this commented code so we should also comment it or delete it

ziccardi: Yes, it was commented on purpose. This PR is about changes to the QUOTA-LIST management, but it required some changes to the shared `QuotaService` interface too, so I added a comment to explain what was the intended usage when implementing the AMS Quota manager too.

I'm going to assign that task to myself, so this comment doesn't make sense anymore: removing.

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: Going to remove it in this PR, and I will rethink the name.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -167,7 +177,15 @@ func (q amsQuotaService) ValidateBillingAccount(organisationId string, instanceT
 	return nil
 }
 
-func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType) (bool, *errors.ServiceError) {
+func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType, billingModel services.KafkaBillingModel) (bool, *errors.ServiceError) {
```
miguelsorianod, line 180: the `billingModel` parameter has been added to the `CheckIfQuotaIsDefinedForInstanceType` method. Should this be removed from here too for now in this PR? (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368/files#r1015433365) or we want to keep it so it is there already?

miguelsorianod: I see now it is due to the interface it implements has that field added too and it is used by the quota management list implementation. Let's add a comment in here anyway as I understand this implementation will need to be changed too

pkg/quota_management/account.go
```diff
@@ -1,20 +1,69 @@
 package quota_management
```
miguelsorianod, line 1: Are `pkg/quota_management/account.go` and `pkg/quota_management/organisation.go` exclusive to quota management list?
If so I think we should rename the package to be more specific to quota management list or move the contents to a more specific to quota management list package(s) as in that case it gives the idea that is where elements to perform quota management, independently of the implementation, are located.

pkg/quota_management/billing_model.go
```diff
@@ -0,0 +1,10 @@
+package quota_management
+
+type BillingModel struct {
```
miguelsorianod, line 3: Is this a KafkaBillingModel or an AMSBillingModel?

I think we should be more specific as it is not clear from just the name.

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -23,20 +23,24 @@ func getValidMaturityStates() []MaturityStatus {
...
-}
-
 type KafkaBillingModel struct {
 	ID               string   `yaml:"id" validate:"required"`
 	AMSResource      string   `yaml:"ams_resource" validate:"required,ams_resource_validator"`
 	AMSProduct       string   `yaml:"ams_product" validate:"required,ams_product_validator"`
 	AMSBillingModels []string `yaml:"ams_billing_models" validate:"min=1,unique,ams_billing_models_validator"`
 }
 
+func (kbm *KafkaBillingModel) GetID() string {
```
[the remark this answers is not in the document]
ziccardi: Adding the interface guard, we will add a dependency from the config package and the quota management package. 
I think that is not ideal.

pkg/quota_management/account.go
```diff
@@ -1,20 +1,69 @@
...
+	MaxAllowedInstances int       `yaml:"max_allowed_instances"`
+	GrantedQuota        QuotaList `yaml:"granted_quota,omitempty"`
 }
 
-func (account Account) IsInstanceCountWithinLimit(count int) bool {
-	return count <= account.GetMaxAllowedInstances()
+var _ QuotaManagementListItem = &Account{}
+
+func (account Account) IsInstanceCountWithinLimit(instanceTypeID string, billingModelName string, count int) bool {
+	if billingModelName == "" {
```
miguelsorianod, line 17: Is there some reason we have a default?

I think we should consider avoiding implicit behavior around this, as with the AMS implementation it's proven to increase complexity substantially

pkg/quota_management/account.go
```diff
@@ -1,20 +1,69 @@
...
+		}
+		return account.MaxAllowedInstances
+	}
+
+	return bm.Allowed
+}
+
+func (account Account) GetGrantedQuota() QuotaList {
+	if len(account.GrantedQuota) == 0 {
+		return defaultInstanceTypes
```
miguelsorianod, line 46: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368/files#r1015633583

miguelsorianod:
Could I get a configuration sample / explanation of the quota management list config format?

I think having that information might help reviewers to get an idea of what is trying to be achieved

pkg/quota_management/account.go
ziccardi: I had exactly the same thought. I already discussed this with Manyanda and we agreed to change that package name.
I will do it in another PR.

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -18,33 +19,49 @@ type QuotaManagementListService struct {
...
+
 	if orgFound && org.IsUserRegistered(username) {
 		userIsRegistered = true
 	} else {
-		_, userFound := q.quotaManagementList.QuotaList.ServiceAccounts.GetByUsername(username)
-		userIsRegistered = userFound
+		account, serviceAccountIsRegistered = q.quotaManagementList.QuotaList.ServiceAccounts.GetByUsername(username)
+	}
+
+	billingModelName := "standard"
```
miguelsorianod, line 43: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368/files#r1015635270

pkg/quota_management/account.go
ziccardi: I totally agree. The idea here was to guarantee compatibility with the current clients. When the clients (UI and CLI) are updated, we can then remove the default.

pkg/quota_management/billing_model.go
miguelsorianod, line 3: Another side comment:
With recent work I've been doing, also looking at our code and the code of this PR I think we have quite a lot of Go types overload of the "billing model" concept and also strings named that way.

I think it would be a good idea to have some discussion around this subject through other channels

ziccardi: Very good point.

pkg/quota_management/account.go
miguelsorianod: Shouldn't we be setting billingModel beforehand in this case?

pkg/quota_management/billing_model.go
miguelsorianod: I understand :+1: . What would be the difference between the QuotaListBillingModel and the KafkaBillingModel?

ziccardi: `KafkaBillingModel` does not contains the values of 'how much quota' is assigned. The `QuotaListBillingModel` includes the values of how much 'is allowed' for that specific 'BillingModel' instance.

In other words, `KafkaBillingModel` is the **definition** of the list of billing models supported by KFM for each instance. At the same time, `QuotaBillingModel` is the **configuration** of the billing models for each instance type in the QUOTA-LIST manager.

miguelsorianod: But even if `QuotaBillingModel` is the configuration of the billing models, the billing models themselves are kafka billing models right?  the configuration only accepts the values that exist for kafka billing models

pkg/quota_management/account.go
ziccardi: I moved it elsewhere (caller).

pkg/quota_management/account.go
ziccardi: This is for compatibility with the previous configuration file: if no quota is specified, it defaults to 'STANDARD' to keep working as before.

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,62 @@
 # In the future it may support service accounts outside RH.
 # The structure of registered service accounts is:
 #       - username: is the account of the user. The username must be unique
-#       - max_allowed_instances: is the maximum number of instances this user can create.
+#       - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
+#         This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
 #         Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+#       - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
+#         - instance_type_id: the id of the instance type for which we are configuring the quota
```
miguelsorianod, line 12: Let's document that the accepted values are the values specified in the instance types configuration file

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,62 @@
...
+#         - billing_models: the list of billing models that we are going to configure. This is defined as follows:
```
miguelsorianod, line 13: when we talk about billing_models here what are we talking about? is it kafka billing models? ams billing models? another billing model concept independent of those?

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,62 @@
...
+#           - id: the id of the billing model we are going to configure
+#           - allowed: the number of instances of this type the user can create
```
miguelsorianod, line 15: can we call this max_allowed_instances?

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,62 @@
...
+#           - expiration: after how many days the instance will expire (it will be suspended). Defaults to 0 (unlimited)
```
miguelsorianod, line 16: Can we make this more granular to seconds? to leave room open to more finer granularities

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,62 @@
...
+#           - grace_period: after how many days after being suspended, the instance will be deleted. Defaults to 0 (unlimited)
```
miguelsorianod, line 17: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368/files#r1022705381

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,62 @@
...
 # If a user is not in this or in the `registered_service_accounts` list, only DEVELOPER kafka instances will be allowed.
 # - "id": is the organisation id
 # - "any_user": "any_user": Controls whether to allow all users to create standard kafka instances with this organisation if "registered_users" list is empty.
-# - max_allowed_instances: is the maximum number of instances this orgnisation. Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+# - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
+#   This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
+#   Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+# - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
+#   - instance_type_id: the id of the instance type for which we are configuring the quota
+#   - billing_models: the list of billing models that we are going to configure. This is defined as follows:
```
miguelsorianod, line 59: Same comments as the other section above at service account level

config/quota-management-list-configuration.yaml
miguelsorianod, line 17: It is also a good idea to specify the units in the name of the attributes when it accepts a specific unit. I suggest expiration_\<accepted_unit\>. Same for grace_period

pkg/quota_management/quota.go
```diff
@@ -0,0 +1,40 @@
+package quota_management
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+)
+
+var defaultBillingModel = BillingModel{
+	ID:          "STANDARD",
```
miguelsorianod, line 9: can we make the id lowercase to match what we have for other kinds of billing models?

pkg/quota_management/quota.go
```diff
@@ -0,0 +1,40 @@
...
+
+type Quota struct {
+	InstanceTypeID string           `yaml:"instance_type_id"`
+	BillingModels  BillingModelList `yaml:"billing_models,omitempty"`
+}
+
+func (quota *Quota) GetBillingModels() BillingModelList {
+	if len(quota.BillingModels) == 0 {
+		ret := defaultBillingModel
+		ret.Allowed = 0 // if 0, defaults to the MaxAllowedInstances
```
miguelsorianod, line 24: This seems a noop. ret is created with the value of `defaultBillingModel` and ret.Allowed is set to 0 but that variable is not returned. or used anywhere

pkg/quota_management/account.go
```diff
@@ -1,20 +1,62 @@
...
+}
+
+func (account Account) GetMaxAllowedInstances(instanceTypeID string, billingModelID string) int {
+	bm, ok := account.getBillingModel(instanceTypeID, billingModelID)
+
+	if !ok {
+		return 0
+	}
+
+	if bm.Allowed <= 0 {
```
miguelsorianod, line 27: Looking at this and at the defined configuration format I find it a little bit confusing:
If a user explicitly sets allowed at billing model to 0 I would expect that it means "I don't want to allow quota for this instance type in this billing model, overwriting whatever there is at the upper level".
Another reason is that it is inconsistent with the behavior of max_allowed_instnaces at the global level, where 0 means not allowed.
I think we should consider allowing differentiating between not being specified with being specified to 0.

pkg/quota_management/account.go
```diff
@@ -1,20 +1,62 @@
...
+	return count <= account.GetMaxAllowedInstances(instanceTypeID, billingModelID)
+}
+
+func (account Account) GetMaxAllowedInstances(instanceTypeID string, billingModelID string) int {
+	bm, ok := account.getBillingModel(instanceTypeID, billingModelID)
+
+	if !ok {
+		return 0
+	}
+
```
miguelsorianod, line 26: I think we can change this a little bit to avoid some conditional nesting.
For example:

if bm.Allowed > 0 then return it

res := MaxAllowedInstances
if account.MaxAllowedInstances > 0 then res = account.MaxAllowedInstances
return res

pkg/quota_management/organisation.go
```diff
@@ -1,10 +1,25 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+)
+
+var defaultQuotaForStandard = Quota{
+	InstanceTypeID: "standard",
+	BillingModels:  nil,
+}
+
+var defaultQuotaList = []Quota{defaultQuotaForStandard}
```
miguelsorianod, line 13: Maybe defaultGrantedQuota would be a more descriptive name

pkg/quota_management/organisation.go
```diff
@@ -19,16 +34,52 @@ func (org Organisation) HasUsersRegistered() bool {
 	return len(org.RegisteredUsers) > 0
 }
 
-func (org Organisation) IsInstanceCountWithinLimit(count int) bool {
-	return count <= org.GetMaxAllowedInstances()
+func (org Organisation) IsInstanceCountWithinLimit(instanceTypeID string, billingModelID string, count int) bool {
+	return count <= org.GetMaxAllowedInstances(instanceTypeID, billingModelID)
+}
+
+func (org Organisation) GetMaxAllowedInstances(instanceTypeID string, billingModelID string) int {
```
miguelsorianod, line 41: Same comments as I left for this method in the Account type apply

pkg/quota_management/organisation.go
```diff
@@ -1,10 +1,25 @@
 package quota_management
 
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+)
+
+var defaultQuotaForStandard = Quota{
```
miguelsorianod, line 8: Does that mean that the default granted quota is for standard instance types but no billing models are supported? what is the meaning of that?

pkg/quota_management/organisation.go
```diff
@@ -19,16 +34,52 @@ func (org Organisation) HasUsersRegistered() bool {
...
+	if len(org.GrantedQuota) == 0 {
+		return defaultQuotaList
+	}
+	return org.GrantedQuota
 }
 
-func (org Organisation) GetMaxAllowedInstances() int {
-	if org.MaxAllowedInstances <= 0 {
-		return MaxAllowedInstances
+func (org Organisation) getBillingModel(instanceTypeId string, billingModelID string) (BillingModel, bool) {
```
miguelsorianod, line 65: depending on some answers on other comments this might be better reworded as getBillingModelQuotaEntitlement or getBillingModelGrantedQuota. Same applies to the `BillingModel` itself

pkg/quota_management/account.go
```diff
@@ -1,20 +1,62 @@
...
+		idx, bm := arrays.FindFirst(instanceType.GetBillingModels(), func(bm BillingModel) bool { return shared.StringEqualsIgnoreCase(bm.ID, billingModelID) })
+		if idx != -1 {
+			return bm, true
+		}
 	}
+	return BillingModel{}, false
+}
 
-	return account.MaxAllowedInstances
+func (account Account) HasQuotaFor(instanceTypeId string, billingModelID string) bool {
```
miguelsorianod, line 57: I find this name is confusing. Does `HasQuotaFor` mean that I can create an instance of that type with that billing model, or that there is a BillingModel defined for it, which might end up resolving to 0 allowed quota?

pkg/quota_management/organisation.go
```diff
@@ -19,16 +34,52 @@ func (org Organisation) HasUsersRegistered() bool {
...
+		}
 	}
+	return BillingModel{}, false
+}
+
+func (org Organisation) HasQuotaFor(instanceTypeId string, billingModelID string) bool {
+	instanceTypes := org.GetGrantedQuota()
 
-	return org.MaxAllowedInstances
+	idx, instanceType := arrays.FindFirst(instanceTypes, func(x Quota) bool { return shared.StringEqualsIgnoreCase(x.InstanceTypeID, instanceTypeId) })
```
miguelsorianod, line 81: Can't getBillingModel be used here as it's done in the account.go code?

internal/kafka/internal/services/kafka.go
```diff
@@ -287,12 +287,19 @@ func (k *kafkaService) AssignInstanceType(owner string, organisationId string) (
...
+			if err != nil {
+				return "", err
+			}
+			if hasRhosakQuota {
+				return types.KafkaInstanceType(instanceType.Id), nil
+			}
+		}
 	}
 
 	return types.DEVELOPER, nil
```
miguelsorianod, line 305: A side comment about this which does not seem strictly related to this PR: It returns `developer` but if `developer` is not defined in the supported kafka instances configuration then this would be incorrect.

internal/kafka/internal/services/kafka.go
```diff
@@ -287,12 +287,19 @@ func (k *kafkaService) AssignInstanceType(owner string, organisationId string) (
...
-		return "", err
-	}
-	if hasRhosakQuota {
-		return types.STANDARD, nil
+	for _, instanceType := range k.kafkaConfig.SupportedInstanceTypes.Configuration.SupportedKafkaInstanceTypes {
+		if instanceType.Id == types.DEVELOPER.String() {
+			continue
+		}
+		for _, bm := range instanceType.SupportedBillingModels {
+			hasRhosakQuota, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(owner, organisationId, types.KafkaInstanceType(instanceType.Id), bm)
```
miguelsorianod, line 295: Side comment: This variable name might not make sense. I am guessing the "Rhosak" part of the name referenced to the ams product

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -12,39 +14,57 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 )
 
+const (
+	billingModelStandard = "standard"
```
miguelsorianod, line 18: Same questions around meaning of "billing model" in the quota management list context that I expressed in other comments

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -12,39 +14,57 @@ import (
...
+
 // don't validate billing accounts when using the quota list
 func (q QuotaManagementListService) ValidateBillingAccount(organisationId string, instanceType types.KafkaInstanceType, billingCloudAccountId string, marketplace *string) *errors.ServiceError {
 	return nil
 }
 
-func (q QuotaManagementListService) CheckIfQuotaIsDefinedForInstanceType(username string, organisationId string, instanceType types.KafkaInstanceType) (bool, *errors.ServiceError) {
+func (q QuotaManagementListService) CheckIfQuotaIsDefinedForInstanceType(username string, organisationId string, instanceType types.KafkaInstanceType, kafkaBillingModel config.KafkaBillingModel) (bool, *errors.ServiceError) {
 	orgId := organisationId
+	var org quota_management.Organisation
```
miguelsorianod, line 37: No need to predeclare the variable. Is there some reason is it done this way? Same applies to the `account` variable

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -12,39 +14,57 @@ import (
...
+	var account quota_management.Account
 	org, orgFound := q.quotaManagementList.QuotaList.Organisations.GetById(orgId)
 	userIsRegistered := false
+	serviceAccountIsRegistered := false
+
 	if orgFound && org.IsUserRegistered(username) {
```
miguelsorianod, line 43: Side comment about this PR: It seems there is some kind of priority hierarchy when a username appears in both svcaccounts section and org users section. We should clearly document that in the configuration file

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -72,11 +92,19 @@ func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest, inst
...
+		return "", err
+	}
+	// TODO: find a better place to set this instead of this side effect
+	kafka.DesiredKafkaBillingModel = billingModelID
+
 	dbConn := q.connectionFactory.New().
 		Model(&dbapi.KafkaRequest{}).
-		Where("instance_type = ?", instanceType.String())
+		Where("instance_type = ?", kafka.InstanceType).
+		Where("actual_kafka_billing_model = ?", kafka.DesiredKafkaBillingModel)
```
miguelsorianod, line 105: Could you explain further why this condition has been added and what would be the difference when not specifying it?

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -95,28 +123,75 @@ func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest, inst
...
 
-	if quotaManagementListItem != nil && instanceType == types.STANDARD {
+	if quotaManagementListItem != nil && kafka.InstanceType != types.DEVELOPER.String() {
 		kafkaInstanceSize, e := q.kafkaConfig.GetKafkaInstanceSize(kafka.InstanceType, kafka.SizeId)
 		if e != nil {
 			return "", errors.NewWithCause(errors.ErrorGeneral, e, "Error reserving quota")
 		}
-		if quotaManagementListItem.IsInstanceCountWithinLimit(totalInstanceCount + kafkaInstanceSize.CapacityConsumed) {
+		if quotaManagementListItem.IsInstanceCountWithinLimit(kafka.InstanceType, billingModelID, totalInstanceCount+kafkaInstanceSize.CapacityConsumed) {
+			kafka.ActualKafkaBillingModel = kafka.DesiredKafkaBillingModel
```
miguelsorianod, line 132: We can add a TODO here too of pending to look for a better place / improve this

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -95,28 +123,75 @@ func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest, inst
...
 			return "", errors.MaximumAllowedInstanceReached(message)
 		}
 	}
 
-	if instanceType == types.DEVELOPER && quotaManagementListItem == nil {
+	if kafka.InstanceType == types.DEVELOPER.String() && quotaManagementListItem == nil {
 		if totalInstanceCount >= quota_management.GetDefaultMaxAllowedInstances() {
 			return "", errors.MaximumAllowedInstanceReached(message)
 		}
+		kafka.ActualKafkaBillingModel = kafka.DesiredKafkaBillingModel
```
miguelsorianod, line 143: We can add a TODO here too of pending to look for a better place / improve this

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -95,28 +123,75 @@ func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest, inst
...
 
 	return "", errors.InsufficientQuotaError("Insufficient Quota")
 }
 
+func (q QuotaManagementListService) detectBillingModel(kafka *dbapi.KafkaRequest) (string, *errors.ServiceError) {
+	if kafka.DesiredKafkaBillingModel != "" {
+		return kafka.DesiredKafkaBillingModel, nil
+	}
+	if kafka.InstanceType == types.DEVELOPER.String() {
+		return billingModelStandard, nil
```
miguelsorianod, line 155: Side comment: This assumes that developer is defined as an instance type in the instances type configuration, and also assumes than standard exists for developer types

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -12,39 +14,57 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 )
 
+const (
+	billingModelStandard = "standard"
+	defaultBillingModel  = billingModelStandard
+)
+
 type QuotaManagementListService struct {
```
miguelsorianod, line 22: Given that the behavior of this is not trivial can we add method documentation for the non-trivial methods?

pkg/quota_management/account.go
ziccardi: I agree, but this is how it worked until now: you can define max_allowed_instance at `org` level, but if it is `0`, the default value is used. Now we just added one more level (billing model).

pkg/quota_management/account.go
```diff
@@ -1,20 +1,62 @@
...
+func (account Account) GetMaxAllowedInstances(instanceTypeID string, billingModelID string) int {
+	bm, ok := account.getBillingModel(instanceTypeID, billingModelID)
+
+	if !ok {
+		return 0
+	}
+
+	if bm.Allowed <= 0 {
+		if account.MaxAllowedInstances <= 0 {
+			return MaxAllowedInstances
```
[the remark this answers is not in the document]
ziccardi: Agree, but that would be a breaking change. Let's create a JIRA for this.

internal/kafka/internal/services/quota/quota_management_list_service.go
ziccardi: In the quota_list configuration we specify how many instances we want to allow for a specific `instance_type/pair`.
If I don't specify the `actual_kafka_billing_model` I will sound all the created instances without considering the billing_model.

I'm now thinking if I shouldn't change it to:
```go
Where("actual_kafka_billing_model = ? or desired_billing_model = ?", kafka.DesiredKafkaBillingModel, kafka.DesiredKafkaBillingModel)
```

to count the already accepted requests too.

pkg/quota_management/account.go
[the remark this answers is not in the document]
miguelsorianod: Changing the name of this variable wouldn't be a breaking change right?
What would be a breaking change would be changing the behavior described in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368/files#r1022782423

internal/kafka/internal/services/kafka.go
ziccardi: Agree. Currently, the `developer` instance type is an alien beast: if you don't have any quota configured, you can create a `developer`. No configuration needs to be present, and no quota either.

internal/kafka/internal/services/quota/quota_management_list_service.go
ziccardi: Removed.

config/quota-management-list-configuration.yaml
ziccardi: This is the quota_list, ams is not related to this.
As in AMS we have a billin_model in the AMS configuration (saying how much of a resource with a specified billing model is allowed), here we have a billing_model in the quota_list saying how many `instance_type` instances are allowed for a specific billing_model.
The set of accepted billing models here is the list of configured kafka_billing_models for the given instance type.

miguelsorianod: Could we call them then kafka_billing_models instead?

pkg/quota_management/account.go
miguelsorianod: It is not adding one more level, but different behavior:
At the top level, 0 or unset means "use a default value"
In this new level, 0 or unset means "inherit from the parent level"

pkg/quota_management/account.go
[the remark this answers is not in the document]
miguelsorianod: Even if we don't change the behavior for now, would it be possible to change the variable name to one more descriptive? something like `DefaultMaxAllowedInstances`, `DefaultAccountMaxAllowedInstances` or something similar

pkg/quota_management/organisation.go
miguelsorianod: I am not sure making the rewording is redundant. Due to the term "billing model" is overloaded through the code, and the source of truth of the accepted values belong to different packages I think it might be good to be more explicit.

From the comment https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368#discussion_r1023146896 I understand this would be kafka billing model.

Unless we don't want to tie the name to it. But then at least we should be very clear that on the current implementation that's the expectation

internal/kafka/internal/services/quota/quota_management_list_service.go
miguelsorianod: This is a tricky question because of potential race conditions.

* If we only consider desired:
  *  It might happen that we count as being in use a kafka that ends up failing and therefore counting more than what will actually be used
  * It might happen that we don't count a kafka instance that is using that billing model (actual) but is in the process of being promoted to another type
* If we only consider actual:
  * Depending on the order of events we will not detect a kafka instance that had desired and no actual but it was in the process of setting actual. Ending up with more used kafka instances than the ones allowed
  * It might happen that we count as being in use a kafka that ends up being promoted to another type, counting more than what will actually be used
* If we consider both desired an actual then the previously explained issues of "counting more than what will end up being used" might happen

I think the approach of using an "or" is reasonable. It will potentially give us "false positives" where we will count more than what will end up being used but opting for the side of caution is what I would go for.

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -12,39 +15,58 @@ import (
...
 		return true, nil
 	}
 
 	return false, nil
 }
 
-func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest, instanceType types.KafkaInstanceType) (string, *errors.ServiceError) {
+// ReserveQuota - tries to reserve the quota for the received kafka request
+func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *errors.ServiceError) {
 	if !q.quotaManagementList.EnableInstanceLimitControl {
```
JameelB, line 70: Note that this flag controls whether the `desired_billing_model` is set in the KafkaRequest object in the database if the `billing_model`  is not specified in the Kafka request payload as its set further down this function.

A bit of an edge case if KFM was ran initially with this disabled and then re-ran it with this enabled. It could affect the calculations below. This probably is unlikely to happen though.

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,64 @@
 # In the future it may support service accounts outside RH.
 # The structure of registered service accounts is:
 #       - username: is the account of the user. The username must be unique
-#       - max_allowed_instances: is the maximum number of instances this user can create.
+#       - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
+#         This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
```
miguelsorianod, line 9: All occurrences of `allowed` in this file should be updated to the new name

pkg/quota_management/quota.go
```diff
@@ -0,0 +1,38 @@
+package quota_management
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+)
+
+var defaultBillingModel = BillingModel{
+	Id:              "STANDARD",
```
miguelsorianod, line 9: Let's change this to lowercase

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -72,11 +94,19 @@ func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest, inst
 
 	var kafkas []*dbapi.KafkaRequest
 
+	billingModelID, err := q.detectBillingModel(kafka)
+	if err != nil {
+		return "", err
+	}
+	// TODO: find a better place to set this instead of this side effect
+	kafka.DesiredKafkaBillingModel = billingModelID
```
JameelB, line 102: Is it possible to move this at the start of this function or before this function is called? The error messages above will always return `0` max allowed instances if the billing model id is not specified in the request payload which can be a bit confusing when the quota configuration has the `max_allowed_instances` set to a non-zero number. 

Example error message received when testing this locally even though the `max_allowed_instances` of the 'standard' billing model was set to 2
```
"organization '13640203' has reached a maximum number of 0 allowed streaming units"
```

It will also avoid the scenario as described in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368/files#r1024204938

Otherwise, we probably need to update the error message further down this function after the billing model has been detected.

JameelB, line 102: Is it possible to move this at the start of this function or before this function is called? The error messages above will always return `0` max allowed instances if the billing model id is not specified in the request payload which can be a bit confusing when the quota configuration has the `max_allowed_instances` set to a non-zero number. 

Example error message received when testing this locally even though the `max_allowed_instances` of the 'standard' billing model was set to 2
```
"organization '13640203' has reached a maximum number of 0 allowed streaming units"
```

It will also avoid the scenario as described in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368/files#r1024204938

Otherwise, we probably need to update the error message further down this function after the billing model has been detected.

JameelB (commented):
Verified changes locally, works as expected. Max allowed instance was in preference of billing model level > org/user level > overall `max-allowed-instance`.

Also tested with two users belonging to the same org with various scenarios. I was able to create the number of Kafka instances as per the quota configuration.

Only found some minor issue as defined below.

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,64 @@
...
 #         Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+#       - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
+#         - instance_type_id: the id of the instance type for which we are configuring the quota
```
JameelB, line 12: We should note here that the instance type id must match the id of an instance type configured in the `supported_instance_types` configuration

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,64 @@
...
+#         - billing_models: the list of billing models that we are going to configure.
+#           The set of accepted billing models, depends on the KafkaBillingModel configuration.
+#           This is defined as follows:
+#           - id: the id of the billing model we are going to configure. The set of valid IDs here is the set of configured instance_types in KFM.
```
JameelB, line 16:
```suggestion
#           - id: the id of the billing model we are going to configure. The set of valid IDs here is the set of the KafkaBillingModel IDs available in the 'supported_billing_models' set per instance type in the 'supported_instance_types' configuration.
```

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,64 @@
...
 # The structure of registered service accounts is:
 #       - username: is the account of the user. The username must be unique
-#       - max_allowed_instances: is the maximum number of instances this user can create.
+#       - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
+#         This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
 #         Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+#       - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
+#         - instance_type_id: the id of the instance type for which we are configuring the quota
+#         - billing_models: the list of billing models that we are going to configure.
+#           The set of accepted billing models, depends on the KafkaBillingModel configuration.
```
JameelB, line 14: There's an open comment by @miguelsorianod around the naming of this field (see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1368#discussion_r1022703770).

Since this is talking about the KafkaBillingModel it might be better to just rename this as `kafka_billing_model` , wdyt?

JameelB, line 14:
```suggestion
#         - billing_models: the list of billing models that we are going to configure.
#           The set of accepted billing models, depends on the KafkaBillingModels defined per instance type in the 'supported_instance_types' configuration. 
```

config/quota-management-list-configuration.yaml
```diff
@@ -5,25 +5,64 @@
 # In the future it may support service accounts outside RH.
 # The structure of registered service accounts is:
 #       - username: is the account of the user. The username must be unique
-#       - max_allowed_instances: is the maximum number of instances this user can create.
+#       - max_allowed_instances: is the maximum number of instances of a given instance_type/billing_model pair this user can create.
+#         This value is used if `allowed` field (inside `billing_model`) is 0 or omitted.
 #         Defaults to the global value of `max-allowed-instances` which has different values for distinct environments.
+#       - granted_quota: the configuration of the quota granted to the user. The configuration is defined as follows:
```
JameelB, line 11: Can we make a note here on the defaults given to this when its set to an empty array?

config/quota-management-list-configuration.yaml
JameelB: Can we also make a note here on the defaults given to the `billing_models` when its set to an empty array

internal/kafka/internal/services/quota/quota_management_list_service.go
ziccardi: Added the or clause

JameelB (approved):

PR #1369, opened 2022-11-01
fix: removed dependency on the bash timeout command

ziccardi:
## Description
The bash timeout command is not available on all systems. Replaced the `timeout` with a `while` loop.

The Keycloak config script was checking if keycloak was available using the `timeout command` to interrupt an infinite loop used to check the keycloak status.
However, the timeout command is not available by default on many systems (mac, for example), and the keycloak setup failed.
Moreover, if keycloak was not ready, the script didn't stop and all subsequent `curl` commands were failing.

With this PR we won't have that dependency anymore and, if keycloak is not ready, the script doesn't attempt the cur commands.

- [ ] Code Review completed

### DIFF
--- a/scripts/mas_sso_config.sh
+++ b/scripts/mas_sso_config.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+MAX_KEYCLOAK_ATTEMPT_COUNT=24
+
 TOKEN_PATH="/auth/realms/master/protocol/openid-connect/token"
 
 CLIENT_ID=admin-cli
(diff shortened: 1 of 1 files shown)


scripts/mas_sso_config.sh
```diff
@@ -8,155 +9,172 @@ FLEET_OPERATOR_ROLE=kas_fleetshard_operator
...
+while [[ $COUNT -lt $MAX_KEYCLOAK_ATTEMPT_COUNT ]]; do
+  ((COUNT+=1))
+  if [[ $(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8180/auth/realms/master) == "200" ]]; then
+    echo "[$COUNT/$MAX_KEYCLOAK_ATTEMPT_COUNT] - Keycloak ready"
+    KEYCLOAK_READY=1
+    break
+  fi
+
+  echo "[$COUNT/$MAX_KEYCLOAK_ATTEMPT_COUNT] - Waiting for keycloak server at localhost:8180"
+  sleep 10
```
miguelsorianod, line 23: Maybe we can decrease it a little bit more so we don't need to wait 10 seconds in the worst case?

scripts/mas_sso_config.sh
```diff
@@ -8,155 +9,172 @@ FLEET_OPERATOR_ROLE=kas_fleetshard_operator
...
+done
+
+if [[ $KEYCLOAK_READY ]]; then
```
miguelsorianod, line 26: Instead of an if-else at that level we can do the opposite:
If it is not ready then we show that error message and we exit with err code (exit 1)
Otherwise we just continue running the rest of the tasks

scripts/mas_sso_config.sh
```diff
@@ -8,155 +9,172 @@ FLEET_OPERATOR_ROLE=kas_fleetshard_operator
...
+      "id": "'${ADMIN_FULL_ROLE_ID}'",
+      "name": "'${FLEETMANAGER_ADMIN_ROLE}'"
+    }]')
+  if [ $? -ne 0 ] ; then
+      exit 1
+  fi
+
+  echo "Associated role ${FLEETMANAGER_ADMIN_ROLE}(${ADMIN_FULL_ROLE_ID}) with client ${FLEETMANAGER_ADMIN_CLIENT}(${ADMIN_CLIENT_ID}) service account"
+else
+  echo "Keycloak server not running at localhost:8180. No realm configuration will be applied"
```
miguelsorianod, line 177: Can we update this message to add `error: ...` at the beginning to make more explicitly that this is unexpected?

miguelsorianod: Or instead of error something like `max attempts reached: ...`

scripts/mas_sso_config.sh
ziccardi: done

scripts/mas_sso_config.sh
ziccardi: done

scripts/mas_sso_config.sh
ziccardi: done

miguelsorianod (approved):
:+1:

PR #1388, opened 2022-11-08
feat: validate kafka billing models in kafka creation from instance types config

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-10014.

This PR updates the kafka billing models validation in the kafka create endpoint. The set of valid values is now gathered from the configured supported kafka billing models in `config/kafka-instance-types-configuration.yaml`.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/handlers/kafka.go
+++ b/internal/kafka/internal/handlers/kafka.go
@@ -50,8 +50,8 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
 			ValidateKafkaClaims(ctx, ValidateUsername(), ValidateOrganisationId()),
 			ValidateCloudProvider(ctx, h.service, &kafkaRequestPayload, h.providerConfig, "creating kafka requests"),
 			ValidateKafkaPlan(ctx, h.service, h.kafkaConfig, &kafkaRequestPayload),
+			validateKafkaBillingModel(ctx, h.service, h.kafkaConfig, &kafkaRequestPayload),
 			ValidateBillingCloudAccountIdAndMarketplace(ctx, h.service, &kafkaRequestPayload),
-			ValidateBillingModel(&kafkaRequestPayload),
 		},
 		Action: func() (interface{}, *errors.ServiceError) {
 			convKafka := presenters.ConvertKafkaRequest(kafkaRequestPayload)
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -9,7 +9,6 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
-	v1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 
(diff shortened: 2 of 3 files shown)


miguelsorianod:
The PR is still not finished (pending tests) but I want a review first and discussion on the following point:
* By merging this PR the new kafka billing models defined for long lived eval instances will be accepted when performing kafka request creation. Do we want to keep the old behavior for now or do we just include the new code assuming no one will create it in stage/prod with one of the new values?

ziccardi:
> The PR is still not finished (pending tests) but I want a review first and discussion on the following point:
> 
> * By merging this PR the new kafka billing models defined for long lived eval instances will be accepted when performing kafka request creation. Do we want to keep the old behavior for now or do we just include the new code assuming no one will create it in stage/prod with one of the new values?

To be extremely sure, I believe we can merge this PR after all the other PRs for long lived trial instances are merged

ziccardi:
@miguelsorianod I believe this should be retargeted to the feature branch. It would be great if we could get it merged so that we can start creating `evals` without workarounds

internal/kafka/internal/handlers/validation.go
```diff
@@ -29,14 +29,30 @@ var ValidKafkaClusterNameRegexp = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])
...
+		if err != nil {
+			return errors.ToServiceError(err)
+		}
+
+		kafkaBillingModel := shared.SafeString(kafkaRequestPayload.BillingModel)
+		// No explicitly set kafka billing mode is allowed for now, in which case
+		// an implementation-defined default is chosen
+		if kafkaBillingModel == "" {
+			return nil
+		}
```
ziccardi, line 49:
```suggestion
		// No explicitly set kafka billing mode is allowed for now, in which case
		// an implementation-defined default is chosen
		if shared.StringEmpty(kafkaRequestPayload.BillingModel) {
			return nil
		}
```

miguelsorianod:
> @miguelsorianod I believe this should be retargeted to the feature branch. It would be great if we could get it merged so that we can start creating `evals` without workarounds

Rebased on top of the new feature branch and PR retargeted to it.

Need to check if the changes in this PR are still relevant with the recent changes made around AMS quota in the feature branch.

internal/kafka/internal/handlers/validation.go
```diff
@@ -36,14 +35,29 @@ var ClusterIdLength = 32
 
 const minimunNumberOfNodesForTheKafkaMachinePool = 3
 
-func ValidateBillingModel(kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
+func validateKafkaBillingModel(ctx context.Context, kafkaService services.KafkaService, kafkaConfig *config.KafkaConfig, kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
```
miguelsorianod, line 38: I renamed the method to `validateKafkaBillingModel` to be more specific about what we validate in here

internal/kafka/internal/handlers/kafka.go
```diff
@@ -50,8 +50,8 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
 			ValidateKafkaClaims(ctx, ValidateUsername(), ValidateOrganisationId()),
 			ValidateCloudProvider(ctx, h.service, &kafkaRequestPayload, h.providerConfig, "creating kafka requests"),
 			ValidateKafkaPlan(ctx, h.service, h.kafkaConfig, &kafkaRequestPayload),
+			validateKafkaBillingModel(ctx, h.service, h.kafkaConfig, &kafkaRequestPayload),
```
miguelsorianod, line 53: I moved the order of the validation to do it before the cloud account id and marketplace validation as it is less resource intensive than the later.

I think it shouldn't impact the behavior. If you think otherwise let me know.

miguelsorianod:
This is ready for review

internal/kafka/internal/handlers/validation.go
miguelsorianod: I also made it unexported as there is no need to expose it outside of the package

internal/kafka/internal/handlers/validation.go
```diff
@@ -36,14 +35,29 @@ var ClusterIdLength = 32
...
-		// the billing model can only be either standard or marketplace
-		billingModel := shared.SafeString(kafkaRequestPayload.BillingModel)
-		if billingModel != "" && billingModel != string(v1.BillingModelStandard) && billingModel != string(v1.BillingModelMarketplace) {
-			return errors.InvalidBillingAccount("invalid billing model: %s, only %v and %v are allowed", billingModel,
-				v1.BillingModelStandard, v1.BillingModelMarketplace)
+		// No explicitly set kafka billing mode is allowed for now, in which case
+		// an implementation-defined default is chosen
+		if shared.StringEmpty(kafkaRequestPayload.BillingModel) {
+			return nil
+		}
```
ziccardi, line 44: I believe the implementation is correct, but I don't understand the comment

ziccardi (approved):
Just a small consideration about a `comment`

internal/kafka/internal/handlers/validation.go
miguelsorianod: It just means that kafka billing model does not need to be provided explicitely. In case it is not provided then we choose a default (in the implementation)

PR #1390, opened 2022-11-10
Added support to the AMSQuota manager for the new KafkaBillingModel

ziccardi:
## Description
Added support for Kafka Billing Models to AMS so that EVAL instances can now be requested.

- [ ] Code Review completed


internal/kafka/internal/services/quota/internal/utils/ams_quota_config_provider.go
```diff
@@ -0,0 +1,27 @@
...
+import amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+
+type AMSQuotaConfigProvider interface {
+	GetQuotaCostsForProduct(organizationID, resourceName, product string) ([]*amsv1.QuotaCost, error)
+}
+
+// It can happen that multiple resolvers will ask for the same quota costs. To avoid multiple
+// calls to AMS, we cache the result here
+type cachingQuotaConfigProvider struct {
+	cache         map[string][]*amsv1.QuotaCost
```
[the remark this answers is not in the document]
ziccardi: I don't think there is any cache in the client:

- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ae2cbdb867743fa0fabf09d139f8b650ccb999fb/pkg/client/ocm/client.go#L577
- https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ae2cbdb867743fa0fabf09d139f8b650ccb999fb/pkg/client/ocm/client.go#L536

internal/kafka/internal/services/quota.go
```diff
@@ -16,5 +16,5 @@ type QuotaService interface {
 	// DeleteQuota deletes a reserved quota
 	DeleteQuota(subscriptionId string) *errors.ServiceError
 	// ValidateBillingAccount validates if a billing account is contained in the quota cost response
-	ValidateBillingAccount(organisationId string, instanceType types.KafkaInstanceType, billingCloudAccountId string, marketplace *string) *errors.ServiceError
+	ValidateBillingAccount(organisationId string, instanceType types.KafkaInstanceType, billingModelID string, billingCloudAccountId string, marketplace *string) *errors.ServiceError
```
machi1990, line 19: Might be good to extract the parameters for this method and possibly other quota related methods into a dedicated type e.g `BillingInfo` or `QuotaParameters`, `SomeThingElse` etc...

internal/kafka/internal/services/quota/internal/utils/misc.go
```diff
@@ -0,0 +1,40 @@
+package utils
```
machi1990, line 1: love the file name `misc.go` in this package ;-). I cannot think of a better name to suggest but I find the filename not clear about the utility functions that are being offered. 
If the utility functions are not common enough for us to come with a common and clear name then an alternative would be to have a utility file per function.

internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+)
+
+/********************************************************************************************************************
+ * This resolver is used when the user didn't specify the `billing model` and the `marketplace` in the kafka request.
+ * It implements the legacy logic: if the user has quota for `standard`, it returns `standard`. If he has quota for
```
machi1990, line 13: `It implements the legacy logic`

Will this legacy logic go away? 
1. If not, then it's not really a legacy logic and it'll be good to update the comment to avoid future confusions
2. If yes, then let's keep track of this in an JIRA issue as well to not forget about this

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
```diff
@@ -0,0 +1,76 @@
...
+		return BillingModelDetails{}, errors.NewWithCause(errors.ErrorGeneral, err, "error reserving quota")
+	}
+
+	quotaCosts, err := r.quotaConfigProvider.GetQuotaCostsForProduct(orgId, kafkaBillingModel.AMSResource, kafkaBillingModel.AMSProduct)
+	if err != nil {
+		return BillingModelDetails{}, errors.InsufficientQuotaError("%v: error getting quotas for product %s", err, kafkaBillingModel.AMSProduct)
+	}
+	if idx, cloudAccount := arrays.FindFirst(getCloudAccounts(quotaCosts), func(cloudAccount *v1.CloudAccount) bool {
+		return cloudAccount.CloudAccountID() == kafka.BillingCloudAccountId && (cloudAccount.CloudProviderID() == kafka.Marketplace || kafka.Marketplace == "")
+	}); idx != -1 {
```
machi1990, line 53: I find this part a bit long and hard to read.
Maybe it can be splitted?

internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
ziccardi: We will probably simplify the logic in the future. A Jira already exists for that. I will however remove the 'legacy logic' text.

internal/kafka/internal/services/quota/internal/utils/billing_model_resolver.go
```diff
@@ -0,0 +1,71 @@
...
+
+/********************************************************************************************************************
+ * This is the root of the billing resolver chain.
+ * It demands the resolution of the billing model to the defined resolvers and stops the chain as soon as one of
+ * the resolvers is able to manage the request.
+ ********************************************************************************************************************/
+
+type BillingModelResolver interface {
+	SupportRequest(kafka *dbapi.KafkaRequest) bool
+	Resolve(orgID string, kafka *dbapi.KafkaRequest) (BillingModelDetails, error)
```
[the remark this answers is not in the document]
ziccardi: The organisationID in the kafka request is the `external organisation id`. The resolver needs the organisation ID (the one returned by `GetOrganisationIdFromExternalId`)

internal/kafka/internal/services/quota.go
ziccardi: will do in a future PR.

JameelB (approved):

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
```diff
@@ -0,0 +1,118 @@
...
+
+/********************************************************************************************************************
+ * This resolver is used when the user specifies a `DesiredBillingModel` in the request.
+ * If the specified `DesiredBillingModel` is a `marketplace`, attempts are done to detect the marketplace type (if
+ * not specified)
+ ********************************************************************************************************************/
+
+var _ BillingModelResolver = &kafkaBillingModelResolver{}
+
+type kafkaBillingModelResolver struct {
```
miguelsorianod, line 18: Is there a difference in behavior here?

Before this PR checking whether there is quota available is performed whereas in kafka billing model resolver it is unconditionally returned without checking it

internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
```diff
@@ -0,0 +1,145 @@
...
+	var kafkaBillingModels []config.KafkaBillingModel
+	standardBillingModel, err := r.kafkaConfig.GetBillingModelByID(kafka.InstanceType, "standard")
+	if err == nil {
+		kafkaBillingModels = append(kafkaBillingModels, standardBillingModel)
+	}
+	marketplaceBillingModel, err := r.kafkaConfig.GetBillingModelByID(kafka.InstanceType, "marketplace")
+	if err == nil {
+		kafkaBillingModels = append(kafkaBillingModels, marketplaceBillingModel)
+	}
+	trialBillingModel, err := r.kafkaConfig.GetBillingModelByID(kafka.InstanceType, "trial")
```
miguelsorianod, line 45: This is referencing `trial`. Is that correct after recent changes where trial supported billing model does not exist anymore?

internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (*simpleBillingModelResolver) SupportRequest(kafka *dbapi.KafkaRequest) bool {
+	return kafka.BillingCloudAccountId == "" && kafka.Marketplace == ""
+}
+
+// Resolve - tries to resolve the KafkaBillingModel and amsBillingModel by analysing the parameter received by the user.
+// This function checks if for the requested instance type the `standard` and `marketplace` billing model are defined
+// and produces a sorted array contained those billing models (if defined) in the priority order (standard first,
+// marketplace second), then demands the resolving logic to the `resolve` method.
```
miguelsorianod, line 33: `trial` is not mentioned here when describing the order. However I see it instantiated below

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
```diff
@@ -0,0 +1,80 @@
...
+			(kafka.Marketplace == "" || cloudAccount.CloudProviderID() == kafka.Marketplace)
+	})
+
+	if idx != -1 {
+		if !hasSufficientMarketplaceQuota(quotaCosts, kafkaInstanceSize.QuotaConsumed) {
+			return BillingModelDetails{}, errors.InsufficientQuotaError("quota does not support marketplace billing or has insufficient quota")
+		}
+
+		// assign marketplace in case it was not provided in the request
+		kafka.Marketplace = cloudAccount.CloudProviderID()
```
miguelsorianod, line 63: Shouldn't this be done only when it was not provided in the request?

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
```diff
@@ -0,0 +1,80 @@
...
+		return BillingModelDetails{}, err
+	}
+
+	return r.resolve(orgId, kafka, kafkaBillingModel)
+}
+
+// resolve - tries to resolve the AMS Billing Model by performing the following checks:
+// 1) Tries to retrieve the cloud account. A failure here will result in an error.
+// 2) Checks if the user has sufficient marketplace quota. If not, an error is returned.
+// 3) Tries to get the marketplace associated with the cloud provider.
```
miguelsorianod, line 39: Shouldn't 3) be reworded adding "if not provided in the request"  at the end?

miguelsorianod (commented):
Reviewed now as I was not available last week.

Great work :+1: 

I left some comments / feedback

internal/kafka/internal/services/quota/internal/utils/billing_model_resolver.go
```diff
@@ -0,0 +1,71 @@
...
+var supportedCloudProviders = []string{CloudProviderAWS, CloudProviderRHM, CloudProviderAzure}
+
+/********************************************************************************************************************
+ * This is the root of the billing resolver chain.
+ * It demands the resolution of the billing model to the defined resolvers and stops the chain as soon as one of
+ * the resolvers is able to manage the request.
+ ********************************************************************************************************************/
+
+type BillingModelResolver interface {
+	SupportRequest(kafka *dbapi.KafkaRequest) bool
```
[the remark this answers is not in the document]
ziccardi: Depending on which values the user has specified in the request, a `Resolver` can be able to try or not to resolve the billing model. For example, one resolver could require the `DesiredBillingModel` to be valued, while another resolver could require the `marketplace` to be valued.

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
ziccardi: This is done only if a `cloudAccount` is found (`idx != -1`) where 
```
cloudAccount.CloudAccountID() == kafka.BillingCloudAccountId &&
			(kafka.Marketplace == "" || cloudAccount.CloudProviderID() == kafka.Marketplace)
```

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+)
+
+/********************************************************************************************************************
+ * This resolver is used when the user specifies a `DesiredBillingModel` in the request.
```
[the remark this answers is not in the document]
ziccardi: It means `when DesiredKafkaBillingModel != ""`

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
ziccardi: Can you elaborate more? Before this PR, the code to resolve the billing model was in 3 main `if` clauses.
I split that code into three objects (resolvers), but the main logic should be the same.
What is the check that is missing now?

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
```diff
@@ -0,0 +1,118 @@
...
+			kafkaConfig:         k.kafkaConfig,
+		}
+		billingModelDetails, err := accountBillingModelResolver.resolve(orgId, kafka, kafkaBillingModel)
+		if err != nil {
+			return "", err
+		}
+		return billingModelDetails.AMSBillingModel, nil
+	}
+
+	var billingModelsToCheck []config.KafkaBillingModel
```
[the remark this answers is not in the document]
ziccardi: If the resolver has not been able to resolve the AMS billing model, it falls back to the `simple resolver` (the one that checks for `standard` and then for `marketplace`)
The `resolve` method of the `simple resolver` (maybe a better name should be given to that resolver) takes an array of `billing models` ordered by priority and returns the first one the user has quota for

However, it doesn't make sense to ask the `simple resolver` to check for `standard or marketplace` if the user specified a `DesiredBillingModel` that does not support it, so before adding it to the array, I check if the `DesiredBillingModel` supports it or not.

internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
ziccardi: Yeah. We have to remove it, since we removed it from the config. Good catch.

internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
ziccardi: Yeah. We have to remove it.

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
```diff
@@ -0,0 +1,118 @@
...
+	}, nil
+}
+
+// detectAmsBillingModel - tries to detect the `ams billing model`
+// 1) If the user has specified a cloud account id, it falls back to the `cloud account billing model resolver`
+// 2) If no cloud account has been specified, it falls back to the `simple billing model resolver`
+func (k *kafkaBillingModelResolver) detectAmsBillingModel(orgId string, kafka *dbapi.KafkaRequest, kafkaBillingModel config.KafkaBillingModel) (string, error) {
+	validAmsBillingModels := kafkaBillingModel.AMSBillingModels
+	if kafka.BillingCloudAccountId != "" {
+		validAmsBillingModels = arrays.Filter(validAmsBillingModels, arrays.StringHasPrefixIgnoreCasePredicate("marketplace-"))
```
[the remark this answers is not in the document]
ziccardi: Good catch. This should be replaced with a call to `HasSupportForMarketplace`

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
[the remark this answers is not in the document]
miguelsorianod: :+1: . When thinking about request I was thinking the API request, no the `KafkaRequest` type. It is in the context of the `KafkaRequest` type so it makes sense then

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
miguelsorianod: The check that might be missing is whether there is quota availale

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
miguelsorianod: Yes.

But we should respect what's provided in the request  regarding the marketplace.

If the user has provided a CloudAccountID AND a Marketplace in the API request then the marketplace value used should be the one provided in the request and not the one found in the QuotaCost response.
If it turns out that the Cloud Account ID does not exist in that Cloud Account provider then an error should be returned.

The mentioned check also checks that the Cloud Provider ID has to match the marketplace provided in the kafka request when marketplace is not empty so it is ok to unconditionally assign the returned cloudprovider id in the quotacost. It is unnecessary but it would be correct and it avoids using one additional conditional in the code :+1:

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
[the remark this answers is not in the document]
miguelsorianod: I understand that.

My comment was more on the side of that it seems we are using a `config.KafkaBillingModel` to represent an AMS billing model and not a a Kafka Billing Model.

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
ziccardi: The kafka billing model resolver is used when the `user` explicitly asks for a spcific kafka billing model (DesiredBillingModel is not nil).

If the requested billing model supports only one ams billing model, there is no need to make any detection, since that is exactly what the user asked for. If there is no quota for that, it will fail to reserve the quota.

If the requested billing model supports more than one ams billing model, then the detection is triggered, and the first ams billing model for whom the user has quota for is returned.

There is no change in behaviour: if the user has not enough quota, it will return a no quota error in both cases.

miguelsorianod: > If the requested billing model supports only one ams billing model, there is no need to make any detection, since that is exactly what the user asked for. If there is no quota for that, it will fail to reserve the quota.

That is the change in behavior I referred to.

Maybe "change in behavior" was not specific enough: The user would experience the same behavior although a different error would be returned.

Before, the quota check was explicitly performed before reserving quota and now it is delegated to the AMS API.

In other resolvers this check continues to be explicitly performed before sending the request to the AMS API

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -170,17 +123,46 @@ func (q amsQuotaService) ValidateBillingAccount(organisationId string, instanceT
...
+		history[kbm.AMSResource+kbm.AMSProduct] = 1
+		return q.validateBillingAccount(organisationId, instanceType, kbm, billingCloudAccountId, marketplace) == nil
+	}) {
+		return nil
+	}
+
+	return errors.InvalidBillingAccount("we have not been able to validate your billingAccountID")
+}
+
 func (q amsQuotaService) CheckIfQuotaIsDefinedForInstanceType(username string, externalId string, instanceType types.KafkaInstanceType, kafkaBillingModel config.KafkaBillingModel) (bool, *errors.ServiceError) {
```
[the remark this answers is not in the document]
ziccardi: The method comes from the QuotaService interface. The Quota List implementation of that interface still uses it.

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
ziccardi: Nope. This resolver runs when the `BillingCloudAccountId` is empty (see `SupportRequest`)

internal/kafka/internal/services/quota/ams_quota_service.go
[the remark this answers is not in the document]
miguelsorianod: :+1:

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
miguelsorianod: The check in `SupportRequest` is when BillingCloudAccountId is not empty :thinking: :

```
	return kafka.BillingCloudAccountId != ""
```

ziccardi: Sorry, I got confused. You are right. Moreover, the check on BillingCloudAccountId is not related to what you were asking (the marketplace).
I will update the comment.

internal/kafka/internal/services/quota/internal/utils/billing_model_details.go
```diff
@@ -0,0 +1,8 @@
+package utils
```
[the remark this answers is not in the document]
ziccardi: I just didn't like the package to be called `internal`

internal/kafka/internal/services/quota/quota_management_list_service_test.go
```diff
@@ -289,6 +289,26 @@ var kafkaSupportedInstanceTypesConfig = config.KafkaSupportedInstanceTypesConfig
 			{
 				Id:          "standard",
 				DisplayName: "Standard",
+				SupportedBillingModels: []config.KafkaBillingModel{
```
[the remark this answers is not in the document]
ziccardi: I was probably doing something I didn't do anymore in the end. They are leftovers. Removing.

PR #1391, opened 2022-11-10
Add kafka desired and existing billingmodel

miguelsorianod:
## Description
Related to https://issues.redhat.com/browse/MGDSTRM-10006.

As part of the long-lived eval instances functionality a long-lived eval instance will be able to be promoted to a non-eval instance. Due to this, we need to keep track of both the desired and actual kafka billing models being used, as they might not be in-sync at all times.

This PR performs two main changes:
1. A new column `desired_kafka_billing_model` is added to the kafka request DB type
2. The column `billing_model` is renamed to `actual_kafka_billing_model`

DB migrations have been included to transition from the old to the new columns model.

This PR does not perform any changes at openapi definition level as those are not strictly needed. If we deem we want to perform any modification at that level another PR will be opened to deal with that part.

Some additional remarks:
* The kafka get endpoint definition has not been modified. The `billing_model` value returned in the endpoint is the "actual billing model"
* The kafka create endpoint definition has not been modified. The `billing_model` field sets the "desired billing model"
* For AMS, the "actual billing model" is set only when the cluster authorization ends successfully and with enough quota

## Verification Steps

* Tests pass

Verify the following scenarios:
* Run kas-fleet-manager with code before this PR. Make sure that kas-fleet-manager is run with AMS enabled. Then create a kafka request and wait until it is ready. Take note of the `billing_model` column value set in the DB
* Run db migrations with the code from this PR. Then verify that:
  1. A new `desired_kafka_billing_model` column exists, and it contains the value that `billing_model` originally contained
  1. The `billing_model` column does not exist anymore and now there is a column named `actual_kafka_billing_model`. Verify that this new column also contains the value that `billing_model` originally contained
  1.  Perform a GET kafka request to the kfm api. Verify that the field `billing_model` contains the value of `actual_kafka_billing_model`
* Run kas-fleet-manager with code from this PR. Make sure that kas-fleet-manager is run with AMS enabled. Then create a kafka request and wait until it is ready. Then verify that:
  1. The `desired_kafka_billing_model` column contains the value of the expected kafka billing model
  1. The `actual_kafka_billing_model` column contains the value of the expected kafka billing model
  1. Perform a GET kafka request to the kfm api. Verify that the field `billing_model` contains the value of `actual_kafka_billing_model`

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/migrations/20221109150000_rename_kafka_billing_model_column.go
+++ b/internal/kafka/internal/migrations/20221109150000_rename_kafka_billing_model_column.go
@@ -1,0 +1,32 @@
+package migrations
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func renameKafkaBillingModelColumn() *gormigrate.Migration {
+	const oldKafkaBillingModelFieldName string = "billing_model"
+	const newKafkaBillingModelFieldName string = "actual_kafka_billing_model"
+
+	type KafkaRequest struct {
+		BillingModel            string `json:"billing_model"`
+		ActualKafkaBillingModel string `json:"actual_kafka_billing_model"`
+	}
+
+	return &gormigrate.Migration{
+		ID: "20221109150000",
+		Migrate: func(tx *gorm.DB) error {
+			if !tx.Migrator().HasColumn(KafkaRequest{}, newKafkaBillingModelFieldName) {
+				return tx.Migrator().RenameColumn(KafkaRequest{}, oldKafkaBillingModelFieldName, newKafkaBillingModelFieldName)
+			}
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
+			if !tx.Migrator().HasColumn(KafkaRequest{}, oldKafkaBillingModelFieldName) {
+				return tx.Migrator().RenameColumn(KafkaRequest{}, newKafkaBillingModelFieldName, oldKafkaBillingModelFieldName)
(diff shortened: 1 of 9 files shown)


pawelpaszki (commented):
the billing model assignment, retrieval and migrations work as expected. I'll let other folks have a look as well

pawelpaszki (approved):

internal/kafka/internal/migrations/20221109150000_rename_kafka_billing_model_column.go
```diff
@@ -0,0 +1,32 @@
+package migrations
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func renameKafkaBillingModelColumn() *gormigrate.Migration {
```
machi1990, line 8: This migration could result in intermittent `500` internal server errors if an incoming requests is served by an old pod. 
This is because the old pod will still be using the old column which won't exist by the time this migration has been executed: This is because since we do an equivalent of `select *` in most of the code bases. 

We could do a two phases rollout or just simply be aware of the error when this is being deployed and an incoming requests to list `/api/kafkas/v1/kafkas..` is made.

miguelsorianod: We are aware of this but we decided to go ahead. The potential cases we can think is what you said about creating or listing kafka instances

ziccardi (approved):
LGTM

PR #1399, opened 2022-11-16
feat(connectors): fixes MGDCTRS-1635, support for loading connectors metadata from yaml config files

dhirajsb:
## Description
Load connectors metadata from flat yaml config files instead of pushing it upstream in connector catalog projects.

## Verification Steps
CI tests should pass.

- [x] Code Review completed

### DIFF
--- a/internal/connector/internal/config/connectors.go
+++ b/internal/connector/internal/config/connectors.go
@@ -11,8 +11,6 @@ import (
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/files"
 
-	gherrors "github.com/pkg/errors"
-
 	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/api/public"
@@ -72,18 +70,20 @@ func (c *ConnectorsConfig) ReadFiles() error {
 		return err
 	}
 
-	c.CatalogEntries, err = c.readConnectorCatalog(connectorMetadata)
+	// read catalogs and merge metadata, removing entries from the map
+	err = c.readConnectorCatalog(connectorMetadata)
 	if err != nil {
 		return err
 	}
 
+	// check if there are any unused metadata entries left
 	remainingIds := len(connectorMetadata)
 	if remainingIds > 0 {
 		ids := make([]string, 0, remainingIds)
 		for id := range connectorMetadata {
 			ids = append(ids, id)
 		}
-		return gherrors.Errorf("Found %d unrecognized connector metadata with ids: %s", remainingIds, ids)
+		return fmt.Errorf("Found %d unrecognized connector metadata with ids: %s", remainingIds, ids)
 	}
 
 	glog.Infof("loaded %d connector types", len(c.CatalogEntries))
@@ -111,7 +111,7 @@ func (c *ConnectorsConfig) readConnectorMetadata() (connectorMetadata map[string
 			var metas []ConnectorMetadata
 			err = shared.ReadYamlFile(path, &metas)
 			if err != nil {
-				return gherrors.Errorf("error reading connector metadata from %s: %s", path, err)
+				return fmt.Errorf("error reading connector metadata from %s: %s", path, err)
 			}
 			for _, m := range metas {
 				connectorMetadata[m.ConnectorTypeId] = m
@@ -121,14 +121,14 @@ func (c *ConnectorsConfig) readConnectorMetadata() (connectorMetadata map[string
 		})
 
 		if err != nil {
-			return nil, gherrors.Errorf("error listing connector metadata in %s: %s", metaDir, err)
+			return nil, fmt.Errorf("error listing connector metadata in %s: %s", metaDir, err)
 		}
 	}
 
 	return
 }
 
-func (c *ConnectorsConfig) readConnectorCatalog(connectorMetadata map[string]ConnectorMetadata) (values []ConnectorCatalogEntry, err error) {
+func (c *ConnectorsConfig) readConnectorCatalog(connectorMetadata map[string]ConnectorMetadata) (err error) {
 	typesLoaded := map[string]string{}
 	for _, dir := range c.ConnectorCatalogDirs {
 		dir = shared.BuildFullFilePath(dir)
@@ -148,13 +148,13 @@ func (c *ConnectorsConfig) readConnectorCatalog(connectorMetadata map[string]Con
 			// Read the file
 			buf, err := os.ReadFile(path)
 			if err != nil {
-				return gherrors.Errorf("error reading catalog file %s: %s", path, err)
+				return fmt.Errorf("error reading catalog file %s: %s", path, err)
 			}
 
 			entry := ConnectorCatalogEntry{}
 			err = json.Unmarshal(buf, &entry)
 			if err != nil {
-				return gherrors.Errorf("error unmarshaling catalog file %s: %s", path, err)
+				return fmt.Errorf("error unmarshaling catalog file %s: %s", path, err)
 			}
 
 			// set catalog metadata from metadata config read earlier
@@ -164,13 +164,13 @@ func (c *ConnectorsConfig) readConnectorCatalog(connectorMetadata map[string]Con
 				entry.ConnectorType.Labels = meta.Labels
 				// TODO annotations
 			} else {
-				return gherrors.Errorf("Missing metadata for connector %s", id)
+				return fmt.Errorf("Missing metadata for connector %s", id)
 			}
 
 			// compute checksum for catalog entry to look for updates
 			sum, err := checksum(entry)
 			if err != nil {
-				return gherrors.Errorf("error computing checksum for catalog file %s: %s", path, err)
+				return fmt.Errorf("error computing checksum for catalog file %s: %s", path, err)
 			}
 
 			// when walking directories with symlink such as what kubernetes does when mounting
@@ -208,7 +208,7 @@ func (c *ConnectorsConfig) readConnectorCatalog(connectorMetadata map[string]Con
 			c.CatalogChecksums[id] = sum
 			typesLoaded[id] = path
 
-			values = append(values, entry)
+			c.CatalogEntries = append(c.CatalogEntries, entry)
 
 			glog.Infof("loaded connector %s from file %s", id, path)
 
@@ -216,21 +216,20 @@ func (c *ConnectorsConfig) readConnectorCatalog(connectorMetadata map[string]Con
 		})
 
 		if err != nil {
-			err = gherrors.Errorf("error listing connector catalogs in %s: %s", dir, err)
-			return nil, err
+			return fmt.Errorf("error listing connector catalogs in %s: %s", dir, err)
 		}
 	}
 
 	// remove all processed metadata entries
-	for _, entry := range values {
+	for _, entry := range c.CatalogEntries {
 		delete(connectorMetadata, entry.ConnectorType.Id)
 	}
 
-	sort.Slice(values, func(i, j int) bool {
-		return values[i].ConnectorType.Id < values[j].ConnectorType.Id
+	sort.Slice(c.CatalogEntries, func(i, j int) bool {
+		return c.CatalogEntries[i].ConnectorType.Id < c.CatalogEntries[j].ConnectorType.Id
 	})
 
-	return values, nil
+	return nil
 }
 
 func checksum(spec interface{}) (string, error) {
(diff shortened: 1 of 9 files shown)


internal/connector/internal/config/connectors.go
```diff
@@ -79,22 +148,29 @@ func (c *ConnectorsConfig) ReadFiles() error {
...
-				err = gherrors.Errorf("error unmarshaling catalog file %s: %s", path, err)
-				return err
+				return gherrors.Errorf("error unmarshaling catalog file %s: %s", path, err)
+			}
+
+			// set catalog metadata from metadata config read earlier
+			id := entry.ConnectorType.Id
+			if meta, found := connectorMetadata[id]; found {
+				entry.ConnectorType.FeaturedRank = meta.FeaturedRank
+				entry.ConnectorType.Labels = meta.Labels
```
lburgazzoli, line 164: should we merge labels from connector and the metadata ? there are some connectors that [have labels already ](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleet-catalog-debezium/blob/270d14761e2a49c1e990f0f7a062f5a1efcc3c40/descriptors/mongodb/debezium-mongodb-1.9.6.Final.json#L10)

dhirajsb: Yeah, I considered that option. But, besides the code complexity in merging two lists (will it have to check for case/similar names?? to avoid duplicates) it would mean there are two sources of truth for labels. 
So I figured it's better to just make the catalog the source of non-metadata/core properties and make the metadata config file the exclusive source for the rest. 

It will also be easier to keep the labels consistent across in config files in one location rather than spreading them out over multiple pom files in the catalog. 

What do you think?

lburgazzoli: It is fine with me

we just need to have some coordination as I don't know if the current labels were meant to be used so in such case we need to be sure they are added to the yaml file.

dhirajsb: You're right. So I'm going to generate the first version of the metadata file from the current catalog. That way we keep what we have right now. Also, for categories and internationalization, UI is going to use labels with the prefix `category-`, e.g. `category-aws`. That way they can ignore any other labels when displaying their list of categories.

dhirajsb:
The intial config maps for connector metadata are in the draft [PR](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleet-manager/pull/204).

internal/connector/internal/config/connectors.go
```diff
@@ -43,6 +44,12 @@ type ConnectorCatalogEntry struct {
 	ConnectorType public.ConnectorType              `json:"connector_type"`
 }
 
+type ConnectorMetadata struct {
+	ConnectorTypeId string   `json:"id" yaml:"id"`
+	FeaturedRank    int32    `json:"featured-rank" yaml:"featured-rank"`
+	Labels          []string `json:"labels" yaml:"labels"`
```
lgarciaaco, line 50: There is no need to use StructTag for `Labels` since the default conversion (for both json and yaml packages) is to `labels`.

Also it is the common approach to use underscore instead of minus, any reason why minus for FeatureRank here makes more sense?

internal/connector/internal/config/connectors.go
```diff
@@ -51,20 +58,82 @@ func NewConnectorsConfig() *ConnectorsConfig {
...
+				return nil
+			}
+
+			glog.Infof("loading connector metadata from file %s", path)
 
+			// Read the file
+			var metas []ConnectorMetadata
+			err = shared.ReadYamlFile(path, &metas)
+			if err != nil {
+				return gherrors.Errorf("error reading connector metadata from %s: %s", path, err)
```
lgarciaaco, line 114: You should be using `%w` verb instead of `%s` so that the underlaying error is correctly wrapped.

internal/connector/internal/config/connectors.go
```diff
@@ -51,20 +58,82 @@ func NewConnectorsConfig() *ConnectorsConfig {
...
+			}
+			for _, m := range metas {
+				connectorMetadata[m.ConnectorTypeId] = m
+			}
+
+			return nil
+		})
+
+		if err != nil {
+			return nil, gherrors.Errorf("error listing connector metadata in %s: %s", metaDir, err)
```
lgarciaaco, line 124: You should be using `%w` verb instead of `%s` so that the underlaying error is correctly wrapped.

internal/connector/internal/config/connectors.go
```diff
@@ -79,22 +148,29 @@ func (c *ConnectorsConfig) ReadFiles() error {
 			// Read the file
 			buf, err := os.ReadFile(path)
 			if err != nil {
-				err = gherrors.Errorf("error reading catalog file %s: %s", path, err)
-				return err
+				return gherrors.Errorf("error reading catalog file %s: %s", path, err)
```
lgarciaaco, line 151: You should be using `%w` verb instead of `%s` so that the underlaying error is correctly wrapped.

internal/connector/internal/config/connectors.go
```diff
@@ -79,22 +148,29 @@ func (c *ConnectorsConfig) ReadFiles() error {
...
 			}
 
 			entry := ConnectorCatalogEntry{}
 			err = json.Unmarshal(buf, &entry)
 			if err != nil {
-				err = gherrors.Errorf("error unmarshaling catalog file %s: %s", path, err)
-				return err
+				return gherrors.Errorf("error unmarshaling catalog file %s: %s", path, err)
```
lgarciaaco, line 157: You should be using `%w` verb instead of `%s` so that the underlaying error is correctly wrapped.

internal/connector/internal/config/connectors.go
```diff
@@ -79,22 +148,29 @@ func (c *ConnectorsConfig) ReadFiles() error {
...
+			} else {
+				return gherrors.Errorf("Missing metadata for connector %s", id)
 			}
 
 			// compute checksum for catalog entry to look for updates
 			sum, err := checksum(entry)
 			if err != nil {
-				err = gherrors.Errorf("error computing checksum for catalog file %s: %s", path, err)
-				return err
+				return gherrors.Errorf("error computing checksum for catalog file %s: %s", path, err)
```
lgarciaaco, line 173: You should be using `%w` verb instead of `%s` so that the underlaying error is correctly wrapped.

internal/connector/internal/config/connectors.go
```diff
@@ -51,20 +58,82 @@ func NewConnectorsConfig() *ConnectorsConfig {
...
 func (c *ConnectorsConfig) ReadFiles() error {
-	typesLoaded := map[string]string{}
-	var values []ConnectorCatalogEntry
+	// read metadata first to merge with catalog next
+	connectorMetadata, err := c.readConnectorMetadata()
+	if err != nil {
+		return err
+	}
+
+	c.CatalogEntries, err = c.readConnectorCatalog(connectorMetadata)
```
lgarciaaco, line 75: It seems to that this assignment is the only reason why ConnectorsConfig is implementing this function by reference instead of value, by the look of it a value implementation will be enough and yield the same results if only we could get rid of this assignment.

Maybe we can return CatalogEntries instead of assigning it?

internal/connector/internal/config/connectors.go
```diff
@@ -119,40 +195,42 @@ func (c *ConnectorsConfig) ReadFiles() error {
...
 		if err != nil {
 			err = gherrors.Errorf("error listing connector catalogs in %s: %s", dir, err)
-			return err
+			return nil, err
 		}
 	}
 
+	// remove all processed metadata entries
+	for _, entry := range values {
+		delete(connectorMetadata, entry.ConnectorType.Id)
```
lgarciaaco, line 226: This is obscure. It would add clarity to return a new list with the remaining ids, rather than modifying the original parameter.

internal/connector/internal/config/connectors.go
```diff
@@ -51,20 +58,82 @@ func NewConnectorsConfig() *ConnectorsConfig {
...
+	if err != nil {
+		return err
+	}
+
+	c.CatalogEntries, err = c.readConnectorCatalog(connectorMetadata)
+	if err != nil {
+		return err
+	}
+
+	remainingIds := len(connectorMetadata)
```
lgarciaaco, line 80: The fact that `connectorMetadata` gets modified inside `readConnectorCatalog` doesn't help for clarity, see other comment a but further.

lgarciaaco (changes requested):

internal/connector/internal/config/connectors.go
dhirajsb: I am not sure I understand what you saying. But I'll optimize it to directly assign `values` to `c.CatalogEntries` in the function. I think that's what you are suggesting.

internal/connector/internal/config/connectors.go
dhirajsb: The parameter `connectorMetadata` is being merged with the catalog entries. Nothing obscure about it. Maybe you are thinking functional programming. But it's common to modify parameters, see `delete(map,key)`. 
A new list would be sub-optimal use of memory, since neither the original map nor the new list are useful beyond the calling function.

internal/connector/internal/config/connectors.go
dhirajsb: See previous comment.

dhirajsb:
> Also, this is not directly related with your work, but I wonder why we started using `gherrors "github.com/pkg/errors"`. See `https://github.com/pkg/errors/` is not the same as what go bundles (for 1.19: https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/errors/errors.go). I don't know why we are using a package that doesn't offer any benefits over the `src/pkg/errors` that comes with go, on the contrary because we are using `return gherrors.` we can't properly wrap the errors this way. A better alternative would be to return `fmt.Errorf("Some error %w", err)`.

The use of gherrors started before I joined the project. Anyway, it has some extra functionality, which will be absorbed in go 2.0 it seems. There is some stuff about recursively printing causes, etc. It also supports wrapping errors, but wrapping should only be done when you want to expose the underlying error and it's stacktrace etc. 

> 
> Also, `https://github.com/pkg/errors/` is an archived readonly project, it seems dead to me.

I wouldn't say the project is dead. It's just that no more development is needed for a basic module like this.

lgarciaaco:
> > Also, this is not directly related with your work, but I wonder why we started using `gherrors "github.com/pkg/errors"`. See `https://github.com/pkg/errors/` is not the same as what go bundles (for 1.19: https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/errors/errors.go). I don't know why we are using a package that doesn't offer any benefits over the `src/pkg/errors` that comes with go, on the contrary because we are using `return gherrors.` we can't properly wrap the errors this way. A better alternative would be to return `fmt.Errorf("Some error %w", err)`.
> 
> The use of gherrors started before I joined the project. Anyway, it has some extra functionality, which will be absorbed in go 2.0 it seems. There is some stuff about recursively printing causes, etc. It also supports wrapping errors, but wrapping should only be done when you want to expose the underlying error and it's stacktrace etc.
> 
> > Also, `https://github.com/pkg/errors/` is an archived readonly project, it seems dead to me.
> 
> I wouldn't say the project is dead. It's just that no more development is needed for a basic module like this.

It seems to me this project made some sense before go 1.13 when wrap wasn't available.

internal/connector/internal/config/connectors.go
lgarciaaco: No really. My worry is that I see `(c *ConnectorsConfig)` being the receiver instead of `(c ConnectorsConfig)`. I don't see a reason why a pointer should be the receiver instead of a variable and the only change required to change the receiver from reference to value would be to get rid of the line `c.CatalogEntries, err = c.readConnectorCatalog(connectorMetadata)` and returning the value instead.

internal/connector/internal/config/connectors.go
lgarciaaco: It is obscure. There is a syntactical difference from a built in function called `delete` that returns `nil`, and a function called `readConnectorCatalog` returning values. There is no indicator that your function is doing some magic and modifying one of the parameters, I didn't find any indicator and unless I read into the function I don't know what it was doing. If you want to keep it that way it is fine, but returning a new map would add in clarity.

internal/connector/internal/config/connectors.go
dhirajsb: Fixed.

internal/connector/internal/config/connectors.go
dhirajsb: I added a comment to explain what the function does with the map.

dhirajsb:
All PR comments have been resolved.

lgarciaaco (approved):

dhirajsb:
Thanks for the detailed review @lgarciaaco

PR #1414, opened 2022-11-22
add kafka billing model promotion endpoint

miguelsorianod:
## Description
Related to https://issues.redhat.com/browse/MGDSTRM-10007

As part of the functionality to support long lived evaluation instances we want to support being able to promote long lived standard evaluation instances to non evaluation standard instances with kafka billing model `standard` or `marketplace`.

This PR takes care of providing an endpoint to allow promotion of long lived standard evaluation instances.

This PR does not take care of implementing promotion itself. That will be done as part of another PR.

### Endpoint description

A new endpoint dedicated to promotion has been added. The endpoint URL path is:
`/api/kafkas_mgmt/v1//kafkas/{id}/promote`

where `{id}` is the id of the kafka instance to promote.

Only kafka instances with an `eval` kafka billing model are supported for promotion.

`POST` is the method used for the endpoint.

The endpoint is asynchronous: Promotion itself is performed asynchronously in reconcilers code.

#### Request description

The request of the endpoint accepts the following fields:
```
{
  desired_kafka_billing_model
  desired_marketplace
  desired_billing_cloud_account_id
}
```

* `desired_kafka_billing_model` is mandatory whereas `desired_marketplace` and `desired_billing_cloud_account_id` are optional
* When `desired_kafka_billing_model` is `marketplace` then `desired_marketplace` and `desired_billing_cloud_account_id` become mandatory. This is a difference from the kafka create endpoint where some implicit defaults are calculated. This difference in behavior intentional in order to simplify the behavior
* If `desired_kafka_billing_model` is not `marketplace` then `desired_marketplace` and `desired_billing_cloud_account_id` are ignored

The supported promotion path would be from:
* kafka instance type: `standard`
* kafka billing model: `eval`
* ams billing model: `standard`

to:
* kafka instance type: `standard`
* kafka billing model: `standard` or `marketplace`
* ams billing model: `standard` or any of the marketplace combinations supported (marketplace-rhm, marketplace-aws, ...)

All other combinations won't be supported.

#### Response description

No body is returned in the HTTP response in case of success.
HTTP status code is the element signaling the success/failure of the operation.
HTTP 202 Accepted is returned if the promotion request has been accepted. Promotion itself will be performed asynchronously on the reconcilers code.

### Other changes

* The Kafka GET endpoint has been updated to now return two new attributes: `promotion_status` and `promotion_details`
  * When they are not set or set to the empty string the attributes themselves are not returned
  * `promotion_status` returns the promotion status of the kafka request. The following values are possible:
    * `promoting`: The Kafka request is being promoted
    * `failed`: The Kafka request has failed to be promoted
    * The attribute is not returned: No kafka request is being promoted
  * `promotion_details`: Details of the promotion. Useful when promotion has failed

### Design decisions

* A decision has been taken of creating a new endpoint dedicated to perform promotion instead of reusing the already existing kafka update endpoint. The reasons for that are:
  * Gives a clear meaning to the endpoint vs just a generic "update kafka" endpoint name
  * Kafka update endpoints deals with multiple orthogonal sets of concerns:
    *  By adding another orthogonal concern like promotion we make the endpoint harder to understand for end users: The update endpoint performs promotion only if a certain set of fields on the request are performed. It performs ownership change if another set of fields are set, ...
    *  Increased implementation complexity: We would  need to deal with potential changes being done in parallel at the same time where order would be potentially an issue
    * Some changes might be synchronous and some asynchronous making it even more confusing for end users

* Two new columns `promotion_status` and `promotion_details have been added at DB level.  The reason for doing that instead of adding a new `promoting` status in the kafka request `status` DB column are:
  * If a promotion happens and we update the `status` value to `promoting` then we can't know to what status we should go back when promotion finishes or fails
  * An instance in ready status being promoted is still ready from the operational point of view, including reconciling operations in that ready status
* Due to the asynchronous nature of the promote endpoint the `promotion_details` has been added to be able to give feedback to the user about the promotion when some issue happens
* The Kafka GET endpoint now returns the `promotion_status` and `promotion_details` attributes. If they are empty the keys themselves are not even returned as part of the response. This is to be consistent with most of the other fields, like with failed_reason
* Promotion is performed asynchronously on the reconciler. and not during the request to the promote endpoint. The motivation for that is:
  * To improve reliability when some of the multiple steps needed to accomplish promotion fail. It does not fully solve all scenarios but we think it solves more than the synchronous option
  * The user does not need to retry the request again if promotion fails for some scenarios
* A promotion request is accepted in the following kafka request `status` values only: ready, suspended, resuming

TODO:

- [x] Add tests
- [x] Update OpenAPI file API version

## Verification Steps

* Tests pass
* Perform the following manual testing:
  1. Run KFM with with `--quota-type=ams` and configure an OCM cluster that ends up in ready state
  2. Create a standard kafka instance with `standard` billing_model. Wait until it ends up in ready state
  3. Login to the KFM DB and perform the following update in it to update the actual_billing_model and desired_kafka_billing_model to 'eval':  `update kafka_requests set actual_kafka_billing_model='eval', desired_kafka_billing_model='eval' where id='<kafka-id>'`;
  4. Call the promote endpoint with desired kafka billing model "standard"
  6. Perform a Kafka GET call to the KFM API and verify that:
    * promotion_status attribute has the value `promoting`
  7. In the DB, check that the kafka request entry contains the new specified desired_kafka_billing_model set to `standard` 
 
Repeat the previous steps but on point 3 set `marketplace` as desired with a valid desired_marketplace and desired_billing_cloud_account_id. In that case additionally verify that the marketplace and billing_cloud_account_id requested as part of the promote call end up being in the response of the kafka request get and in the DB.

Try other possible scenarios taking as the model the description on how the endpoint should work.

Repeat the tests but running KFM with `--quota-type=quota-management-list`.

- [ ] Code Review completed


miguelsorianod:
The currently submitted WIP implementation performs some changes but when working on it some  non trivial questions or observations related to the design of the functionality have arisen that I would like to discuss:
* The issue describes the promotion to be implemented as part of the kafka update endpoint. I am wondering whether implementing promotion in a different endpoint (for example `.../promote`) dedicated exclusively to that would be a better idea. The reason I am wondering about that is due to kafka update endpoint deals with multiple orthogonal sets of concerns. Maybe doing it in its own endpoint might be an alternative to consider. Using the kafka update endpoint for it makes:
  * The endpoint hard to understand for end users.
  * It also increases the implementation complexity of it, as now we have to deal with potential changes being done in parallel at the same time. It might be that applying different combinations of orthogonal changes in parallel would not be desirable
* What should be the format of the request? take into account that it is not only the desired kafka billing model but also potentially the marketplace and cloud account id. The answer to might also depend on the answer to the previous point
*  Is it supported to change the marketplace or billing cloud account id without changing the kafka billing model? This is, is this functionality exclusive to promote, or we want to support changing marketplaces / billing accounts in general without doing promotion?
*  Should we perform validation that there is quota available during request time or should we delegate it to the reconciler?
   * If we decide we have to perform validation at request time, how do we do it? right now there's only reserve method in quota service which does: validation + implicit setting + reservation and it also assumes to receive a kafkarequest object
   * If we decide we have to perform validation at request time there's a chance we have enough quota at that time but by the time the reservation is performed at reconciler time we don't have quota anymore. Do we just accept that? What do we do if that scenario occurs?
* What happens with the promotion logic when KFM is running in Quota Management List mode? It seems to me the concept of promotion only makes sense when AMS is being used, as it seems to be about billing.
* The issue mentions updating the status of the kafka instance to `promoting`. We now have the concepts of actual and desired billing models
   * Is that concept of this new `promoting` status still relevant? is just checking the difference between actual and desired enough and therefore having a new explicit status value is not needed?
   * If we decide the concept of an explicit `promoting` status is needed, do we set it at request time or we delegate it at reconcile time?
   * If we decide the concept of an explicit `promoting` status is needed, do we need to update other parts of the code that reference sets of status for some other functionalities?
* On promotion, should we check whether kafka is in some acceptable state? if so, which are considered as acceptable states for the promotion?

cc @ziccardi @JameelB .

We can do it through other channels to have more direct communication.

openapi/kas-fleet-manager.yaml
```diff
@@ -256,6 +256,81 @@ paths:
                   $ref: '#/components/examples/500Example'
     parameters:
       - $ref: "#/components/parameters/id"
+  /api/kafkas_mgmt/v1/kafkas/{id}/promote:
+    parameters:
+      - $ref: "#/components/parameters/id"
```
machi1990, line 261: @miguelsorianod @JameelB @ziccardi in some endpoints e.g https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/0c9948ab90b1e1eb45b032402b6a6b948f67ddf3/openapi/kas-fleet-manager.yaml#L338-L339 there is a usage of `async=true` to indicate that the operation is going to happen in async way. Is this something that can be used here as well?

openapi/kas-fleet-manager.yaml
```diff
@@ -256,6 +256,81 @@ paths:
                   $ref: '#/components/examples/500Example'
     parameters:
       - $ref: "#/components/parameters/id"
+  /api/kafkas_mgmt/v1/kafkas/{id}/promote:
```
machi1990, line 259: love the clarity that this new endpoint sends! Kudos @miguelsorianod @ziccardi @JameelB

miguelsorianod: I updated this PR description to give more details about this PR functionality and design decisions about it

openapi/kas-fleet-manager.yaml
miguelsorianod: I am aware of that and I also thought about it.
This endpoint is going to be asynchronous and there won't be a way to do it in a synchronous way. Is there any particular reason where to add the `async` query parameter? for consistency with the other endpoints? To make explicit the asynchronous nature of the endpoint? In any case, even with adding the parameter the endpoint documentation should document that the endpoint is asynchronous (as it is already done)

cc @ziccardi @JameelB for other views on this

miguelsorianod:
> The currently submitted WIP implementation performs some changes but when working on it some non trivial questions or observations related to the design of the functionality have arisen that I would like to discuss:
> 
>     * The issue describes the promotion to be implemented as part of the kafka update endpoint. I am wondering whether implementing promotion in a different endpoint (for example `.../promote`) dedicated exclusively to that would be a better idea. The reason I am wondering about that is due to kafka update endpoint deals with multiple orthogonal sets of concerns. Maybe doing it in its own endpoint might be an alternative to consider. Using the kafka update endpoint for it makes:
>       
>       * The endpoint hard to understand for end users.
>       * It also increases the implementation complexity of it, as now we have to deal with potential changes being done in parallel at the same time. It might be that applying different combinations of orthogonal changes in parallel would not be desirable
> 
>     * What should be the format of the request? take into account that it is not only the desired kafka billing model but also potentially the marketplace and cloud account id. The answer to might also depend on the answer to the previous point
> 
>     * Is it supported to change the marketplace or billing cloud account id without changing the kafka billing model? This is, is this functionality exclusive to promote, or we want to support changing marketplaces / billing accounts in general without doing promotion?
> 
>     * Should we perform validation that there is quota available during request time or should we delegate it to the reconciler?
>       
>       * If we decide we have to perform validation at request time, how do we do it? right now there's only reserve method in quota service which does: validation + implicit setting + reservation and it also assumes to receive a kafkarequest object
>       * If we decide we have to perform validation at request time there's a chance we have enough quota at that time but by the time the reservation is performed at reconciler time we don't have quota anymore. Do we just accept that? What do we do if that scenario occurs?
> 
>     * What happens with the promotion logic when KFM is running in Quota Management List mode? It seems to me the concept of promotion only makes sense when AMS is being used, as it seems to be about billing.
> 
>     * The issue mentions updating the status of the kafka instance to `promoting`. We now have the concepts of actual and desired billing models
>       
>       * Is that concept of this new `promoting` status still relevant? is just checking the difference between actual and desired enough and therefore having a new explicit status value is not needed?
>       * If we decide the concept of an explicit `promoting` status is needed, do we set it at request time or we delegate it at reconcile time?
>       * If we decide the concept of an explicit `promoting` status is needed, do we need to update other parts of the code that reference sets of status for some other functionalities?
> 
>     * On promotion, should we check whether kafka is in some acceptable state? if so, which are considered as acceptable states for the promotion?
> 
>     * Other comments related to design I left as TODO comments in this PR's code
> 
> 
> cc @ziccardi @JameelB .
> 
> We can do it through other channels to have more direct communication.

I updated the original description in this PR to reflect the new design decisions that have been taken

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -48,14 +49,44 @@ type KafkaRequest struct {
...
-	DesiredKafkaBillingModel string `json:"desired_kafka_billing_model"`
+	Namespace                string              `json:"namespace"`
+	ReauthenticationEnabled  bool                `json:"reauthentication_enabled"`
+	RoutesCreationId         string              `json:"routes_creation_id"`
+	SizeId                   string              `json:"size_id"`
+	BillingCloudAccountId    string              `json:"billing_cloud_account_id"`
+	Marketplace              string              `json:"marketplace"`
+	ActualKafkaBillingModel  string              `json:"actual_kafka_billing_model"`
+	DesiredKafkaBillingModel string              `json:"desired_kafka_billing_model"`
+	PromotionState           KafkaPromotionState `json:"promotion_state"`
```
miguelsorianod, line 60: Should we use the word `State` or `Status`? this also applies to the attributes in the kafka get response

openapi/kas-fleet-manager.yaml
machi1990: > Is there any particular reason where to add the async query parameter? for consistency with the other endpoints? To make explicit the asynchronous nature of the endpoint? 

+1 on these two points. I think all async endpoints have this, so it'll be consistency with other endpoints. And the way this query parameter was mainly used (from my understanding) is to make it explicit the async nature of the endpoint.

internal/kafka/internal/handlers/kafka_promote.go
```diff
@@ -0,0 +1,76 @@
...
+	validateKafkaFound := func() handlers.Validate {
+		return func() *errors.ServiceError {
+			return kafkaGetError
+		}
+	}
+	cfg := &handlers.HandlerConfig{
+		MarshalInto: &kafkaPromotionRequest,
+		Validate: []handlers.Validate{
+			handlers.ValidateMinLength(&kafkaPromotionRequest.DesiredKafkaBillingModel, "desired_kafka_billing_model", 1),
+			// ValidateKafkaUserFacingUpdateFields(ctx, h.authService, kafkaRequest, &kafkaUpdateReq), // TODO should some of this checks in here related to owner be performed?
```
miguelsorianod, line 48: Should we keep some of the behavior of this method for this endpoint?

internal/kafka/internal/handlers/kafka_promote.go
```diff
@@ -0,0 +1,76 @@
...
+			validateRequestedKafkaPromotionHasDifferentKafkaBillingModel(&kafkaPromotionRequest, kafkaRequest),
+			validateNoKafkaPromotionInProgress(kafkaRequest),
+			validateKafkaPromotionRequestFields(&kafkaPromotionRequest, kafkaRequest, h.service, h.kafkaConfig, h.kafkaPromotionValidatorFactory),
+		},
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+			kafkaFieldsToUpdate := map[string]interface{}{}
+			kafkaFieldsToUpdate["desired_kafka_billing_model"] = kafkaPromotionRequest.DesiredKafkaBillingModel
+			// We clear any potential promotion details being set due to a previous
+			// Kafka promotion failure
+			if kafkaRequest.PromotionDetails != "" {
```
miguelsorianod, line 61: We cleanup the `promotion_details` attribute in the DB to cleanup any potential previous message due to a previous failure

internal/kafka/internal/handlers/kafka_promote.go
```diff
@@ -0,0 +1,76 @@
...
+				kafkaFieldsToUpdate["promotion_details"] = ""
+			}
+			kafkaFieldsToUpdate["promotion_state"] = dbapi.KafkaPromotionStatePromoting
```
miguelsorianod, line 64: If a kafka request is previously in `failed` state we allow performing a new promotion to allow API users to perform a promotion again

internal/kafka/internal/handlers/kafka_promote.go
```diff
@@ -0,0 +1,76 @@
...
+			if kafkaPromotionRequest.DesiredMarketplace != "" {
```
miguelsorianod, line 65: Should we unset the fields instead if marketplace is not provided? That should never happen as promotion is only possible from eval where we know for sure we should never have a marketplace nor billing_cloud_account_id set beforehand as eval instances are standard kafka billing model

miguelsorianod: Looking for some other views about this too @ziccardi @JameelB

internal/kafka/internal/handlers/validation.go
```diff
@@ -370,3 +373,93 @@ func ValidateKafkaStorageSize(kafkaRequest *dbapi.KafkaRequest, kafkaUpdateReq *
 		return nil
 	}
 }
+
+func validateNoKafkaPromotionInProgress(kafkaRequest *dbapi.KafkaRequest) handlers.Validate {
+	return func() *errors.ServiceError {
+		// TODO is this check desirable? what happens if a promotion has previously failed
```
miguelsorianod, line 379: See this comment

openapi/kas-fleet-manager.yaml
ziccardi: Agree with @machi1990 about adding it for consistency.

miguelsorianod: added

internal/kafka/internal/api/dbapi/kafka_request_types.go
ziccardi: Maybe Status would be better

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1791,6 +1879,32 @@ components:
       required:
       - name
       type: object
+    KafkaPromotionRequest:
+      example:
+        desired_billing_cloud_account_id: desired_billing_cloud_account_id
+        desired_marketplace: desired_marketplace
+        desired_kafka_billing_model: desired_kafka_billing_model
+      properties:
```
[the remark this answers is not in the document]
ziccardi: I'd document it as it is done now: we list all the accepted values in the documentation.
IMO, the fact that by configuration (or by quota allowance) the user is not enabled to use some of them is a different topic

internal/kafka/internal/handlers/kafka_promote.go
ziccardi: Yes, I believe we should: when a promotion requests arrive, we should use only the received values, regardless of any promotion attempt history

internal/kafka/internal/handlers/validation.go
ziccardi: I believe this check should be removed: we just need to check the promotion status and allow the promotion if it is different than 'promoting' and 'promoted'

miguelsorianod: I agree.
For example, if a previous promotion has previously failed desired could be different than actual in the DB and no promotion being in progress at all, as it had failed.

I will remove this check and keep the check that checks the promoting state field

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -283,6 +283,89 @@ paths:
           description: Unexpected error occurred
       security:
       - Bearer: []
+  /api/kafkas_mgmt/v1/kafkas/{id}/promote:
```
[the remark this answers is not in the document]
ziccardi: This brings us back to a discussion we had a few days ago. We are not creating a resource with this endpoint: we are updating the `kafka` resource.

We should use the `update Kafka endpoint from a REST POV (IMO).
That, however, led to the issues you already exposed in the PR description.

On the other hand, since we don't have a `promotion` resource, I also believe that using a `promotion` endpoint is not the best option.

Following those reasonings, I believe that the `promote` endpoint is the best option.
I'd be interested to know what others think about it: @machi1990 @JameelB

internal/kafka/internal/handlers/validation.go
miguelsorianod: > I believe this check should be removed: we just need to check the promotion status and allow the promotion if it is different than 'promoting' and 'promoted'

Just a minor comment about this. There is no `promoted` value at all for promoting state field

internal/kafka/internal/handlers/kafka_promote.go
ziccardi: Hmmm... Let's keep this part: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/867982657339b85a34555e5bff3d5b34eaea7856/internal/kafka/internal/handlers/validation.go#L277-L289.

The only doubt I have is: shouldn't we do that in some auth middleware?

internal/kafka/internal/handlers/kafka_promote.go
miguelsorianod: changed

internal/kafka/internal/handlers/validation.go
miguelsorianod: Updated

internal/kafka/internal/handlers/kafka_promote.go
miguelsorianod: Regarding doing that in some auth middleware. It is a possibilty.

I suspect one of the possible reasons to not being done in a middleware is due to some endpoints not needing it. That means that the endpoints in our API at code level should be restructured at subrouter level so that only applies to certain endpoints.

For now I will create a validation that performs that part. We can look at a general restructuring of that for KFM in general.

miguelsorianod: I added a new commit with the changes.
The new commit:
1. Refactors the `ValidateKafkaUserFacingUpdateFields` by extracting the code that performs ownership check into its own new validator. The new validator is called `validateUserIsKafkaOwnerOrOrgAdmin`. `ValidateKafkaUserFacingUpdateFields` calls that other validator as part of its code
2. The kafka promote endpoint now calls `validateUserIsKafkaOwnerOrOrgAdmin` to ensure promotion is only allowed by the kafka owner or for an org admin of the kafka instance's organization

openapi/kas-fleet-manager.yaml
machi1990: Thanks @miguelsorianod @ziccardi

openapi/kas-fleet-manager.yaml
```diff
@@ -256,6 +256,87 @@ paths:
...
+        "404":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              examples:
+                404Example:
+                  $ref: '#/components/examples/404Example'
+          description: The requested resource doesn't exist
+        "409":
```
JameelB, line 320: Will this error ever occur for this endpoint?

internal/kafka/internal/api/public/api/openapi.yaml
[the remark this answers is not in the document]
JameelB: Agree with @ziccardi, I think it makes more sense to have `promote` for this endpoint. the cluster service of ocm does have some verbs in their api as well (i.e. /resume and /hibernate)

openapi/kas-fleet-manager.yaml
miguelsorianod: I think at the moment no. Originally I added it in case we might need it

internal/kafka/internal/api/dbapi/kafka_request_types.go
miguelsorianod: Consistency reasons seems a good argument that I already thought and I think makes sense.
I will update it to `PromotionStatus`. Also for the corresponding API attributes

internal/kafka/internal/api/public/api/openapi.yaml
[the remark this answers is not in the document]
miguelsorianod: I think going for `promote` also makes sense. Although not "pure" from a REST point of view I think it denotes better that it is an action more than a resource in this case and it also leaves space to become a resource in the future if needed by adding the endpoint name `promotion` and deprecating the `promote` one.

I will keep the `promote` name for the endpoint

internal/kafka/internal/api/public/api/openapi.yaml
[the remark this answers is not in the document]
miguelsorianod: :+1:

internal/kafka/internal/api/dbapi/kafka_request_types.go
miguelsorianod: Updated

miguelsorianod:
Updated public KFM OpenAPI minor version

miguelsorianod:
I updated the original description with instructions on how to perform manual verifications

miguelsorianod:
This PR is ready for a final review and verifications. See comments and the original description again to review

miguelsorianod:
Changed the base branch of the PR to be against a feature branch `longlived-instances-featurebranch` so we can built other functionality on top of this one before merging to main

ziccardi (approved):
It works as expected.
I tried with valid and invalid states.

Great job!

PR #1420, opened 2022-11-29
feat: add expires_at column to the kafka_requests table

JameelB:
## Description
Adds an `expires_at` column which contains the timestamp of when a Kafka instance is scheduled to expire. On expiration, Kafka instances will be marked for deletion by having its status set to 'deprovision'. 

`expires_at` should only be set for Kafka instances that have a limited lifespan (i.e. `developer` and `eval` instances). This field should not be set for `standard` instances as these should never expire. 

Related JIRA issue: [MGDSTRM-10010](https://issues.redhat.com/browse/MGDSTRM-10010)

## Verification Steps
Run `make db/migration`
  - [ ] Ensure the migration runs successfully (67 migration should've been applied)

Login to the database using `make db/login`
   - [ ] Ensure that the kafka_request table has the `expires_at` column with an empty value.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/migrations/20221129121515_add_expires_at_to_kafka_request.go
+++ b/internal/kafka/internal/migrations/20221129121515_add_expires_at_to_kafka_request.go
@@ -1,0 +1,24 @@
+package migrations
+
+import (
+	"time"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func addExpiresAtToKafkaRequest() *gormigrate.Migration {
+	type KafkaRequest struct {
+		ExpiresAt time.Time
+	}
+
+	return &gormigrate.Migration{
+		ID: "20221129121515",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&KafkaRequest{})
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return tx.Migrator().DropColumn(&KafkaRequest{}, "expires_at")
+		},
+	}
+}
--- a/internal/kafka/internal/api/dbapi/kafka_request_types.go
+++ b/internal/kafka/internal/api/dbapi/kafka_request_types.go
@@ -56,6 +56,9 @@ type KafkaRequest struct {
 	Marketplace              string `json:"marketplace"`
 	ActualKafkaBillingModel  string `json:"actual_kafka_billing_model"`
 	DesiredKafkaBillingModel string `json:"desired_kafka_billing_model"`
+	// ExpiresAt contains the timestamp of when a Kafka instance is scheduled to expire.
+	// On expiration, the Kafka instance will be marked for deletion, its status will be set to 'deprovision'.
+	ExpiresAt time.Time `json:"expires_at"`
 }
 
 type KafkaList []*KafkaRequest
--- a/internal/kafka/internal/migrations/migrations.go
+++ b/internal/kafka/internal/migrations/migrations.go
@@ -88,6 +88,7 @@ var migrations = []*gormigrate.Migration{
 	removeTheWronglyAutoCreatedClusterInStageEnvironmentWithID_cdhunvd8igjhbi0nmtt0(),
 	addDesiredKafkaBillingModel(),
 	renameKafkaBillingModelColumn(),
+	addExpiresAtToKafkaRequest(),
 }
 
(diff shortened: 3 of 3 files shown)


internal/kafka/internal/migrations/20221129121515_add_expires_at_to_kafka_request.go
```diff
@@ -0,0 +1,24 @@
...
+
+func addExpiresAtToKafkaRequest() *gormigrate.Migration {
+	type KafkaRequest struct {
+		ExpiresAt time.Time `json:"expires_at"`
+	}
+
+	return &gormigrate.Migration{
+		ID: "20221129121515",
+		Migrate: func(tx *gorm.DB) error {
+			return tx.AutoMigrate(&KafkaRequest{})
```
JameelB, line 18: The lifespanSeconds that this value will be based on for developer instances is in the supported_instance_types config and values currently differ in stage and production.  Didn't include it as part of this migration as this will be set by one of the Kafka reconcilers anyway ([MGDSTRM-10201](https://issues.redhat.com/browse/MGDSTRM-10201)). Does this sound ok to you @miguelsorianod @ziccardi?

miguelsorianod: It will be set by one of the reconcilers for **new** developer instances but not for the already existing ones

JameelB: I would've thought the reconciler should handle setting that field as long as its not set for developer instances :thinking:  My understanding is that even if the lifespanSeconds or grace period was changed, KFM should not update Kafka instances if the `expires_at` was already set.

I've just realized that we return the expires_at field with a value for developers in the response from Kafka create. 
image

If we want to keep the same behaviour, we might want to also set this on Kafka creation for developer instances and a reconciler would probably not be needed (can close issue [MGDSTRM-10201](https://issues.redhat.com/browse/MGDSTRM-10201))

 In this case, setting this for existing kafka instances in this migration would make sense.

miguelsorianod: > My understanding is that even if the lifespanSeconds or grace period was changed, KFM should not update Kafka instances if the expires_at was already set

That's my understanding too

machi1990: >If we want to keep the same behaviour, we might want to also set this on Kafka creation for developer instances and a reconciler would probably not be needed (can close issue [MGDSTRM-10201](https://issues.redhat.com/browse/MGDSTRM-10201))
> In this case, setting this for existing kafka instances in this migration would make sense if possible.

It is possible that a developer kafka could be created just after the migration has rolled out and just before all the new pods (with the new logic) are rolled out. And this developer Kafka might not have its `expires_at` column set by the migration. If we do not have a reconciler logic to handle for this case, then we could be in an inconsitent state for this kafka.

JameelB: @machi1990 that's a good point. I think we may need to do this migration in multiple steps. 

1. Add the `expires_at` column to the kafka_requests table
    - Field should be set empty (should be ok since it's not used at the moment)
2. Add logic to set the `expires_at` value for all instances with a limited lifespan (currently developer and long lived instances)
    - It should only set it if expires_at is empty.
    - Done as part of [MGDSTRM-10201](https://issues.redhat.com/browse/MGDSTRM-10201)
3. Refactor logic to use this `expires_at` value. Following steps can be done simultaneously.
    - Refactor reconcile logic for deleting expired instances
    - Refactor KafkaRequest presenter so that it takes the `expires_at` value from the db instead of dynamically calculating it.
    - Refactor Kafka creation so that it sets the `expires_at` value on create if lifespanSeconds was defined for that instance type.
       - This keeps the existing behaviour of returning the `expires_at` with a value for developer instances on create. If we don't do this here, `expires_at` value will be empty for a few moments until the reconciler sets it.  
4. Remove reconcile logic to set `expires_at` value for developer instances if no longer needed.

wdyt? does this make sense? :thinking:

machi1990: The steps above should be able to handle this migration perfectly. 
Thanks for describing the proposal and welcome back :-)

ziccardi (approved):

PR #1431, opened 2022-12-06
fix(connectors): fixes MGDCTRS-1755, added migration to add annotations for existing clusters and connectors

dhirajsb:
## Description
Added migration to add annotations for existing clusters and connectors

## Verification Steps
CI including migration tests should pass.

- [x] Code Review completed

### DIFF
--- a/internal/connector/internal/migrations/202212050000_add_orgid_annotations.go
+++ b/internal/connector/internal/migrations/202212050000_add_orgid_annotations.go
@@ -1,0 +1,29 @@
+package migrations
+
+// Migrations should NEVER use types from other packages. Types can change
+// and then migrations run on a _new_ database will fail or behave unexpectedly.
+// Instead of importing types, always re-create the type in the migration, as
+// is done here, even though the same type is defined in pkg/api
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/go-gormigrate/gormigrate/v2"
+)
+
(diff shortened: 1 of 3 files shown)


dhirajsb:
@valdar this PR adds missing annotations for existing clusters and annotations.

internal/connector/internal/migrations/202212050000_add_orgid_annotations.go
```diff
@@ -0,0 +1,29 @@
...
+// is done here, even though the same type is defined in pkg/api
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/go-gormigrate/gormigrate/v2"
+)
+
+func addOrgIDAnnotations(migrationID string) *gormigrate.Migration {
+
+	return db.CreateMigrationFromActions(migrationID,
```
valdar, line 15: Is it intended that there is no rollback to these actions?

valdar (changes requested):

internal/connector/internal/migrations/202212050000_add_orgid_annotations.go
dhirajsb: Correct, but we missed that and now it's in dev already. So the solution is to create a new migration to just add the missing data. 

Not a huge problem, but we'll have to keep that in mind, that when new tables are added, we also need some SQL to create records for old resources.

valdar (approved):

PR #1436, opened 2022-12-09
Add reconciler for kafkas promotions

ziccardi:
## Description

This PR adds the promotion reconcile.
A kafka is chosen for promotion when all of the following conditions are verified:

- `ActualBillingModel` is not empty
- `DesiredBillingModel` is different than `ActualBillingModel`

- [ ] Code Review completed

### DIFF
--- a/config/quota-management-list-configuration.yaml
+++ b/config/quota-management-list-configuration.yaml
@@ -91,6 +91,8 @@ registered_users_per_organisation:
         kafka_billing_models:
           - id: standard
             max_allowed_instances: 5
+          - id: marketplace
+            max_allowed_instances: 5
     # The App SRE team
   - id: 12147054
     any_user: true
--- a/internal/kafka/internal/migrations/20221219154830_add_kafka_promote_worker_in_leader_leases.go
+++ b/internal/kafka/internal/migrations/20221219154830_add_kafka_promote_worker_in_leader_leases.go
@@ -1,0 +1,25 @@
+package migrations
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func addKafkaPromoteWorkerInLeaderLeases() *gormigrate.Migration {
+	leaderLeaseType := "promoting_kafka"
+	return &gormigrate.Migration{
+		ID: "20221219154830",
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.Create(&api.LeaderLease{Expires: &db.KafkaAdditionalLeasesExpireTime, LeaseType: leaderLeaseType, Leader: api.NewID()}).Error; err != nil {
+				return err
+			}
+
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
+			return tx.Unscoped().Where("lease_type = ?", leaderLeaseType).Delete(&api.LeaderLease{}).Error
+		},
+	}
+}
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -88,6 +88,7 @@ type KafkaService interface {
 	List(ctx context.Context, listArgs *services.ListArguments) (dbapi.KafkaList, *api.PagingMeta, *errors.ServiceError)
 	// Lists all kafkas. As this returns all Kafka requests without need for authentication, this should only be used for internal purposes
 	ListAll() (dbapi.KafkaList, *errors.ServiceError)
+	ListKafkasToBePromoted() ([]*dbapi.KafkaRequest, *errors.ServiceError)
 	GetManagedKafkaByClusterID(clusterID string) ([]managedkafka.ManagedKafka, *errors.ServiceError)
 	// GenerateReservedManagedKafkasByClusterID returns a list of reserved managed
 	// kafkas for a given clusterID. The number of generated reserved managed
@@ -518,6 +519,28 @@ func (k *kafkaService) ListByStatus(status ...constants.KafkaStatus) ([]*dbapi.K
 
 	if err := dbConn.Model(&dbapi.KafkaRequest{}).Where("status IN (?)", status).Scan(&kafkas).Error; err != nil {
 		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to list by status")
+	}
+
+	return kafkas, nil
+}
+
+func (k *kafkaService) ListKafkasToBePromoted() ([]*dbapi.KafkaRequest, *errors.ServiceError) {
+	dbConn := k.connectionFactory.New()
+
+	validPromotionStatuses := []string{
+		constants.KafkaRequestStatusReady.String(),
+		constants.KafkaRequestStatusSuspended.String(),
+		constants.KafkaRequestStatusResuming.String(),
+	}
+
+	var kafkas []*dbapi.KafkaRequest
+
+	if err := dbConn.Model(&dbapi.KafkaRequest{}).
+		Where("actual_kafka_billing_model <> desired_kafka_billing_model").
+		Where("promotion_status <> ?", dbapi.KafkaPromotionStatusFailed).
+		Where("status in ?", validPromotionStatuses).
+		Scan(&kafkas).Error; err != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to list kafkas to be promoted")
 	}
 
 	return kafkas, nil
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -268,13 +268,21 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
 	// will be empty if no marketplace account is used
 	rr.BillingMarketplaceAccount(kafka.BillingCloudAccountId)
 
+	// TODO: temporary workaround. This is to be removed when this issue is implemented: https://issues.redhat.com/browse/SDB-3157
+	getClusterID := func(kafka *dbapi.KafkaRequest) string {
+		if kafka.DesiredKafkaBillingModel == "eval" {
+			return kafkaId + "-eval"
+		}
+		return kafkaId
+	}
+
 	cb, _ := amsv1.NewClusterAuthorizationRequest().
 		AccountUsername(kafka.Owner).
 		CloudProviderID(kafka.CloudProvider).
 		ProductID(kafkaBillingModel.AMSProduct).
 		Managed(true).
-		ClusterID(kafkaId).
-		ExternalClusterID(kafkaId).
+		ClusterID(getClusterID(kafka)).
+		ExternalClusterID(getClusterID(kafka)).
 		Disconnected(false).
 		BYOC(false).
 		AvailabilityZone(q.getAMSClusterAuthorizationRequestAvailabilityZone(kafka.MultiAZ)).
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -2,6 +2,8 @@ package quota
 
 import (
 	"fmt"
+	v1 "github.com/openshift-online/ocm-sdk-go/authorizations/v1"
+	"net/http"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/cloudproviders"
@@ -295,14 +297,94 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
 	return resp.Subscription().ID(), nil
 }
 
-func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError {
-	if subscriptionId == "" {
+func (q amsQuotaService) ReserveQuotaIfNotAlreadyReserved(kafka *dbapi.KafkaRequest) (string, *errors.ServiceError) {
+	instanceType, err := q.kafkaConfig.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(kafka.InstanceType)
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+	kafkaBillingModel, err := instanceType.GetKafkaSupportedBillingModelByID(kafka.DesiredKafkaBillingModel)
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+
+	subscriptions, err := q.amsClient.FindSubscriptions(fmt.Sprintf("cluster_id='%s'", kafka.ClusterID))
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+	for _, subscription := range subscriptions {
+		// get the reserved resources
+		if subscription.Plan().ID() == kafkaBillingModel.AMSProduct && subscription.Status() == string(v1.SubscriptionStatusActive) {
+			reservedResources, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscription.ID())
+			if err != nil {
+				return "", errors.GeneralError("failed checking current quota: %v", err)
+			}
+
+			if arrays.AnyMatch(reservedResources, func(rr *amsv1.ReservedResource) bool { return rr.ResourceName() == kafkaBillingModel.AMSResource }) {
+				return subscription.ID(), nil
+			}
+		}
+	}
+
+	// quota is not defined. We need to reserve a new one
+	return q.ReserveQuota(kafka)
+}
+
+func (q amsQuotaService) DeleteQuota(subscriptionID string) *errors.ServiceError {
+	if subscriptionID == "" {
 		return nil
 	}
 
-	_, err := q.amsClient.DeleteSubscription(subscriptionId)
+	_, err := q.amsClient.DeleteSubscription(subscriptionID)
 	if err != nil {
 		return errors.GeneralError("failed to delete the quota: %v", err)
 	}
 	return nil
+}
+
+// DeleteQuotaForBillingModel deletes the quota identified by `subscriptionID` only if it is related to received config.KafkaBillingModel
+func (q amsQuotaService) DeleteQuotaForBillingModel(subscriptionID string, kafkaBillingModel config.KafkaBillingModel) *errors.ServiceError {
+	if subscriptionID == "" {
+		return nil
+	}
+
+	subscription, found, err := q.GetSubscriptionByID(subscriptionID)
+	if err != nil {
+		return errors.GeneralError("failed to delete the quota: %v", err)
+	}
+
+	if !found {
+		return nil
+	}
+
+	if subscription.Plan().ID() == kafkaBillingModel.AMSProduct {
+		rr, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscriptionID)
+		if err != nil {
+			return errors.GeneralError("failed to delete the quota: %v", err)
+		}
+		resourceMatches := arrays.AnyMatch(rr, func(resource *amsv1.ReservedResource) bool {
+			return resource.ResourceName() == kafkaBillingModel.AMSResource
+		})
+		if resourceMatches {
+			// this subscription is related to this billing model. We can delete it.
+			return q.DeleteQuota(subscriptionID)
+		}
+	}
+
+	// the specified subscription id is not related to the specified billing model
+	return nil
+}
+
+// GetSubscriptionByID gets a subscription by ID
+// return an error if an error happens calling the endpoint, otherwise returns whether it found the subscription(subscription, true, nil)
+// or not (subscription, false, nil)
+func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subscription, bool, *errors.ServiceError) {
+	resp, err := q.amsClient.GetSubscriptionByID(subscriptionID)
+	if resp != nil && resp.Status() == http.StatusNotFound {
+		return resp.Body(), false, nil
+	}
+	if err != nil {
+		return nil, false, errors.NewWithCause(errors.ErrorGeneral, err, "error getting the subscriptions for id '%s'", subscriptionID)
+	}
+
+	return resp.Body(), true, nil
 }
--- a/internal/kafka/internal/services/quota/ams_quota_service_test.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service_test.go
@@ -2,6 +2,7 @@ package quota
 
 import (
 	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services/quota/internal/mocks/ams"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services/quota/internal/test"
 	"testing"
 
@@ -981,8 +982,8 @@ func Test_AMSCheckQuota(t *testing.T) {
 			sq, err := quotaService.CheckIfQuotaIsDefinedForInstanceType(kafka.Owner, kafka.OrganisationId, types.STANDARD, bm)
 			g.Expect(err).ToNot(gomega.HaveOccurred())
 			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()).To(gomega.HaveLen(1))
-			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[0].ResourceName).To(gomega.Equal("rhosak"))
-			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[0].Product).To(gomega.Equal("RHOSAK"))
+			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[0].ResourceName).To(gomega.Equal(ocm.RHOSAKResourceName))
+			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[0].Product).To(gomega.BeEquivalentTo(ocm.RHOSAKProduct))
 			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[0].Product).To(gomega.Equal(bm.AMSProduct))
 			g.Expect(sq).To(gomega.Equal(tt.args.hasStandardQuota))
 			bm, err1 = tt.fields.kafkaConfig.GetBillingModelByID(types.DEVELOPER.String(), "standard")
@@ -992,14 +993,203 @@ func Test_AMSCheckQuota(t *testing.T) {
 			fmt.Printf("eq is %v\n", eq)
 			g.Expect(eq).To(gomega.Equal(tt.args.hasDeveloperQuota))
 			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()).To(gomega.HaveLen(2))
-			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[1].ResourceName).To(gomega.Equal("rhosak"))
-			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[1].Product).To(gomega.Equal("RHOSAKTrial"))
+			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[1].ResourceName).To(gomega.Equal(ocm.RHOSAKResourceName))
+			g.Expect(tt.fields.ocmClient.GetQuotaCostsForProductCalls()[1].Product).To(gomega.BeEquivalentTo(ocm.RHOSAKTrialProduct))
 			_, err = quotaService.ReserveQuota(kafka)
 			g.Expect(err != nil).To(gomega.Equal(tt.wantErr))
 		})
 	}
 }
 
+func Test_AMSReserveQuotaIfNotAlreadyReserved(t *testing.T) {
+	var amsDefaultKafkaConf = config.KafkaConfig{
+		Quota:                  config.NewKafkaQuotaConfig(),
+		SupportedInstanceTypes: test.NewAMSTestKafkaSupportedInstanceTypesConfig(),
+	}
+	type fields struct {
+		ocmClient   *ocm.ClientMock
+		kafkaConfig config.KafkaConfig
+	}
+	type args struct {
+		kafkaID                           string
+		kafkaInstanceType                 types.KafkaInstanceType
+		kafkaRequestMarketplace           string
+		kafkaRequestDesiredBillingModel   string
+		kafkaRequestBillingCloudAccountID string
+		kafkaRequestCloudProvider         string
+
+		wantsReserveCalled bool
+		wantsProductID     string
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+	}{
+		{
+			name: "Reserve quota for new cluster",
+			fields: fields{
+				kafkaConfig: amsDefaultKafkaConf,
+				ocmClient: &ocm.ClientMock{
+					FindSubscriptionsFunc: func(query string) ([]*v1.Subscription, error) {
+						return ams.FindSubscriptions(query), nil
+					},
+					GetOrganisationIdFromExternalIdFunc: func(externalId string) (string, error) {
+						return fmt.Sprintf("fake-org-id-%s", externalId), nil
+					},
+					ClusterAuthorizationFunc: func(cb *v1.ClusterAuthorizationRequest) (*v1.ClusterAuthorizationResponse, error) {
+						sub := v1.SubscriptionBuilder{}
+						sub.ID(cb.ClusterID() + "subscription")
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "1234",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "eval",
+
+				wantsReserveCalled: true,
+				wantsProductID:     string(ocm.RHOSAKEvalProduct),
+			},
+		},
+		{
+			name: "Reserve quota for existing cluster and existing subscription",
+			fields: fields{
+				kafkaConfig: amsDefaultKafkaConf,
+				ocmClient: &ocm.ClientMock{
+					FindSubscriptionsFunc: func(query string) ([]*v1.Subscription, error) {
+						return ams.FindSubscriptions(query), nil
+					},
+					GetReservedResourcesBySubscriptionIDFunc: func(subscriptionID string) ([]*v1.ReservedResource, error) {
+						return ams.GetReservedResourcesBySubscriptionID(subscriptionID), nil
+					},
+					GetOrganisationIdFromExternalIdFunc: func(externalId string) (string, error) {
+						return fmt.Sprintf("fake-org-id-%s", externalId), nil
+					},
+					ClusterAuthorizationFunc: func(cb *v1.ClusterAuthorizationRequest) (*v1.ClusterAuthorizationResponse, error) {
+						sub := v1.SubscriptionBuilder{}
+						sub.ID(cb.ClusterID() + "subscription")
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-1",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "standard",
+
+				wantsReserveCalled: false,
+				wantsProductID:     string(ocm.RHOSAKProduct),
+			},
+		},
+		{
+			// In this test we try to reserve a standard for a cluster that has `eval` assigned
+			name: "Reserve standard quota for existing cluster with `eval` subscription",
+			fields: fields{
+				kafkaConfig: amsDefaultKafkaConf,
+				ocmClient: &ocm.ClientMock{
+					FindSubscriptionsFunc: func(query string) ([]*v1.Subscription, error) {
+						return ams.FindSubscriptions(query), nil
+					},
+					GetReservedResourcesBySubscriptionIDFunc: func(subscriptionID string) ([]*v1.ReservedResource, error) {
+						return ams.GetReservedResourcesBySubscriptionID(subscriptionID), nil
+					},
+					GetOrganisationIdFromExternalIdFunc: func(externalId string) (string, error) {
+						return fmt.Sprintf("fake-org-id-%s", externalId), nil
+					},
+					ClusterAuthorizationFunc: func(cb *v1.ClusterAuthorizationRequest) (*v1.ClusterAuthorizationResponse, error) {
+						sub := v1.SubscriptionBuilder{}
+						sub.ID(cb.ClusterID() + "subscription")
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-7",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "standard",
+
+				wantsReserveCalled: true,
+				wantsProductID:     string(ocm.RHOSAKProduct),
+			},
+		},
+		{
+			// In this test we try to reserve an `eval` for a cluster that has `eval` assigned
+			name: "Reserve `eval` quota for existing cluster with `eval` subscription",
+			fields: fields{
+				kafkaConfig: amsDefaultKafkaConf,
+				ocmClient: &ocm.ClientMock{
+					FindSubscriptionsFunc: func(query string) ([]*v1.Subscription, error) {
+						return ams.FindSubscriptions(query), nil
+					},
+					GetReservedResourcesBySubscriptionIDFunc: func(subscriptionID string) ([]*v1.ReservedResource, error) {
+						return ams.GetReservedResourcesBySubscriptionID(subscriptionID), nil
+					},
+					GetOrganisationIdFromExternalIdFunc: func(externalId string) (string, error) {
+						return fmt.Sprintf("fake-org-id-%s", externalId), nil
+					},
+					ClusterAuthorizationFunc: func(cb *v1.ClusterAuthorizationRequest) (*v1.ClusterAuthorizationResponse, error) {
+						sub := v1.SubscriptionBuilder{}
+						sub.ID(cb.ClusterID() + "subscription")
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-7",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "eval",
+
+				wantsReserveCalled: false,
+				wantsProductID:     string(ocm.RHOSAKProduct),
+			},
+		},
+	}
+
+	for _, testcase := range tests {
+		tt := testcase
+
+		kafka := &dbapi.KafkaRequest{
+			Meta: api.Meta{
+				ID: tt.args.kafkaID,
+			},
+			ClusterID:                tt.args.kafkaID,
+			CloudProvider:            tt.args.kafkaRequestCloudProvider,
+			Owner:                    "owner",
+			SizeId:                   "x1",
+			InstanceType:             string(tt.args.kafkaInstanceType),
+			Marketplace:              tt.args.kafkaRequestMarketplace,
+			BillingCloudAccountId:    tt.args.kafkaRequestBillingCloudAccountID,
+			DesiredKafkaBillingModel: tt.args.kafkaRequestDesiredBillingModel,
+		}
+
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			factory := NewDefaultQuotaServiceFactory(tt.fields.ocmClient, nil, nil, &tt.fields.kafkaConfig)
+			quotaService, _ := factory.GetQuotaService(api.AMSQuotaType)
+
+			_, err := quotaService.ReserveQuotaIfNotAlreadyReserved(kafka)
+			g.Expect(err).ToNot(gomega.HaveOccurred(), "Expecting error not to happen:", err)
+
+			if tt.args.wantsReserveCalled {
+				g.Expect(tt.fields.ocmClient.ClusterAuthorizationCalls()).To(gomega.HaveLen(1))
+				g.Expect(tt.fields.ocmClient.ClusterAuthorizationCalls()[0].Cb.ProductID()).To(gomega.Equal(tt.args.wantsProductID))
+			} else {
+				g.Expect(tt.fields.ocmClient.ClusterAuthorizationCalls()).To(gomega.HaveLen(0), "Should not reserve a new quota")
+			}
+		})
+	}
+
+}
 func Test_AMSReserveQuota(t *testing.T) {
 	var amsDefaultKafkaConf = config.KafkaConfig{
 		Quota:                  config.NewKafkaQuotaConfig(),
@@ -1933,7 +2123,7 @@ func Test_amsQuotaService_newBaseQuotaReservedResourceBuilder(t *testing.T) {
 			want := tt.wantFactory()
 			kafkaBillingModel := config.KafkaBillingModel{
 				ID:          "standard",
-				AMSResource: "rhosak",
+				AMSResource: ocm.RHOSAKResourceName,
 			}
 			res := amsQuotaService.newBaseQuotaReservedResourceBuilder(tt.args.kafkaRequest, kafkaBillingModel)
 			g.Expect(res).To(gomega.Equal(want))
--- a/internal/kafka/internal/services/quota/internal/mocks/ams/subscriptions_mock.go
+++ b/internal/kafka/internal/services/quota/internal/mocks/ams/subscriptions_mock.go
@@ -1,0 +1,47 @@
+package ams
+
+import (
+	v1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+	v12 "github.com/openshift-online/ocm-sdk-go/authorizations/v1"
+	"regexp"
+)
+
+func newSubscription(subscriptionID string, clusterID string, planID string) *v1.Subscription {
+	subscription, _ := v1.NewSubscription().
+		ID(subscriptionID).
+		ClusterID(clusterID).
+		Status(string(v12.SubscriptionStatusActive)).
+		Plan(v1.NewPlan().ID(planID).Name(planID)).Build()
+	return subscription
+}
+
+var subscriptions = map[string]*v1.Subscription{
+	"subscription-1":  newSubscription("subscription-1", "cluster-1", "RHOSAK"),
+	"subscription-2":  newSubscription("subscription-2", "cluster-2", "RHOSAK"),
+	"subscription-3":  newSubscription("subscription-3", "cluster-3", "RHOSAKTrial"),
+	"subscription-4":  newSubscription("subscription-4", "cluster-4", "RHOSAK"),
+	"subscription-5":  newSubscription("subscription-5", "cluster-5", "RHOSAK"),
+	"subscription-6":  newSubscription("subscription-6", "cluster-6", "RHOSAKTrial"),
+	"subscription-7":  newSubscription("subscription-7", "cluster-7", "RHOSAKEval"),
+	"subscription-8":  newSubscription("subscription-8", "cluster-8", "RHOSAKEval"),
+	"subscription-9":  newSubscription("subscription-9", "cluster-9", "RHOSAKEval"),
+	"subscription-10": newSubscription("subscription-10", "cluster-10", "RHOSAK"),
+}
+
+func FindSubscriptions(query string) []*v1.Subscription {
+	r := regexp.MustCompile("^cluster_id='(?P<clusterid>.*)'$")
+
+	switch true {
+	case r.MatchString(query):
+		clusterID := r.FindStringSubmatch(query)[1]
+
+		for _, v := range subscriptions {
+			if v.ClusterID() == clusterID {
+				return []*v1.Subscription{v}
+			}
+		}
+	default:
+		panic("Query non supported by test:" + query)
+	}
+	return nil
+}
--- a/internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
+++ b/internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
@@ -59,22 +59,6 @@ func (r *simpleBillingModelResolver) resolve(orgID string, kafka *dbapi.KafkaReq
 		return BillingModelDetails{}, errors.NewWithCause(errors.ErrorGeneral, err, "Error reserving quota")
 	}
 
-	return r.resolve(orgId, kafka, kafkaBillingModels)
-}
-
-// resolve - This method tries to resolve the billing model and, eventually, the marketplace type for the received
-// kafka request among the received kafkaBillingModels. The billing models in the `kafkaBillingModel` array are evaluated
-// in the received order, so that an order of preference can be honored.
-// Parameters:
-// orgId: The organisation ID
-// kafka: The kafka request received by the user
-// kafkaBillingModels: The list of kafka billing models to be considered, ordered by preference
-func (r *simpleBillingModelResolver) resolve(orgId string, kafka *dbapi.KafkaRequest, kafkaBillingModels []config.KafkaBillingModel) (BillingModelDetails, error) {
-	kafkaInstanceSize, err := r.kafkaConfig.GetKafkaInstanceSize(kafka.InstanceType, kafka.SizeId)
-	if err != nil {
-		return BillingModelDetails{}, errors.NewWithCause(errors.ErrorGeneral, err, "Error reserving quota")
-	}
-
 	for _, kbm := range kafkaBillingModels {
 		available, amsBillingModel, err := r.isBillingModelAvailable(orgID, kbm, kafkaInstanceSize.CapacityConsumed)
 		if err != nil {
--- a/internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
+++ b/internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
@@ -59,6 +59,22 @@ func (r *simpleBillingModelResolver) resolve(orgID string, kafka *dbapi.KafkaReq
 		return BillingModelDetails{}, errors.NewWithCause(errors.ErrorGeneral, err, "Error reserving quota")
 	}
 
+	return r.resolve(orgId, kafka, kafkaBillingModels)
+}
+
+// resolve - This method tries to resolve the billing model and, eventually, the marketplace type for the received
+// kafka request among the received kafkaBillingModels. The billing models in the `kafkaBillingModel` array are evaluated
+// in the received order, so that an order of preference can be honored.
+// Parameters:
+// orgId: The organisation ID
+// kafka: The kafka request received by the user
+// kafkaBillingModels: The list of kafka billing models to be considered, ordered by preference
+func (r *simpleBillingModelResolver) resolve(orgId string, kafka *dbapi.KafkaRequest, kafkaBillingModels []config.KafkaBillingModel) (BillingModelDetails, error) {
+	kafkaInstanceSize, err := r.kafkaConfig.GetKafkaInstanceSize(kafka.InstanceType, kafka.SizeId)
+	if err != nil {
+		return BillingModelDetails{}, errors.NewWithCause(errors.ErrorGeneral, err, "Error reserving quota")
+	}
+
 	for _, kbm := range kafkaBillingModels {
 		available, amsBillingModel, err := r.isBillingModelAvailable(orgID, kbm, kafkaInstanceSize.CapacityConsumed)
 		if err != nil {
--- a/internal/kafka/internal/services/quota/quota_management_list_service.go
+++ b/internal/kafka/internal/services/quota/quota_management_list_service.go
@@ -25,6 +25,14 @@ type QuotaManagementListService struct {
 	kafkaConfig         *config.KafkaConfig
 }
 
+func (q QuotaManagementListService) ReserveQuotaIfNotAlreadyReserved(kafka *dbapi.KafkaRequest) (string, *errors.ServiceError) {
+	return q.ReserveQuota(kafka)
+}
+
+func (q QuotaManagementListService) DeleteQuotaForBillingModel(subscriptionId string, kafkaBillingModel config.KafkaBillingModel) *errors.ServiceError {
+	return nil // NOOP
+}
+
 var _ services.QuotaService = &QuotaManagementListService{}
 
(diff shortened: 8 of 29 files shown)


internal/kafka/internal/services/quota/internal/utils/billing_model_resolver.go
```diff
@@ -0,0 +1,71 @@
...
+
+/********************************************************************************************************************
+ * This is the root of the billing resolver chain.
+ * It demands the resolution of the billing model to the defined resolvers and stops the chain as soon as one of
+ * the resolvers is able to manage the request.
+ ********************************************************************************************************************/
+
+type BillingModelResolver interface {
+	SupportRequest(kafka *dbapi.KafkaRequest) bool
+	Resolve(orgID string, kafka *dbapi.KafkaRequest) (BillingModelDetails, error)
```
machi1990, line 27: What's the difference between the orgID parameter and the organization id field in the `kafka` parameter?
Is the first the AMS org id and the later the RH SSO external_id?

internal/kafka/internal/services/quota/internal/utils/simple_billing_model_resolver.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+)
+
+/********************************************************************************************************************
+ * This resolver is used when the user didn't specify the `billing model` and the `marketplace` in the kafka request.
+ * It implements the legacy logic: if the user has quota for `standard`, it returns `standard`. If he has quota for
```
machi1990, line 13: > It implements the legacy logic

Why legacy? Is this code meant to be removed at some point in the future?

ziccardi: See previous comment

internal/kafka/internal/services/quota/internal/utils/cloudaccount_billing_model_resolver.go
```diff
@@ -0,0 +1,76 @@
...
+		return BillingModelDetails{}, errors.NewWithCause(errors.ErrorGeneral, err, "error reserving quota")
+	}
+
+	quotaCosts, err := r.quotaConfigProvider.GetQuotaCostsForProduct(orgId, kafkaBillingModel.AMSResource, kafkaBillingModel.AMSProduct)
+	if err != nil {
+		return BillingModelDetails{}, errors.InsufficientQuotaError("%v: error getting quotas for product %s", err, kafkaBillingModel.AMSProduct)
+	}
+	if idx, cloudAccount := arrays.FindFirst(getCloudAccounts(quotaCosts), func(cloudAccount *v1.CloudAccount) bool {
+		return cloudAccount.CloudAccountID() == kafka.BillingCloudAccountId && (cloudAccount.CloudProviderID() == kafka.Marketplace || kafka.Marketplace == "")
+	}); idx != -1 {
```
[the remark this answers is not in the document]
ziccardi: see previous comment

internal/kafka/internal/services/quota/internal/utils/misc.go
```diff
@@ -0,0 +1,40 @@
+package utils
```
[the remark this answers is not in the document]
ziccardi: see previous comment

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/delete_actual_quota_action.go
```diff
@@ -0,0 +1,53 @@
...
+		return res, true, err
+	}
+
+	// WARNING: all the other methods called before are returning an `error`. Quota service methods
+	// are returning a *ServiceError object. If we reuse the `err` variable here, err != nil will be
+	// always `false`.
+	if err := quotaService.DeleteQuotaForBillingModel(kafkaRequest.SubscriptionId, *bm); err != nil {
+		// at this stage we have already reserved the new quota, so we need the reconciler to keep retrying
+		// mark the error as recoverable
+		return res, true, errors.NewServiceErrorBuilder().Wrap(*err).Recoverable().Build()
```
ziccardi, line 49: Should we log this to sentry?

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/update_kafka_request_action.go
```diff
@@ -0,0 +1,37 @@
...
+	updates["PromotionStatus"] = ""
+	updates["PromotionDetails"] = ""
+	updates["ActualKafkaBillingModel"] = kafkaRequest.DesiredKafkaBillingModel
+	// current result must have the subscription id at this point
+	updates["SubscriptionId"] = currentResult.Value().SubscriptionID
+
+	err := u.kafkaService.Updates(kafkaRequest, updates)
+	if err != nil {
+		// we need to mark the error as recoverable so that the reconciler will keep on retrying
+		return currentResult, true, errors.NewServiceErrorBuilder().Wrap(*err).Recoverable().Build()
```
ziccardi, line 34: Should we log this to sentry?

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,6 +296,39 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+	}
+
+	subscriptions, err := q.amsClient.FindSubscriptions(fmt.Sprintf("cluster_id='%s'", kafka.ClusterID))
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+	for _, subscription := range subscriptions {
+		// TODO: should we check subscription status?
+		// get the reserved resources
+		if subscription.Plan().ID() == kafkaBillingModel.AMSProduct {
```
ziccardi, line 316: Should we check the subscription status and ignore it if it is not 'ACTIVE'?

config/quota-management-list-configuration.yaml
```diff
@@ -91,6 +91,8 @@ registered_users_per_organisation:
         kafka_billing_models:
           - id: standard
             max_allowed_instances: 5
+          - id: marketplace
```
miguelsorianod, line 94: Was that missing from another PR? it doesn't seem related to this PR

internal/kafka/internal/migrations/20221215161330_add_kafka_promote_worker_in_leader_leases.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+var promotingLeaderLeaseTypes = []string{"promoting_kafka"}
+
+func addKafkaPromoteWorkerInLeaderLeases() *gormigrate.Migration {
+	return &gormigrate.Migration{
+		ID: "20221215161330",
```
miguelsorianod, line 14: We will need to change the ID and reorder the migration as in `main` branch there is a newer dated migration already

internal/kafka/internal/migrations/20221215161330_add_kafka_promote_worker_in_leader_leases.go
```diff
@@ -0,0 +1,28 @@
...
+		Migrate: func(tx *gorm.DB) error {
+			for _, leaderLeaseType := range promotingLeaderLeaseTypes {
```
miguelsorianod, line 16: Is there some reason we put it in an array? there does not seem to be a need for it

pkg/errors/errors.go
```diff
@@ -361,15 +361,75 @@ func ToServiceError(err error) *ServiceError {
...
+	return seb
+}
+
+func (seb *serviceErrorBuilder) Recoverable() ServiceErrorBuilder {
+	seb.recoverable = true
+	return seb
+}
+
+func (seb *serviceErrorBuilder) Build() *ServiceError {
+	err := NewWithCause(seb.Code, seb.cause, seb.Error(), seb.Reason)
```
miguelsorianod, line 417: Shouldn't the builder perform some verifications or at least set some defaults?

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,55 @@
+package chain
```
miguelsorianod, line 1: Is there some reason this has been implemented with generics over just interfaces?

I suspect it is due to having potentially different return types in the results between reconcile actions.

What would be the alternative without using generics? I think it is good we evaluate the pros/cons of using each approach

config/quota-management-list-configuration.yaml
ziccardi: I added that so that I could add tests to promote from `eval` to `marketplace`

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
miguelsorianod: > Yes, that is the reason. Different chains could produce different results. Using an interface, we would end up using the any value.

If `any` is used with interface that means that the receiver of the result should perform downcasting to get the real type right?

ziccardi: yes

config/quota-management-list-configuration.yaml
ziccardi: Is that used for something other than a test? It contains user `testuser1@example.com`, `testuser2@example.com`, etc so I thought it was used only for tests

miguelsorianod: `config/quota-management-list-configuration.yaml` is used as the configuration of the quota management list.
It takes effect even on non-test runs.
I was assuming the existence of `testuser1...` was to give an example of the format, although that could be commented out. We should avoid setting the default configuration files with test data. Although this seems an issue that was already happening with this file as you mentioned. We can deal with it separately.

internal/kafka/internal/migrations/20221215161330_add_kafka_promote_worker_in_leader_leases.go
ziccardi: done

pkg/errors/errors.go
ziccardi: The checks are done in the `NewWithCause` function.
The `recoverable` field defaults to `false`

internal/kafka/internal/migrations/20221219154830_add_kafka_promote_worker_in_leader_leases.go
```diff
@@ -0,0 +1,26 @@
+package migrations
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+var leaderLeaseType = "promoting_kafka"
```
miguelsorianod, line 10: Can we make it a local variable of the addKafkaPromoteWorkerInLeaderLeases method to avoid polluting the package-level namespace?

pkg/errors/errors_test.go
```diff
@@ -145,23 +145,23 @@ func Test_NewWithCause(t *testing.T) {
 			args: args{
 				reason: internalServerCause,
 			},
```
miguelsorianod, line 147: Should we add tests for the builder?

ziccardi: I was going to add the tests, but I stopped: the builder doesn't contain any logic. It just assigns received values to the embedded instance, and for the BUILD it relies on the `NewWithCause` function.

I believe adding tests would be overkill for this object.

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
```diff
@@ -0,0 +1,97 @@
+package promotion
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/workers/kafka_mgrs/promotion/chain"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions"
+	errors2 "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
```
miguelsorianod, line 9: Let's use a more descriptive package alias than `errors2`. For example `apiErrors` or `svcErrors`, `kasfleetmanagererrors`, `kerrors`, `fleetErrors`, ...

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return nil
+}
+
+func (k *PromotionKafkaManager) Reconcile() []error {
+	glog.Infoln("reconciling kafkas to be promoted")
+	kafkasToPromote, err := k.kafkaService.ListKafkasToBePromoted()
+	if err != nil {
+		return []error{errors.Wrap(err, "failed to list kafkas to promote")}
+	} else {
```
miguelsorianod, line 70: No need for an `else`.
If there is no error we can unconditionally log as if there was an error we return before

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
```diff
@@ -0,0 +1,97 @@
...
+	promotionChain := chain.NewReconcileActionRunner(
+		actions.NewAssignDesiredQuotaAction(*k.kafkaConfig, k.quotaServiceFactory),
+		actions.NewDeleteActualQuotaAction(*k.kafkaConfig, k.quotaServiceFactory),
+		actions.NewUpdateKafkaRequestAction(k.kafkaService),
+	)
+
+	var promotionErrors []error
+
+	for _, kafka := range kafkasToPromote {
+		if res, promotionError := promotionChain.Run(kafka); promotionError != nil {
```
miguelsorianod, line 84: Can we extract the individualized kafka logic into its own method?

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
```diff
@@ -0,0 +1,97 @@
...
+func (k *PromotionKafkaManager) Start() {
+	k.StartWorker(k)
+}
+
+// Stop causes the process for reconciling accepted kafka requests to stop.
+func (k *PromotionKafkaManager) Stop() {
+	k.StopWorker(k)
+}
+
+func (k *PromotionKafkaManager) updatePromotionDetails(kafkaRequest *dbapi.KafkaRequest, promotionError error) error {
```
miguelsorianod, line 49: This assumes the kafka promotion has failed. Let's rename the method to a one that reflects that like for example updateFailedPromotionDetails

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/assign_desired_quota_action.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (d *AssignActualQuotaAction) PerformJob(kafkaRequest *dbapi.KafkaRequest, currentResult chain.Result[PromotionContext]) (chain.Result[PromotionContext], bool, error) {
+	res := chain.Result[PromotionContext]{}
+	quotaService, factoryErr := d.quotaServiceFactory.GetQuotaService(api.QuotaType(d.kafkaConfig.Quota.Type))
+	if factoryErr != nil {
+		return res, true, errors.NewWithCause(errors.ErrorGeneral, factoryErr, "unable to check quota")
+	}
+
+	subscriptionId, err := quotaService.ReserveQuotaIfNotAlreadyReserved(kafkaRequest)
```
miguelsorianod, line 33: This should be called `subscriptionID`

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/update_kafka_request_action.go
```diff
@@ -0,0 +1,37 @@
...
+func NewUpdateKafkaRequestAction(kafkaService services.KafkaService) chain.ReconcileAction[PromotionContext] {
+	return &UpdateKafkaRequestAction{
+		kafkaService: kafkaService,
+	}
+}
+
+func (u UpdateKafkaRequestAction) PerformJob(kafkaRequest *dbapi.KafkaRequest, currentResult chain.Result[PromotionContext]) (chain.Result[PromotionContext], bool, error) {
+	// gorm ignores zero values, so to zero `PromotionStatus` and `PromotionDetails` we need to use a map
+	updates := map[string]any{}
+	updates["PromotionStatus"] = ""
```
miguelsorianod, line 25: This does not seem accurate.

The fields in the DB are `promotion_status`,  `promotion_details`, `actual_kafka_billing_model` and `subscription_id`. This also makes me think we need an integration test for this to verify it with a real DB.

Does gorm update it correctly with the current implementation? Maybe it is flexible enough to detect it but I think we should use the values in the DB as that's what the key is intended to be

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/update_kafka_request_action.go
```diff
@@ -0,0 +1,37 @@
...
+	kafkaService services.KafkaService
+}
+
+func NewUpdateKafkaRequestAction(kafkaService services.KafkaService) chain.ReconcileAction[PromotionContext] {
+	return &UpdateKafkaRequestAction{
+		kafkaService: kafkaService,
+	}
+}
+
+func (u UpdateKafkaRequestAction) PerformJob(kafkaRequest *dbapi.KafkaRequest, currentResult chain.Result[PromotionContext]) (chain.Result[PromotionContext], bool, error) {
```
miguelsorianod, line 22: We should document what is this action intended to perform along with the other actions.
Important in this cases to document what are the expected previous results, the assumptions made (like in this one it is that the kafka has successfully been reserved at ams level without error?), etc...

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/delete_actual_quota_action.go
```diff
@@ -0,0 +1,53 @@
...
+		// instance type was validated at creation stage. This should never happen.
+		return res, true, err
+	}
+	bm, err := instanceType.GetKafkaSupportedBillingModelByID(kafkaRequest.ActualKafkaBillingModel)
+	if err != nil {
+		// actual billing model was validated at creation stage. This should never happen.
+		return res, true, err
+	}
+
+	// WARNING: all the other methods called before are returning an `error`. Quota service methods
```
miguelsorianod, line 43: In this cases I find helpful to explicitely assign a non standard name to the error. For example `svcErr`

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/assign_desired_quota_action.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/workers/kafka_mgrs/promotion/chain"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+)
+
+var _ chain.ReconcileAction[PromotionContext] = &AssignActualQuotaAction{}
+
+type AssignActualQuotaAction struct {
```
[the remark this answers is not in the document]
ziccardi: Nope. Fixed the namings! Thanks!

pkg/client/ocm/client.go
```diff
@@ -41,7 +41,8 @@ type Client interface {
 	DeleteCluster(clusterID string) (int, error)
 	ClusterAuthorization(cb *amsv1.ClusterAuthorizationRequest) (*amsv1.ClusterAuthorizationResponse, error)
 	DeleteSubscription(id string) (int, error)
-	FindSubscriptions(query string) (*amsv1.SubscriptionsListResponse, error)
+	FindSubscriptions(query string) ([]*amsv1.Subscription, error)
```
miguelsorianod, line 44: What is the criteria on returning an AMSXXXResponse object vs an AMSXXX object?

internal/kafka/internal/services/kafka.go
```diff
@@ -498,6 +499,28 @@ func (k *kafkaService) ListByStatus(status ...constants.KafkaStatus) ([]*dbapi.K
...
+	}
+
+	var kafkas []*dbapi.KafkaRequest
+
+	if err := dbConn.Model(&dbapi.KafkaRequest{}).
+		Where("actual_kafka_billing_model <> desired_kafka_billing_model").
+		Where("promotion_status <> ?", dbapi.KafkaPromotionStatusFailed).
+		Where("status in ?", validPromotionStatuses).
+		Scan(&kafkas).Error; err != nil {
+		return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to list by status")
```
miguelsorianod, line 518: rename the error to failed to list kafkas to be promoted or something else

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -161,6 +170,14 @@ func (q QuotaManagementListService) ReserveQuota(kafka *dbapi.KafkaRequest) (str
 // 7) if [6] fails, return the first defined billing model
 func (q QuotaManagementListService) detectBillingModel(kafka *dbapi.KafkaRequest) (string, *errors.ServiceError) {
 	if kafka.DesiredKafkaBillingModel != "" {
+		instanceType, err := q.kafkaConfig.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(kafka.InstanceType)
```
miguelsorianod, line 173: Could you explain why is this added?

pkg/client/ocm/client.go
ziccardi: I believe we should always return the result instead of the response.
Anyway, here I had to change it because `amsv1.SubscriptionsListResponse` is not mockable, and that was preventing me from creating unit tests.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +340,46 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
 	}
 	return nil
 }
+
+func (q amsQuotaService) DeleteQuotaForBillingModel(subscriptionId string, kafkaBillingModel config.KafkaBillingModel) *errors.ServiceError {
```
miguelsorianod, line 344: This should be `subscriptionID`

internal/kafka/internal/services/quota/quota_management_list_service.go
ziccardi: because without that any `DesiredBillingModel` was valid, even if not present into the configuration

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +340,46 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
...
+	if subscriptionId == "" {
```
miguelsorianod, line 345: What's the reason this is checked?

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +340,46 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
...
+		return nil
+	}
+
+	subscription, found, err := q.GetSubscriptionByID(subscriptionId)
+	if err != nil {
+		return errors.GeneralError("failed to delete the quota: %v", err)
+	}
+
+	if !found {
```
miguelsorianod, line 354: Let's document this method's behavior

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +340,46 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
...
+		return nil
+	}
+
+	if subscription.Plan().ID() == kafkaBillingModel.AMSProduct {
```
miguelsorianod, line 358: Is there the possibility that the given provided subscription ID is not the one we want to delete? Is this check needed?

miguelsorianod: Same question applies to the AMS Resource check

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/update_kafka_request_action.go
ziccardi: An integration test has already been added, and it looks like gorm correctly updates the fields.
I will, however, change this.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +340,46 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
...
+			// this subscription is related to this billing model. We can delete it.
+			return q.DeleteQuota(subscriptionId)
+		}
+	}
+
+	// the specified subscription id is not related to the specified billing model
+	return nil
+}
+
+func (q amsQuotaService) GetSubscriptionByID(subscriptionId string) (*amsv1.Subscription, bool, *errors.ServiceError) {
```
miguelsorianod, line 375: Let's document the behavior of this method in the different cases: is not finding it an error? a non-error but with the subscription nil? etc...

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,6 +296,39 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+
+	subscriptions, err := q.amsClient.FindSubscriptions(fmt.Sprintf("cluster_id='%s'", kafka.ClusterID))
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+	for _, subscription := range subscriptions {
+		// TODO: should we check subscription status?
```
miguelsorianod, line 314: I think so. What happens if there's an old subscription that is not used anymore?

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,6 +296,39 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+	instanceType, err := q.kafkaConfig.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(kafka.InstanceType)
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+	kafkaBillingModel, err := instanceType.GetKafkaSupportedBillingModelByID(kafka.DesiredKafkaBillingModel)
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+
+	subscriptions, err := q.amsClient.FindSubscriptions(fmt.Sprintf("cluster_id='%s'", kafka.ClusterID))
```
miguelsorianod, line 309: Can't we check the subscription_id? this would require storing it in the DB in this step itself and not at the end :thinking:

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: I think so. But this check was already done for `DeleteQuota` so I preferred to keep it.

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: Hmm, we can probably get rid of this `DeleteQuotaForBillingModel` method.
It was meant to ensure that it didn't delete the subscription if, after a promotion, the `subscription_id` was updated, but the `actual billing model` was still pointing to the old one.

I believe with the latest code it can't happen.

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: Nope. This method is used by the reconciler so that it can recover even in case of errors updating the database.

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/delete_actual_quota_action.go
ziccardi: @miguelsorianod @JameelB ^

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/update_kafka_request_action.go
ziccardi: @miguelsorianod @JameelB ^

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+)
+
+type ReconcileAction[T any] interface {
+	PerformJob(kafkaRequest *dbapi.KafkaRequest, currentResult Result[T]) (Result[T], bool, error)
```
JameelB, line 8: Can we document this method and its return values?

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/delete_actual_quota_action.go
JameelB: I think we can log this in Sentry. It should already be without any additional changes, any errors returned by reconcilers are logged to Sentry I believe. 

[optional] It would be great to note that this error is recoverable in the Sentry logs though if possible.

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/update_kafka_request_action.go
JameelB: Same comment as above

ziccardi:
> [optional] It would be great to note that this error is recoverable in the Sentry logs though if possible.

@JameelB : Agree. It should be done here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/de3c8de28fba0d5031cfc2e1cf8d701bf3df4ccd/pkg/logger/logger.go#L210

I don't have a solution off the top of my mind. I would leave that for a future PR.

internal/kafka/internal/services/kafka.go
```diff
@@ -498,6 +499,28 @@ func (k *kafkaService) ListByStatus(status ...constants.KafkaStatus) ([]*dbapi.K
 	return kafkas, nil
 }
 
+func (k *kafkaService) ListKafkasToBePromoted() ([]*dbapi.KafkaRequest, *errors.ServiceError) {
+	dbConn := k.connectionFactory.New()
+
+	validPromotionStatuses := []string{
```
miguelsorianod, line 505: In the `github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi` package we now have the `KafkaPromotionStatus` type with a set of methods.
Can we add a method that returns the "valid promotion statuses / promotable states" and reuse it here?

miguelsorianod, line 505: In `internal/kafka/internal/handlers/validation.go` the `validateKafkaRequestToPromoteHasAPromotableStatus` also has an instantiation of a variable called `acceptedKafkaStatusesForPromotion`.

Maybe it would be a good idea to somehow reuse the list of states for that between them

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,69 @@
...
+	// * currentResult - object containing the results that have been produced up to this point into the pipeline
+	// Return values
+	// * res Result[T] - the result produced by this action
+	// * finished bool - returning `true` here will interrupt the pipeline execution
+	// * err error - an eventual error. Returning an error always interrupts the pipeline execution
+	PerformJob(kafkaRequest *dbapi.KafkaRequest, currentResult Result[T]) (res Result[T], finished bool, err error)
+}
+
+// Result - This struct contains the result produced by each action
+type Result[T any] struct {
```
miguelsorianod, line 21: Maybe a better name could be `ActionResult` as it is intended to use in the context of reconcile actions?

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (r *Result[T]) Value() T {
+	return r.value
+}
+
+func (r *Result[T]) HasValue() bool {
+	return r.hasValue
+}
+
+// The ReconcileActionRunner is the engine of the action chain. It runs each action one after the other
```
miguelsorianod, line 39: The `The` word should be removed. Idiomatic Go documentation contains the name of the method / type to be documented as the first word in the document comment.

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,69 @@
+package chain
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+)
+
+// ReconcileAction - This is the interface that must be implemented by each action that the reconciler will execute.
```
miguelsorianod, line 7: To be idiomatic with Go documentation style this should be reworded as:
`ReconcileAciton is the interface that ... `

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,69 @@
...
+	// * kafkaRequest - the kafka request to reconcile
+	// * currentResult - object containing the results that have been produced up to this point into the pipeline
+	// Return values
+	// * res Result[T] - the result produced by this action
+	// * finished bool - returning `true` here will interrupt the pipeline execution
+	// * err error - an eventual error. Returning an error always interrupts the pipeline execution
+	PerformJob(kafkaRequest *dbapi.KafkaRequest, currentResult Result[T]) (res Result[T], finished bool, err error)
+}
+
+// Result - This struct contains the result produced by each action
```
miguelsorianod, line 20: To be idiomatic with Go documentation style this should be reworded as:
`Result contains ...`

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,69 @@
...
+	return r.value
+}
+
+func (r *Result[T]) HasValue() bool {
+	return r.hasValue
+}
+
+// The ReconcileActionRunner is the engine of the action chain. It runs each action one after the other
+// passing to each of them the current result so that all the action can concur to create a piece of the
+// result. When the chain execution is finished, the result will be complete.
```
miguelsorianod, line 41: Let's document the meaning of `T` for ReconcileActionRunner`

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
```diff
@@ -0,0 +1,69 @@
...
+	actions []ReconcileAction[T]
+}
+
+func NewReconcileActionRunner[T any](actions ...ReconcileAction[T]) ReconcileActionRunner[T] {
+	return ReconcileActionRunner[T]{
+		actions: actions,
+	}
+}
+
+// Run the chain end return the final result of the execution
```
miguelsorianod, line 52: Let's document the behavior of how the intermediate results are managed and also the return value. By that I mean: are intermediate results merged? overwritten? is the final result a merge of all the results? the result of the latest executed action?

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,14 +297,93 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
 
-	_, err := q.amsClient.DeleteSubscription(subscriptionId)
+	_, err := q.amsClient.DeleteSubscription(subscriptionID)
 	if err != nil {
 		return errors.GeneralError("failed to delete the quota: %v", err)
 	}
 	return nil
 }
+
+// DeleteQuotaForBillingModel - Deletes the quota identified by `subscriptionID` only if it is related to received KafkaBillingModel
```
miguelsorianod, line 344: This should be rewritten to be idiomatic with Go documentation:
`DeleteQuotaForBillingModel deletes the quota ...`

internal/kafka/internal/services/quota/ams_quota_service.go
miguelsorianod: :+1: Let's review this

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,14 +297,93 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+	}
+
+	if subscription.Plan().ID() == kafkaBillingModel.AMSProduct {
+		rr, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscriptionID)
+		if err != nil {
+			return errors.GeneralError("failed to delete the quota: %v", err)
+		}
+		if arrays.AnyMatch(rr, func(resource *amsv1.ReservedResource) bool {
+			return resource.ResourceName() == kafkaBillingModel.AMSResource
+		}) {
```
miguelsorianod, line 366: This could be rewritten to store the result of anymatch in a variable and then check the value of the variable.
I think that would improve readability as right now it is a conditional where the arugment is a method call that includes a method as parameter adding multiline and nesting into the condition itself

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,14 +297,93 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+			// this subscription is related to this billing model. We can delete it.
+			return q.DeleteQuota(subscriptionID)
+		}
+	}
+
+	// the specified subscription id is not related to the specified billing model
+	return nil
+}
+
+// GetSubscriptionByID - Gets a subscription by ID
```
miguelsorianod, line 376: This should be rewritten to be idiomatic with Go documentation:
`GetSubscriptionByID gets a subscription by ID ...

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,14 +297,93 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+// return an error if an error happens calling the endpoint, otherwise returns weather it found the subscription(subscription, true, nil)
```
miguelsorianod, line 377: weather -> whether

internal/kafka/internal/workers/kafka_mgrs/promotion/internal/actions/reserve_desired_quota_action.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func NewReserveDesiredQuotaAction(kafkaConfig config.KafkaConfig, quotaServiceFactory services.QuotaServiceFactory) chain.ReconcileAction[PromotionContext] {
+	return &ReserveDesiredQuotaAction{
+		quotaServiceFactory: quotaServiceFactory,
+		kafkaConfig:         kafkaConfig,
+	}
+}
+
+func (d *ReserveDesiredQuotaAction) PerformJob(kafkaRequest *dbapi.KafkaRequest, currentResult chain.Result[PromotionContext]) (chain.Result[PromotionContext], bool, error) {
```
miguelsorianod, line 26: Would be good to document the behavior of the different implementations of quota actions. Specially if they are idempotent

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
```diff
@@ -0,0 +1,105 @@
...
+			WorkerType: "promoting_kafka",
+			Reconciler: reconciler,
+		},
+		kafkaService:        kafkaService,
+		quotaServiceFactory: quotaServiceFactory,
+		kafkaConfig:         kafkaConfig,
+	}
+}
+
+// Start initializes the kafka manager to reconcile accepted kafka requests.
```
miguelsorianod, line 39: Comment is outdated. We should update it

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
```diff
@@ -0,0 +1,105 @@
...
+func (k *PromotionKafkaManager) Start() {
+	k.StartWorker(k)
+}
+
+// Stop causes the process for reconciling accepted kafka requests to stop.
```
miguelsorianod, line 44: Comment is outdated. We should update it

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (k *PromotionKafkaManager) Reconcile() []error {
+	glog.Infoln("reconciling kafkas to be promoted")
+	kafkasToPromote, err := k.kafkaService.ListKafkasToBePromoted()
+	if err != nil {
+		return []error{errors.Wrap(err, "failed to list kafkas to promote")}
+	}
+	glog.Infof("found %d kafkas to promote", len(kafkasToPromote))
+
+	var promotionErrors []error
```
miguelsorianod, line 73: A `ErrorList` can be used instead

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "1234",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "EVAL",
+
+				wantsReserveCalled: true,
+				wantsProductID:     "RHOSAKEval",
```
miguelsorianod, line 1055: The `RHOSAKEvalProduct` constant available in the `pkg/client/ocm` package can be used instead

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "1234",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "EVAL",
```
miguelsorianod, line 1052: Let's use `eval` instead to match with how it would be used in reality

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-1",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "STANDARD",
+
+				wantsReserveCalled: false,
+				wantsProductID:     "RHOSAK",
```
miguelsorianod, line 1087: The RHOSAKProduct constant available in the pkg/client/ocm package can be used instead

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-1",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "STANDARD",
```
miguelsorianod, line 1084: Let's use `standard` to match with the value it would be used in reality

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+
+				wantsReserveCalled: false,
+				wantsProductID:     "RHOSAK",
+			},
+		},
+		{
+			// In this test we try to reserve a standard for a cluster that has EVAL assigned
+			name: "Reserve standard quota for existing cluster with EVAL subscription",
```
miguelsorianod, line 1092: Let's use `eval` to match with the value it would be used in reality

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-7",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "STANDARD",
```
miguelsorianod, line 1117: Let's use `standard` to match with reality

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+
+				wantsReserveCalled: true,
+				wantsProductID:     "RHOSAK",
+			},
+		},
+		{
+			// In this test we try to reserve an EVAL for a cluster that has EVAL assigned
+			name: "Reserve EVAL quota for existing cluster with EVAL subscription",
```
miguelsorianod, line 1125: Let's use `eval` to match with reality

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-7",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "EVAL",
+
+				wantsReserveCalled: false,
+				wantsProductID:     "RHOSAK",
```
miguelsorianod, line 1153: The `RHOSAKProduct` constant available in the pkg/client/ocm package can be used instead

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+						sub.Status("Active")
+						ca, _ := v1.NewClusterAuthorizationResponse().Allowed(true).Subscription(&sub).Build()
+						return ca, nil
+					},
+				},
+			},
+			args: args{
+				kafkaID:                         "cluster-7",
+				kafkaInstanceType:               types.STANDARD,
+				kafkaRequestDesiredBillingModel: "EVAL",
```
miguelsorianod, line 1150: let's use `eval` to match with reality

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+			DesiredKafkaBillingModel: tt.args.kafkaRequestDesiredBillingModel,
+		}
+
+		t.Run(tt.name, func(t *testing.T) {
+			g := gomega.NewWithT(t)
+			factory := NewDefaultQuotaServiceFactory(tt.fields.ocmClient, nil, nil, &tt.fields.kafkaConfig)
+			quotaService, _ := factory.GetQuotaService(api.AMSQuotaType)
+
+			_, err := quotaService.ReserveQuotaIfNotAlreadyReserved(kafka)
+			g.Expect(err).ToNot(gomega.HaveOccurred(), "Expecting error not to happen:", err)
```
miguelsorianod, line 1181: We could add a test that tests some failure

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -1000,6 +1001,195 @@ func Test_AMSCheckQuota(t *testing.T) {
...
+		fields fields
+		args   args
+	}{
+		{
+			name: "Reserve quota for new cluster",
+			fields: fields{
+				kafkaConfig: amsDefaultKafkaConf,
+				ocmClient: &ocm.ClientMock{
+					FindSubscriptionsFunc: func(query string) ([]*v1.Subscription, error) {
+						return ams.FindSubscriptions(query), nil
```
miguelsorianod, line 1035: I understand the motivation of using the mock which is re-usability but it has the implication that it is effectively using some global state which means that changing the mock impacts all tests that use it. This makes in turn hard to perform changes when new needs arise for the mock, with effects of the change impacting several tests in the code

internal/kafka/internal/services/quota/internal/mocks/ams/subscriptions_mock.go
```diff
@@ -0,0 +1,47 @@
...
+	"subscription-9":  newSubscription("subscription-9", "cluster-9", "RHOSAKEval"),
+	"subscription-10": newSubscription("subscription-10", "cluster-10", "RHOSAK"),
+}
+
+func FindSubscriptions(query string) []*v1.Subscription {
+	r := regexp.MustCompile("^cluster_id='(?P<clusterid>.*)'$")
+
+	switch true {
+	case r.MatchString(query):
+		clusterId := r.FindStringSubmatch(query)[1]
```
miguelsorianod, line 36: clusterId -> clusterID

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr_test.go
```diff
@@ -0,0 +1,325 @@
...
+							SupportedKafkaInstanceTypes: []config.KafkaInstanceType{
+								{
+									Id:          "standard",
+									DisplayName: "standard",
+									Sizes:       nil,
+									SupportedBillingModels: []config.KafkaBillingModel{
+										{
+											ID:               "eval",
+											AMSResource:      "rhosak",
+											AMSProduct:       "RHOSAKEval",
```
miguelsorianod, line 104: For the `AMSProduct` the constants defined in the `pkg/client/ocm` package can be used instead

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr_test.go
```diff
@@ -0,0 +1,325 @@
...
+						Configuration: config.SupportedKafkaInstanceTypesConfig{
+							SupportedKafkaInstanceTypes: []config.KafkaInstanceType{
+								{
+									Id:          "standard",
+									DisplayName: "standard",
+									Sizes:       nil,
+									SupportedBillingModels: []config.KafkaBillingModel{
+										{
+											ID:               "eval",
+											AMSResource:      "rhosak",
```
miguelsorianod, line 103: The `RHOSAKResourceName` constant available in the `pkg/client/ocm` package can be used instead

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr_test.go
```diff
@@ -0,0 +1,325 @@
...
+					GetQuotaServiceFunc: func(quotaType api.QuotaType) (services.QuotaService, *errors.ServiceError) {
+						return tt.fields.quotaService, nil
+					},
+				},
+			)
+
+			errs := k.Reconcile()
+			g.Expect(errs).To(gomega.HaveLen(tt.wantErrCount))
+
+			//if len(errs) == 0 {
```
miguelsorianod, line 306: This comment does not seem accurate.
If len(errs) == 0 the following code would still be executed, assuming that the gomega expectation executed above matches

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
ziccardi: An optional leading article is admitted AFAIK

miguelsorianod: My comment is based on information seen around godoc's approach to documentation, https://go.dev/blog/godoc, where beginning with the name of the element it describes seems to be the convention: 
```
Notice this comment is a complete sentence that begins with the name of the element it describes. This important convention allows us to generate documentation in a variety of formats, from plain text to HTML to UNIX man pages, and makes it read better when tools truncate it for brevity, such as when they extract the first line or sentence.
```

ziccardi: Look at [here](https://tip.golang.org/doc/comment):
```
package bytes

// A Buffer is a variable-sized buffer of bytes with Read and Write methods.
// The zero value for Buffer is an empty buffer ready to use.
type Buffer struct {
    ...
}
```

```
package io

// A LimitedReader reads from R but limits the amount of
// data returned to just N bytes. Each call to Read
// updates N to reflect the new amount remaining.
// Read returns EOF when N <= 0.
type LimitedReader struct {
    R   Reader // underlying reader
    N   int64  // max bytes remaining
}
```

and many others.

miguelsorianod: However, the usage of ` - ` does not seem that conventional . For example, in :
`// PerformJob - ...`

ziccardi: Yes. I'm going through the code and removing all of them.

internal/kafka/internal/services/quota/ams_quota_service_test.go
ziccardi: That wouldn't be useful.
The only cases the function can fail are:

1. If there is a configuration error and the instance type of an existing Kafka doesn't exist in the configuration
2. If there is a configuration error and the billing model type of an existing Kafka doesn't exist in the configuration
3. If there are communication errors with AMS
4. If it fails to reserve the quota. This relies on the `ReserveQuota` method, that is already been tested elsewhere

There is no need to test cases 1 to 3.

internal/kafka/internal/services/quota/ams_quota_service_test.go
ziccardi: I see your point. However, I put that mock in an internal package confined to the quota service test. The risk is limited enough to be acceptable.

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: done

internal/kafka/internal/workers/kafka_mgrs/promotion/promotion_kafkas_mgr.go
ziccardi: Done. Mind checking if I used it correctly?

internal/kafka/internal/services/quota/ams_quota_service_test.go
miguelsorianod: Even between two tests in a single file it is potentially fragile and I've seen that already happening in several places with package-global variables used between 1..N tests. I think we should aim to have per-test instances usually.
Let's keep this in mind when writing new tests where we need to introduce new elements

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -2,6 +2,8 @@ package quota
 
 import (
 	"fmt"
+	v1 "github.com/openshift-online/ocm-sdk-go/authorizations/v1"
```
miguelsorianod, line 5: Side comment:

Do you use some kind of auto importer / auto imports organizer?

Usually the imports are reordered automatically on save by standard libraries followed by non-standard external libraries imports and other packages in the project.

internal/kafka/internal/services/quota/ams_quota_service.go
miguelsorianod: Was this reviewed?

internal/kafka/internal/services/quota/ams_quota_service_test.go
miguelsorianod: I think a valid test is testing what happens if fails to reserve quota, not because of being a bug in the ReserveQuota happening but because this happening during reconciliation time

internal/kafka/internal/workers/kafka_mgrs/promotion/chain/reconciler_chain.go
miguelsorianod: I still see some instances. For example:
```
	// PerformJob - This is the method where the Action implementation will perform the real job
```

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: it could be intellij

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,14 +297,94 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+	}
+
+	subscriptions, err := q.amsClient.FindSubscriptions(fmt.Sprintf("cluster_id='%s'", kafka.ClusterID))
+	if err != nil {
+		return "", errors.GeneralError("failed checking current quota: %v", err)
+	}
+	for _, subscription := range subscriptions {
+		// get the reserved resources
+		if subscription.Plan().ID() == kafkaBillingModel.AMSProduct && subscription.Status() == string(v1.SubscriptionStatusActive) {
+			reservedResouces, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscription.ID())
```
JameelB, line 317:
```suggestion
			reservedResources, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscription.ID())
```

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -295,14 +297,94 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
...
+			if err != nil {
+				return "", errors.GeneralError("failed checking current quota: %v", err)
+			}
+
+			if arrays.AnyMatch(reservedResouces, func(rr *amsv1.ReservedResource) bool { return rr.ResourceName() == kafkaBillingModel.AMSResource }) {
```
JameelB, line 322:
```suggestion
			if arrays.AnyMatch(reservedResources, func(rr *amsv1.ReservedResource) bool { return rr.ResourceName() == kafkaBillingModel.AMSResource }) {
```

pkg/client/ocm/client.go
JameelB: Is it worth changing the response type of the GetSubscriptionByID below to return the resource object instead in that case?

ziccardi:
> Is it worth changing the response type of the GetSubscriptionByID below to return the resource object instead in that case?

@JameelB I believe all methods should return the `resource` object instead of the `response` object. However, most of them are currently producing the `response` one.

Let's discuss this in a tech meeting to get a decision and fix all of them if needed.

internal/kafka/internal/services/quota/ams_quota_service_test.go
ziccardi: Yes, but this test is testing the `ReserveQuotaIfNotAlreadyReserved` and in that method, the call to `ReserveQuota` is the last call of the method, so testing it when there is no quota doesn't add value to this test.

pkg/client/ocm/client.go
ziccardi: I believe all methods should return the resource object instead of the response object. However, most of them are currently producing the response one.

Let's discuss this in a tech meeting to get a decision and fix all of them if needed.

miguelsorianod:
One of the commits is written in past. It is a good practice to write the commits in imperative style

internal/kafka/internal/services/quota/ams_quota_service_test.go
miguelsorianod: That is accurate for now, not if the implementation is modified in the future.
I think testing that would be more robust but it is not a big concern now so ok if we don't want to add it now.

ziccardi:
> One of the commits is written in past tense. It is a good practice to write the commits in imperative style

@miguelsorianod Fixed

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -266,13 +268,21 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
 	// will be empty if no marketplace account is used
 	rr.BillingMarketplaceAccount(kafka.BillingCloudAccountId)
 
+	// TODO: temporary workaround. This is to be removed when this issue is implemented: https://issues.redhat.com/browse/SDB-3157
+	getClusterId := func(kafka *dbapi.KafkaRequest) string {
+		if kafka.DesiredKafkaBillingModel == "eval" {
+			return kafkaId + "-eval"
```
miguelsorianod, line 274: this can be rewritten as `return fmt.Sprintf("%s-eval", kafkaId)`

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -266,13 +268,21 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
 	// will be empty if no marketplace account is used
 	rr.BillingMarketplaceAccount(kafka.BillingCloudAccountId)
 
+	// TODO: temporary workaround. This is to be removed when this issue is implemented: https://issues.redhat.com/browse/SDB-3157
+	getClusterId := func(kafka *dbapi.KafkaRequest) string {
```
miguelsorianod, line 272: The id accronym should be named `ID`. Therefore the variable name should be `getClusterID`

miguelsorianod: Alternatively, `getAuthRequestClusterID`

ziccardi: Fixed!

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: Yeah, but in this case I find concatenating the string more readable.

JameelB (approved):
Verified locally, works as expected. Eval subscriptions are created with a suffix of `eval` on cluster id and external_cluster_id. On promotion, a new subscription was created without this suffix and the old one was deleted. 

Kafka request in the database was updated with the correct values 👍

PR #1455, opened 2022-12-16
feat: add grace_period_days to kafka billing model configuration

JameelB:
## Description
Adds `grace_period_days` to the Kafka billing model configuration. 

Note that this PR only contains changes to add this field to the configuration. Any changes to use this field will be done in follow on JIRA issues/PRs.

## Verification Steps
- [ ] CI checks passing
- [ ] Run `make deploy/service` and ensure the default configuration is applied correctly.

- [ ] Code Review completed


miguelsorianod:
Maybe having the destination of the PR to the feature branch might be better?

internal/kafka/internal/config/kafka_supported_instance_types.go
```diff
@@ -35,6 +35,7 @@ type KafkaBillingModel struct {
 	AMSResource      string   `yaml:"ams_resource" validate:"required,ams_resource_validator"`
 	AMSProduct       string   `yaml:"ams_product" validate:"required,ams_product_validator"`
 	AMSBillingModels []string `yaml:"ams_billing_models" validate:"min=1,unique,ams_billing_models_validator"`
+	GracePeriodDays  int      `yaml:"grace_period_days"`
```
miguelsorianod, line 38: Here we can add a minimum validation of 0

config/kafka-instance-types-configuration.yaml
```diff
@@ -12,6 +12,23 @@
 #   - [required] ams_resource: AMS resource to be used. Accepted values: ['rhosak']
 #   - [required] ams_product: AMS product to be used. Accepted values: ['RHOSAK', 'RHOSAKTrial', 'RHOSAKEval']
 #   - [required] ams_billing_models: List of AMS available billing models: Accepted values: ['marketplace', 'marketplace-rhm', 'marketplace-aws']
+#   - [optional] grace_period_days: The length of time in days when a Kafka instance is set to be in a 'grace period'. This configuration only applies
```
miguelsorianod, line 15: Let's document the default too and the meaning of it

miguelsorianod: I see it's documented a little bit below with another wording

internal/kafka/internal/config/kafka_supported_instance_types.go
miguelsorianod: Unless we want to use `-1` to signal another meaning?

JameelB: don't think we need a `-1` if we have 0, i'll add `validate:"min=0"` to this, thanks!

config/kafka-instance-types-configuration.yaml
```diff
@@ -12,6 +12,23 @@
...
+#                                   If the Kafka instance has not been promoted by the end of the grace period, the Kafka expires and is marked for deletion.
+# 
+#                                   If this field is not configured or is set to 0, Kafka instances with this instance type and billing model are not given
+#                                   any grace period. In this case, when the user or organisation's subscription is no longer valid, the Kafka instance 
+#                                   expires (marked for deletion) immediately.
+#                                   
+#                                   Please note that lifespanSeconds takes precedence over this configuration if its specified for the instance type. 
+#                                   E.g. An instance type has a lifespanSeconds of 172800 (2 days) and the kafka billing model has a grace period of 10 days.
+#                                   If the subscription expires 1 day after the Kafka instance was created, it will be set to expire (marked for deletion)
+#                                   2 days after its creation date without fully completing the 10 day grace period.
```
JameelB, line 31: @miguelsorianod @ziccardi do you agree with this behaviour? We should never have lifespanSeconds for `standard` instances, but in case someone tries to add this grace_period to `developer` instances or any other instances with this config, lifespanSeconds should always take precedence. wdyt?

config/kafka-instance-types-configuration.yaml
JameelB: I've also added the default and minimum value at the top to make those clearer.

config/kafka-instance-types-configuration.yaml
```diff
@@ -55,6 +72,7 @@ supported_instance_types:
       ams_product: RHOSAKEval
       ams_billing_models:
       - standard
+      grace_period_days: 10
```
ziccardi, line 75: This should be 4

config/kafka-instance-types-configuration.yaml
ziccardi: I agree about the `lifespanSeconds` precedence.
Currently, the description is accurate. In the future, we could however expand the `grace_period_days` to a generalised usage for any expired instance.

miguelsorianod: +1 to the defined precedence here.

config/kafka-instance-types-configuration.yaml
JameelB: updated there, thanks :)

config/kafka-instance-types-configuration.yaml
JameelB: thanks for confirmation! :)

miguelsorianod (approved):

PR #1458, opened 2022-12-16
fix: remove wrongly created enterprise cluster in prod

pb82:
## Description

I accidentally hit the enterprise cluster registration endpoint in prod. This migration removes the cluster from the database.

## Verification Steps

1. Check the KAS Fleet Manager Metrics dashboard in grafana prod and make sure that the cluster stuck in waiting for fleetshard operator is removed.

### DIFF
--- a/internal/kafka/internal/migrations/20221216120000_remove_wrongly_created_enterprise_cluster.go
+++ b/internal/kafka/internal/migrations/20221216120000_remove_wrongly_created_enterprise_cluster.go
@@ -1,0 +1,24 @@
+package migrations
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func removeWronglyCreatedEnterpriseClusterInProd() *gormigrate.Migration {
+	return &gormigrate.Migration{
+		ID: "20221216120000",
+		Migrate: func(tx *gorm.DB) error {
+			// remove the cluster with cluster id=20l15s8hsji3190maa4ut40p5561k87q which was wrongly created by a enterprise cluster registration call
+			if err := tx.Exec("DELETE FROM clusters where cluster_id='20l15s8hsji3190maa4ut40p5561k87q';").Error; err != nil {
+				return err
+			}
+
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
+			// no need to rollback a one shot operation
+			return nil
+		},
+	}
+}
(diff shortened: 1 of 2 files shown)


internal/kafka/internal/migrations/20221216120000_remote_wrongly_created_enterprise_cluster.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func removeWronglyCreatedEnterpriseClusterInProd() *gormigrate.Migration {
+	return &gormigrate.Migration{
+		ID: "20221216120000",
+		Migrate: func(tx *gorm.DB) error {
+			// remove the cluster with cluster id=20l15s8hsji3190maa4ut40p5561k87q which was wrongly created by a enterprise cluster registration call
+			if err := tx.Exec("DELETE FROM clusters where status='waiting_for_kas_fleetshard_operator' AND cluster_id='20l15s8hsji3190maa4ut40p5561k87q';").Error; err != nil {
```
miguelsorianod, line 13: What happens if the cluster id does not exist? Would an error be returned?

pb82: we did something similar before: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/internal/migrations/20221104090000_remove_the_wrongly_auto_created_cluster_from_stage_environment_with_id_cdhunvd8igjhbi0nmtt0.go

this migration is still present, so I assume it will just dry run if the cluster is not found.
@JameelB are you familiar with how that case in stage was  handled?

JameelB: if the cluster id doesn't exist, it returns no error. It returns delete applied to 0 clusters.

image

I believe gorm doesn't re-apply migrations that already ran (already recorded in the migrations table). However, the previous db migration we did on Stage you linked above was also released to prod which didn't have the cluster it was trying to delete. That migration was successful. 

I've tested this change locally with nothing in my clusters table and it ran successfully. In the logs I see 
```
I1216 16:59:38.361477   81872 cmd.go:22] Database has 70 migrations applied
```

One thing though, do we want to keep the status here? Will this cluster change status before we can apply this migration? @pb82

pb82: The cluster should not change status. However, it also doesn't need to be there, the cluster_id is unique. I can remove it.

miguelsorianod (approved):

PR #1464, opened 2022-12-19
chore: addressed code comments

ziccardi:
## Description

Addressed code comments

- [ ] Code Review completed


internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -87,7 +87,7 @@ func (q amsQuotaService) validateBillingAccount(organisationId string, instanceT
 
 	quotaCosts, err := q.amsClient.GetQuotaCostsForProduct(orgId, kafkaBillingModel.AMSResource, kafkaBillingModel.AMSProduct)
 	if err != nil {
-		return errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("error checking quota: failed to get assigned quota of type %s/%s for organization with id %v", instanceType.String(), kafkaBillingModel.ID, orgId))
+		return errors.NewWithCause(errors.ErrorGeneral, err, fmt.Sprintf("error checking quota: failed to get assigned quota for %s/%s to organization with id %v", instanceType.String(), kafkaBillingModel.ID, orgId))
```
miguelsorianod, line 90: `orgId` should be named `orgID`

internal/kafka/internal/services/quota/internal/test/kafka_supported_instance_types_config.go
```diff
@@ -0,0 +1,131 @@
...
+	cfg := quotaListKafkaSupportedInstanceTypesConfig
+	return &cfg
+}
+
+func NewAMSTestKafkaSupportedInstanceTypesConfig() *config.KafkaSupportedInstanceTypesConfig {
+	cfg := amsKafkaSupportedInstanceTypesConfig
+	return &cfg
+}
+
+var quotaListKafkaSupportedInstanceTypesConfig = config.KafkaSupportedInstanceTypesConfig{
```
miguelsorianod, line 15: Can we add the word `test` somewhere in between this name? It is a package-level variable which is also reachable from other files in this package

internal/kafka/internal/services/quota/internal/test/kafka_supported_instance_types_config.go
```diff
@@ -0,0 +1,131 @@
...
+						DeprecatedQuotaType:         "rhosak",
+						CapacityConsumed:            1,
+					},
+				},
+			},
+		},
+	},
+}
+
+var amsKafkaSupportedInstanceTypesConfig = config.KafkaSupportedInstanceTypesConfig{
```
miguelsorianod, line 60: Same as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1464/files#r1052415724

internal/kafka/internal/services/quota/internal/utils/kafka_billing_model_resolver.go
```diff
@@ -72,19 +72,17 @@ func (k *kafkaBillingModelResolver) Resolve(orgId string, kafka *dbapi.KafkaRequ
 // detectAmsBillingModel - tries to detect the `ams billing model`
 // 1) If the user has specified a cloud account id, it falls back to the `cloud account billing model resolver`
 // 2) If no cloud account has been specified, it falls back to the `simple billing model resolver`
-func (k *kafkaBillingModelResolver) detectAmsBillingModel(orgId string, kafka *dbapi.KafkaRequest, kafkaBillingModel config.KafkaBillingModel) (string, error) {
-	validAmsBillingModels := kafkaBillingModel.AMSBillingModels
+func (k *kafkaBillingModelResolver) detectAmsBillingModel(orgID string, kafka *dbapi.KafkaRequest, kafkaBillingModel config.KafkaBillingModel) (string, error) {
```
miguelsorianod, line 75: AMS should be uppercased as it is an acronym

internal/kafka/internal/services/quota/internal/test/kafka_supported_instance_types_config.go
miguelsorianod: Not necessary as I see it is in a package named `test`.
Why are we creating our own `test` package for that? It does not seem a common practice in Go

ziccardi: It is a shared configuration to be used only for testing. I couldn't find a better name.

JameelB: There is a [mocks](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/tree/main/internal/kafka/test/mocks) package, is it possible to move this to that package instead?

ziccardi: that package is an `internal` package of Kafka. If we put this there, it won't be accessible.

JameelB: But this package is also within `internal/kafka` so we should be able to import it in the tests within `internal/kafka` which I assume this is what its going to be used for? 

I'm wondering if we can move everything in this file to the mocks package and follow the mock builder pattern that's there?

ziccardi: Oh, you are right. They are in the same root. That would work.

I wouldn't say I like putting everything inside the same folder: I'm more for self-containing code than spaghetti dependencies.
When I look at a test, I expect that everything I need is there and if I change something I use in the test, the worst that can happen is that I break that test. If we put everything in a centralised package, you have to worry that it could be used somewhere in the code every time you change it.

miguelsorianod: The decision seems to be between putting it in the internal package or in this test package.
No strong opinion about one way or the other. Having this `test` package is not very idiomatic of Go as it is usually put together in the same package as non-test code.

miguelsorianod (approved):

PR #1471, opened 2023-01-05
Add long lived eval instances expiration deletion logic

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-10012

This PR adds logic that takes care of marking long lived eval instances whose expiration (ExpiresAt) has been reached as deprovisioning.

The PR assumes that for developer instances the ExpiredAt column is also set. The code for that is still not implemented but merging this should not be an issue as this PR is to be merged against the `longlived-instances-featurebranch` and this PR and the one about setting ExpiredAt column for developer instances should both be merged together in main.

The logic for deprovisioning instances now evaluates whether kafka instances should be deprovisioned based on the ExpiresAt column, no matter the kafka instance type.

Additionally, the ExpiresAt type at DB type level has been changed from time.Time to sql.NullTime. The reason for this is to signal `null` as meaning "No expiration time has been set for the kafka instance". A new migration has been added to update the zero-value of the old time.Time being used for ExpiresAt to null for already existing kafka instances. To avoid potential race conditions the same logic has been applied to the kafkas_mgr reconciler temporarily.

To do:
- [ ] Tests
- [x] Decide how to combine and structure the code for both developer and long lived eval instances
- [x] Fully decide which type to use forKafkaRequest's ExpiresAt attribute at DB level
- [ ] Cleanup code with the definitive approach

## Verification Steps

1. Tests pass
1. Verify that newly created instances are created with the expires_at field set to NULL instead of to the zero-value of time (0001-01-01 ...)
1. Perform migration verifications:
  3.1. Run kfm from main branch and apply migrations
  3.2. Create a kafka instance and verify that it is created with expires_at with the zero-value of time (0001-01-01 ...)
  3.3. Stop KFM and move to this PR's branch
  3.4. Delete the migration id created in this PR from the DB
  3.5. Run db/migrate
  3.6. Verify that now that kafka instance has the expires_at field has changed to NULL
  3.7. Run the previous steps but now after creating the kafka instance manually  update the deleted_at value (not to be confused with the expires_at column) to NULL. Then verify that when running the migration the expires_at field has not been changed to NULL
  3.8. Run the previous steps but now after creating the kafka instance manually update the expires_at value to a non-null value. Then verify that when running the migration the expires_at field has not been changed to NULL
1. Perform reconcile verifications:
  4.1. Run kfm from main branch and apply migrations
  4.2. Create a kafka instance and verify that it is created with expires_at with the zero-value of time (0001-01-01 ...)
  4.3 Stop KFM and move to this PR's branch
  4.4 Run kas fleet manager and verify that after some time the kafka instance has the expires_at field changed to NULL
  4.5 Run the previous steps but after creating the kafka manually update the deleted_at value (not to be confused with the expires_at column) to NULL. Then verify that when running kfm the expires_at field has not been changed to NULL
  4.6 Run the previous steps but now after creating the kafka instance manually update the expires_at value to a non-null value. Then verify that when running the kfm the expires_at field has not been changed to NULL

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/migrations/20230109140000_update_expires_at_zero_value.go
+++ b/internal/kafka/internal/migrations/20230109140000_update_expires_at_zero_value.go
@@ -1,0 +1,34 @@
+package migrations
+
+import (
+	"database/sql"
+	"time"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func updateExpiresAtZeroValueFromKafkaRequests() *gormigrate.Migration {
+	nullTime := sql.NullTime{Time: time.Time{}, Valid: false}
+	zeroTime := time.Time{}
+
+	return &gormigrate.Migration{
+		ID: "20230109140000",
+		Migrate: func(tx *gorm.DB) error {
+			err := tx.Table("kafka_requests").Where("expires_at = ?", zeroTime).Where("deleted_at IS NULL").Update("expires_at", nullTime).Error
+			if err != nil {
+				return err
+			}
+
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
+			err := tx.Table("kafka_requests").Where("expires_at = ?", nullTime).Where("deleted_at IS NULL").Update("expires_at", zeroTime).Error
+			if err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+}
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -625,24 +625,24 @@ func (k *kafkaService) DeprovisionKafkaForUsers(users []string) *errors.ServiceE
 	return nil
 }
 
+func (k *kafkaService) kafkaWithExpiresAtShouldBeDeprovisioned(kafkaRequest *dbapi.KafkaRequest, currentTime time.Time) bool {
+	glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', size ID '%s' and status '%s'", kafkaRequest.ID, kafkaRequest.InstanceType, kafkaRequest.SizeId, kafkaRequest.Status)
+	if currentTime.After(kafkaRequest.ExpiresAt.Time) {
+		glog.V(10).Infof("Kafka ID '%s' has expired", kafkaRequest.ID)
+		return true
+	}
+
+	glog.V(10).Infof("Kafka ID '%s' has not expired", kafkaRequest.ID)
+
+	return false
+}
+
 func (k *kafkaService) DeprovisionExpiredKafkas() *errors.ServiceError {
 	dbConn := k.connectionFactory.New().Model(&dbapi.KafkaRequest{}).Session(&gorm.Session{})
 
-	var typesWithLifespan []string
-	for _, kafkaInstanceType := range k.kafkaConfig.SupportedInstanceTypes.Configuration.SupportedKafkaInstanceTypes {
-		if kafkaInstanceType.HasAnInstanceSizeWithLifespan() {
-			typesWithLifespan = append(typesWithLifespan, kafkaInstanceType.Id)
-		}
-	}
-
-	if len(typesWithLifespan) == 0 {
-		return nil
-	}
-	glog.V(10).Infof("Kafka instance types with lifespan set: %+v", typesWithLifespan)
-
 	var existingKafkaRequests []dbapi.KafkaRequest
-	db := dbConn.Where("instance_type IN (?)", typesWithLifespan).
-		Where("status NOT IN (?)", kafkaDeletionStatuses).
+	db := dbConn.Where("status NOT IN (?)", kafkaDeletionStatuses).
+		Where("expires_at IS NOT NULL").
 		Scan(&existingKafkaRequests)
 	err := db.Error
 	if err != nil {
@@ -651,22 +651,14 @@ func (k *kafkaService) DeprovisionExpiredKafkas() *errors.ServiceError {
 
 	var kafkasToDeprovisionIDs []string
 	timeNow := time.Now()
-	for _, existingKafkaRequest := range existingKafkaRequests {
-		glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', ID '%s' and status '%s'", existingKafkaRequest.ID, existingKafkaRequest.InstanceType, existingKafkaRequest.SizeId, existingKafkaRequest.Status)
-		kafkaInstanceSize, err := k.kafkaConfig.GetKafkaInstanceSize(existingKafkaRequest.InstanceType, existingKafkaRequest.SizeId)
-		if err != nil {
-			return errors.NewWithCause(errors.ErrorGeneral, err, "unable to deprovision expired kafkas")
-		}
-		if kafkaInstanceSize.LifespanSeconds != nil {
-			glog.V(10).Infof("Kafka size associated to kafka ID '%s' has '%d' lifespanSeconds", existingKafkaRequest.ID, *kafkaInstanceSize.LifespanSeconds)
-			expTime := existingKafkaRequest.GetExpirationTime(*kafkaInstanceSize.LifespanSeconds)
-			glog.V(10).Infof("Expiration time of kafka ID '%s' is '%s'", existingKafkaRequest.ID, expTime)
-			if timeNow.After(*expTime) {
-				glog.V(10).Infof("Kafka ID '%s' has expired", existingKafkaRequest.ID)
-				kafkasToDeprovisionIDs = append(kafkasToDeprovisionIDs, existingKafkaRequest.ID)
-			} else {
-				glog.V(10).Infof("Kafka ID '%s' still has not expired", existingKafkaRequest.ID)
-			}
+
+	for idx := range existingKafkaRequests {
+		existingKafkaRequest := &existingKafkaRequests[idx]
+		glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', size ID '%s' and status '%s'", existingKafkaRequest.ID, existingKafkaRequest.InstanceType, existingKafkaRequest.SizeId, existingKafkaRequest.Status)
+
+		shouldBeDeprovisioned := k.kafkaWithExpiresAtShouldBeDeprovisioned(existingKafkaRequest, timeNow)
+		if shouldBeDeprovisioned {
+			kafkasToDeprovisionIDs = append(kafkasToDeprovisionIDs, existingKafkaRequest.ID)
 		}
 	}
 
@@ -679,7 +671,7 @@ func (k *kafkaService) DeprovisionExpiredKafkas() *errors.ServiceError {
 			return errors.NewWithCause(errors.ErrorGeneral, err, "unable to deprovision expired kafkas")
 		}
 		if db.RowsAffected >= 1 {
-			glog.Infof("%v kafka_request's lifespans are over their lifespan and have had their status updated to deprovisioning", db.RowsAffected)
+			glog.Infof("%v expired kafka_request's have had their status updated to deprovisioning", db.RowsAffected)
 			var counter int64 = 0
 			for ; counter < db.RowsAffected; counter++ {
 				metrics.IncreaseKafkaTotalOperationsCountMetric(constants.KafkaOperationDeprovision)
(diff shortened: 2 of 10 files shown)


miguelsorianod:
Updated the original description to include verifications for reviewers

internal/kafka/internal/services/kafka.go
```diff
@@ -619,24 +625,25 @@ func (k *kafkaService) DeprovisionKafkaForUsers(users []string) *errors.ServiceE
...
-	for _, kafkaInstanceType := range k.kafkaConfig.SupportedInstanceTypes.Configuration.SupportedKafkaInstanceTypes {
-		if kafkaInstanceType.HasAnInstanceSizeWithLifespan() {
-			typesWithLifespan = append(typesWithLifespan, kafkaInstanceType.Id)
-		}
+func (k *kafkaService) kafkaWithExpiredAtShouldBeDeprovisioned(kafkaRequest *dbapi.KafkaRequest, currentTime time.Time) (bool, error) {
+	glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', size ID '%s' and status '%s'", kafkaRequest.ID, kafkaRequest.InstanceType, kafkaRequest.SizeId, kafkaRequest.ID, kafkaRequest.Status)
+	shouldDeprovision := false
+	if currentTime.After(kafkaRequest.ExpiresAt.Time) {
+		glog.V(10).Infof("Kafka ID '%s' has expired", kafkaRequest.ID)
+		shouldDeprovision = true
```
ziccardi, line 633:
```suggestion
		return true, nil
```

internal/kafka/internal/services/kafka.go
```diff
@@ -619,24 +625,25 @@ func (k *kafkaService) DeprovisionKafkaForUsers(users []string) *errors.ServiceE
...
+	} else {
+		glog.V(10).Infof("Kafka ID '%s' has not expired", kafkaRequest.ID)
 	}
 
-	if len(typesWithLifespan) == 0 {
-		return nil
-	}
-	glog.V(10).Infof("Kafka instance types with lifespan set: %+v", typesWithLifespan)
+	return shouldDeprovision, nil
```
ziccardi, line 638:
```suggestion
	return false, nil
```

ziccardi:
> cc @JameelB @ziccardi ready for a review. If the approach looks good to you then I will proceed to update the testing if/where needed

@miguelsorianod The approach looks ok to me. Let's go on with this.

JameelB (commented):
Approach looks good to me too

internal/kafka/internal/services/kafka.go
miguelsorianod: :+1: . This was a leftover of a refactoring. I will change it

internal/kafka/internal/services/kafka.go
miguelsorianod: :+1:

internal/kafka/internal/services/kafka.go
```diff
@@ -645,22 +652,22 @@ func (k *kafkaService) DeprovisionExpiredKafkas() *errors.ServiceError {
...
 	timeNow := time.Now()
-	for _, existingKafkaRequest := range existingKafkaRequests {
-		glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', ID '%s' and status '%s'", existingKafkaRequest.ID, existingKafkaRequest.InstanceType, existingKafkaRequest.SizeId, existingKafkaRequest.Status)
-		kafkaInstanceSize, err := k.kafkaConfig.GetKafkaInstanceSize(existingKafkaRequest.InstanceType, existingKafkaRequest.SizeId)
+
+	for idx, _ := range existingKafkaRequests {
+		existingKafkaRequest := &existingKafkaRequests[idx]
+		glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', size ID '%s' and status '%s'", existingKafkaRequest.ID, existingKafkaRequest.InstanceType, existingKafkaRequest.SizeId, existingKafkaRequest.ID, existingKafkaRequest.Status)
+
+		if !existingKafkaRequest.ExpiresAt.Valid {
```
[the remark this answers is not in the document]
miguelsorianod: This check is not required. I will remove it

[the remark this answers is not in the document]
miguelsorianod: Changed

internal/kafka/internal/services/kafka.go
miguelsorianod: updated

internal/kafka/internal/services/kafka.go
miguelsorianod: updated

miguelsorianod:
Thanks for the review. I will update the tests

miguelsorianod:
@JameelB @ziccardi this PR is ready for the final review and verifications.

Please pay special attention to the verifications steps. Specially around the migration edge cases in both the migration step and the continuous reconciliation step for the migration.

internal/kafka/internal/migrations/20230109140000_update_expires_at_zero_value.go
```diff
@@ -0,0 +1,34 @@
...
+		Migrate: func(tx *gorm.DB) error {
+			err := tx.Table("kafka_requests").Where("expires_at = ?", zeroTime).Where("deleted_at IS NULL").Update("expires_at", nullTime).Error
+			if err != nil {
+				return err
+			}
+
+			return nil
+		},
+		Rollback: func(tx *gorm.DB) error {
+			err := tx.Table("kafka_requests").Where("expires_at = ?", nullTime).Where("deleted_at IS NULL").Update("expires_at", zeroTime).Error
```
ziccardi, line 26: I believe this is not an _actual_ rollback since it will update **all** the rows where `expires_at` is `null` to the `zeroTime` value (even the rows where `expires_at` was already null before this migration started.

However, it should only be an issue if I'm missing something.

internal/kafka/internal/services/kafka.go
```diff
@@ -619,24 +625,24 @@ func (k *kafkaService) DeprovisionKafkaForUsers(users []string) *errors.ServiceE
...
-			typesWithLifespan = append(typesWithLifespan, kafkaInstanceType.Id)
-		}
+func (k *kafkaService) kafkaWithExpiresAtShouldBeDeprovisioned(kafkaRequest *dbapi.KafkaRequest, currentTime time.Time) (bool, error) {
+	glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', size ID '%s' and status '%s'", kafkaRequest.ID, kafkaRequest.InstanceType, kafkaRequest.SizeId, kafkaRequest.Status)
+	if currentTime.After(kafkaRequest.ExpiresAt.Time) {
+		glog.V(10).Infof("Kafka ID '%s' has expired", kafkaRequest.ID)
+		return true, nil
+	} else {
+		glog.V(10).Infof("Kafka ID '%s' has not expired", kafkaRequest.ID)
 	}
```
ziccardi, line 635: This else clause can be removed since the `if` is returning

internal/kafka/internal/services/kafka.go
```diff
@@ -619,24 +625,24 @@ func (k *kafkaService) DeprovisionKafkaForUsers(users []string) *errors.ServiceE
...
-	dbConn := k.connectionFactory.New().Model(&dbapi.KafkaRequest{}).Session(&gorm.Session{})
-
-	var typesWithLifespan []string
-	for _, kafkaInstanceType := range k.kafkaConfig.SupportedInstanceTypes.Configuration.SupportedKafkaInstanceTypes {
-		if kafkaInstanceType.HasAnInstanceSizeWithLifespan() {
-			typesWithLifespan = append(typesWithLifespan, kafkaInstanceType.Id)
-		}
+func (k *kafkaService) kafkaWithExpiresAtShouldBeDeprovisioned(kafkaRequest *dbapi.KafkaRequest, currentTime time.Time) (bool, error) {
+	glog.V(10).Infof("Evaluating expiration time of kafka request '%s' with instance type '%s', size ID '%s' and status '%s'", kafkaRequest.ID, kafkaRequest.InstanceType, kafkaRequest.SizeId, kafkaRequest.Status)
+	if currentTime.After(kafkaRequest.ExpiresAt.Time) {
```
ziccardi, line 630: I believe we should check if `ExpiresAt` is `nil`.

I think we can use a method like this here: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/be6dc11b91d2a4d4e0015c9b623e8febe3dc0d47/internal/kafka/internal/api/dbapi/kafka_request_types.go#L145

internal/kafka/internal/services/kafka.go
```diff
@@ -619,24 +625,24 @@ func (k *kafkaService) DeprovisionKafkaForUsers(users []string) *errors.ServiceE
...
 
-func (k *kafkaService) DeprovisionExpiredKafkas() *errors.ServiceError {
-	dbConn := k.connectionFactory.New().Model(&dbapi.KafkaRequest{}).Session(&gorm.Session{})
-
-	var typesWithLifespan []string
-	for _, kafkaInstanceType := range k.kafkaConfig.SupportedInstanceTypes.Configuration.SupportedKafkaInstanceTypes {
-		if kafkaInstanceType.HasAnInstanceSizeWithLifespan() {
-			typesWithLifespan = append(typesWithLifespan, kafkaInstanceType.Id)
-		}
+func (k *kafkaService) kafkaWithExpiresAtShouldBeDeprovisioned(kafkaRequest *dbapi.KafkaRequest, currentTime time.Time) (bool, error) {
```
ziccardi, line 628: This method never returns errors

internal/kafka/internal/services/kafka_test.go
```diff
@@ -2519,20 +2530,63 @@ func Test_kafkaService_DeprovisionExpiredKafkas(t *testing.T) {
...
+					return dbTime
+				},
 			},
 			wantErr: false,
 			setupFn: func() {
-				mocket.Catcher.Reset().NewMock().WithQuery(`SELECT * FROM "kafka_requests" WHERE instance_type IN ($1) AND status NOT IN ($2,$3)`).WithReply([]map[string]interface{}{{"id": "kafkainstance1", "instance_type": instanceType, "size_id": instanceSize}})
-				mocket.Catcher.NewMock().WithQuery(`UPDATE "kafka_requests" SET "status"=$1,"updated_at"=$2 WHERE id IN ($3)`)
+				mocket.Catcher.Reset().NewMock().
+					WithQuery(`SELECT * FROM "kafka_requests" WHERE status NOT IN ($1,$2) AND expires_at IS NOT NULL`).
+					WithReply([]map[string]interface{}{{"id": "kafkainstance1", "instance_type": instanceType, "size_id": instanceSize, "expires_at": &expiredTime}})
```
ziccardi, line 2550: I believe using `any` instead of `interface{}` would be more readable (`any` is just an alias for `interface{}`)

internal/kafka/internal/services/kafka.go
miguelsorianod: There is no need to check it. This method is a private method and it is only called when the time is a valid time. It assumes a precondition of the time being valid

internal/kafka/internal/migrations/20230109140000_update_expires_at_zero_value.go
miguelsorianod: Before the migration no columns with the NULL value should exist. If the migration fails then those columns would be reverted. There might be some edge case for an instance being created but in the middle but there is not much we can do. The alternative would be not doing any rollback.

internal/kafka/internal/services/kafka.go
miguelsorianod: :+1:  I will change it

internal/kafka/internal/services/kafka.go
miguelsorianod: Changed :+1:

internal/kafka/internal/services/kafka_test.go
miguelsorianod: Not going to change it to be consistent with the other elements in the file. At some point if we conclude that `any` is preferred we can change sets of files as a general practice throughout the project. Maybe a good idea however would be to reference the same type as the one defined in the signatures of the library being used in the case of 3rd party libraries to have a 1:1 visual mapping. In any case not something to be dealt with in this PR

internal/kafka/internal/services/kafka.go
miguelsorianod: Changed

JameelB (approved):
Verified locally, works as expected 👍

ziccardi (approved):
Tested. Works as expected.

Great work!

PR #1473, opened 2023-01-06
feat: deregister enterprise cluster endpoint

pawelpaszki:
## Description
feat: deregister enterprise cluster endpoint

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -445,9 +445,11 @@ func validateEnterpriseClusterEligibleForDeregistration(ctx context.Context, clu
 			if err != nil {
 				return errors.GeneralError(err.Error())
 			}
-			if len(clusterKafkaCount) > 0 {
-				return errors.Forbidden("unable to deregister %s cluster with id: %s without explicitly setting 'force:true' due to kafka requests present on this cluster",
-					api.EnterpriseDataPlaneClusterType.String(), clusterID)
+			for _, count := range clusterKafkaCount {
+				if count.Clusterid == clusterID && count.Count > 0 {
+					return errors.Forbidden("unable to deregister %s cluster with id: %s without explicitly setting 'force:true' due to kafka requests present on this cluster",
+						api.EnterpriseDataPlaneClusterType.String(), clusterID)
+				}
 			}
 		}
 		return nil
--- a/internal/kafka/internal/handlers/validation_test.go
+++ b/internal/kafka/internal/handlers/validation_test.go
@@ -1928,6 +1928,30 @@ func Test_validateEnterpriseClusterEligibleForDeregistration(t *testing.T) {
 				api.EnterpriseDataPlaneClusterType.String(), clusterID),
 		},
 		{
+			name: "should not return an error if force is not set to true and FindKafkaInstanceCount contains clusterID and count is 0",
+			args: args{
+				ctx:       ctxWithClaims,
+				clusterID: clusterID,
+				clusterService: &services.ClusterServiceMock{
+					FindClusterByIDFunc: func(clusterID string) (*api.Cluster, *errors.ServiceError) {
+						return &api.Cluster{
+							OrganizationID: mocks.DefaultOrganisationId,
+							ClusterType:    "enterprise",
+						}, nil
+					},
+					FindKafkaInstanceCountFunc: func(clusterIDs []string) ([]services.ResKafkaInstanceCount, error) {
+						return []services.ResKafkaInstanceCount{
+							{
+								Clusterid: clusterID,
+								Count:     0,
+							},
+						}, nil
+					},
+				},
+			},
+			want: nil,
+		},
+		{
 			name: "should return no error if force is not set to true and FindKafkaInstanceCount returns an empty array",
 			args: args{
 				ctx:       ctxWithClaims,


internal/kafka/internal/handlers/cluster.go
```diff
@@ -132,3 +134,38 @@ func (h clusterHandler) List(w http.ResponseWriter, r *http.Request) {
...
+
+	cfg := &handlers.HandlerConfig{
+		Validate: []handlers.Validate{
+			handlers.ValidateAsyncEnabled(r, "deleting enterprise cluster"),
+			validateForceParam(),
+			ValidateKafkaClaims(ctx, ValidateOrganisationId()),
+			validateEnterpriseClusterEligibleForDeregistration(ctx, clusterID, forceBoolValue, h.clusterService),
+		},
+		Action: func() (i interface{}, serviceError *errors.ServiceError) {
+			// cluster deletion from db and cleanup will be handled in a follow up Jira
```
pawelpaszki, line 166: note that this will be done as a follow up. in this PR no db entry deletion takes place and its deliberate

internal/kafka/internal/handlers/cluster.go
```diff
@@ -132,3 +134,38 @@ func (h clusterHandler) List(w http.ResponseWriter, r *http.Request) {
...
+			return nil
+		}
+	}
+
+	cfg := &handlers.HandlerConfig{
+		Validate: []handlers.Validate{
+			handlers.ValidateAsyncEnabled(r, "deleting enterprise cluster"),
+			validateForceParam(),
+			ValidateKafkaClaims(ctx, ValidateOrganisationId()),
+			validateEnterpriseClusterEligibleForDeregistration(ctx, clusterID, forceBoolValue, h.clusterService),
```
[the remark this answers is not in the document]
machi1990: The admin has to be of the same organization as the one who registered the cluster i.e we also should have a check for org id `context.orgID === cluster.organizationID`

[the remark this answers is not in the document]
pawelpaszki: I was thinking about it, but currently we don't have it for registration endpoint, so if we add it here, we should also add it in the other endpoint

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1300,6 +1300,92 @@ paths:
...
+          When provided with value: true - enterprise cluster will be deleted alongside all kafkas present on the cluster.
+          When skipped and enterprise cluster has any kafkas associated with it, the request will fail.
+        explode: true
+        in: query
+        name: force
+        required: false
+        schema:
+          type: boolean
+        style: form
+      - description: The ID of record
```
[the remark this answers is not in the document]
pawelpaszki: btw. this is taken from generic param. I will replace it

internal/kafka/internal/handlers/cluster.go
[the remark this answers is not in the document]
machi1990: +1 on adding it everywhere. 

For the registration endpoint I seem to remember there was a JIRA for this. It could be done as part of this PR or a follow up one but for this PR it is essential that we add those two checks.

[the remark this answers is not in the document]
pawelpaszki: @machi1990 - I have addressed orgAdmin comment. orgID match is already implemented AFAIK

internal/kafka/internal/handlers/cluster.go
pawelpaszki: FYI @pb82

internal/kafka/internal/handlers/cluster.go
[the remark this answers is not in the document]
machi1990: @pawelpaszki the is org admin check is not there. I think you added it only for the registration endpoint.

> orgID match is already implemented AFAIK

I do not see this check. Can you point me to where it is done? Thanks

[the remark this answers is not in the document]
pawelpaszki: in the validation: `internal/kafka/internal/handlers/validation.go#L435` you check ID match and L425 checks for org admin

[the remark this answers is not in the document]
machi1990: Thanks. I see all the checks now, earlier on I was on an outdated patch.

internal/kafka/internal/handlers/validation.go
```diff
@@ -407,3 +407,51 @@ func validateKafkaMachinePoolNodeCount(clusterPayload *public.EnterpriseOsdClust
...
+				clusterKafkaCount, err := clusterService.FindKafkaInstanceCount([]string{clusterID})
+				if err != nil {
+					return errors.GeneralError(err.Error())
+				}
+				for _, count := range clusterKafkaCount {
+					if count.Clusterid == clusterID && count.Count > 0 {
+						return errors.Forbidden("unable to deregister %s cluster with id: %s without explicitly setting 'force:true' due to kafka requests present on this cluster",
+							api.EnterpriseDataPlaneClusterType.String(), clusterID)
+					}
+				}
```
machi1990, line 451: The loop is not required. We could just check if `len(clusterKafkaCount) > 0` and return the error returned in line 448.

machi1990 (changes requested):

internal/kafka/internal/handlers/validation.go
pawelpaszki: but clusterKafkaCount might be greater than 0 and don't contain entry for the cluster in question

openapi/kas-fleet-manager.yaml
```diff
@@ -1019,7 +1019,87 @@ paths:
...
+            type: boolean
+          required: true
+        - in: query
+          name: force
+          description: |-
+            When provided with value: true - enterprise cluster will be deleted alongside all kafkas present on the cluster.
+            When skipped and enterprise cluster has any kafkas associated with it, the request will fail.
+          schema:
+            type: boolean
+          required: false
```
machi1990, line 1042: @pb82 @pawelpaszki 

I understand the motivation behind the `force` flag but I am wondering if the alternative of not deleting the data plane cluster if it contains kafka was considered and only letting the user explicitly delete the Kafka.

pawelpaszki: I'll let Peter answer. I am fine either way. I see this endpoint being used mostly via Ui. if thats the case, there should be a warning stating that `force:true` will remove all kafkas

internal/kafka/internal/handlers/validation.go
machi1990: From here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/8cb312233571ae936ea9643379969fb30642634c/internal/kafka/internal/handlers/validation.go#L442 we are only filtering for the given cluster. The output is either empty if the cluster in question does not contain any kafka, otherwise it returns a list of kafkas of the given cluster and nothing else.

openapi/kas-fleet-manager.yaml
machi1990: Thanks @pawelpaszki let's see what @pb82 says. Perhaps this was discussed already in some other channels and I could missing some context.

internal/kafka/internal/handlers/validation.go
```diff
@@ -407,3 +407,51 @@ func validateKafkaMachinePoolNodeCount(clusterPayload *public.EnterpriseOsdClust
...
+		if !claims.IsOrgAdmin() {
+			return errors.New(errors.ErrorUnauthorized, "non admin user not authorized to perform this action")
+		}
+
+		cluster, err := clusterService.FindClusterByID(clusterID)
+		if err != nil {
+			return err
+		}
+
+		if cluster != nil {
```
machi1990, line 434: we could inverse the condition here to 
```suggestion
		if cluster == nil {
```

and to and early return for the error case returned by line 455

pawelpaszki: changed

internal/kafka/internal/handlers/validation.go
pawelpaszki: good point. changed

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1300,6 +1300,92 @@ paths:
...
+          When provided with value: true - enterprise cluster will be deleted alongside all kafkas present on the cluster.
+          When skipped and enterprise cluster has any kafkas associated with it, the request will fail.
+        explode: true
+        in: query
+        name: force
+        required: false
+        schema:
+          type: boolean
+        style: form
+      - description: ID of the OSD enterprise cluster
```
machi1990, line 1325:
```suggestion
      - description: ID of the enterprise data plane cluster
```

internal/kafka/internal/handlers/validation.go
machi1990: @pawelpaszki I noticed that the CI was failing due to my suggestion. I was mistaken https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1473#discussion_r1064574003 and my statement below 

> The output is either empty if the cluster in question does not contain any kafka, otherwise it returns a list of kafkas of the given cluster and nothing else.

is incorrect and misleading. 

The correct behaviour is:
1. if the cluster in question contains Kafka, a non empty list is returned. The list has one element whose Count reflects the total streaming unit count.
2. if the cluster in question doesn't contain any kafka, a non empty list is returned. The list has one element whose Count is `0`.

machi1990: Apologies for the wrong suggestion earlier on.
 
We could either keep the loop knowing that the list will always contain one element.
Or we could just fetch the first item of the list and work with it knowing that the list will always contain one element.
we should be guarded from future regressions by the tests.

internal/kafka/internal/handlers/cluster.go
pb82: what is the difference between a hard- and a soft delete in the database? Do we only set a deleted flag?

PR #1481, opened 2023-01-10
feat: add logic to suspend expiring instances

ziccardi:
## Description
Add logic to suspend expiring kafka instances.

A Kafka instance is suspended only if it is in one of the following statuses:

- KafkaRequestStatusAccepted
- KafkaRequestStatusPreparing
- KafkaRequestStatusReady
- KafkaRequestStatusProvisioning
- KafkaRequestStatusResuming,

Expired instances are ignored (we deal with them in #1471 )

## Verification Steps
1. Run KFM locally
2. Create a Kafka cluster. This can be simulated by adding the corresponding row into the database. If the reconcile for `active` state bother, disable it.
3. Update the Kafka request and set the ExpireAt column with a value sooner than your configured grace period
4. Verify that the cluster is moved to `suspending`

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/api/dbapi/kafka_request_types.go
+++ b/internal/kafka/internal/api/dbapi/kafka_request_types.go
@@ -4,6 +4,8 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
@@ -140,4 +142,85 @@ func (k *KafkaRequest) SetRoutes(routes []DataPlaneKafkaRoute) error {
 func (k *KafkaRequest) GetExpirationTime(lifespanSeconds int) *time.Time {
 	expireTime := k.CreatedAt.Add(time.Duration(lifespanSeconds) * time.Second)
 	return &expireTime
+}
+
+// The RemainingLifeSpan is the remaining life a kafka instance have before it expires and gets deleted
+type RemainingLifeSpan interface {
+	// Infinite returns whether the Kafka instance has infinite (no expiration) remaining life span
+	Infinite() bool
+
+	// Days returns the number of remaining life span days a Kafka instance has.
+	// The value can be negative if the instance has expired by 1 day or more to indicate how many days ago the instance
+	// has expired, or it can be `-1` if the remaining life is `infinite`.
+	// Caller shouldn't rely on this function returning `-1` to check for infinity, since this function will return `-1`
+	// even when the instance has expired one day ago.
+	// To reliable check for infinity, check the value returned by `Infinite`
+	Days() float64
+
+	// LessThanOrEqual returns whether the remaining number of days is less or equal then the specified one.
+	// If the remaining life is `infinte` this method always returns `false`.
+	LessThanOrEqual(days float64) bool
+}
+
+var _ RemainingLifeSpan = &remainingLifeSpan{}
+
+// remainingLifeSpan is an implementation of the RemainingLifeSpan interface
+type remainingLifeSpan struct {
+	// remainingLifeSpanDays is the number of days before the Kafka instance will be deleted. It is a `float64` to be able to manage fraction of days.
+	remainingLifeSpanDays float64
+	// infinite represent whether the Kafka instance has an expiration
+	infinite bool
+}
+
+func (l *remainingLifeSpan) Infinite() bool {
+	return l.infinite
+}
+
+func (l *remainingLifeSpan) Days() float64 {
+	return l.remainingLifeSpanDays
+}
+
+func (l *remainingLifeSpan) LessThanOrEqual(days float64) bool {
+	if l.infinite {
+		return false
+	}
+	return l.remainingLifeSpanDays <= days
+}
+
+// IsExpired returns whether a kafka is expired and how many days the instance will live before expiring
+func (k *KafkaRequest) IsExpired() (bool, RemainingLifeSpan) {
+	if !k.ExpiresAt.Valid {
+		return false, &remainingLifeSpan{
+			remainingLifeSpanDays: -1,
+			infinite:              true,
+		}
+	}
+
+	now := time.Now()
+	if now.After(k.ExpiresAt.Time) {
+		return true, &remainingLifeSpan{
+			remainingLifeSpanDays: 0,
+			infinite:              false,
+		}
+	}
+
+	remainingLife := k.ExpiresAt.Time.Sub(now)
+	return false, &remainingLifeSpan{
+		remainingLifeSpanDays: remainingLife.Hours() / 24,
+		infinite:              false,
+	}
+}
+
+// CanBeAutomaticallySuspended returns whether the kafka instance can be suspended or not
+// This method is used when the Kafka instance enters its grace period
+func (k *KafkaRequest) CanBeAutomaticallySuspended() bool {
+	validSuspensionStatuses := []string{
+		constants.KafkaRequestStatusAccepted.String(),
+		constants.KafkaRequestStatusPreparing.String(),
+		constants.KafkaRequestStatusReady.String(),
+		constants.KafkaRequestStatusProvisioning.String(),
+		constants.KafkaRequestStatusResuming.String(),
+	}
+
+	return arrays.Contains(validSuspensionStatuses, k.Status)
 }
--- a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
+++ b/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
@@ -120,6 +120,37 @@ func (k *KafkaManager) Reconcile() []error {
 		return encounteredErrors
 	}
 
+	for _, kafka := range kafkas {
+		if !kafka.CanBeAutomaticallySuspended() {
+			// this kafka is not in a state that can be suspended
+			continue
+		}
+
+		expired, remainingLifespan := kafka.IsExpired()
+		if expired {
+			// Expired kafkas will be deleted elsewhere
+			continue
+		}
+
+		// get the billing model
+		bm, err := k.kafkaConfig.GetBillingModelByID(kafka.InstanceType, kafka.ActualKafkaBillingModel)
+		if err != nil {
+			wrappedError := errors.Wrap(err, "failed to suspend expired Kafka instances")
+			encounteredErrors = append(encounteredErrors, wrappedError)
+			continue
+		}
+
+		if remainingLifespan.LessThanOrEqual(float64(bm.GracePeriodDays)) {
+			glog.Infof("cluster with ID '%s' entered its grace period. Suspending", kafka.ID)
+			// the instance is in grace period
+			_, err := k.kafkaService.UpdateStatus(kafka.ID, constants.KafkaRequestStatusSuspending)
+			if err != nil {
+				wrappedError := errors.Wrap(err, "failed to suspend expired Kafka instances")
+				encounteredErrors = append(encounteredErrors, wrappedError)
+			}
+		}
+	}
+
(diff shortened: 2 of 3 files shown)


internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -140,3 +140,48 @@ func (k *KafkaRequest) GetExpirationTime(lifespanSeconds int) *time.Time {
...
+	if now.After(k.ExpiresAt) {
+		return true, &lifeSpan{
+			lifeSpanDays: 0,
+			infinite:     false,
+		}
+	}
+
+	remainingLife := k.ExpiresAt.Sub(now)
+	return false, &lifeSpan{
+		lifeSpanDays: remainingLife.Hours() / 24,
```
miguelsorianod, line 184: This will result in inaccuracies of up to an hour.
A kafka instance that expires in 59 minutes will return a lifeSpanDays of 0 however it will return that it is not expired (false) at the same time.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -112,7 +113,42 @@ func (k *KafkaManager) Reconcile() []error {
 		}
 	}
 
-	// cleaning up expired qkafkas
+	validSuspensionStatuses := []constants.KafkaStatus{
```
miguelsorianod, line 116: Could we refactor the logic in its own method?

miguelsorianod:
Side note: It is a good practice to write commits in present-tense imperative style and not in past tense. See https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181 imperative-mood section. When writing commits you can think as: "If applied, this commit will ...." . Therefore, following imperative style the commit in the PR would be named:
`feat: add logic to suspend expiring instances`

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -140,3 +140,48 @@ func (k *KafkaRequest) GetExpirationTime(lifespanSeconds int) *time.Time {
 	expireTime := k.CreatedAt.Add(time.Duration(lifespanSeconds) * time.Second)
 	return &expireTime
 }
+
+type LifeSpan interface {
```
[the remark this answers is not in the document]
ziccardi: Done! Renamed to ResidualLifeSpan

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: Done! Added a `CanBeSuspended` to the `KafkaRequest` object

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -140,3 +142,60 @@ func (k *KafkaRequest) GetExpirationTime(lifespanSeconds int) *time.Time {
...
 }
+
+type ResidualLifeSpan interface {
+	Infinite() bool
+	Days() float64
+}
+
+var _ ResidualLifeSpan = &residualLifeSpan{}
+
+type residualLifeSpan struct {
```
JameelB, line 153: Can we add comments on what this struct and its properties represents?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
```diff
@@ -156,6 +158,226 @@ func TestKafkaManager_Reconcile(t *testing.T) {
...
+					return *cfg
+				}(),
+			},
+			wantErr: false,
+			updateStatusCall: updateStatusCall{
+				count: 0,
+			},
+		},
+		{
+			name: "should ignore non expiring kafkas",
```
JameelB, line 312:
```suggestion
			name: "should ignore kafkas that are not in a grace period",
```

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -120,6 +120,37 @@ func (k *KafkaManager) Reconcile() []error {
...
+
+		// get the billing model
+		bm, err := k.kafkaConfig.GetBillingModelByID(kafka.InstanceType, kafka.ActualKafkaBillingModel)
+		if err != nil {
+			wrappedError := errors.Wrap(err, "failed to suspend expired Kafka instances")
+			encounteredErrors = append(encounteredErrors, wrappedError)
+			continue
+		}
+
+		if !residualLifespan.Infinite() && residualLifespan.Days() < float64(bm.GracePeriodDays) {
```
machi1990, line 143:
```suggestion
		if !residualLifespan.Infinite() && residualLifespan.IsInGracePeriod(bm.GracePeriodDays) {
```

What do you think of having a `IsInGracePeriod()` method?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -120,6 +120,37 @@ func (k *KafkaManager) Reconcile() []error {
...
+			// the instance is in grace period
+			_, err := k.kafkaService.UpdateStatus(kafka.ID, constants.KafkaRequestStatusSuspending)
```
machi1990, line 145: Isn't there a possibility that a developer kafka will be seen in `suspsending` state just before (a few nanoseconds) before it moves to deprovision?

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -141,3 +143,67 @@ func (k *KafkaRequest) GetExpirationTime(lifespanSeconds int) *time.Time {
...
+
+	remainingLife := k.ExpiresAt.Time.Sub(now)
+	return false, &remainingLifeSpan{
+		remainingLifeSpanDays: remainingLife.Hours() / 24,
+		infinite:              false,
+	}
+}
+
+// CanBeSuspended returns whether the kafka instance can be suspended or not
+func (k *KafkaRequest) CanBeSuspended() bool {
```
miguelsorianod, line 199: Maybe we should rename this to `CanBeAutomaticallySuspended` to differentiate it with a manual suspension? The set of states would be different

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -141,3 +143,67 @@ func (k *KafkaRequest) GetExpirationTime(lifespanSeconds int) *time.Time {
...
+	validSuspensionStatuses := []constants.KafkaStatus{
+		constants.KafkaRequestStatusAccepted,
+		constants.KafkaRequestStatusPreparing,
+		constants.KafkaRequestStatusReady,
+		constants.KafkaRequestStatusProvisioning,
+		constants.KafkaRequestStatusResuming,
+	}
+
+	return arrays.AnyMatch(validSuspensionStatuses, func(x constants.KafkaStatus) bool { return x.String() == k.Status })
```
miguelsorianod, line 208: An alternative would be using Contains but that would require validSuspensionStatuses being a []string.
Both approaches look good to me

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -141,3 +143,67 @@ func (k *KafkaRequest) GetExpirationTime(lifespanSeconds int) *time.Time {
...
+
+	now := time.Now()
+	if now.After(k.ExpiresAt.Time) {
+		return true, &remainingLifeSpan{
+			remainingLifeSpanDays: 0,
+			infinite:              false,
+		}
+	}
+
+	remainingLife := k.ExpiresAt.Time.Sub(now)
```
miguelsorianod, line 191: How does `Sub` behaves if expires_at is earlier than now?

For example:
What happens if it expires 20 october and now is 22 october
What happens if it expires expires 20 october and now is 10 october

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: The residual life span object has no concept of `grace period`: it just returns the lifespan of the Kafka instance.
The concept of 'grace period' exists in the caller. However I could add a `LessThanOrEqual(days int) bool` method and make the code:
```go
residualLifespan.LessThanOrEqual(bm.GracePeriodDays)
```

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: Can you elaborate a bit more? I'm not sure I get what you mean.

internal/kafka/internal/api/dbapi/kafka_request_types.go
ziccardi: Returns a negative result: https://go.dev/play/p/I1oUTnIDqep

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
machi1990: The loop seems to be treating all the Kafkas the same irrespective of their instance type. 
What I am asking is, would it be possible (and acceptable?) that a developer cluster that's going to expire (in a few nanoseconds i.e just after line 129 is evaluated) to be have the status of `suspending`?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: done

machi1990: Thanks

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: The developer instance has a grace period set to 0 (zero). It won't enter the suspending state: it will be deleted immediately upon expiration.

machi1990: Thank you.

internal/kafka/internal/api/dbapi/kafka_request_types.go
miguelsorianod: Then what happens if it goes late by 1 day? The result would be -1 which is considered infinite :thinking:

ziccardi: -1 is not considered infinite. Infinite is when `Infinite` returns `true` and that happens only when `ExpiredAt` is `nil`

miguelsorianod: In the method documentation the following is said:
```
	// Days returns the number of remaining life span days a Kafka instance has. If infinite, returns -1
```

It is not considered infinite but -1 is returned if it is infinite.
Maybe we could add that it also can return negative number when it is not infinite meaning the days after the expiration that have elapsed?

ziccardi: Documentation updated!

miguelsorianod: :+1:

miguelsorianod:
Could you add/document the verifications needed with their steps so it can be reliably verified?

If it is possible in the verification to avoid external systems it would be even better (like for example not needing AMS)

Thanks

ziccardi:
@miguelsorianod done!

miguelsorianod:
When verifying the issue I wasn't able to easily find what was happening on KFM.
Could we add some logging?

miguelsorianod:
Following test done:

1. Create a standard kafka instance with standard billing model
2. Verify it is not marked as suspended
3. Set an expires_at time for the kafka instance 9 days in the future
4. Verify that the kafka instance is not suspended (as grace_period is not set for standard kafka instance)
5. Set grace period days setting for standard kafka instances with standard billing model with a value of 2 days
6. Verify that the kafka instance is not suspended (as grace_period has still not been reached)
7. Set grace period days setting for standard kafka instances with standard billing model with a value of 10 days
8. Verify that the kafka instance is set as suspended (as the instance has reached the grace_period)

ziccardi:
@miguelsorianod log added

PR #1482, opened 2023-01-11
feat(clusters): capture whether Kafkas have to be accessed via private network on enterprise clusters

machi1990:
## Description

capture whether Kafkas have to be accessed via private network on enterprise clusters

Closes https://issues.redhat.com/browse/MGDSTRM-10286

## Verification Steps

Unit and integration tests have to pass
 
- [ ] Code Review completed


miguelsorianod:
Could you explain what do we mean with `have to be accessed via private network`?

Does it mean that the Data Plane Cluster general endpoints (dns, api server, ingress controller) are not exposed to the internet?
Does it mean that the Pods corresponding to the kafka instances use internal addressing to expose its services?
Or does it have another meaning?

machi1990:
> > > Could you explain what do we mean with `have to be accessed via private network`?
> > > Does it mean that the Data Plane Cluster general endpoints (dns, api server, ingress controller) are not exposed to the internet? Does it mean that the Pods corresponding to the kafka instances use internal addressing to expose its services? Or does it have another meaning?
> > 
> > 
> > Yes. I'll send the corresponding epic brief via chat to you which contains a summary of the architecture design on the data plane side.
> 
> I am asking this questions because I think we should be accurate with the naming on our API side and code. `private` Might be too generic depending on the meaning

@miguelsorianod  thanks for the review. 

`private` field is only returned for the MKAgent CR, and it is the name that the data plane have implemented in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/840 , adding in @shawkins for additional thoughts regarding the naming choice.

For public facing when a user registers a cluster, the proposal in the PR is to use `access_kafkas_via_private_network`. Let me know what you think about that naming.

shawkins:
> adding in @shawkins for additional thoughts regarding the naming choice

We may eventually expand on the private option in platform specific ways similar to confluent cloud and make the choice of private vs public options per ManagedKafka - however none of that is currently supported by the data plane.  So for now this private flag means that the kafka instance endpoints only accessible from within the vpc / private network.

openapi/kas-fleet-manager-private.yaml
```diff
@@ -483,6 +483,15 @@ components:
           description: "Data plane cluster agent spec"
           type: object
           properties:
```
machi1990, line 485: @shawkins this is the spec change to the managedkafkaagent. It should corresponds to what's in  https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/840 but I'd like to have another pair of eyes on this.

openapi/kas-fleet-manager.yaml
```diff
@@ -1789,8 +1794,12 @@ components:
         - cluster_external_id
         - cluster_ingress_dns_name
         - kafka_machine_pool_node_count
+        - access_kafkas_via_private_network
       type: object
       properties:
+        access_kafkas_via_private_network:
```
machi1990, line 1800: @pawelpaszki @pb82 this is the field that's going to be exposed to the user. Let me know what you think of the naming.

/cc @dimakis this might be another information that's need to be supplied by the user and sent to KFM when a cluster is being registered.

openapi/kas-fleet-manager-private.yaml
shawkins: Yes that matches.

openapi/kas-fleet-manager.yaml
dimakis: @machi1990 is it just this value that needs to be returned? does this effect the CLI in any other way? 🤔

machi1990: @dimakis once this is in, it is a value that will need to be asked to the user, just like the machine pool node count and then passed in as a request payload field when registering an enterprise cluster in KFM. 

Additionally, this value will be returned in the clusters list endpoint, but you do not need to worry about that.

dimakis: That's perfect, @machi1990, thank you for the clarification. I'll see the changes in the sdk once it goes into kfm anyway. I'll know to include it then! 👍

internal/kafka/internal/services/clusters.go
```diff
@@ -32,7 +32,7 @@ type ClusterService interface {
 	Create(cluster *api.Cluster) (*api.Cluster, *apiErrors.ServiceError)
 	GetClusterDNS(clusterID string) (string, *apiErrors.ServiceError)
 	GetExternalID(clusterID string) (string, *apiErrors.ServiceError)
-	// ListEnterpriseClustersOfAnOrganization - returns a list of enterprise clusters (ClusterID and Status fields only) which belong to organization obtained from the context
+	// ListEnterpriseClustersOfAnOrganization - returns a list of enterprise clusters (ClusterID, AccessKafkasViaPrivateNetwork and Status fields only) which belong to organization obtained from the context
```
ziccardi, line 35: The `-` should be removed to comply with godoc standards.
```suggestion
	// ListEnterpriseClustersOfAnOrganization returns a list of enterprise clusters (ClusterID, AccessKafkasViaPrivateNetwork and Status fields only) which belong to organization obtained from the context
```

internal/kafka/internal/services/clusters.go
```diff
@@ -108,7 +108,7 @@ func (c clusterService) RegisterClusterJob(clusterRequest *api.Cluster) *apiErro
 	return nil
 }
 
-// ListEnterpriseClustersOfAnOrganization - returns a list of clusters (ClusterID and Status fields only) which belong to organization obtained from the context
+// ListEnterpriseClustersOfAnOrganization - returns a list of clusters (ClusterID, AccessKafkasViaPrivateNetwork and Status fields only) which belong to organization obtained from the context
```
ziccardi, line 111:
```suggestion
// ListEnterpriseClustersOfAnOrganization returns a list of clusters (ClusterID, AccessKafkasViaPrivateNetwork and Status fields only) which belong to organization obtained from the context
```

openapi/kas-fleet-manager.yaml
machi1990: Thanks.

internal/kafka/internal/services/clusters.go
machi1990: thanks

internal/kafka/internal/services/clusters.go
machi1990: thanks

openapi/kas-fleet-manager.yaml
pawelpaszki: @machi1990 - I think the name is fine

pawelpaszki (approved):

PR #1483, opened 2023-01-11
feat: set used/max metrics for enterprise dataplane clusters

pawelpaszki:
## Description
feat: set used/max metrics for enterprise dataplane clusters

- [x] Code Review completed
- [x] Required metrics/dashboards/alerts have been added (or PR created).

### DIFF
--- a/config/dataplane-cluster-configuration.yaml
+++ b/config/dataplane-cluster-configuration.yaml
@@ -1,8 +1,8 @@
 ---
 #- A list of clusters for kas fleet manager
 #- The `cluster_id` field can not be empty
-#- All clusters in kas fleet manager DB already but are missing in the list will be marked as
-#  'deprovisioning' and will later be deleted.
+#- All non-enterprise (cluster_type != "enterprise") clusters in kas fleet manager DB already but are missing in the list will be marked as
+#  'deprovisioning' and will later be deleted. Enterprise clusters are not included in handling this configuration
 #- This list is ordered, any new cluster should be appended at the end.
 #e.g.:
 #clusters:
--- a/internal/kafka/internal/services/cluster_placement_strategy.go
+++ b/internal/kafka/internal/services/cluster_placement_strategy.go
@@ -165,17 +165,24 @@ func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cl
 		return nil, errors.Wrapf(getInstanceSizeErr, "failed to get kafka instance size for cluster with criteria '%v'", criteria)
 	}
 
-	// Find first ready cluster that has remaining capacity (total streaming unit used by existing and requested kafka is within the cluster maxUnit limit)
 	for _, cluster := range clusters {
-		currentStreamingUnitsUsed := streamingUnitCountPerRegionList.GetStreamingUnitCountForClusterAndInstanceType(cluster.ClusterID, kafka.InstanceType)
-		capacityInfo := cluster.RetrieveDynamicCapacityInfo()
-		maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
-
-		if currentStreamingUnitsUsed+instanceSize.CapacityConsumed <= int(maxStreamingUnits) {
-			return cluster, nil
+		if cluster.ClusterType == api.ManagedDataPlaneClusterType.String() {
+			availableCluster := f.isManagedClusterNotFull(cluster, streamingUnitCountPerRegionList, kafka, instanceSize)
+			if availableCluster {
+				return cluster, nil
+			}
 		}
 	}
 
 	// no cluster found
 	return nil, nil
+}
+
+func (f *FirstReadyWithCapacity) isManagedClusterNotFull(cluster *api.Cluster, streamingUnitCountPerRegionList KafkaStreamingUnitCountPerClusterList,
+	kafka *dbapi.KafkaRequest, instanceSize *config.KafkaInstanceSize) bool {
+	currentStreamingUnitsUsed := streamingUnitCountPerRegionList.GetStreamingUnitCountForClusterAndInstanceType(cluster.ClusterID, kafka.InstanceType)
+	capacityInfo := cluster.RetrieveDynamicCapacityInfo()
+	maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
+
+	return currentStreamingUnitsUsed+instanceSize.CapacityConsumed <= int(maxStreamingUnits)
 }
--- a/internal/kafka/internal/services/cluster_placement_strategy.go
+++ b/internal/kafka/internal/services/cluster_placement_strategy.go
@@ -167,8 +167,8 @@ func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cl
 
 	for _, cluster := range clusters {
 		if cluster.ClusterType == api.ManagedDataPlaneClusterType.String() {
-			availableCluster := f.isManagedClusterNotFull(cluster, streamingUnitCountPerRegionList, kafka, instanceSize)
-			if availableCluster {
+			clusterNotFull := f.isManagedClusterNotFull(cluster, streamingUnitCountPerRegionList, kafka, instanceSize)
+			if clusterNotFull {
 				return cluster, nil
 			}
 		}
--- a/internal/kafka/internal/services/clusters.go
+++ b/internal/kafka/internal/services/clusters.go
@@ -712,6 +712,7 @@ type KafkaStreamingUnitCountPerCluster struct {
 	CloudProvider string
 	MaxUnits      int32
 	Status        string
+	ClusterType   string
 }
 
 func (k KafkaStreamingUnitCountPerCluster) isSame(kafkaPerRegionFromDB *KafkaPerClusterCount) bool {
@@ -763,6 +764,7 @@ type ClusterSelection struct {
 	SupportedInstanceType string
 	DynamicCapacityInfo   api.JSON
 	Status                string
+	ClusterType           string
 }
 
 func (c *clusterService) FindStreamingUnitCountByClusterAndInstanceType() (KafkaStreamingUnitCountPerClusterList, error) {
@@ -770,8 +772,7 @@ func (c *clusterService) FindStreamingUnitCountByClusterAndInstanceType() (Kafka
 	var clusters []*ClusterSelection
 	dbConn := c.connectionFactory.New().
 		Model(&api.Cluster{}).
-		Where("status != ?", api.ClusterFailed).
-		Where("cluster_type != ?", api.EnterpriseDataPlaneClusterType.String())
+		Where("status != ?", api.ClusterFailed)
 
 	if err := dbConn.Scan(&clusters).Error; err != nil {
 		return nil, errors.Wrap(err, "failed to list data plane clusters")
@@ -808,6 +809,7 @@ func (c *clusterService) FindStreamingUnitCountByClusterAndInstanceType() (Kafka
 				Count:         0,
 				MaxUnits:      maxUnits,
 				Status:        clusterSelection.Status,
+				ClusterType:   clusterSelection.ClusterType,
 			})
 		}
 	}
--- a/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
+++ b/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
@@ -163,6 +163,9 @@ func (m *DynamicScaleDownManager) createAMapOfProcessedClusters(kafkaStreamingUn
 	var processedClusters map[string]processed = map[string]processed{}
 
 	for i, kafkaStreamingUnitCountPerCluster := range kafkaStreamingUnitCountPerClusterList {
+		if kafkaStreamingUnitCountPerCluster.ClusterType != api.ManagedDataPlaneClusterType.String() {
+			continue
+		}
 		clusterID := kafkaStreamingUnitCountPerCluster.ClusterId
 		existing, ok := processedClusters[clusterID]
 
@@ -328,6 +331,7 @@ func (p *standardDynamicScaleDownProcessor) isScaleUpNeededAfterCandidateCluster
 			provider:         suCount.CloudProvider,
 			region:           suCount.Region,
 			instanceTypeName: suCount.InstanceType,
+			clusterType:      api.ManagedDataPlaneClusterType.String(),
 		}
 
 		instanceTypeConfig, ok := p.regionsSupportedInstanceType[suCount.InstanceType]
--- a/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
+++ b/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
@@ -77,6 +77,8 @@ func (m *DynamicScaleUpManager) Reconcile() []error {
 	return errList.ToErrorSlice()
 }
 
+// Note. "enterprise" clusters are not taken into consideration when
+// scaling up/ down with dynamic scaling turned on
 func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 	var errList fleeterrors.ErrorList
 	kafkaStreamingUnitCountPerClusterList, err := m.ClusterService.FindStreamingUnitCountByClusterAndInstanceType()
@@ -92,6 +94,7 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 					provider:         provider.Name,
 					region:           region.Name,
 					instanceTypeName: supportedInstanceTypeName,
+					clusterType:      api.ManagedDataPlaneClusterType.String(),
 				}
 				supportedInstanceTypeConfig := region.SupportedInstanceTypes[supportedInstanceTypeName]
 				var dynamicScaleUpProcessor dynamicScaleUpProcessor = &standardDynamicScaleUpProcessor{
@@ -129,17 +132,19 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 
 // supportedInstanceTypeLocator is a data structure
 // that contains all the information to help locate a
-// supported instance type in a region's cluster
+// supported instance type in a region's cluster of a given cluster type
 type supportedInstanceTypeLocator struct {
 	provider         string
 	region           string
 	instanceTypeName string
+	clusterType      string
 }
 
 func (l *supportedInstanceTypeLocator) Equal(other supportedInstanceTypeLocator) bool {
 	return l.provider == other.provider &&
 		l.region == other.region &&
-		l.instanceTypeName == other.instanceTypeName
+		l.instanceTypeName == other.instanceTypeName &&
+		l.clusterType == other.clusterType
 }
 
 // dynamicScaleUpExecutor is able to perform dynamic ScaleUp execution actions
@@ -364,7 +369,10 @@ type instanceTypeConsumptionSummaryCalculator struct {
 }
 
 // Calculate returns a instanceTypeConsumptionSummary containing a consumption
-// summary for the provided supportedInstanceTypeLocator
+// summary for the provided supportedInstanceTypeLocator.
+// NOTE - this function will only take into account clusters whose locator matches
+// specified supportedInstanceTypeLocator
+//
 // For the calculation of the max streaming units capacity:
 //   - Clusters in deprovisioning and cleanup state are excluded, as
 //     clusters into those states don't accept kafka instances anymore.
@@ -398,6 +406,7 @@ func (i *instanceTypeConsumptionSummaryCalculator) Calculate() (instanceTypeCons
 			provider:         kafkaStreamingUnitCountPerCluster.CloudProvider,
 			region:           kafkaStreamingUnitCountPerCluster.Region,
 			instanceTypeName: kafkaStreamingUnitCountPerCluster.InstanceType,
+			clusterType:      kafkaStreamingUnitCountPerCluster.ClusterType,
 		}
 
 		if !i.locator.Equal(currLocator) {
--- a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
+++ b/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
@@ -96,7 +96,7 @@ func (k *KafkaManager) Reconcile() []error {
 		encounteredErrors = append(encounteredErrors, statusErrors...)
 	}
 
-	capacityError := k.setClusterStatusCapacityMetrics()
+	capacityError := k.updateClusterStatusCapacityMetrics()
 	if capacityError != nil {
 		encounteredErrors = append(encounteredErrors, capacityError)
 	}
@@ -147,7 +147,7 @@ func (k *KafkaManager) setKafkaStatusCountMetric() []error {
 	return nil
 }
 
-func (k *KafkaManager) setClusterStatusCapacityMetrics() error {
+func (k *KafkaManager) updateClusterStatusCapacityMetrics() error {
 	usedStreamingUnitsCountByRegion, err := k.clusterService.FindStreamingUnitCountByClusterAndInstanceType()
 	if err != nil {
 		return errors.Wrap(err, "failed to count Kafkas by region")
@@ -162,13 +162,13 @@ func (k *KafkaManager) setClusterStatusCapacityMetrics() error {
 
 	if k.dataplaneClusterConfig.IsDataPlaneAutoScalingEnabled() {
 		for _, usedStreamingUnitCount := range usedStreamingUnitsCountByRegion {
-			if usedStreamingUnitCount.Status != api.ClusterAccepted.String() {
+			if usedStreamingUnitCount.Status != api.ClusterAccepted.String() && usedStreamingUnitCount.ClusterType != api.EnterpriseDataPlaneClusterType.String() {
 				availableAndMaxCapacityCounts, err := k.calculateAvailableAndMaxCapacityForDynamicScaling(usedStreamingUnitCount)
 				if err != nil {
 					return err
 				}
-				k.setClusterStatusCapacityAvailableMetric(availableAndMaxCapacityCounts)
-				k.setClusterStatusCapacityMaxMetric(availableAndMaxCapacityCounts)
+				k.updateClusterStatusCapacityAvailableMetric(availableAndMaxCapacityCounts)
+				k.updateClusterStatusCapacityMaxMetric(availableAndMaxCapacityCounts)
 			}
 		}
 	}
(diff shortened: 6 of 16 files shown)


internal/kafka/internal/config/dataplane_cluster_config.go
```diff
@@ -143,17 +143,18 @@ func NewDataplaneClusterConfig() *DataplaneClusterConfig {
...
+	CloudProvider         string                   `yaml:"cloud_provider"`
+	Region                string                   `yaml:"region"`
+	MultiAZ               bool                     `yaml:"multi_az"`
+	Schedulable           bool                     `yaml:"schedulable"`
+	KafkaInstanceLimit    int                      `yaml:"kafka_instance_limit"`
+	Status                api.ClusterStatus        `yaml:"status"`
+	ProviderType          api.ClusterProviderType  `yaml:"provider_type"`
+	ClusterDNS            string                   `yaml:"cluster_dns"`
+	SupportedInstanceType string                   `yaml:"supported_instance_type"`
+	ClusterType           api.DataPlaneClusterType `yaml:"cluster_type"`
```
machi1990, line 157: We do not need to add it here

config/dataplane-cluster-configuration.yaml
```diff
@@ -17,4 +17,5 @@
 #    provider_type: "ocm" #Valid values are `ocm` and `standalone`. `ocm` will be used if not specified.
 #    cluster_dns: apps.example.com #Valid cluster DNS. This will be used to build kafka bootstrap url and to communicate with standalone clusters. Required when "provider_type" is "standalone" 
 #    supported_instance_type: "developer" # could be "developer", "standard" or both i.e "standard,developer" or "developer,standard". Defaults to "standard,developer" if not set 
+#    cluster_type: "managed" # could be "managed" or "enterprise"
```
machi1990, line 20:
```suggestion
```

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -163,19 +163,21 @@ func (m *DynamicScaleDownManager) createAMapOfProcessedClusters(kafkaStreamingUn
...
-			processedClusters[clusterID] = processed{
-				// consider non ready cluster as already processed for scale down evaluation and thus they'll be skipped from scale down consideration
-				processed:                              kafkaStreamingUnitCountPerCluster.Status != api.ClusterReady.String(),
-				indexesOfStreamingUnitForSameClusterID: []int{i},
-			}
-		} else {
-			processedClusters[clusterID] = processed{
-				indexesOfStreamingUnitForSameClusterID: append(existing.indexesOfStreamingUnitForSameClusterID, i),
-				processed:                              existing.processed,
+		if kafkaStreamingUnitCountPerCluster.ClusterType != api.EnterpriseDataPlaneClusterType.String() {
```
machi1990, line 166:
```suggestion
		if kafkaStreamingUnitCountPerCluster.ClusterType == api.ManagedDataPlaneClusterType.String() {
```

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
```diff
@@ -394,35 +394,37 @@ func (i *instanceTypeConsumptionSummaryCalculator) Calculate() (instanceTypeCons
 	clusterStatesTowardDeletion := []string{api.ClusterDeprovisioning.String(), api.ClusterCleanup.String()}
 
 	for _, kafkaStreamingUnitCountPerCluster := range i.kafkaStreamingUnitCountPerClusterList {
-		currLocator := supportedInstanceTypeLocator{
-			provider:         kafkaStreamingUnitCountPerCluster.CloudProvider,
-			region:           kafkaStreamingUnitCountPerCluster.Region,
-			instanceTypeName: kafkaStreamingUnitCountPerCluster.InstanceType,
-		}
+		if kafkaStreamingUnitCountPerCluster.ClusterType != api.EnterpriseDataPlaneClusterType.String() {
```
machi1990, line 397: I was wondering if we could build this onto the locator.

in line https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/11d2609bfca4435ded886e1899c68e438a96bae6/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go#L91 we could construct a locator that knows how to find `managed` clusters.

```go
currLocator := supportedInstanceTypeLocator{
					provider:         provider.Name,
					region:           region.Name,
					instanceTypeName: supportedInstanceTypeName,
clusterType: "managed",
				}
```
Then in this method at line number 398 we could have
 ```go
 currLocator := supportedInstanceTypeLocator{
				provider:         kafkaStreamingUnitCountPerCluster.CloudProvider,
				region:           kafkaStreamingUnitCountPerCluster.Region,
				instanceTypeName: kafkaStreamingUnitCountPerCluster.InstanceType,
                               clusterType: kafkaStreamingUnitCountPerCluster.ClusterType
			}
```			

And then if we enhance https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/11d2609bfca4435ded886e1899c68e438a96bae6/internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go#L139 to also checks for the `clusterType`, it'll do the job and we remove this level of nesting introduced here while keep the code in the same original spirit.

File: internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go [hunk 5]
```diff
@@ -207,6 +207,11 @@ func (k *KafkaManager) calculateAvailableAndMaxCapacityForDynamicScaling(streami
 	// of Kafkas have already been created. In this case we limit the metric to 0 anyways.
 	available := math.Max(0, maxAvailable-float64(streamingUnitsByRegion.Count))
 
+	if streamingUnitsByRegion.ClusterType == api.EnterpriseDataPlaneClusterType.String() {
+		maxAvailable = float64(streamingUnitsByRegion.MaxUnits)
+		available = maxAvailable - float64(streamingUnitsByRegion.MaxUnits)
+	}
```
machi1990, line 213: As discussed, when we encounter an enterprise cluster, we shouldn't compute the region limits at all: this is current happening in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/11d2609bfca4435ded886e1899c68e438a96bae6/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L210

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -273,6 +278,11 @@ func (k *KafkaManager) calculateCapacityByRegionAndInstanceTypeForManualClusters
 				availableByInstanceType = 0
 			}
 
+			if cluster.ClusterType == api.EnterpriseDataPlaneClusterType {
+				maxAvailable = float64(cluster.KafkaInstanceLimit)
+				availableByInstanceType = maxAvailable - float64(totalUsed)
+			}
```
machi1990, line 284: enterprise clusters will never be part of the manual list. They'll co-exists with clusters added in the manual list but they'll only be in the data base. This check being englobed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/11d2609bfca4435ded886e1899c68e438a96bae6/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L247 will mean that if we switch to manual scaling and we've enterprise clusters, then we'll never see metrics for enterprise clusters. 

I suggest we treat enterprise clusters outside the loop in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/11d2609bfca4435ded886e1899c68e438a96bae6/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L247 

It'll be another loop that goes through `streamingUnitsByRegion` and finds only enterprise clusters, treats them and appending them onto the `result` list.

pawelpaszki:
@machi1990 I think I addressed your comments

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -239,6 +244,20 @@ func (k *KafkaManager) calculateCapacityByRegionAndInstanceTypeForManualClusters
...
 	}
 
+	// enterprise data plane clusters
+	for _, cluster := range streamingUnitsByRegion {
+		if cluster.ClusterType == api.EnterpriseDataPlaneClusterType.String() {
+			result = append(result, services.KafkaStreamingUnitCountPerCluster{
+				Region:        cluster.Region,
+				InstanceType:  cluster.InstanceType,
+				ClusterId:     cluster.ClusterId,
+				Count:         int32(cluster.Count),
```
machi1990, line 254:
```suggestion
				Count:         int32(cluster.MaxUnits - cluster.Count),
```

I think this count is meant to represent `available` streaming unit Count.

machi1990: e.g here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/23d0d0031fd3e60848740dbd5cafad45afff232a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L299

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
```diff
@@ -92,6 +92,7 @@ func (m *DynamicScaleUpManager) processDynamicScaleUpReconcileEvent() error {
 					provider:         provider.Name,
 					region:           region.Name,
 					instanceTypeName: supportedInstanceTypeName,
+					clusterType:      api.ManagedDataPlaneClusterType.String(),
```
machi1990, line 95: might be good to update the unit test to cover for this line

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -207,6 +207,11 @@ func (k *KafkaManager) calculateAvailableAndMaxCapacityForDynamicScaling(streami
 	// of Kafkas have already been created. In this case we limit the metric to 0 anyways.
 	available := math.Max(0, maxAvailable-float64(streamingUnitsByRegion.Count))
 
+	if streamingUnitsByRegion.ClusterType == api.EnterpriseDataPlaneClusterType.String() {
+		maxAvailable = float64(streamingUnitsByRegion.MaxUnits)
+		available = maxAvailable - float64(streamingUnitsByRegion.Count)
+	}
```
machi1990, line 213: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1483/files#r1067884760

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -239,6 +244,20 @@ func (k *KafkaManager) calculateCapacityByRegionAndInstanceTypeForManualClusters
 		return totalUsed, instanceTypeUsed
 	}
 
+	// enterprise data plane clusters
+	for _, cluster := range streamingUnitsByRegion {
+		if cluster.ClusterType == api.EnterpriseDataPlaneClusterType.String() {
```
machi1990, line 249: same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1483/files#r1067884760

machi1990 (changes requested):
thanks @pawelpaszki 

The code change looks good to me. I left a few comments to update the unit tests + it'll be good to add an integration test for the enterprise cluster metrics.

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -167,12 +167,14 @@ func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cl
 
 	// Find first ready cluster that has remaining capacity (total streaming unit used by existing and requested kafka is within the cluster maxUnit limit)
 	for _, cluster := range clusters {
-		currentStreamingUnitsUsed := streamingUnitCountPerRegionList.GetStreamingUnitCountForClusterAndInstanceType(cluster.ClusterID, kafka.InstanceType)
-		capacityInfo := cluster.RetrieveDynamicCapacityInfo()
-		maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
-
-		if currentStreamingUnitsUsed+instanceSize.CapacityConsumed <= int(maxStreamingUnits) {
-			return cluster, nil
+		if cluster.ClusterType == api.ManagedDataPlaneClusterType.String() {
```
miguelsorianod, line 170: Could we refactor this logic to its own method?

internal/kafka/internal/services/clusters.go
```diff
@@ -712,6 +712,7 @@ type KafkaStreamingUnitCountPerCluster struct {
 	CloudProvider string
 	MaxUnits      int32
 	Status        string
+	ClusterType   string
```
miguelsorianod, line 715: How does this impact dynamic scaling evaluation? Are enterprise clusters explicitly excluded there?

internal/kafka/internal/services/clusters.go
```diff
@@ -763,15 +764,15 @@ type ClusterSelection struct {
 	SupportedInstanceType string
 	DynamicCapacityInfo   api.JSON
 	Status                string
+	ClusterType           string
```
miguelsorianod, line 767: Side comment: What is the ClusterType when a cluster is not enterprise?

internal/kafka/internal/services/clusters.go
miguelsorianod: edit: I see code in scaleup and scaledown mgr has been modified to take it into account

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
miguelsorianod, line 95: Could we update the documentation of the `supportedInstanceTypeLocator` type?

internal/kafka/internal/services/clusters.go
pawelpaszki: `managed`

internal/kafka/internal/services/clusters.go
pawelpaszki: it should not affect it

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_down_mgr.go
```diff
@@ -163,19 +163,21 @@ func (m *DynamicScaleDownManager) createAMapOfProcessedClusters(kafkaStreamingUn
...
-			processedClusters[clusterID] = processed{
-				// consider non ready cluster as already processed for scale down evaluation and thus they'll be skipped from scale down consideration
-				processed:                              kafkaStreamingUnitCountPerCluster.Status != api.ClusterReady.String(),
-				indexesOfStreamingUnitForSameClusterID: []int{i},
-			}
-		} else {
-			processedClusters[clusterID] = processed{
-				indexesOfStreamingUnitForSameClusterID: append(existing.indexesOfStreamingUnitForSameClusterID, i),
-				processed:                              existing.processed,
+		if kafkaStreamingUnitCountPerCluster.ClusterType == api.ManagedDataPlaneClusterType.String() {
```
miguelsorianod, line 166: To avoid extra indentation can we change the approach to this?:

 If clusterType is not managed then `continue`. Otherwise execute next statements in this iteration as normal

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -207,6 +207,11 @@ func (k *KafkaManager) calculateAvailableAndMaxCapacityForDynamicScaling(streami
 	// of Kafkas have already been created. In this case we limit the metric to 0 anyways.
 	available := math.Max(0, maxAvailable-float64(streamingUnitsByRegion.Count))
 
+	if streamingUnitsByRegion.ClusterType == api.EnterpriseDataPlaneClusterType.String() {
```
miguelsorianod, line 210: Could you explain / expand why this is needed?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -239,6 +244,20 @@ func (k *KafkaManager) calculateCapacityByRegionAndInstanceTypeForManualClusters
 		return totalUsed, instanceTypeUsed
 	}
 
+	// enterprise data plane clusters
```
miguelsorianod, line 247: Is an enterprise cluster considered a "manual" cluster? Maybe this is not the most appropriate place where to put this logic?

internal/kafka/internal/services/clusters.go
miguelsorianod: Is a manual cluster always managed?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
pawelpaszki: I'll move it out this function

internal/kafka/internal/services/clusters.go
pawelpaszki: right now we only have two types of clusters regardless of scaling policy (enterprise or non-enterprise (managed))

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
machi1990: @pawelpaszki we also do not need to execute https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/23d0d0031fd3e60848740dbd5cafad45afff232a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L196 if the cluster is enterprise as the region configuration shouldn't be taken onto account for these type of clusters

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
pawelpaszki: I believe because region limits shouldn't be taken into consideration for enterprise clusters capacity. @machi1990 - can you confirm?

machi1990: +1 that's correct.

Regions limits are meant to be for managed clusters i.e our own clusters.
For the enterprise case, it could well be possible to have a cluster in a region that KFM does not have configuration for in its supported cloud provider configuration.

internal/kafka/internal/services/clusters.go
machi1990: Yes. Clusters registered via the manual list are managed clusters i.e our own clusters and not user supplied clusters.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
miguelsorianod: Could we add a comment mentioning that?

miguelsorianod: Also does that mean that the region limits do not apply to enterprise clusters? If so, can we document it too in the config?

pawelpaszki:
@miguelsorianod @machi1990 - I think I addressed your comments. would you mind taking another look before I start adding/ changing tests please?

internal/kafka/internal/workers/cluster_mgrs/dynamic_scale_up_mgr.go
```diff
@@ -364,7 +367,10 @@ type instanceTypeConsumptionSummaryCalculator struct {
 }
 
 // Calculate returns a instanceTypeConsumptionSummary containing a consumption
-// summary for the provided supportedInstanceTypeLocator
+// summary for the provided supportedInstanceTypeLocator.
+// NOTE - only supportedInstanceTypeLocator instances with clusterType = "managed"
```
miguelsorianod, line 371: I think this note is not really accurate in the context of this method.

The `Calculate` method will only take into account clusters whose locator matches. It can be used for any kind of clusters. It happens to be that in the context of scaleup mgr the only locators we set set the cluster type to `managed`  but that's independent.

As I mentioned in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1483/files#r1067898699 I think we should update the documentation of the `supportedInstanceTypeLocator` type itself to document that now it includes the cluster type

internal/kafka/internal/services/cluster_placement_strategy.go
pb82: do you mean something like `cluster.IsEnterpriseType()`?

miguelsorianod: No. This code is inside a loop that processes N clusters each one individually. What I am referring is to extract the logic that deals with the cluster individually to a method that is called from the loop

internal/kafka/internal/services/clusters.go
```diff
@@ -763,15 +764,15 @@ type ClusterSelection struct {
...
 }
 
 func (c *clusterService) FindStreamingUnitCountByClusterAndInstanceType() (KafkaStreamingUnitCountPerClusterList, error) {
 
 	var clusters []*ClusterSelection
 	dbConn := c.connectionFactory.New().
 		Model(&api.Cluster{}).
-		Where("status != ?", api.ClusterFailed).
-		Where("cluster_type != ?", api.EnterpriseDataPlaneClusterType.String())
+		Where("status != ?", api.ClusterFailed)
```
ziccardi, line 775: Does this work? Usually `<>` is used in SQL (see here too: https://gorm.io/docs/query.html)

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -187,9 +187,34 @@ func (k *KafkaManager) setClusterStatusCapacityMetrics() error {
 		}
 	}
 
+	// enterprise clusters max and available is handled separately
+	enterpriseClustersAvailableAndMaxCapacityCounts := k.calculateAvailableAndMadCapacityForEnterpriseClusters(usedStreamingUnitsCountByRegion)
+	for _, availableAndMaxCapacityCounts := range enterpriseClustersAvailableAndMaxCapacityCounts {
+		k.setClusterStatusCapacityAvailableMetric(availableAndMaxCapacityCounts)
+		k.setClusterStatusCapacityMaxMetric(availableAndMaxCapacityCounts)
+	}
```
ziccardi, line 195: Is this correct? Are we interested only in the values of the latest element of the array?

internal/kafka/internal/services/clusters.go
pawelpaszki: it worked so far. I actually restored the original code in this PR

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
pawelpaszki: hi Massimo. can you clarify what you mean?

pb82: I believe the issue is that you're overwriting the values in every iteration of the loop. Only the last iteration's values are exported.

pb82: nvm. on second look the code is correct: the `setClusterStatusCapacityMetric` functions take an object which has the region, cloud provider, cluster_id etc.
No values will be overwritten.

ziccardi: You call `setClusterStatusCapacityAvailableMetric` for every element in the array.
Unless I'm missing something I'd say you are overwriting the values on each iteration

pawelpaszki: I don't think I am overwriting anything. the loop iterates over each enterprise cluster and sets metrics with the values of that enterprise cluster

ziccardi: If those values are 'added' or 'updated' based on some value inside the `availableAndMaxCapacityCounts`, then I believe the method names should be changed to something like `addClusterStatusCapacityAvailableMetric` and `addClusterStatusCapacityMaxMetric`.

Calling a `set` method multiple times makes me think you are updating the same thing multiple times (i.e., overwriting)

pawelpaszki: I get it now. renamed the functions

miguelsorianod: I think there is still some confusion @pawelpaszki @ziccardi 

the original name was `setClusterStatusCapacityAvailableMetric`. The new name is `sendClusterStatusCapacityAvailableMetric`.

What the method does is that it SETS the corresponding cluster status capacity available prometheus metric. If you dig into the code you will see that it is a Prometheus Gauge.
What it is being done is setting the gauge, not sending anything nor adding anything. Prometheus model is that Prometheus scrapes the endpoint. In this case the endpoint is KFM. In KFM we are setting the gauge but not sending it anywhere.
Therefore I think this should still be named setXXXX.

Aside from that, in each iteration we are setting the gauge. This is, overwriting the previous value. This might be intentional or not, depending on what is the meaning of that gauge and what we want to do.

This should be clarified.

miguelsorianod: I reviewed the code and the methods set the same metric name but with different labels so it would be correct

miguelsorianod (approved):

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: I wonder if maybe `update` metrics would be a better prefix than `set`  to make it clear that each call doesn't overwrite the previous one (not always).

pawelpaszki: @miguelsorianod - WDYT about Massimo's suggestion?

pawelpaszki:
> thanks @pawelpaszki
> 
> The code change looks good to me. I left a few comments to update the unit tests + it'll be good to add an integration test for the enterprise cluster metrics.

I added max and available metrics check to integration test. will add used metrics check along with the create enterprise kafka work

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
miguelsorianod: I am fine with both approaches

pawelpaszki: changed to `update`

internal/kafka/test/integration/cluster_list_enterprise_test.go
```diff
@@ -84,4 +90,10 @@ func TestEnterpriseClustersList(t *testing.T) {
 	if resp2 != nil {
 		defer resp2.Body.Close()
 	}
+
+	checkMetricsError := common.WaitForMetricToBePresent(h, t, metrics.ClusterStatusCapacityAvailable, "3", api.StandardTypeSupport.String(), cluster.ClusterID, cluster.Region, cluster.CloudProvider)
+	g.Expect(checkMetricsError).NotTo(gomega.HaveOccurred())
+
+	checkMetricsError = common.WaitForMetricToBePresent(h, t, metrics.ClusterStatusCapacityMax, "3", api.StandardTypeSupport.String(), cluster.ClusterID, cluster.Region, cluster.CloudProvider)
+	g.Expect(checkMetricsError).NotTo(gomega.HaveOccurred())
```
machi1990, line 98: @pawelpaszki we could add this check here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/46b36c569f59e89a45f0609bb2623dbe7b453ca6/internal/kafka/test/integration/cluster_metrics_test.go#L26

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -165,17 +165,24 @@ func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cl
...
-	// Find first ready cluster that has remaining capacity (total streaming unit used by existing and requested kafka is within the cluster maxUnit limit)
 	for _, cluster := range clusters {
-		currentStreamingUnitsUsed := streamingUnitCountPerRegionList.GetStreamingUnitCountForClusterAndInstanceType(cluster.ClusterID, kafka.InstanceType)
-		capacityInfo := cluster.RetrieveDynamicCapacityInfo()
-		maxStreamingUnits := capacityInfo[kafka.InstanceType].MaxUnits
-
-		if currentStreamingUnitsUsed+instanceSize.CapacityConsumed <= int(maxStreamingUnits) {
-			return cluster, nil
+		if cluster.ClusterType == api.ManagedDataPlaneClusterType.String() {
+			availableCluster := f.isManagedClusterNotFull(cluster, streamingUnitCountPerRegionList, kafka, instanceSize)
```
machi1990, line 170:
```suggestion
			clusterNotFull := f.isManagedClusterNotFull(cluster, streamingUnitCountPerRegionList, kafka, instanceSize)
```

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -165,17 +165,24 @@ func (f *FirstReadyWithCapacity) FindCluster(kafka *dbapi.KafkaRequest) (*api.Cl
...
+			if availableCluster {
```
machi1990, line 171:
```suggestion
			if clusterNotFull {
```

machi1990 (commented):
Thank @pawelpaszki 

left a comment on integration test location and one minor suggestion

internal/kafka/test/integration/cluster_list_enterprise_test.go
machi1990: Thanks. We can delete the check from here as well as it is duplicated.

machi1990 (approved):
left one nitpick, otherwise the change looks good.

Thanks @pawelpaszki

PR #1486, opened 2023-01-12
Set expires_at on kafka creation when lifespanSeconds configuration is defined

miguelsorianod:
## Description

Related to https://issues.redhat.com/browse/MGDSTRM-10249.

As part of the design of the long lived eval instances functionality and as implemented in https://issues.redhat.com/browse/MGDSTRM-10012 it was decided that the logic that verifies whether a kafka instance should be marked for provisioning due to reaching its expiration date should make use of the `expires_at` value in the DB. This applies to all instance types including developer instances, where up until now had calculated its expiration time dynamically from the config and not being persisted in the database.

This PR adds logic to set the expires_at DB field to created_at + configured lifespanSeconds when creating kafka instances whose corresponding instance size has a configured lifespanSeconds set. 

TODO:

- [x] Update tests

## Verification Steps

1. Tests pass
1. Verify that when creating registering kafka instances the created_at and updated_at values are the same always
1. Verify that when creating a kafka instance whose size has lifespanSeconds configured in the instances configuration file then its expires_at value is set to created_at + lifespanSeconds in the DB
1. Verify that when creating a kafka instance whose size doesn't have lifespanSeconds set in the instances configuration file then its expires_at value is set to null in the DB

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -427,6 +427,18 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
 
 	kafkaRequest.KafkaStorageSize = size.MaxDataRetentionSize.String()
 
+	// We intentionally manually set CreatedAt and UpdatedAt instead of letting
+	// gorm do it. The reason for that is that otherwise the ExpiresAt value
+	// would be calculated from a start date potentially earlier than the CreatedAt
+	// time a different value than those. An alternative would be performing
+	// two different database updates but it would be less performant
+	timeNow := dbConn.NowFunc()
+	kafkaRequest.CreatedAt = timeNow
+	kafkaRequest.UpdatedAt = timeNow
+	if size.LifespanSeconds != nil {
+		kafkaRequest.ExpiresAt = sql.NullTime{Time: timeNow.Add(time.Duration(*size.LifespanSeconds) * time.Second), Valid: true}
+	}
+
 	// Persist the QuotaTyoe to be able to dynamically pick the right Quota service implementation even on restarts.
 	// A typical usecase is when a kafka A is created, at the time of creation the quota-type was ams. At some point in the future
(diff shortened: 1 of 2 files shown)


internal/kafka/internal/services/kafka.go
```diff
@@ -427,6 +427,22 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
 
 	kafkaRequest.KafkaStorageSize = size.MaxDataRetentionSize.String()
 
+	// We intentionally manually set CreatedAt and UpdatedAt instead of letting
```
miguelsorianod, line 430: See all this comments and the TODO comment

internal/kafka/internal/services/kafka.go
```diff
@@ -427,6 +427,22 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
...
+	timeNow := dbConn.NowFunc()
+	// TODO should we manually set CreatedAt and UpdatedAt or just
+	// carry with the implication that ExpiresAt would be not calculated
+	// from the exact creation time but from a time potentially earlier than the
+	// creation time?
+	kafkaRequest.CreatedAt = timeNow
+	kafkaRequest.UpdatedAt = timeNow
+	if size.LifespanSeconds != nil {
+		kafkaRequest.ExpiresAt = sql.NullTime{Time: timeNow.Add(time.Duration(*size.LifespanSeconds) * time.Second), Valid: true}
+	}
```
ziccardi, line 444: Whether we do it ourselves or let gorm do it, the difference will be milliseconds (or even less): it would be no harm.

I don't see any harm in this approach too. I've no preference for the two methods.

miguelsorianod: :+1: I will leave it as it is now. This is, manually setting the createdat and updated at so they are consistent

miguelsorianod:
Updated tests.

This is ready for final review and verifications

ziccardi (approved):
LGTM

PR #1499, opened 2023-01-18
feat: reconcile expires at value for kafka instances

JameelB:
## Description
The following changes adds logic to reconcile the `expires_at` value of Kafka instances in KFM.

The `expires_at` value determines when a Kafka instance is to expire, marked for deletion. It can either contain a timestamp, of when the Kafka instance will be marked for deletion, or null which means the Kafka instance will never expire.

The value is determined whether LifespanSeconds in the Kafka instance types configuration is defined for that instance type. If LifespanSeconds is defined, the value should be calculated with `Kafka.CreatedAt + LifespanSeconds`. 

If it's not defined, then the value is based on whether the organisation/user has an active subscription (SKU allowance given to the org) for the particular billing model used by the Kafka instance. We can't get information on the subscriptions status directly from AMS. Instead we can take a look at the quota allowed and consumed for that SKU. When a subscription in AMS expires, the allowed value is reset to 0. If the `Consumed` value is greater than the `Allowed` value, that indicates that the subscription is no longer active/is expired. Once the subscription expires, the `expires_at` value should be set and calculated with `Current Time + Grace Period Days`.

There is also a chance that the subscription is extended. In this case, `Allowed` is set to a non-zero value. In this case, we need to ensure that any Kafka instances using that subscription has their `expires_at` set to null.

The subscription check in AMS should always be based on the `actual_kafka_billing_model`. If a user/organisation decide to purchase a paid subscription after their evaluation period they will be given allowance for a new SKU. Kafka instance will remain in the `eval` billing model until the user promotes it to use a different billing model. The `expires_at` value will not be changed to null until its promoted successfully. Any Kafka instances that were not promoted before the expiration time will be deleted.

Related JIRA Issue: https://issues.redhat.com/browse/MGDSTRM-10201

## Verification Steps
<TODO>

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafka.go
+++ b/internal/kafka/internal/services/kafka.go
@@ -129,6 +129,12 @@ type KafkaService interface {
 	GetAvailableSizesInRegion(criteria *FindClusterCriteria) ([]string, *errors.ServiceError)
 	ValidateBillingAccount(externalId string, instanceType types.KafkaInstanceType, kafkaBillingModelID string, billingCloudAccountId string, marketplace *string) *errors.ServiceError
 	AssignBootstrapServerHost(kafkaRequest *dbapi.KafkaRequest) error
+	// IsQuotaEntitlementActive checks if the user/organisation have an active entitlement to the quota
+	// used by the given Kafka instance.
+	//
+	// It returns true if the user has an active quota entitlement and false if not.
+	// It returns false and an error if it encounters any issues while trying to check the quota entitlement status.
+	IsQuotaEntitlementActive(kafkaRequest *dbapi.KafkaRequest) (bool, error)
 
 	// MGDSTRM-10012 temporarily add method to reconcile updating the zero-value
 	// of ExpiredAt. Remove this method when functionality has been rolled out
@@ -1362,4 +1368,13 @@ func (k *kafkaService) UpdateZeroValueOfKafkaRequestsExpiredAt() error {
 	glog.Infof("%v kafka_requests had expires_at with the zero-value of time.Time '%v' and have been updated to NULL", db.RowsAffected, zeroTime)
 
 	return db.Error
+}
+
+func (k *kafkaService) IsQuotaEntitlementActive(kafkaRequest *dbapi.KafkaRequest) (bool, error) {
+	quotaService, factoryErr := k.quotaServiceFactory.GetQuotaService(api.QuotaType(k.kafkaConfig.Quota.Type))
+	if factoryErr != nil {
+		return false, factoryErr
+	}
+
+	return quotaService.IsQuotaEntitlementActive(kafkaRequest)
 }
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -71,8 +71,8 @@ func (q amsQuotaService) newBaseQuotaReservedResourceBuilder(kafka *dbapi.KafkaR
 // with an AMS ReservedResource. The set of Billing models shown/accepted is
 // different between them
 var supportedAMSRelatedResourceBillingModels = map[string]struct{}{
-	string(amsv1.BillingModelMarketplace): {},
-	string(amsv1.BillingModelStandard):    {},
+	ocm.AMSRelatedResourceBillingModelMarketplace.String(): {},
+	ocm.AMSRelatedResourceBillingModelStandard.String():    {},
 }
 
 // checks if the requested billing model (pre-paid vs. consumption based) matches with the model returned from AMS
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -2,8 +2,9 @@ package quota
 
 import (
 	"fmt"
-	v1 "github.com/openshift-online/ocm-sdk-go/authorizations/v1"
 	"net/http"
+
+	v1 "github.com/openshift-online/ocm-sdk-go/authorizations/v1"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/cloudproviders"
@@ -13,6 +14,7 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services/quota/internal/utils"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/ocm"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
@@ -395,4 +397,97 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
 	}
 
 	return resp.Body(), true, nil
+}
+
+// Checks if an organisation has a SKU entitlement and that entitlement is still active in AMS
+func (q amsQuotaService) IsQuotaEntitlementActive(kafka *dbapi.KafkaRequest) (bool, error) {
+	kafkaBillingModel, err := q.kafkaConfig.GetBillingModelByID(kafka.InstanceType, kafka.ActualKafkaBillingModel)
+	if err != nil {
+		return false, err
+	}
+
+	amsRelatedResourceBillingModel, err := q.getAMSRelatedResourceBillingModel(kafka.SubscriptionId, kafkaBillingModel)
+	if err != nil || shared.StringEmpty(amsRelatedResourceBillingModel) {
+		return false, err
+	}
+
+	orgID, err := q.amsClient.GetOrganisationIdFromExternalId(kafka.OrganisationId)
+	if err != nil {
+		return false, err
+	}
+
+	quotaCostRelatedResourceFilter := ocm.QuotaCostRelatedResourceFilter{
+		ResourceName: &kafkaBillingModel.AMSResource,
+		Product:      &kafkaBillingModel.AMSProduct,
+		BillingModel: &amsRelatedResourceBillingModel,
+	}
+	quotaCosts, err := q.amsClient.GetQuotaCosts(orgID, true, false, quotaCostRelatedResourceFilter)
+	if err != nil {
+		return false, err
+	}
+
+	// SKU not entitled to the organisation
+	if len(quotaCosts) == 0 {
+		logger.Logger.Infof("ams quota cost not found for Kafka instance %q in organisation %q with the following filter: {ResourceName: %q, Product: %q, BillingModel: %q}",
+			kafka.ID, orgID, *quotaCostRelatedResourceFilter.ResourceName, *quotaCostRelatedResourceFilter.Product, *quotaCostRelatedResourceFilter.BillingModel)
+		return false, nil
+	}
+
+	if len(quotaCosts) > 1 {
+		return false, fmt.Errorf("more than 1 quota cost was returned for organisation %q with the following filter: {ResourceName: %q, Product: %q, BillingModel: %q}",
+			orgID, *quotaCostRelatedResourceFilter.ResourceName, *quotaCostRelatedResourceFilter.Product, *quotaCostRelatedResourceFilter.BillingModel)
+	}
+
+	// result will always return one quota for the given org, ams related resource billing model, resource and product for rhosak quotas
+	quotaCost := quotaCosts[0]
+
+	// when a SKU entitlement expires in AMS, the allowed value for that quota cost is set back to 0
+	// if the allowed value is 0 and consumed is greater than this, that denotes that the SKU entitlement
+	// has expired and is no longer active.
+	if quotaCost.Consumed() > quotaCost.Allowed() {
+		if quotaCost.Allowed() == 0 {
+			logger.Logger.Infof("quota no longer entitled for organisation %q (quotaid: %q, consumed: %q, allowed: %q)",
+				orgID, quotaCost.QuotaID(), quotaCost.Consumed(), quotaCost.Allowed())
+			return false, nil
+		}
+		logger.Logger.Warningf("Organisation %q has exceeded their quota allowance (quotaid: %q, consumed %q, allowed: %q",
+			orgID, quotaCost.QuotaID(), quotaCost.Consumed(), quotaCost.Allowed())
+	}
+	return true, nil
+}
+
+// Determine the AMS QuotaCost related resource billing model from the AMS Subscription reserved resource billing model
+// based on the given subscription id and kafka billing model.
+func (q amsQuotaService) getAMSRelatedResourceBillingModel(subscriptionID string, kafkaBillingModel config.KafkaBillingModel) (string, error) {
+	reservedResources, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscriptionID)
+	if err != nil {
+		return "", err
+	}
+
+	for _, rr := range reservedResources {
+		if kafkaBillingModel.AMSResource == rr.ResourceName() {
+			amsBillingModel := string(rr.BillingModel())
+			switch true {
+			case shared.StringEqualsIgnoreCase(amsBillingModel, ocm.AMSRelatedResourceBillingModelMarketplace.String()):
+				// legacy ams billing model marketplace, this will be removed in the future and replaced with cloud specific marketplace billing models
+				// if the kafka billing model supports 'marketplace' or any cloud specific marketplace billing model, return 'marketplace'
+				if kafkaBillingModel.HasSupportForMarketplace() {
+					return ocm.AMSRelatedResourceBillingModelMarketplace.String(), nil
+				}
+			case shared.StringHasPrefixIgnoreCase(amsBillingModel, fmt.Sprintf("%s-", ocm.AMSRelatedResourceBillingModelMarketplace)):
+				// ams billing model is a cloud specific marketplace, i.e. 'marketplace-aws'
+				// if the kafka billing model supports this cloud specific marketplace billing model, return 'marketplace'
+				if kafkaBillingModel.HasSupportForAMSBillingModel(amsBillingModel) {
+					return ocm.AMSRelatedResourceBillingModelMarketplace.String(), nil
+				}
+			case shared.StringEqualsIgnoreCase(amsBillingModel, ocm.AMSRelatedResourceBillingModelStandard.String()):
+				if kafkaBillingModel.HasSupportForStandard() {
+					return ocm.AMSRelatedResourceBillingModelStandard.String(), nil
+				}
+			}
+		}
+	}
+
+	logger.Logger.V(10).Infof("ams related resource billing model not found for subscription %q and kafka billing model %q", subscriptionID, kafkaBillingModel.ID)
+	return "", nil
 }
--- a/internal/kafka/internal/services/quota/quota_management_list_service.go
+++ b/internal/kafka/internal/services/quota/quota_management_list_service.go
@@ -2,7 +2,9 @@ package quota
 
 import (
 	"fmt"
+
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/quota_management"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
@@ -223,4 +225,37 @@ func (q QuotaManagementListService) detectBillingModel(kafka *dbapi.KafkaRequest
 
 func (q QuotaManagementListService) DeleteQuota(SubscriptionId string) *errors.ServiceError {
 	return nil // NOOP
+}
+
+// Checks if quota used by the given Kafka instance is granted to the organisation/user and
+// if it is active, not expired, in the quota management list configuration
+// Note that organisation will always take priority over individual accounts to mimic the behaviour of
+// quota allowance checks during Kafka creation.
+func (q QuotaManagementListService) IsQuotaEntitlementActive(kafka *dbapi.KafkaRequest) (bool, error) {
+	var billingModel *quota_management.BillingModel
+
+	org, orgFound := q.quotaManagementList.QuotaList.Organisations.GetById(kafka.OrganisationId)
+	if orgFound && org.IsUserRegistered(kafka.Owner) {
+		logger.Logger.Infof("user registered by organisation, checking quota entitlement for organisation %q", org.Id)
+		bm, ok := org.GetBillingModel(kafka.InstanceType, kafka.ActualKafkaBillingModel)
+		if ok {
+			billingModel = &bm
+		}
+	} else {
+		logger.Logger.Infof("user is not registered by organisation, checking quota entitlement for %q as an individual account", kafka.Owner)
+		account, accountFound := q.quotaManagementList.QuotaList.ServiceAccounts.GetByUsername(kafka.Owner)
+		if accountFound {
+			bm, ok := account.GetBillingModel(kafka.InstanceType, kafka.ActualKafkaBillingModel)
+			if ok {
+				billingModel = &bm
+			}
+		}
+	}
+
+	// unable to find billing model for the instance type indicates that quota is no longer granted to the org/user
+	if billingModel == nil || billingModel.HasExpired() {
+		return false, nil
+	}
+
+	return true, nil
 }
--- a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
+++ b/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
@@ -1,17 +1,24 @@
 package kafka_mgrs
 
 import (
+	"database/sql"
+	"fmt"
 	"math"
 	"strings"
+	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/acl"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	serviceErr "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/metrics"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/workers"
+
 	"github.com/golang/glog"
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
@@ -118,6 +125,13 @@ func (k *KafkaManager) Reconcile() []error {
 	if updateErr != nil {
 		encounteredErrors = append(encounteredErrors, updateErr)
 		return encounteredErrors
+	}
+
+	// reconciles expires_at field for kafka instances
+	updateExpiresAtErrors := k.reconcileKafkaExpiresAt(kafkas)
+	if updateExpiresAtErrors != nil {
+		wrappedError := errors.Wrap(updateExpiresAtErrors, "failed to update expires_at for kafka instances")
+		encounteredErrors = append(encounteredErrors, wrappedError)
 	}
 
 	for _, kafka := range kafkas {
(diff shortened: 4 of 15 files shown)


internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +306,37 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
...
+	if len(quotas) == 0 {
+		return &[]bool{false}[0], nil
+	}
+
+	// result will always return one quota for the given org, resource and product for rhosak quotas
+	quota := quotas[0]
+
+	// when a subscription expires in AMS, the allowed value for that quota is set back to 0
+	// if the allowed value is 0 and consumed is greater than this, that denotes that the subscription
+	// has expired and is no longer active.
```
machi1990, line 336: What happens to the underlying subscription denoted by the subscriptionID that we save in the database? Does that status of that subscription changes? If so, it is possible to re-use it?

ziccardi: This PR is just about updating the `expires_at` column but doesn't do anything to the subscription.
A dedicated reconciler manages the real cluster expiration.
That reconciler will do the following:

- If expires_at< today : delete the cluster (and its subscription)
- If expires_at > today && (expires_at - grace_period) < today : suspend the cluster and leave the subscription untouched

internal/kafka/internal/services/kafka.go
```diff
@@ -1342,3 +1347,12 @@ func (k *kafkaService) UpdateZeroValueOfKafkaRequestsExpiredAt() error {
 
 	return db.Error
 }
+
+func (k *kafkaService) IsSubscriptionActive(kafkaRequest *dbapi.KafkaRequest) (*bool, error) {
```
ziccardi, line 1351: Why `*bool`? I believe this should be 
```suggestion
func (k *kafkaService) IsSubscriptionActive(kafkaRequest *dbapi.KafkaRequest) (bool, error) {
```

In case of error we return the `zero value` for `bool` (false) and the error

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +306,37 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
 	}
 	return nil
 }
+
+func (q amsQuotaService) IsSubscriptionActive(kafka *dbapi.KafkaRequest) (*bool, error) {
```
ziccardi, line 310: See previous comment

internal/kafka/internal/services/quota/quota_management_list_service.go
```diff
@@ -207,3 +208,9 @@ func (q QuotaManagementListService) detectBillingModel(kafka *dbapi.KafkaRequest
 func (q QuotaManagementListService) DeleteQuota(SubscriptionId string) *errors.ServiceError {
 	return nil // NOOP
 }
+
+// noop method, subscriptions do not apply to quota management list.
+// This function will always return true
+func (q QuotaManagementListService) IsSubscriptionActive(kafka *dbapi.KafkaRequest) (*bool, error) {
+	return &[]bool{true}[0], nil
```
ziccardi, line 215: We have an expiration configuration in the quota list as well. Can we use that to decide if the quota has expired or not?

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: I believe that subscription remains untouched, the status will be the same as it was when the subscription is still active. Maybe @ziccardi can confirm. 

> If so, it is possible to re-use it?

If the AMS subscription (SKU allowance) was extended after it expires, then yes we can re-use it. Once the Kafka instance is promoted however, it will create a new subscription and the existing one deleted.

internal/kafka/internal/services/kafka.go
JameelB: I was trying to be cautious in case someone doesn't check the error and to avoid `false` having two meanings. If an error is returned, `false` here could be a false negative, it may not mean that the subscription is actually expired.

internal/kafka/internal/services/quota/quota_management_list_service.go
JameelB: oh good point! forgot about that 😃 will update this to use that.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -343,3 +355,95 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+				subscriptionStatusByOrgAndBillingModel[orgBillingModelID] = *isActive
+				active = *isActive
+			}
+
+			if active {
+				// if subscription is active, set expires_at to null
+				if err := expiresAt.Scan(nil); err != nil {
+					svcErrors = append(svcErrors, errors.Wrapf(err, "failed to set expires_at value for kafka instance %q to NULL", kafka.ID))
+					continue
+				}
```
ziccardi, line 398: Not sure it make sense to make this check. I think we can just do:
```suggestion
				expiresAt = NullTime{ 
				     Time: nil,
				     Valid: false,
				}
```

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: The word 'Subscription' can mean the Subscription (SKU allowance) and AMS Subscription resource. It's a bit confusing, if anyone has any other name suggestions to make the separation a bit clearer between these two let me know.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -343,3 +355,95 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+					duration, err := time.ParseDuration(fmt.Sprintf("%dh", billingModel.GracePeriodDays*24))
+					if err != nil {
+						svcErrors = append(svcErrors, errors.Wrapf(err, "failed to parse expiration duration for kafka instance %q", kafka.ID))
+						continue
+					}
+
+					if err := expiresAt.Scan(time.Now().Add(duration)); err != nil {
+						svcErrors = append(svcErrors, errors.Wrapf(err, "failed to set expires_at value for kafka instance %q", kafka.ID))
+						continue
+					}
```
ziccardi, line 418:
```suggestion
					expiresAt = NullTime{
					    Time, time.Now().Add(duration),
					    Valid: true,
					}
```

internal/kafka/internal/services/quota/ams_quota_service.go
machi1990: Thank you. 

> This PR is just about updating the expires_at column but doesn't do anything to the subscription.

Correct. What I was asking is specifically in relation to the method naming:
```
IsSubscriptionActive(kafkaRequest)
```

Reading the method signature, I thought that it was checking that the Kafka subscription is still active or not? Is that correct?

> I believe that subscription remains untouched, the status will be the same as it was when the subscription is still active. Maybe @ziccardi can confirm.

Right. This is very odd - ideally, I would have expected some fields to change in the subscription to indicate that it has expired. 

This way we precisely know that the subscription for the given Kafka is active or not and not having. By the looks of the current implementation
```
if quota.Allowed() == 0 && quota.Consumed() > quota.Allowed() {
		return &[]bool{false}[0], nil
	}
```
doesn't tell us that a given kafka subscription has expired but it just tells us that for a given billing model and product, there are some expired subscriptions, correct?

ziccardi: Correct. What the method check id if the user is still entitled to have the instance, hence, he still has an active subscription. As Jameel said before, the `subscription` name is reused in many different topics so it can be confusing.

Maybe we should rename the function to '`IsEntitled` or something similar? WDYT?

machi1990: > As Jameel said before, the subscription name is reused in many different topics so it can be confusing.

I only saw Jameel's reply after I posted my comment above, but I agree with what's said there. The name is confusing.

> Maybe we should rename the function to 'IsEntitled or something similar? WDYT?

+1 on renaming it. What do you think of `IsSKUEntitlementStillActive(...)` ?

ziccardi: +1 for me

JameelB: Thanks for the suggestions! Not sure about `SKU` because that term isn't really used in the quota-management-list 🤔  Would `IsQuotaEntitlementActive` make sense? the method call would then be `quotaService.IsQuotaEntitlementActive()`? 

We also need to clarify this separation in our docs, I can update them in this PR as well.

machi1990: +1 from me @JameelB Thanks.

Agreeing on the docs modification as well.

internal/kafka/internal/services/kafka.go
miguelsorianod: +1 to change it to `bool, error`.
In Go in general it is assumed that if an error is returned the value of the result is not valid / reliable unless explicitly stated.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +306,37 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
...
+	if err != nil {
+		return nil, err
+	}
+
+	// quota not found, subscription is not valid
+	if len(quotas) == 0 {
+		return &[]bool{false}[0], nil
+	}
+
+	// result will always return one quota for the given org, resource and product for rhosak quotas
```
miguelsorianod, line 331: Is this correct?

If a user has consumption-based and prepaid subscriptions wouldn't we be getting more than 1 quotacost?

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -306,3 +306,37 @@ func (q amsQuotaService) DeleteQuota(subscriptionId string) *errors.ServiceError
...
+	if err != nil {
+		return nil, err
+	}
+
+	orgID, err := q.amsClient.GetOrganisationIdFromExternalId(kafka.OrganisationId)
+	if err != nil {
+		return nil, err
+	}
+
+	quotas, err := q.amsClient.GetQuotaCostsForProduct(orgID, billingModel.AMSResource, billingModel.AMSProduct)
```
miguelsorianod, line 321: can we call the variable quotaCosts?
The Quota concept is different than the QuotaCost concept in AMS

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -343,3 +355,95 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+			svcErrors = append(svcErrors, errors.Wrapf(err,
+				"failed to get kafka instance size for %q with instance type %q and size id %q",
+				kafka.ID, kafka.InstanceType, kafka.SizeId,
+			))
+			continue
+		}
+
+		expiresAt := kafka.ExpiresAt
+		// If lifespan seconds is not defined, expiration is determined based on the organisation's active subscription
+		if instanceSize.LifespanSeconds == nil {
```
miguelsorianod, line 380: Can we refactor this two big blocks into its own methods?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -343,3 +355,95 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+				}
+
+				if err := expiresAt.Scan(kafka.CreatedAt.Add(time.Duration(duration))); err != nil {
+					svcErrors = append(svcErrors, errors.Wrapf(err, "failed to set expires_at value for kafka instance %q", kafka.ID))
+					continue
+				}
+			}
+		}
+
+		if expiresAt != kafka.ExpiresAt {
```
miguelsorianod, line 439: This is doing struct comparison. Are we aware of this?

ziccardi: This was another think I though as well. But with the change I suggested, we don't need this anymore.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -343,3 +355,95 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+				// if subscription is not valid and expires_at is not already set, set its value based on the current time and grace period allowance
+				if !kafka.ExpiresAt.Valid {
+					// set expires_at to now + grace period days
+					billingModel, err := k.kafkaConfig.GetBillingModelByID(kafka.InstanceType, kafka.ActualKafkaBillingModel)
+					if err != nil {
+						svcErrors = append(svcErrors, errors.Wrapf(err, "failed to get billing model of kafka instance %q", kafka.ID))
+						continue
+					}
+
+					duration, err := time.ParseDuration(fmt.Sprintf("%dh", billingModel.GracePeriodDays*24))
```
miguelsorianod, line 409: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1499/files#r1086451249 but using different units.

For example:
```
duration := time.Duration(gracePeriodDays*86400) * time.Second
```

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -120,6 +125,13 @@ func (k *KafkaManager) Reconcile() []error {
 		return encounteredErrors
 	}
 
+	// reconciles expires_at field for kafka instances
+	updateExpiresAtErrors := k.reconcileKafkaExpiresAt(kafkas)
```
miguelsorianod, line 129: Here there is a potential race condition where `updateZeroValueOfKafkaRequestsExpiredAt` updates the value of some kafka instances in the DB but the `kafkas` slice in memory does not contain the changes

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -343,3 +355,95 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+			}
+		}
+
+		if expiresAt != kafka.ExpiresAt {
+			if err := k.kafkaService.Updates(kafka, map[string]interface{}{
+				"expires_at": expiresAt,
+			}); err != nil {
+				svcErrors = append(svcErrors, errors.Wrapf(err, "failed to update expires_at value for kafka instance %q", kafka.ID))
+			}
+		}
```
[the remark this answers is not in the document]
miguelsorianod: I am not sure I understand it.

This will perform the update all the time instead of conditionally.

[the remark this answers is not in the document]
ziccardi: Nope. When the expireAt is set, the current code exits and then the `if` checks if `expireAt` has been changed.
With this change we update it and continue to the next cluster instead of using a temporary `expireAt` and then checking if it has been changed.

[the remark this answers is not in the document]
miguelsorianod: I think I understand the proposal now.

The method will perform the update all the time instead of conditionally, but the method will only be called under certain conditions in the blocks above.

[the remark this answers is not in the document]
ziccardi: yup. exactly.

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: Hm maybe you're right. I've only managed to test this with our organisation and I think we're only entitled prepaid. @ziccardi is there any way we can verify this?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -343,3 +355,95 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+				if err := expiresAt.Scan(kafka.CreatedAt.Add(time.Duration(duration))); err != nil {
+					svcErrors = append(svcErrors, errors.Wrapf(err, "failed to set expires_at value for kafka instance %q", kafka.ID))
+					continue
+				}
+			}
+		}
+
+		if expiresAt != kafka.ExpiresAt {
+			if err := k.kafkaService.Updates(kafka, map[string]interface{}{
+				"expires_at": expiresAt,
```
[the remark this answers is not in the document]
JameelB: yes i've tested it with nil and actual value for Time, from what I've seen so far it stores it as expected.

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: not really, but I believe Miguel is correct. You should check the ams_billing_model too to be sure to get one result.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
JameelB: do you think it's worth calling ListAll within this method again to avoid that?

miguelsorianod: With the current logic we might get a scenario where the ExpiresAt of some of the kafka instances would be the zero-value of time.Time which is 1 Jan of year 1. The `Valid` attribute would also be set to `true`.

I think this might have an impact on the logic that marks instances as suspended cc @ziccardi.

Regarding the new logic, it also references ExpiresAt but I am not fully sure if it might cause any issue. Could you review it? I think we should be ok from that side but more people double checking it would be useful

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: After discussions with @miguelsorianod and @ziccardi, we've reached an agreement on filtering the quota costs based on: {AMSResourceName, AMSProduct and AMSBillingModel} which should always return one quota cost. 

This works at the moment as this combination is currently unique for each quota cost but note that we may need to change this in the future depending on how quota costs are managed in AMS for RHOSAK SKUs in the future.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -396,3 +398,79 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
...
+	// result will always return one quota for the given org, ams related resource billing model, resource and product for rhosak quotas
+	quotaCost := quotaCosts[0]
+
+	// when a SKU entitlement expires in AMS, the allowed value for that quota cost is set back to 0
+	// if the allowed value is 0 and consumed is greater than this, that denotes that the SKU entitlement
+	// has expired and is no longer active.
+	if quotaCost.Allowed() == 0 && quotaCost.Consumed() > quotaCost.Allowed() {
+		glog.V(10).Infof("quota %q has expired for organisation %q", quotaCost.QuotaID(), orgID)
+		return false, nil
+	}
```
ziccardi, line 448: I think we should add another check with a log here

Something like:
```go
if quotaCost.Consumed() > quotaCost.Allowed() {
	glog.V(10).Warnf("Organisation %q is overquota (quotaid: %q, consumed: %q, allowed: %q)",
		orgID, quotaCost.QuotaID(), quotaCost.Consumed(), quotaCost.Allowed())
	return false, nil
}
```

It should be logged into sentry too.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -396,3 +398,79 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
...
+	for _, rr := range reservedResources {
+		if kafkaBillingModel.AMSResource == rr.ResourceName() && kafkaBillingModel.HasSupportForAMSBillingModel(string(rr.BillingModel())) {
+			if arrays.StringEqualsIgnoreCasePredicate(string(rr.BillingModel()))("marketplace") ||
+				arrays.StringHasPrefixIgnoreCasePredicate(string(rr.BillingModel()))("marketplace-") {
+				return string(ocm.AMSRelatedResourceBillingModelMarketplace), nil
+			}
+
+			if arrays.AnyMatch([]string{string(rr.BillingModel())}, arrays.StringEqualsIgnoreCasePredicate("standard")) {
+				return string(ocm.AMSRelatedResourceBillingModelStandard), nil
+			}
```
ziccardi, line 470: See previous comment about the `arrays` package.
```suggestion
			switch true {
			case shared.StringEqualsIgnoreCase(amsBillingModel, "marketplace"):
				return string(ocm.AMSRelatedResourceBillingModelMarketplace), nil
			case shared.StringHasPrefixIgnoreCase(amsBillingModel, "marketplace-"):
				return string(ocm.AMSRelatedResourceBillingModelMarketplace), nil
			case shared.StringEqualsIgnoreCase(amsBillingModel, "standard"):
				return string(ocm.AMSRelatedResourceBillingModelStandard), nil
			}
```

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -374,3 +387,116 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+	if isQuotaEntitlementActive && kafka.ExpiresAt.Valid {
+		glog.V(10).Infof("updating expiration date of kafka instance %q to NULL", kafka.ID)
+		return k.updateKafkaExpirationDate(kafka, nil)
+	}
+
+	// if quota entitlement is not valid and expires_at is not already set, set its value based on the current time and grace period allowance
+	// note that there is a temporary check for zero value of time.Time for cases where the kafka instance expires_at value
+	// hasn't been updated null yet after migration/rollout. This additional check can be removed once all Kafka instances
+	// has been successfully migrated to the new expires_at type.
+	if !isQuotaEntitlementActive && (!kafka.ExpiresAt.Valid || kafka.ExpiresAt.Time.IsZero()) {
```
ziccardi, line 456: Should we check if `kafka.ExpireAt.Time.After(Now + gracePeriod)` too?

For example, if an instance was set to expire in 10 days, but the subscription has expired today.

I'm not sure that will ever happen though. WDYT?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -374,3 +387,116 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+	return nil
+}
+
+// Updates expires_at field based on the lifespan seconds configured per Kafka instance type/size
+func (k *KafkaManager) updateExpiresAtBasedOnLifespanSeconds(kafka *dbapi.KafkaRequest, instanceSize config.KafkaInstanceSize) error {
+	// note that there is a temporary check for zero value of time.Time in cases where the kafka instance expires_at value
+	// hasn't been updated to null yet after migration/rollout. This additional check can be removed once all Kafka instances
+	// has been successfully migrated to the new expires_at type
+	if !kafka.ExpiresAt.Valid || kafka.ExpiresAt.Time.IsZero() {
+		glog.V(10).Infof("lifespanSeconds defined for Kafka %q, updating value for expires_at as it is not yet set", kafka.ID)
```
ziccardi, line 478: Let's add the new value to this log

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -396,3 +398,79 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
...
+	}
+
+	// SKU not entitled to the organisation
+	if len(quotaCosts) == 0 {
+		glog.V(10).Infof("ams quota cost not found for Kafka instance %q in organisation %q: %+v", kafka.ID, orgID, quotaCostRelatedResourceFilter)
+		return false, nil
+	}
+
+	if len(quotaCosts) > 1 {
+		glog.Warning("more than 1 quota cost was returned with the following related resource filter %+v for organisation %q", orgID, quotaCostRelatedResourceFilter)
```
ziccardi, line 436: Should we log this to sentry?

ziccardi (changes requested):

internal/kafka/internal/services/quota/ams_quota_service.go
miguelsorianod, line 436: Why is this a warning? I think in that case we should return an error as it is unexpected and we don't know what numbers we would be comparing

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -396,3 +398,79 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
...
+// Determine the AMS QuotaCost related resource billing model from the AMS Subscription reserved resource billing model
+// based on the given subscription id and kafka billing model.
+func (q amsQuotaService) getAMSRelatedResourceBillingModel(subscriptionID string, kafkaBillingModel config.KafkaBillingModel) (string, error) {
+	reservedResources, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscriptionID)
+	if err != nil {
+		return "", err
+	}
+
+	for _, rr := range reservedResources {
+		if kafkaBillingModel.AMSResource == rr.ResourceName() && kafkaBillingModel.HasSupportForAMSBillingModel(string(rr.BillingModel())) {
```
miguelsorianod, line 462: The check of `kafkaBillingModel.HasSupportForAMSBillingModel(string(rr.BillingModel()` doesn't seem correct.
HasSupportForAMSBillingModel checks for the billing models in the config which are AMS Reserved Resource Billing Models, and not AMS Related Resource billing models

pkg/client/ocm/types.go
```diff
@@ -23,3 +23,10 @@ const (
 const (
 	RHOSAKResourceName string = "rhosak"
 )
+
+type AMSRelatedResourceBillingModel string
```
miguelsorianod, line 27: Can we make `var supportedAMSRelatedResourceBillingModels = map[string]struct{}{` consistent with this?

Also a `String` method can be added to this type to avoid doing manual type conversion

internal/kafka/internal/services/quota/ams_quota_service.go
miguelsorianod, line 462: Have we verified that the ReservedResource's BillingModel attribute from the Subscription contains the compound name `marketplace-aws`? I think we should make sure that happens and there is no aggregation performed in the response as opposed to the request time

miguelsorianod, line 462: Also an additional note: this is also going to be an issue if at some moment we want to remove `marketplace` from the config file as a supported AMS billing model: For old instances where reservation is made with `marketplace` it will start failing until they don't exist anymore

miguelsorianod: Maybe we should start reserving with the appropriate non-deprecated values and remove it from the config before we introduce all the changes?
This would require verifying that it is being billed correctly with the new values

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -396,3 +398,79 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
...
+			if arrays.StringEqualsIgnoreCasePredicate(string(rr.BillingModel()))("marketplace") ||
+				arrays.StringHasPrefixIgnoreCasePredicate(string(rr.BillingModel()))("marketplace-") {
+				return string(ocm.AMSRelatedResourceBillingModelMarketplace), nil
+			}
+
+			if arrays.AnyMatch([]string{string(rr.BillingModel())}, arrays.StringEqualsIgnoreCasePredicate("standard")) {
```
miguelsorianod, line 468: In here we can use constants from ocm  for standard

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -396,3 +398,79 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
...
+// based on the given subscription id and kafka billing model.
+func (q amsQuotaService) getAMSRelatedResourceBillingModel(subscriptionID string, kafkaBillingModel config.KafkaBillingModel) (string, error) {
+	reservedResources, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscriptionID)
+	if err != nil {
+		return "", err
+	}
+
+	for _, rr := range reservedResources {
+		if kafkaBillingModel.AMSResource == rr.ResourceName() && kafkaBillingModel.HasSupportForAMSBillingModel(string(rr.BillingModel())) {
+			if arrays.StringEqualsIgnoreCasePredicate(string(rr.BillingModel()))("marketplace") ||
```
miguelsorianod, line 463: In here we can use constants from ocm I think

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: isn't `quotaCost.Consumed() > quotaCost.Allowed()` already being checked there? 🤔  I could change the log to that if that's preferred

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: yes it does. When I was testing through this, i was getting `billing_model: marketplace-aws` from ams.

JameelB: I think i'll add some additional checks here to accommodate for that, I'm not sure how long it would take to verify those changes. 

AMS is not going to handle the migration of the billing model when they remove it? What happens if it was a long lived standard instance that wouldn't be removed? Do we need to handle moving the billing model from `marketplace` to a cloud provider specific marketplace?

internal/kafka/internal/services/quota/ams_quota_service.go
miguelsorianod: I think the comment is in relation to log those values so we can see what's the current state  of consumed vs allowed from the logs

internal/kafka/internal/services/quota/ams_quota_service.go
miguelsorianod: I am not sure if they will be able to ever remove it.
We were not given that information.
What I know is that we should stop using it, and if we can avoid introducing the legacy it would be better so we wouldn't need to worry about it

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
JameelB: I don't think we need this check 🤔 The only way that the instance expiration is set before the subscription expires is if there was lifespanSeconds defined for that instance type/size. In that case, we always go by the lifespanSeconds lifetime and not change the expiration date even if the grace period doesn't fully complete.

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: I understood that comment as adding additional log if the org has exceeded their quotas on top of what's already there. Is that correct @ziccardi?  🤔 

So something like this? 
```suggestion
        if quotaCost.Consumed() > quotaCost.Allowed() {
		if quotaCost.Allowed() == 0 {
			glog.V(10).Infof("quota no longer entitled for organisation %q (quotaid: %q, consumed: %q, allowed: %q)",
				orgID, quotaCost.QuotaID(), quotaCost.Consumed(), quotaCost.Allowed())
			return false, nil
		}
		glog.Warning("Organisation %q has exceeded their quota allowance (quotaid: %q, consumed %q, allowed: %q",
			orgID, quotaCost.QuotaID(), quotaCost.Consumed(), quotaCost.Allowed())
	}
```

ziccardi: @JameelB Yes. Even in quota.Allowed() is not zero, I still think it would be useful to log if quota.Consumed() > quota.Allowed()

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
JameelB: I double checked the suspension logic to see how it would be affected if the expires_at has a zero time value. Looks like it should be ok because of this check:

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/b4e39ab567773b7446925d67c4f4905e9ddb5cb4/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L129-L133

The Kafka will be deemed as 'expired' because `Valid` will be true but it won't get suspended.

The problem lies in the deprovision logic because it doesn't account for a zero time value if by chance a Kafka instance expiration date hasn't been updated at this point.
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/b4e39ab567773b7446925d67c4f4905e9ddb5cb4/internal/kafka/internal/services/kafka.go#L647-L657

I think we can add a check here to accommodate for that until this transition to the null value is complete. wdyt? @miguelsorianod @ziccardi

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: I added some additional checks here to accommodate for removing marketplace in the config for now. We can then remove it once we start removing references to the 'marketplace' billing model.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -396,3 +398,96 @@ func (q amsQuotaService) GetSubscriptionByID(subscriptionID string) (*amsv1.Subs
...
+	if err != nil {
+		return false, err
+	}
+
+	amsRelatedResourceBillingModel, err := q.getAMSRelatedResourceBillingModel(kafka.SubscriptionId, kafkaBillingModel)
+	if err != nil || shared.StringEmpty(amsRelatedResourceBillingModel) {
+		return false, err
+	}
+
+	orgID, err := q.amsClient.GetOrganisationIdFromExternalId(kafka.OrganisationId)
```
ziccardi, line 414: side comment: I find extremely confusing the fact that `kafka.OrganisationId` is not an organisation id but an external id. It is even more confusing to see code like `GetOrganisationIdFromExternalId(kafka.OrganisationId)`.

I think we should evaluate the idea of renaming `OrganisationId` to `ExternalID`

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
miguelsorianod: I think there shouldn't be any issue because the Kafkas are retrieved from the DB in `DeprovisionExpiredKafkas` right?
If we want we can still be extra sure and temporarily add what you said to avoid any surprise during 2 pods running at the same time due to some race condition. I can't remember if that's even possible at the moment

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -374,3 +388,116 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+
+// Updates expires_at field of the given Kafka instance based on the user/organisation's quota entitlement status
+func (k *KafkaManager) updateExpiresAtBasedOnQuotaEntitlement(kafka *dbapi.KafkaRequest, isQuotaEntitlementActive bool) error {
+	// if quota entitlement is active, ensure expires_at is set to null
+	if isQuotaEntitlementActive && kafka.ExpiresAt.Valid {
+		logger.Logger.Infof("updating expiration date of kafka instance %q to NULL", kafka.ID)
+		return k.updateKafkaExpirationDate(kafka, nil)
+	}
+
+	// if quota entitlement is not valid and expires_at is not already set, set its value based on the current time and grace period allowance
```
miguelsorianod, line 453: I guess here you meant `if quota entitlement is not active` instead of `if quota entitlement is not valid`

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -374,3 +388,116 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+
+	logger.Logger.Infof("lifespanSeconds defined for Kafka %q but expires_at is already set, skipping update", kafka.ID)
+	return nil
+}
+
+// updates the expires_at field for the given Kafka instance
+func (k *KafkaManager) updateKafkaExpirationDate(kafka *dbapi.KafkaRequest, expiresAtTime *time.Time) error {
+	var expiresAt sql.NullTime
+	if expiresAtTime != nil {
+		_ = expiresAt.Scan(*expiresAtTime)
```
miguelsorianod, line 493: I think there is no need to use scan, and thus we can avoid the possibility of an error

The following would be valid:
expiresAt = sql.NullTime{Time: *expiresAtTime, Valid: true}

miguelsorianod: cc @JameelB @ziccardi to double check what I said is accurate

ziccardi: Agree

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
JameelB: > I think there shouldn't be any issue because the Kafkas are retrieved from the DB in DeprovisionExpiredKafkas right?

Yeah i don't think there will be any issues. As far as I know there can only be one leader pod so it shouldn't be possible.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
JameelB: good spot, yep i meant `is not active` here!

internal/kafka/internal/services/quota/ams_quota_service.go
JameelB: yeah got caught with that the first time round i was doing this check, I had it in my mind that the `kafka.OrganisationId` was the actual org id. +1 to revising this in the future.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
JameelB: yep that works 👍  updated it to use that instead.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
```diff
@@ -1469,3 +1527,609 @@ func TestKafkaManager_calculateAvailableAndMadCapacityForEnterpriseClusters(t *t
...
+			t.Parallel()
+			updatedExpiresAt := sql.NullTime{}
+			updateCallCount := 0
+			k := &KafkaManager{
+				kafkaService: tt.fields.kafkaService(&updatedExpiresAt, &updateCallCount),
+				kafkaConfig:  tt.fields.kafkaConfig,
+			}
+			err := k.reconcileKafkaExpiresAt(tt.args.kafkas)
+			g.Expect(len(err)).To(gomega.Equal(tt.wantErrCount))
+			g.Expect(updateCallCount).To(gomega.Equal(tt.wantUpdateCallCount), "expected update call count does not match actual")
```
miguelsorianod, line 2127: I think the Mock type provides a way to check how many times the Updates call is called. Can you review that?
I think we can avoid using that counter

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
```diff
@@ -1469,3 +1527,609 @@ func TestKafkaManager_calculateAvailableAndMadCapacityForEnterpriseClusters(t *t
...
+
+			g.Expect(updatedExpiresAt.Valid).To(gomega.Equal(tt.wantNewExpiresAtValue.Valid), "expected expires_at.valid does not match actual")
```
miguelsorianod, line 2129: Is there some reason we compare the year, month an day? Can't we do a equality check?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
JameelB: you're right, i completely missed that. Thanks! I'll update this to use that instead

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
JameelB: It's because the expires_at update based on quota entitlement status is based on time.Now() + grace period. I won't be able to know the exact time when that value is taken during the reconcile. Since expiration would be based on days anyway, we can just compare the year, month and day. We can ignore the comparison on the minutes and seconds.

miguelsorianod: An alternative would be refactor the code so it receives a method that retrieves the time.
Then in the tests we could change the method and provide a crafted time now.
What is there now is good enough, but something to keep in mind.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -374,3 +388,116 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+	// hasn't been updated to null yet after migration/rollout. This additional check can be removed once all Kafka instances
+	// has been successfully migrated to the new expires_at type.
+	if !isQuotaEntitlementActive && (!kafka.ExpiresAt.Valid || kafka.ExpiresAt.Time.IsZero()) {
+		billingModel, err := k.kafkaConfig.GetBillingModelByID(kafka.InstanceType, kafka.ActualKafkaBillingModel)
+		if err != nil {
+			return err
+		}
+
+		// set expires_at to now + grace period days
+		expiresAtTime := time.Now().Add(time.Duration(billingModel.GracePeriodDays*86400) * time.Second)
```
ziccardi, line 464:
```suggestion
		expiresAtTime := time.Now().AddDate(0, 0, billingModel.GracePeriodDays)
```

ziccardi (approved):
LGTM. Great work!

PR #1503, opened 2023-01-20
feat: add observatorium proxy token validation endpoint

pawelpaszki:
## Description
feat: add observatorium proxy token validation endpoint

left to do:
- tests
- potentially adding a middleware to the router

- [ ] Code Review completed


internal/kafka/internal/handlers/observatorium_proxy.go
```diff
@@ -0,0 +1,54 @@
...
+			clusters, err := h.clusterService.FindAllClusters(services.FindClusterCriteria{})
+			if err != nil {
+				return nil, errors.GeneralError("failed to validate the request: %v" + err.Error())
+			}
+			for _, cluster := range clusters {
+				if cluster.ExternalID == external_id && cluster.ClientID == clientID {
+					return nil, nil
+				}
+			}
+			return nil, errors.Forbidden("failed to match the client ID in the token against provided external cluster ID: %s" + external_id)
```
ziccardi, line 50:
```suggestion
			if arrays.NoneMatch(clusters, func (cluster *api.Cluster) bool { return cluster.ExternalID == external_id && cluster.ClientID == clientID}) {
				return nil, errors.Forbidden("failed to match the client ID in the token against provided external cluster ID: %s" + external_id)
			}
			
			return nil, nil
```

pkg/auth/kfm_claims.go
```diff
@@ -26,6 +27,13 @@ func (c *KFMClaims) GetAccountId() (string, error) {
 	return "", fmt.Errorf("can't find '%s' attribute in claims", tenantUserIdClaim)
 }
 
+func (c *KFMClaims) GetClientId() (string, error) {
```
ziccardi, line 30:
```suggestion
func (c *KFMClaims) GetClientID() (string, error) {
```

internal/kafka/internal/handlers/observatorium_proxy.go
machi1990: +1 Massimo

My suggestion will be to take `arrays.NoneMatch(clusters, func (cluster *api.Cluster) bool { return cluster.ExternalID == external_id && cluster.ClientID == clientID})` out the if and declare a var that can be used inside the if

internal/kafka/internal/handlers/observatorium_proxy.go
```diff
@@ -0,0 +1,54 @@
...
+			if err != nil {
+				return nil, errors.Unauthenticated("unable to authenticate token provided in the request")
+			}
+
+			clientID, err := claims.GetClientId()
+			if err != nil {
+				return nil, errors.Unauthenticated("unable to get client ID from the token")
+			}
+
+			clusters, err := h.clusterService.FindAllClusters(services.FindClusterCriteria{})
```
machi1990, line 41: It'll be awesome if we could extend the `FindClusterCriteria` so that it accepts ExternalID as the query criteria as well.
This will reduce the result set that has to be processed.

internal/kafka/internal/routes/route_loader.go
```diff
@@ -258,6 +258,13 @@ func (s *options) buildApiBaseRouter(mainRouter *mux.Router, basePath string) er
 	// deliberately returns 404 here if the request doesn't have the required role, so that it will appear as if the endpoint doesn't exist
 	auth.UseOperatorAuthorisationMiddleware(apiV1DataPlaneRequestsRouter, s.Keycloak.GetRealmConfig().ValidIssuerURI, "id", s.ClusterService)
 
+	// observatorium proxy endpoints
+	observatoriumProxyHandler := handlers.NewObservatoriumProxyHandler(s.ClusterService)
+	observatoriumProxyRouter := apiV1Router.PathPrefix("/obs-proxy-verification").Subrouter()
+	observatoriumProxyRouter.HandleFunc("/{cluster_external_id}", observatoriumProxyHandler.ValidateTokenAndExternalClusterID).
+		Name(logger.NewLogEvent("validate-observatorium-proxy-token", "validate observatorium proxy token and cluster external id").ToString()).
+		Methods(http.MethodPost)
+
```
machi1990, line 267: let's use the require issuer middeware for this.

```
observatoriumProxyRouter.Use(auth.NewRequireIssuerMiddleware().RequireIssuer([]string{s.Keycloak.GetRealmConfig().ValidIssuerURI}, errors.ErrorNotFound))
```

openapi/kas-fleet-manager-private.yaml
```diff
@@ -19,8 +19,56 @@ servers:
 tags:
   - name: Agent Clusters
     description: ""
+  - name: observatorium-proxy
+    description: Observatorium proxy related endpoints
 
 paths:
+  # observatorium proxy related endpoints
+  '/api/kafkas_mgmt/v1/obs-proxy-verification/{cluster-external-id}':
+    post:
```
machi1990, line 28: At the moment we do not have payload, I do not think we'll have any.
I suppose we could just make this a `get`. But I am fine with `post` as well since the operation is an action.

pkg/auth/kfm_claims.go
machi1990, line 30: We could use the method here as well: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/22024f07663adf924c1054b3d255899d6b3b2e3b/pkg/auth/operator_authz_middleware.go#L37

internal/kafka/internal/handlers/observatorium_proxy.go
```diff
@@ -22,30 +22,31 @@ func NewObservatoriumProxyHandler(clusterService services.ClusterService) *obser
...
 			}
 
-			clientID, err := claims.GetClientId()
+			clientID, err := claims.GetClientID()
 			if err != nil {
 				return nil, errors.Unauthenticated("unable to get client ID from the token")
 			}
 
-			clusters, err := h.clusterService.FindAllClusters(services.FindClusterCriteria{})
+			clusters, err := h.clusterService.FindAllClusters(services.FindClusterCriteria{ExternalID: external_id})
```
ziccardi, line 43: Here you are calling `FindAllClusters`, but you added support for the `ExternalID` in the `FindClusters` method.

Side comment: FindAllClusters and FindCluster are identical but the return value: the first one returns an array, while the second one returns the first result. I believe `FindCluster` should call `FindAllClusters` to remove duplicated code

internal/kafka/internal/services/clusters.go
```diff
@@ -313,6 +314,7 @@ func (c clusterService) FindCluster(criteria FindClusterCriteria) (*api.Cluster,
 		Region:        criteria.Region,
 		MultiAZ:       criteria.MultiAZ,
 		Status:        criteria.Status,
+		ExternalID:    criteria.ExternalID,
```
ziccardi, line 317: Mind adding a test for this new filter?

internal/kafka/internal/services/clusters.go
```diff
@@ -301,6 +301,7 @@ type FindClusterCriteria struct {
 	MultiAZ               bool
 	Status                api.ClusterStatus
 	SupportedInstanceType string
+	ExternalID            string
 }
 
 func (c clusterService) FindCluster(criteria FindClusterCriteria) (*api.Cluster, error) {
```
ziccardi, line 307: Side comment N.2: This function should be called `FindFirstMatchingCluster` since it is not guaranteed that the passed-in filter will find exactly one cluster.

If OTH, the function's role is always to find exactly one cluster, then the function should check whether the filter returned one result or a list of results instead of blindly returning the first one.

pawelpaszki: I switched to use FindCluster

internal/kafka/internal/services/clusters.go
pawelpaszki: done

internal/kafka/internal/handlers/observatorium_proxy.go
pawelpaszki: I switched to use FindCluster

ziccardi: Mind adding the new `ExternalID` support to `FindAllClusters` too?

pawelpaszki: sure

pkg/environments/environment.go
```diff
@@ -22,7 +22,7 @@ const (
 	IntegrationEnv string = "integration"
 
 	EnvironmentStringKey string = "OCM_ENV"
-	EnvironmentDefault          = DevelopmentEnv
+	EnvironmentDefault          = IntegrationEnv
```
machi1990, line 25: Why this change?

pkg/auth/context_config.go
```diff
@@ -33,6 +34,7 @@ func (c *ContextConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&alternateTenantUsernameClaim, "alternate-tenant-username-claim", alternateTenantUsernameClaim, "Token claims key to retrieve the corresponding user principal using an alternative claim.")
 	fs.StringVar(&tenantUserIdClaim, "tenant-user-id-claim", tenantUserIdClaim, "Token claims key to retrieve the corresponding  Account ID.")
 	fs.StringVar(&alternateTenantIdClaim, "alternate-tenant-id-claim", alternateTenantIdClaim, "Token claims key to retrieve the corresponding organisation ID using an alternative claim.")
+	fs.StringVar(&clientIDclaim, "client-id-claim", clientIDclaim, "Token claims key to retrieve the corresponding client ID.")
```
machi1990, line 37: This is not strictly needed. The client_id claim isn't going to change.

pkg/environments/environment.go
pawelpaszki: I will remove it. I am adding it when debugging integration tests locally

machi1990 (commented):
lgtm

left a comment on adding unit tests for the handler

internal/kafka/internal/handlers/observatorium_proxy.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func NewObservatoriumProxyHandler(clusterService services.ClusterService) *observatoriumProxyHandler {
+	return &observatoriumProxyHandler{
+		clusterService: clusterService,
+	}
+}
+
+// Delete is the handler for deleting a kafka request
+func (h observatoriumProxyHandler) ValidateTokenAndExternalClusterID(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
pawelpaszki: sure. let me create unit test for this

machi1990 (approved):
lgtm

thanks @pawelpaszki

miguelsorianod:
It would be good to add motivations in the description of the PR to understand why is this being added

openapi/kas-fleet-manager-private.yaml
```diff
@@ -19,8 +19,56 @@ servers:
 tags:
   - name: Agent Clusters
     description: ""
+  - name: observatorium-proxy
+    description: Observatorium proxy related endpoints
 
 paths:
+  # observatorium proxy related endpoints
+  '/api/kafkas_mgmt/v1/obs-proxy-verification/{cluster_external_id}':
```
miguelsorianod, line 27: Can we add a description to the endpoint?

pawelpaszki: I'll address this and the url format comment in a new PR

internal/kafka/internal/api/private/api/openapi.yaml
```diff
@@ -15,7 +15,56 @@ servers:
   url: /
 tags:
 - name: Agent Clusters
+- description: Observatorium proxy related endpoints
+  name: observatorium-proxy
 paths:
+  /api/kafkas_mgmt/v1/obs-proxy-verification/{cluster_external_id}:
```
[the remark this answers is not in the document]
machi1990: The URL format for private internal endpoints are still using the `-` e.g https://github.com/pawelpaszki/kas-fleet-manager/blob/e63381d51e4eccc50e9934378a2149e316dc5184/openapi/kas-fleet-manager-private.yaml#L73 for `agents-clusters`. 

The move to using `_` where rhoas spectral linting rules are applied was done for the public endpoints.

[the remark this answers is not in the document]
pawelpaszki: so what do you guys suggest? keep the url as is or change it?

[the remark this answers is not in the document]
machi1990: We do not validate the `-private.yaml` openapi file at the moment against any rules. For consistency with existing endpoints in this file, I could keep it the way it is even if it diverges with what's applied for other files public & admin endpoints files where validation is applied.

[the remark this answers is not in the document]
miguelsorianod: In this case I agree

openapi/kas-fleet-manager-private.yaml
```diff
@@ -19,8 +19,56 @@ servers:
 tags:
   - name: Agent Clusters
     description: ""
+  - name: observatorium-proxy
+    description: Observatorium proxy related endpoints
 
 paths:
```
miguelsorianod, line 25: The minor version of the api document should have been increased because of this change

internal/kafka/internal/services/clusters.go
```diff
@@ -503,14 +506,14 @@ func (c clusterService) FindAllClusters(criteria FindClusterCriteria) ([]*api.Cl
 	// They are mostly the same as the library we use (xid) does take the generation timestamp into consideration,
 	// However, it only down to the level of seconds. This means that if a few records are created at almost the same time,
 	// the order is not guaranteed. So use the `created_at` column will provider better consistency.
-	if err := dbConn.Where(clusterDetails).Order("created_at asc").Scan(&cluster).Error; err != nil {
+	if err := dbConn.Where(clusterDetails).Order("created_at asc").Scan(&clusters).Error; err != nil {
```
miguelsorianod, line 509: Side comment:

For this method we should document that attributes in `FindClusterCriteria` whose value is the zero-value of their corresponding type are not included in the search criteria

internal/kafka/internal/handlers/observatorium_proxy.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewObservatoriumProxyHandler(clusterService services.ClusterService) *observatoriumProxyHandler {
+	return &observatoriumProxyHandler{
+		clusterService: clusterService,
+	}
+}
+
+// ValidateTokenAndExternalClusterID validates combination of external cluster ID parameter against client ID from the claims
+func (h observatoriumProxyHandler) ValidateTokenAndExternalClusterID(w http.ResponseWriter, r *http.Request) {
+	external_id := mux.Vars(r)["cluster_external_id"]
```
miguelsorianod, line 25: Is there some reason the ID provided is ExternalID instead of ClusterID? for cluster agent endpoints ClusterID is used

machi1990: It's because the `cluster_id` label metric actually represents the `cluster.ExternalID` (which is the physical cluster spec ID) in KFM database and not the `cluster.ClusterID` which is an OCM id (primary key)

miguelsorianod: I understand your comment, but how is the value being set for the metric label `cluster_id related to what the data plane sends to the kafka request in the URL? They are independent.

machi1990: > but how is the value being set for the metric label `cluster_id related to what the data plane sends to the kafka request in the URL? 

I am not sure I understood this part. But I'll try to respond based on what I've understood from the question.
First of all, this a Kafka metric's only endpoint. It has nothing to do with what the fleetshard sends to us via the agents-endpoints.

This endpoint is called by the observatorium-proxy to verify if a given clientID (in the token) is authorized for a given cluster external id before metrics are forwarded/proxied to observatorium.
The proxy gets the metric (and not the Kafka request) data from Prometheus remote write - this has an override to ensure that the cluster_id label contains the value i.e the external_id.

This id being the cluster uuid (ocm's external_id) it seems consistency for the URL parameter to speak as such to avoid such confusions e.g
The URL parameter is called `cluster_id` but why are we comparing it against `cluster.ExteranalID` and not `cluster.ClusterID` 

Note: The `cluster_id` that's seen in Agents endpoints is a `cluster.ClusterID` as represented in the `Cluster` entity.

miguelsorianod: Can the observatorium proxy send the ClusterID instead of ExternalID?

pb82: @miguelsorianod the proxy gets the clusterId from the label on the samples in the remote write request.

machi1990: I think the question is, can the observability-operator configured to send the ClusterID, instead of fetching ExternalID from the cluster ? I'll leave that to @pb82 to chime in for more details but as per other previously discussions (precisely about this value) this value is used somewhere - to construct the dashboards and/or  sub watch (I can't remember exactly where in dashboards). An additional question, is the change worth it?

miguelsorianod: > @miguelsorianod the proxy gets the clusterId from the label on the samples in the remote write request.

I understand you mean the External Cluster ID.
Is there another label available that contains the Cluster ID? If yes, could that be used alternatively to perform the request? If not, would it make sense to add it?

PR #1504, opened 2023-01-20
feat(certs-management): generate a certificate per data plane cluster

machi1990:
## Description

implement vault storage based on AWS Secret Manager

Closes https://issues.redhat.com/browse/MGDSTRM-10444 and https://issues.redhat.com/browse/MGDSTRM-10443

This is highly based on https://github.com/bf2fc6cc711aee1a0c2a/architecture/tree/main/_adr/89

## Verification Steps

1. For the manual mode, integrations tests should pass
2. For the automatic mode, sync up with me and I'll guide you through the verification steps

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- ~~[ ] Required metrics/dashboards/alerts have been added (or PR created).~~
- [ ] Required Standard Operating Procedure (SOP) is added.
- ~~[ ] JIRA has been created for changes required on the client side~~

PR #1511, opened 2023-01-24
feat: cluster deregistration

pb82:
## Description

Implements the deregistration of an Enterprise cluster. The SyncSet is removed and the cluster is soft-deleted afterwards.

## Verification Steps

1. Create an OCM cluster
2. Run the KFM locally, but don't target this cluster
3. Register the cluster using the enterprise registration endpoint
4. Make sure that registration was successful and that the Observability resources have been created
5. Deregister the cluster by calling the deregistration endpoint
6. Make sure that the Observability resources are removed, the cluster only soft deleted in the database.

### DIFF
--- a/internal/kafka/constants/cluster.go
+++ b/internal/kafka/constants/cluster.go
@@ -10,6 +10,9 @@ const (
 	// ClusterOperationDelete - OpenShift/k8s cluster delete operation
 	ClusterOperationDelete ClusterOperation = "delete"
 
+	// ClusterOperationHardDelete - OpenShift/k8s cluster hard delete operation
+	ClusterOperationHardDelete ClusterOperation = "hard_delete"
+
 	// The DNS prefixes used for traffic ingress
 	ManagedKafkaIngressDnsNamePrefix = "kas"
 	DefaultIngressDnsNamePrefix      = "apps"
--- a/internal/kafka/internal/clusters/ocm_provider.go
+++ b/internal/kafka/internal/clusters/ocm_provider.go
@@ -125,6 +125,25 @@ func (o *OCMProvider) AddIdentityProvider(clusterSpec *types.ClusterSpec, identi
 	return nil, nil
 }
 
+// RemoveResources deletes the SyncSet from a cluster. If the SyncSet (or the cluster) is not found, no error is returned
+func (o *OCMProvider) RemoveResources(clusterSpec *types.ClusterSpec, syncSetName string) error {
+	status, err := o.ocmClient.DeleteSyncSet(clusterSpec.InternalID, syncSetName)
+	if err != nil {
+		// the SyncSet may already have been deleted in a previous run and the cluster is
+		// waiting for the resources to be terminated
+		if status == http.StatusNotFound {
+			return nil
+		}
+		return err
+	}
+
+	if status != http.StatusNoContent {
+		return errors.New(fmt.Sprintf("unexpected response code when deleting sync set: %v", status))
+	}
+
+	return nil
+}
+
 func (o *OCMProvider) ApplyResources(clusterSpec *types.ClusterSpec, resources types.ResourceSet) (*types.ResourceSet, error) {
 	existingSyncset, err := o.ocmClient.GetSyncSet(clusterSpec.InternalID, resources.Name)
 	syncSetFound := true
--- a/internal/kafka/internal/clusters/standalone_provider.go
+++ b/internal/kafka/internal/clusters/standalone_provider.go
@@ -64,6 +64,10 @@ func newStandaloneProvider(connectionFactory *db.ConnectionFactory, dataplaneClu
 // blank assignment to verify that StandaloneProvider implements Provider
 var _ Provider = &StandaloneProvider{}
 
+func (s *StandaloneProvider) RemoveResources(clusterSpec *types.ClusterSpec, syncSetName string) error {
+	return nil
+}
+
 func (s *StandaloneProvider) Create(request *types.ClusterRequest) (*types.ClusterSpec, error) {
 	return nil, nil
 }
--- a/internal/kafka/internal/handlers/cluster.go
+++ b/internal/kafka/internal/handlers/cluster.go
@@ -147,7 +147,7 @@ func (h clusterHandler) DeregisterEnterpriseCluster(w http.ResponseWriter, r *ht
 			handlers.ValidateAsyncEnabled(r, "deleting enterprise cluster"),
 			ValidateKafkaClaims(ctx, ValidateOrganisationId()),
 			validateEnterpriseClusterEligibleForDeregistration(ctx, clusterID, h.clusterService),
-			validateEnterpriseClusterHasNoKafkas(ctx, clusterID, h.clusterService),
+			validateEnterpriseClusterHasNoKafkas(clusterID, h.clusterService),
 		},
 		Action: func() (i interface{}, serviceError *errors.ServiceError) {
 			return nil, h.clusterService.DeregisterClusterJob(clusterID)
--- a/internal/kafka/internal/handlers/cluster.go
+++ b/internal/kafka/internal/handlers/cluster.go
@@ -1,9 +1,6 @@
 package handlers
 
 import (
-	"net/http"
-	"strconv"
-
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/public"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/presenters"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
@@ -11,6 +8,7 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/handlers"
 	"github.com/gorilla/mux"
+	"net/http"
 )
 
 type clusterHandler struct {
@@ -143,33 +141,16 @@ func (h clusterHandler) List(w http.ResponseWriter, r *http.Request) {
 func (h clusterHandler) DeregisterEnterpriseCluster(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	clusterID := mux.Vars(r)["id"]
-	forceDeleteFlag := r.URL.Query().Get("force")
-
-	if forceDeleteFlag == "" {
-		forceDeleteFlag = "false"
-	}
-
-	forceBoolValue, err := strconv.ParseBool(forceDeleteFlag)
-
-	validateForceParam := func() handlers.Validate {
-		return func() *errors.ServiceError {
-			if err != nil {
-				return errors.GeneralError("failed to parse query param force: %s", forceDeleteFlag)
-			}
-			return nil
-		}
-	}
 
 	cfg := &handlers.HandlerConfig{
 		Validate: []handlers.Validate{
 			handlers.ValidateAsyncEnabled(r, "deleting enterprise cluster"),
-			validateForceParam(),
 			ValidateKafkaClaims(ctx, ValidateOrganisationId()),
-			validateEnterpriseClusterEligibleForDeregistration(ctx, clusterID, forceBoolValue, h.clusterService),
+			validateEnterpriseClusterEligibleForDeregistration(ctx, clusterID, h.clusterService),
+			validateEnterpriseClusterHasNoKafkas(ctx, clusterID, h.clusterService),
 		},
 		Action: func() (i interface{}, serviceError *errors.ServiceError) {
-			// cluster deletion from db and cleanup will be handled in a follow up Jira
-			return nil, nil
+			return nil, h.clusterService.DeregisterClusterJob(clusterID)
 		},
 	}
 	handlers.HandleDelete(w, r, cfg, http.StatusAccepted)
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -445,7 +445,7 @@ func validateEnterpriseClusterEligibleForDeregistration(ctx context.Context, clu
 }
 
 // validateEnterpriseClusterHasNoKafkas requires a cluster to be empty, thus having no kafka instances
-func validateEnterpriseClusterHasNoKafkas(ctx context.Context, clusterID string, clusterService services.ClusterService) handlers.Validate {
+func validateEnterpriseClusterHasNoKafkas(clusterID string, clusterService services.ClusterService) handlers.Validate {
 	return func() *errors.ServiceError {
 		instanceCounts, err := clusterService.FindKafkaInstanceCount([]string{clusterID})
 		if err != nil {
--- a/internal/kafka/internal/handlers/validation.go
+++ b/internal/kafka/internal/handlers/validation.go
@@ -408,7 +408,7 @@ func validateKafkaMachinePoolNodeCount(clusterPayload *public.EnterpriseOsdClust
 	}
 }
 
-func validateEnterpriseClusterEligibleForDeregistration(ctx context.Context, clusterID string, forceDeleteFlag bool, clusterService services.ClusterService) handlers.Validate {
+func validateEnterpriseClusterEligibleForDeregistration(ctx context.Context, clusterID string, clusterService services.ClusterService) handlers.Validate {
 	return func() *errors.ServiceError {
 		claims, claimsErr := getClaims(ctx)
 
@@ -440,19 +440,24 @@ func validateEnterpriseClusterEligibleForDeregistration(ctx context.Context, clu
 		if cluster.ClusterType != api.EnterpriseDataPlaneClusterType.String() {
 			return errors.Forbidden("unable to deregister cluster whose type is not: %s", api.EnterpriseDataPlaneClusterType.String())
 		}
-		if !forceDeleteFlag {
-			clusterKafkaCount, err := clusterService.FindKafkaInstanceCount([]string{clusterID})
-			if err != nil {
-				return errors.GeneralError(err.Error())
-			}
-			for _, count := range clusterKafkaCount {
-				if count.Clusterid == clusterID && count.Count > 0 {
-					return errors.Forbidden("unable to deregister %s cluster with id: %s without explicitly setting 'force:true' due to kafka requests present on this cluster",
-						api.EnterpriseDataPlaneClusterType.String(), clusterID)
-				}
+		return nil
+	}
+}
+
+// validateEnterpriseClusterHasNoKafkas requires a cluster to be empty, thus having no kafka instances
+func validateEnterpriseClusterHasNoKafkas(ctx context.Context, clusterID string, clusterService services.ClusterService) handlers.Validate {
+	return func() *errors.ServiceError {
+		instanceCounts, err := clusterService.FindKafkaInstanceCount([]string{clusterID})
+		if err != nil {
+			return errors.GeneralError("error querying kafka instances for clusterID: %v", clusterID)
+		}
+
+		for _, instanceCount := range instanceCounts {
+			if instanceCount.Clusterid == clusterID && instanceCount.Count > 0 {
+				return errors.Forbidden("unable to deregister cluster with kafka instances")
 			}
 		}
-		return nil
 
+		return nil
 	}
 }
--- a/internal/kafka/internal/services/clusters.go
+++ b/internal/kafka/internal/services/clusters.go
@@ -51,7 +51,6 @@ type ClusterService interface {
 	// DeleteByClusterID will delete the cluster from the database
 	DeleteByClusterID(clusterID string) *apiErrors.ServiceError
 	// HardDeleteByClusterID hard deletes a cluster from the database, no delete flag is set
-	// HardDeleteByClusterID hard deletes a cluster from the database, no delete flag is set
 	HardDeleteByClusterID(clusterID string) *apiErrors.ServiceError
 	// FindNonEmptyClusterByID returns a cluster if it present and it is not empty.
 	// Cluster emptiness is determined by checking whether the cluster contains Kafkas that have been provisioned, are being provisioned on it,
--- a/internal/kafka/internal/services/clusters.go
+++ b/internal/kafka/internal/services/clusters.go
@@ -47,8 +47,12 @@ type ClusterService interface {
 	GetClientID(clusterID string) (string, error)
 	ListGroupByProviderAndRegion(providers []string, regions []string, status []string) ([]*ResGroupCPRegion, *apiErrors.ServiceError)
 	RegisterClusterJob(clusterRequest *api.Cluster) *apiErrors.ServiceError
+	DeregisterClusterJob(clusterID string) *apiErrors.ServiceError
 	// DeleteByClusterID will delete the cluster from the database
 	DeleteByClusterID(clusterID string) *apiErrors.ServiceError
+	// HardDeleteByClusterID hard deletes a cluster from the database, no delete flag is set
+	// HardDeleteByClusterID hard deletes a cluster from the database, no delete flag is set
+	HardDeleteByClusterID(clusterID string) *apiErrors.ServiceError
 	// FindNonEmptyClusterByID returns a cluster if it present and it is not empty.
 	// Cluster emptiness is determined by checking whether the cluster contains Kafkas that have been provisioned, are being provisioned on it,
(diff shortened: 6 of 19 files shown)


internal/kafka/internal/clusters/standalone_provider.go
```diff
@@ -64,6 +64,10 @@ func newStandaloneProvider(connectionFactory *db.ConnectionFactory, dataplaneClu
 // blank assignment to verify that StandaloneProvider implements Provider
 var _ Provider = &StandaloneProvider{}
 
+func (s *StandaloneProvider) RemoveResources(clusterSpec *types.ClusterSpec, syncSetName string) error {
+	return nil
```
pb82, line 68: I'm going to add an implementation for the standalone provider as well, although I believe it's sufficient to verify the ocm provider implementation.

machi1990: I wonder how that's going to be implemented as we do not have the syncset name for standalone clusters.

If this was taking a list of resources, we could have used that somehow to issue a delete call for every of the resources: but I can't imagine the complexity involved. 

I am wondering if a NOOP is the best we can do here.

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -95,6 +101,20 @@ func (m *CleanupClustersManager) processCleanupClusters() error {
 	return errList
 }
 
+func (m *CleanupClustersManager) reconcileCleanupEnterpriseCluster(cluster api.Cluster) error {
```
machi1990, line 104: Will need to sync this method with the changes in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1504 to ensure that the certificates are revoked for the given cluster

internal/kafka/internal/services/clusters.go
```diff
@@ -108,6 +110,21 @@ func (c clusterService) RegisterClusterJob(clusterRequest *api.Cluster) *apiErro
 	return nil
 }
 
+// DeregisterClusterJob deregister an existing cluster
+func (c clusterService) DeregisterClusterJob(clusterId string) *apiErrors.ServiceError {
+	cluster, err := c.FindClusterByID(clusterId)
+	if err != nil {
+		return err
+	}
+
```
machi1990, line 119: This is not needed.

internal/kafka/internal/services/clusters.go
```diff
@@ -108,6 +110,21 @@ func (c clusterService) RegisterClusterJob(clusterRequest *api.Cluster) *apiErro
 	return nil
 }
 
+// DeregisterClusterJob deregister an existing cluster
+func (c clusterService) DeregisterClusterJob(clusterId string) *apiErrors.ServiceError {
```
machi1990, line 114:
```suggestion
func (c clusterService) DeregisterClusterJob(clusterID string) *apiErrors.ServiceError {
```

internal/kafka/internal/services/clusters.go
```diff
@@ -108,6 +110,21 @@ func (c clusterService) RegisterClusterJob(clusterRequest *api.Cluster) *apiErro
...
+	cluster, err := c.FindClusterByID(clusterId)
+	if err != nil {
+		return err
+	}
+
+	updateErr := c.UpdateStatus(*cluster, api.ClusterDeprovisioning)
```
machi1990, line 120:
```suggestion
	updateErr := c.UpdateStatus(api.Cluster{ClusterID: clusterID}, api.ClusterDeprovisioning)
```

related to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1511/files#r1085014963

internal/kafka/internal/services/clusters.go
```diff
@@ -108,6 +110,21 @@ func (c clusterService) RegisterClusterJob(clusterRequest *api.Cluster) *apiErro
...
+	if updateErr != nil {
+		return apiErrors.NewWithCause(apiErrors.ErrorGeneral, err, "failed to deregister cluster")
+	}
+
+	return nil
```
machi1990, line 125:
```suggestion
	return updateErr
```

internal/kafka/internal/services/clusters.go
machi1990: a small nit suggestion

internal/kafka/internal/services/clusters.go
machi1990: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1511#discussion_r1085017236 for suggestion

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -95,6 +101,20 @@ func (m *CleanupClustersManager) processCleanupClusters() error {
...
+	glog.Infof("Removing Enterprise Dataplane cluster %s fleetshard service account", cluster.ClusterID)
+	serviceAccountRemovalErr := m.kasFleetshardOperatorAddon.RemoveServiceAccount(cluster)
+	if serviceAccountRemovalErr != nil {
+		return errors.Wrapf(serviceAccountRemovalErr, "failed to removed Dataplane cluster %s fleetshard service account", cluster.ClusterID)
+	}
+	glog.Infof("Soft deleting the Enterprise Dataplane cluster %s from the database", cluster.ClusterID)
+	deleteError := m.clusterService.DeleteByClusterID(cluster.ClusterID)
```
machi1990, line 111: This method only soft delete the cluster - see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c6de1499fafb2d6c86b3d1d93c25404e78addb8f/internal/kafka/internal/services/clusters.go#L393

To allow for a flow like:
- register cluster
- deregister cluster
- register cluster

We might need to hard delete the record from the database.

We'll need to use `db.Unscoped()` to perform a hard delete of enterprise clusters - see https://gorm.io/docs/delete.html#Delete-permanently for more.

pb82: Agree, we might want to hard delete it to support the case where a previously de-registered cluster is registered again. Those clusters are long living, so this could happen.

internal/kafka/internal/services/clusters.go
pb82: What about the case where the cluster doesn't exist? I guess it's ok because then the update would fail?

internal/kafka/internal/clusters/standalone_provider.go
pb82: It's not using a SyncSet in standalone mode, but installs some resources manually through the Kube API. I have to check if the Observability resources are even installed.

machi1990: > It's not using a SyncSet in standalone mode, but installs some resources manually through the Kube API

Yes. 

> I have to check if the Observability resources are even installed.

Yeap, they'll be installed. But it's not only them that can get installed, various other resources e.g operators pull secrets also gets passed down via this way.

internal/kafka/internal/services/clusters.go
machi1990: It depends on the call site of this method:

- There is a check in the deregistration handler for cluster existence.
- If called from within the reconciler / other handler then the check will have to be there. If it is not, this method won't fail as an update of a non existing record in DB is always successful with a mesage like `0 rows affected`.

So far this is only called from the deregistration handler. I think we are good, the less DB round trips the better.

machi1990 (changes requested):
Thanks @pb82 

left some feedback and it'll also be good to update the unit & integration tests

internal/kafka/internal/handlers/cluster.go
```diff
@@ -143,33 +141,17 @@ func (h clusterHandler) List(w http.ResponseWriter, r *http.Request) {
...
 	cfg := &handlers.HandlerConfig{
 		Validate: []handlers.Validate{
 			handlers.ValidateAsyncEnabled(r, "deleting enterprise cluster"),
-			validateForceParam(),
 			ValidateKafkaClaims(ctx, ValidateOrganisationId()),
-			validateEnterpriseClusterEligibleForDeregistration(ctx, clusterID, forceBoolValue, h.clusterService),
+			validateEnterpriseClusterEligibleForDeregistration(ctx, clusterID, h.clusterService),
 		},
 		Action: func() (i interface{}, serviceError *errors.ServiceError) {
 			// cluster deletion from db and cleanup will be handled in a follow up Jira
```
pawelpaszki, line 152: can this comment be removed?

internal/kafka/internal/handlers/validation.go
```diff
@@ -440,19 +440,6 @@ func validateEnterpriseClusterEligibleForDeregistration(ctx context.Context, clu
 		if cluster.ClusterType != api.EnterpriseDataPlaneClusterType.String() {
 			return errors.Forbidden("unable to deregister cluster whose type is not: %s", api.EnterpriseDataPlaneClusterType.String())
 		}
-		if !forceDeleteFlag {
-			clusterKafkaCount, err := clusterService.FindKafkaInstanceCount([]string{clusterID})
```
pawelpaszki, line 444: should the check for non-empty cluster stay here and fail if there are kafkas on the cluster?

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
```diff
@@ -95,6 +101,20 @@ func (m *CleanupClustersManager) processCleanupClusters() error {
...
+	if deleteError != nil {
+		return errors.Wrapf(deleteError, "failed to soft delete Dataplane cluster %s from the database", cluster.ClusterID)
+	}
+	return nil
```
pawelpaszki, line 115: could this be simplified to `m.clusterService.DeleteByClusterID(cluster.ClusterID)` or whatever is gonna be placed here (e.g. hard delete cluster function)?

internal/kafka/internal/handlers/validation.go
machi1990: +1

openapi/kas-fleet-manager.yaml
```diff
@@ -1032,14 +1032,6 @@ paths:
           schema:
             type: boolean
           required: true
-        - in: query
```
miguelsorianod, line 1035: What is the reason this query parameter removed?

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -125,6 +125,24 @@ func (o *OCMProvider) AddIdentityProvider(clusterSpec *types.ClusterSpec, identi
...
+	if err != nil {
+		// the SyncSet may already have been deleted in a previous run and the cluster is
+		// waiting for the resources to be terminated
+		if status == http.StatusNotFound {
+			return nil
+		}
+		return err
+	}
+
+	if status != http.StatusNoContent {
```
miguelsorianod, line 139: When is `http.StatusNoContent` returned?

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -125,6 +125,24 @@ func (o *OCMProvider) AddIdentityProvider(clusterSpec *types.ClusterSpec, identi
 	return nil, nil
 }
 
+func (o *OCMProvider) RemoveResources(clusterSpec *types.ClusterSpec, syncSetName string) error {
```
miguelsorianod, line 128: Can we document this method so consumers of it understand the behavior? Specially the part where if a cluster does not exist it is not considered an error

internal/kafka/internal/services/clusters.go
```diff
@@ -47,6 +47,7 @@ type ClusterService interface {
 	GetClientID(clusterID string) (string, error)
 	ListGroupByProviderAndRegion(providers []string, regions []string, status []string) ([]*ResGroupCPRegion, *apiErrors.ServiceError)
 	RegisterClusterJob(clusterRequest *api.Cluster) *apiErrors.ServiceError
+	DeregisterClusterJob(clusterId string) *apiErrors.ServiceError
```
miguelsorianod, line 50: Accronyms like 'id` should be written in uppercase. Therefore in this case it should be `clusterID string`

internal/kafka/internal/services/clusters.go
miguelsorianod, line 114: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1511/files?show-viewed-files=true&file-filters%5B%5D=#r1086711815

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
miguelsorianod: In that case wouldn't a different `ID` be created?

internal/kafka/internal/services/clusters.go
ziccardi: I prefer the previous approach, so we are sure we return a naked nil.
That will help when/if we will start returning `error` instead of this pointer

internal/kafka/internal/services/clusters.go
```diff
@@ -70,6 +71,7 @@ type ClusterService interface {
 	Delete(cluster *api.Cluster) (bool, *apiErrors.ServiceError)
 	ConfigureAndSaveIdentityProvider(cluster *api.Cluster, identityProviderInfo types.IdentityProviderInfo) (*api.Cluster, *apiErrors.ServiceError)
 	ApplyResources(cluster *api.Cluster, resources types.ResourceSet) *apiErrors.ServiceError
+	RemoveResources(cluster *api.Cluster, syncSetName string) *apiErrors.ServiceError
```
ziccardi, line 74: As a side comment, I believe we should start returning `error` instead of `*ServiceError` to avoid the `nil is not nil` error.

internal/kafka/internal/clusters/standalone_provider.go
miguelsorianod: Another comment about this: I think there is a bit of a naming problem here:
ApplyResources creates a set of 0..N resources putting them in a given ResourceSet name. Notice how `ResourceSet` is a type that we have created in KFM.

In this PR we have created a method named RemoveResources but the method does not delete the resources you specify. You delete all the resources in the a given ResourceSet name. Therefore `RemoveResourceSet` would be a better name which takes a resourceSetName parameter.
If in the future we want to be able to delete specific resources within a given ResourceSet then that would be a different method too.

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
```diff
@@ -78,7 +78,10 @@ func (m *DeprovisioningClustersManager) processDeprovisioningClusters() error {
 				errList.AddErrors(errors.Wrapf(err, "failed to reconcile deprovisioning cluster %s", cluster.ClusterID))
 			}
 		} else {
-			glog.V(10).Infof("skipping deprovisioning of %s cluster with ClusterID = %s", api.EnterpriseDataPlaneClusterType.String(), cluster.ClusterID)
+			glog.V(10).Infof("deregistering enterprise cluster ClusterID = %s", cluster.ClusterID)
```
miguelsorianod, line 81: Side comment:
I keep seeing an increasing number of conditional blocks in the code that do "if it is enterprise do this, otherwise do that". This makes me think we might to abstract this kind of logic at another level.

internal/kafka/internal/services/clusters.go
miguelsorianod: I see this more as a design issue than a specific issue with this method. Which has to do with whether we want the consumers of the service to have accessible details of the ServiceError or perform the conversion by themselves. Whatever the decision is all the methods in the interface should be consistent.

internal/kafka/internal/workers/cluster_mgrs/deprovisioning_clusters_mgr.go
ziccardi: +1000

internal/kafka/internal/services/clusters.go
ziccardi: returning a custom object exposes us to the `nil is not nil` error.
```go

type mycustomerror struct{}

func (mycustomerror) Error() string {
	return ""
}

func myfunc1() *mycustomerror {
	return nil
}

func myfunc2() error {
	return nil
}

func myfunc3() error {
	err := myfunc2()
	err = myfunc1()
	return err
}

func main() {
	if myfunc3() == nil {
                 // Never enters here
		fmt.Println("No error")
	} else {
                // always error even if err value is `nil`!!
		fmt.Println("Error!", err)
	}
}
```

On the other hand if we always return `error` and we always return `naked nils` the issue goes away:
```
type mycustomerror struct{}

func (mycustomerror) Error() string {
	return ""
}

func (mycustomerror) Error() string {
	return ""
}

func myfunc1() error {
	return nil
}

func myfunc2() error {
	return nil
}

func myfunc3() error {
	err := myfunc2()
	err = myfunc1()
	return err
}

func main() {
	err := myfunc3()
	if myfunc3() == nil {
		fmt.Println("No error")
	} else {
		fmt.Println("Error!", err)
	}
}
```

Since returning an `error` is safer, I believe we should move to that approach.

> Whatever the decision is all the methods in the interface should be consistent.
+1

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
pb82: That should probably be a separate PR?

internal/kafka/internal/services/clusters.go
pb82: Agree with @ziccardi 
Returning `updateErr` makes it look like an error is returned even though its assumed to be nil here.

openapi/kas-fleet-manager.yaml
pb82: Revised behaviour of the deregistration endpoint:

Previously: don't delete cluster if there are Kafka instances on it unless `force` is passed.
Now: never delete a cluster when there are kafka instnaces present

internal/kafka/internal/clusters/ocm_provider.go
pb82: On successful delete of the SyncSet. I've tried out the deletion on a cluster and this was what the client returned.

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
pb82: unfortunately no, because the method returns a simple error, while the clusterService returns a service error, which is not quite nil even when there is no error. See @ziccardi comment below on the error vs. service error pains.

internal/kafka/internal/workers/cluster_mgrs/cleanup_clusters_mgr.go
machi1990: Yes. This will be taken care in #1504 

It was more a note for myself to not forget.

pb82:
@machi1990 @pawelpaszki @miguelsorianod @ziccardi I've addressed your feedback, would you mind taking another look?

machi1990 (changes requested):
the code change lgtm. I'll let others have a look as well. 


There are related integration tests that are failing. Can you have a look?

internal/kafka/internal/handlers/validation.go
```diff
@@ -440,19 +440,24 @@ func validateEnterpriseClusterEligibleForDeregistration(ctx context.Context, clu
...
+		instanceCounts, err := clusterService.FindKafkaInstanceCount([]string{clusterID})
+		if err != nil {
+			return errors.GeneralError("error querying kafka instances for clusterID: %v", clusterID)
+		}
+
+		for _, instanceCount := range instanceCounts {
+			if instanceCount.Clusterid == clusterID && instanceCount.Count > 0 {
+				return errors.Forbidden("unable to deregister cluster with kafka instances")
 			}
 		}
```
ziccardi, line 459:
```suggestion
		if arrays.AnyMatch(instanceCounts, func (c services.ResKafkaInstanceCount) bool { return c.Clusterid == clusterID && c.Count > 0}) {
			return errors.Forbidden("unable to deregister cluster with kafka instances")
		}
```

pawelpaszki:
@pb82 - as Manyanda mentioned, unit tests for cluster handler need to be fixed. otherwise I think it looks good

pb82:
@pawelpaszki @machi1990  all tests are fixed now

internal/kafka/test/integration/cluster_deregistration_endpoint_test.go
```diff
@@ -123,49 +121,41 @@ func TestEnterpriseClusterDeregistration(t *testing.T) {
...
+	// should fail when kafkas are on the cluster
+	_, resp, e = client.EnterpriseDataplaneClustersApi.DeleteEnterpriseClusterById(adminCtx, true, id1)
+	g.Expect(e).To(gomega.HaveOccurred(), "error should be thrown when attempting to deregister enterprise cluster with kafkas on it")
 	closeRespBody(resp)
 
-	// force false with empty cluster success
-	_, resp, e = client.EnterpriseDataplaneClustersApi.DeleteEnterpriseClusterById(adminCtx, true, id2, &forceFalse)
+	// should succeed with empty cluster
+	_, resp, e = client.EnterpriseDataplaneClustersApi.DeleteEnterpriseClusterById(adminCtx, true, id2)
 	g.Expect(e).NotTo(gomega.HaveOccurred(), "error should not be thrown when attempting to deregister enterprise cluster with no kafkas on it without force=true")
```
machi1990, line 157: I've two more suggestions in regard to this test:

1. We could add another assertion here that attempts to register the same cluster that was successfully deleted. To test that cluster can be re-registered successfully. For that to work, we'll need to wait first for the cluster to be fully removed - with something like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5545b76d18126733b9648f2c4855430567012c08/internal/kafka/test/integration/cluster_mgr_test.go#L346 

And just after that, we attempt to re-register the cluster agan.

2. It has to be run only against for `integration` env. If run against `development` env it'll fail as it'll attempt to delete an syncset from a dummy cluster that does not exist in OCM

This is a general comment the whole test https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5545b76d18126733b9648f2c4855430567012c08/internal/kafka/test/integration/cluster_deregistration_endpoint_test.go#L19 

Something like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/5545b76d18126733b9648f2c4855430567012c08/internal/kafka/test/integration/cluster_get_enterprise_test.go#L32-L35 can be used.

machi1990:
> @pawelpaszki @machi1990 all tests are fixed now

Thanks @pb82 

Looking good now. I've left two more suggestions for the tests. Thanks for having a look!

internal/kafka/constants/cluster.go
```diff
@@ -10,6 +10,9 @@ const (
 	// ClusterOperationDelete - OpenShift/k8s cluster delete operation
 	ClusterOperationDelete ClusterOperation = "delete"
 
+	// ClusterOperationHardDelete - OpenShift/k8s cluster hard delete operation
+	ClusterOperationHardDelete ClusterOperation = "hard_delete"
```
miguelsorianod, line 14: Adding a new state makes me think we might need to update other places in the code where "deletion-related" states are considered / ignored.

Additionally, see my thoughts related to this in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1511#discussion_r1086722703

internal/kafka/internal/services/clusters.go
```diff
@@ -47,8 +47,12 @@ type ClusterService interface {
 	GetClientID(clusterID string) (string, error)
 	ListGroupByProviderAndRegion(providers []string, regions []string, status []string) ([]*ResGroupCPRegion, *apiErrors.ServiceError)
 	RegisterClusterJob(clusterRequest *api.Cluster) *apiErrors.ServiceError
+	DeregisterClusterJob(clusterID string) *apiErrors.ServiceError
 	// DeleteByClusterID will delete the cluster from the database
 	DeleteByClusterID(clusterID string) *apiErrors.ServiceError
+	// HardDeleteByClusterID hard deletes a cluster from the database, no delete flag is set
+	// HardDeleteByClusterID hard deletes a cluster from the database, no delete flag is set
```
miguelsorianod, line 54: Duplicated line

machi1990 (approved):
thanks @pb82

PR #1512, opened 2023-01-24
fixes MGDCTRS-1824, added support for deprecated types and connectors for types removed from catalog

dhirajsb:
## Description
fixes MGDCTRS-1824, added support for deprecated types and connectors for types removed from catalog
## Verification Steps
Added integration tests and existing tests should pass. 

- [X] Documentation added for the feature
- [ ] Code Review completed


dhirajsb:
@valdar @lburgazzoli I added a `deprecated` flag to mark types that have been deleted from catalog, but are being used by existing connectors.

valdar:
@dhirajsb do we check/enforce that new connectors can not be created with a deprecated connector type?

dhirajsb:
> @dhirajsb do we check/enforce that new connectors can not be created with a deprecated connector type?

The new `deprecated` property is part of `ConnectorMeta` and not in `ConnectorRequestMeta`, so no way for users to set that property on creation or otherwise.

valdar:
> The new `deprecated` property is part of `ConnectorMeta` and not in `ConnectorRequestMeta`, so no way for users to set that property on creation or otherwise.

I would have probably preferred an Annotation for that since that information is already reachable in the api. From the connector itself that is linked to a type and by searching for deprecated types and connectors of a given type. It is also semantically misleading since is not the connector itself to be deprecated.

That said I guess there were UI related concerns that was decided to prefer to be solved this way.

valdar (approved):

lburgazzoli:
> > @dhirajsb do we check/enforce that new connectors can not be created with a deprecated connector type?
> 
> The new `deprecated` property is part of `ConnectorMeta` and not in `ConnectorRequestMeta`, so no way for users to set that property on creation or otherwise.

I think @valdar was asking if the fleet managers fails the request if a connector creation references a deprecated connector type.

dhirajsb:
@lburgazzoli that's a good question. It'd be changing the behavior, but we can throw a bad request error to stop people from creating new connectors for types. I'll add that to this PR.

lburgazzoli:
@dhirajsb @valdar 

Thinking a little bit more about this, we have to cases to handle:
1. a connector_type is marked as deprecated and there is not replacement
2. a connector_type is marked as deprecated and there is a replacement

For case 1, we should obviously prevent customers to create new connector referencing the removed connector_type but for the case 2, which should be the most common case, then it falls into an upgrade/migration path so it should be possible to create a connector with the deprecated connector_type for some time to give customers the time to amend their code/pipelines, hence the validation must depend on the specific case.

This reasoning makes me also thinking that marking a connector as deprecated does not sounds semantically correct as what is is deprecated is the connector_type and not the connector instance so i believe we should address this issue while we talk about upgrade scenarios and status reporting so I would recommend to remove the deprecation flag at the connector level.

dhirajsb:
Changed PR to only flag connector types as deprecated. Creating new connectors for deprecated types is allowed.

dhirajsb:
@lburgazzoli @valdar want to look it over again and merge it?

PR #1530, opened 2023-02-01
feat: get additional cluster info from ocm during registration

pawelpaszki:
## Description
feat: get additional cluster info from ocm during registration

NOTE.
there was one body parameter removed from the request. Normally deprecation policy should be followed, but this endpoint is not yet used by anybody (apart from RHOAS CLI - who are aware of this change). The removed external cluster ID is redundant and it was agreed that its better to remove it now rather than wait for people start using it and remove it later

## Verification Steps
tests should pass

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/api/public/api/openapi.yaml
+++ b/internal/kafka/internal/api/public/api/openapi.yaml
@@ -1970,7 +1970,6 @@ components:
     EnterpriseOsdClusterPayloadExample:
       value:
         cluster_id: 1234abcd1234abcd1234abcd1234abcd
-        cluster_external_id: 69d631de-9b7f-4bc2-bf4f-4d3295a7b25e
         cluster_ingress_dns_name: apps.enterprise-aws.awdk.s1.devshift.org
         kafka_machine_pool_node_count: 9
         access_kafkas_via_private_network: false
@@ -2608,7 +2607,6 @@ components:
       example:
         kafka_machine_pool_node_count: 0
         cluster_id: cluster_id
-        cluster_external_id: cluster_external_id
         access_kafkas_via_private_network: true
         cluster_ingress_dns_name: cluster_ingress_dns_name
       properties:
@@ -2619,10 +2617,6 @@ components:
         cluster_id:
           description: The data plane cluster ID. This is the ID of the cluster obtained
             from OpenShift Cluster Manager (OCM) API
-          type: string
-        cluster_external_id:
-          description: external cluster ID. Can be obtained from the response JSON
-            of OCM get /api/clusters_mgmt/v1/clusters/<cluster_id>
           type: string
         cluster_ingress_dns_name:
           description: dns name of the cluster. Can be obtained from the response
@@ -2639,7 +2633,6 @@ components:
           type: integer
       required:
       - access_kafkas_via_private_network
-      - cluster_external_id
       - cluster_id
       - cluster_ingress_dns_name
       - kafka_machine_pool_node_count
--- a/internal/kafka/internal/api/public/model_enterprise_osd_cluster_payload.go
+++ b/internal/kafka/internal/api/public/model_enterprise_osd_cluster_payload.go
@@ -16,8 +16,6 @@ type EnterpriseOsdClusterPayload struct {
 	AccessKafkasViaPrivateNetwork bool `json:"access_kafkas_via_private_network"`
 	// The data plane cluster ID. This is the ID of the cluster obtained from OpenShift Cluster Manager (OCM) API
 	ClusterId string `json:"cluster_id"`
-	// external cluster ID. Can be obtained from the response JSON of OCM get /api/clusters_mgmt/v1/clusters/<cluster_id>
-	ClusterExternalId string `json:"cluster_external_id"`
 	// dns name of the cluster. Can be obtained from the response JSON of the /api/clusters_mgmt/v1/clusters/<cluster_id>/ingresses (dns_name)
 	ClusterIngressDnsName string `json:"cluster_ingress_dns_name"`
 	// The node count given to the created kafka machine pool.  The machine pool must be created via /api/clusters_mgmt/v1/clusters/<cluster_id>/machine_pools prior to passing this value. The created machine pool must have a `bf2.org/kafkaInstanceProfileType=standard` label and a `bf2.org/kafkaInstanceProfileType=standard:NoExecute` taint. The name of the machine pool must be `kafka-standard`  The node count value has to be a multiple of 3 with a minimum of 3 nodes.
--- a/internal/kafka/internal/clusters/provider.go
+++ b/internal/kafka/internal/clusters/provider.go
@@ -28,8 +28,9 @@ type Provider interface {
 	RemoveResources(clusterSpec *types.ClusterSpec, syncSetName string) error
 	// GetClusterDNS Get the dns of the cluster
 	GetClusterDNS(clusterSpec *types.ClusterSpec) (string, error)
-	// GetCluster
-	GetCluster(clusterID string) (types.ClusterSpec, error)
+	// GetClusterSpec returns the details of the cluster from the cluster provider
+	// It should set the status in the returned `ClusterSpec` to either `provisioning`, `ready` or `failed`.
+	GetClusterSpec(clusterID string) (types.ClusterSpec, error)
 	// GetCloudProviders Get the information about supported cloud providers from the cluster provider
 	GetCloudProviders() (*types.CloudProviderInfoList, error)
 	// GetCloudProviderRegions Get the regions information for the given cloud provider from the cluster provider
--- a/internal/kafka/internal/clusters/provider.go
+++ b/internal/kafka/internal/clusters/provider.go
@@ -28,6 +28,8 @@ type Provider interface {
 	RemoveResources(clusterSpec *types.ClusterSpec, syncSetName string) error
 	// GetClusterDNS Get the dns of the cluster
 	GetClusterDNS(clusterSpec *types.ClusterSpec) (string, error)
+	// GetCluster
+	GetCluster(clusterID string) (types.ClusterSpec, error)
 	// GetCloudProviders Get the information about supported cloud providers from the cluster provider
 	GetCloudProviders() (*types.CloudProviderInfoList, error)
 	// GetCloudProviderRegions Get the regions information for the given cloud provider from the cluster provider
--- a/internal/kafka/internal/clusters/types/types.go
+++ b/internal/kafka/internal/clusters/types/types.go
@@ -88,6 +88,12 @@ type ClusterSpec struct {
 	StatusDetails string `json:"status_details"`
 	// additional information related to the cluster, can vary depending on the provider
 	AdditionalInfo api.JSON `json:"additional_info"`
+	// cloud provider region the cluster
+	Region string `json:"region"`
+	// cloud provider name of the cluster
+	CloudProvider string `json:"cloud_provider"`
+	// multi AZ availability flag of the cluster
+	MultiAZ bool `json:"multi_az"`
 }
 
 type CloudProviderInfo struct {
--- a/internal/kafka/internal/handlers/cluster.go
+++ b/internal/kafka/internal/handlers/cluster.go
@@ -4,23 +4,27 @@ import (
 	"net/http"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/public"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/clusters"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/presenters"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/handlers"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/gorilla/mux"
 )
 
 type clusterHandler struct {
 	kasFleetshardOperatorAddon services.KasFleetshardOperatorAddon
 	clusterService             services.ClusterService
+	providerFactory            clusters.ProviderFactory
 }
 
-func NewClusterHandler(kasFleetshardOperatorAddon services.KasFleetshardOperatorAddon, clusterService services.ClusterService) *clusterHandler {
+func NewClusterHandler(kasFleetshardOperatorAddon services.KasFleetshardOperatorAddon, clusterService services.ClusterService, providerFactory clusters.ProviderFactory) *clusterHandler {
 	return &clusterHandler{
 		kasFleetshardOperatorAddon: kasFleetshardOperatorAddon,
 		clusterService:             clusterService,
+		providerFactory:            providerFactory,
 	}
 }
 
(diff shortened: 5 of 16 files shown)


internal/kafka/internal/handlers/cluster.go
```diff
@@ -30,23 +33,43 @@ func (h clusterHandler) RegisterEnterpriseCluster(w http.ResponseWriter, r *http
...
 		Action: func() (interface{}, *errors.ServiceError) {
 
+			cluster, getClusterErr := provider.GetCluster(clusterPayload.ClusterId)
+			if getClusterErr != nil && cluster == nil {
+				return nil, errors.GeneralError("failed to get cluster by ID: %s", clusterPayload.ClusterId)
+			}
+
+			if !cluster.MultiAZ() {
+				return nil, errors.BadRequest("single AZ clusters are not supported")
+			}
```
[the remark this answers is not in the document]
pawelpaszki: I was thinking about it and couldn't find a field that would indicate it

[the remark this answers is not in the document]
machi1990: The subscription field will contain the info we want.
It'll be a query to the subscription href. Then the info we want we'll be in the `.creator` - so it'll be another query to get the creator details.

internal/kafka/internal/clusters/provider.go
```diff
@@ -28,6 +29,8 @@ type Provider interface {
 	RemoveResources(clusterSpec *types.ClusterSpec, syncSetName string) error
 	// GetClusterDNS Get the dns of the cluster
 	GetClusterDNS(clusterSpec *types.ClusterSpec) (string, error)
+	// GetCluster
+	GetCluster(clusterID string) (*v1.Cluster, error)
```
machi1990, line 33: I think we have to return a different cluster type than the one that comes from ocm sdk.

This way other provider implementation are not tied to ocm provider.

pawelpaszki: is there a viable option here?

machi1990: Yes.

I feel like we already have the struct already. That's the `types.ClusterSpec`, we could enrich it with missing fields e.g cloud provider and region.

The external id is already there.

miguelsorianod: +1 to @machi1990 comment. The cluster Provider has to be fully independent from the underlying provider types

miguelsorianod:
>  it was agreed that its better to remove it now rather than wait for people start using it and remove it later

Fully agree with this:
In general, when designing an API I think it is preferable to add the minimum necessary you know your users need. The API surface should be as minimal as possible. In case in the future it is not enough it is always much easier to add new attributes that you know are needed and for which real requirements have been observed than removing an existing one that you realize is not needed or should be changed to a different structure.

internal/kafka/internal/api/public/model_enterprise_osd_cluster_payload.go
```diff
@@ -16,8 +16,6 @@ type EnterpriseOsdClusterPayload struct {
 	AccessKafkasViaPrivateNetwork bool `json:"access_kafkas_via_private_network"`
 	// The data plane cluster ID. This is the ID of the cluster obtained from OpenShift Cluster Manager (OCM) API
 	ClusterId string `json:"cluster_id"`
```
miguelsorianod, line 18: Does this mean enterprise payloads are only supported on OCM?

internal/kafka/internal/handlers/cluster.go
```diff
@@ -28,23 +31,43 @@ func (h clusterHandler) RegisterEnterpriseCluster(w http.ResponseWriter, r *http
 
 	ctx := r.Context()
 
+	provider, err := h.providerFactory.GetProvider(api.ClusterProviderOCM)
+
+	validateProviderError := func() handlers.Validate {
```
miguelsorianod, line 36: Can we refactor this to its own method?

internal/kafka/internal/clusters/provider.go
miguelsorianod: Even if you make an interface to conform to v1.Cluster v1.Cluster should not be returned.

If you are referring to keeping an interface conforming to v1.Cluster and return that interface, I think that:
1. Would still have details/methods specific to OCM which we don't at the cluster provider level
2. I don't think it would scale over time. v1.Cluster is a big type with several methods.

Creating its own interface might make sense as a possible alternative too, aside from the suggested approach before which is using `types.ClusterSpec`

internal/kafka/internal/api/public/model_enterprise_osd_cluster_payload.go
machi1990: Yes, this is the case at the moment. Enterprise is built on top of ocm. 

But with what's done in this PR, it should be possible to extend to all other cluster providers in the future - e.g a standalone provider for testing or crc clusters etc, by providing the provider type in the request payload. That's not the immediate target for now

internal/kafka/internal/clusters/provider.go
pawelpaszki: I looked into `types.ClusterSpec` usage and my new changes shouldn't affect anything. I addressed your comments above. let me know what you think

internal/kafka/internal/handlers/cluster.go
pawelpaszki: moved

internal/kafka/internal/api/public/model_enterprise_osd_cluster_payload.go
miguelsorianod: Setting the expectations on this is relevant.
For example, I see in the line I referenced that ClusterId is "the ID of the cluster obtained from OpenShift Cluster Manager (OCM) API". This basically limits that endpoint to OCM

internal/kafka/internal/handlers/cluster.go
miguelsorianod: Can we scope it further?

The change that has been made is create a function called `validateProvider` in the `internal/kafka/internal/handlers/validation.go` file.

That comes up with a set of issues:
1. It is a global-package variable: All files in that package will have access to that function. This pollutes the package namespace as I don't think all the code in the package needs to be aware of it 
2. As I see `internal/kafka/internal/handlers/validation.go` there should go validation functions that are useful to be used by multiple handlers. At the moment I don't think that's the case. Even if it were reusable it might make sense only in one specific handler which to me would indicate it wouldn't belong to that file either
3. Given that the function is a function at package level in a quite generic package name, it is not clear what `Provider` means in that context. It validates a "provider". What is a provider?

I think we should limit the scope of the validators as much as possible.

internal/kafka/internal/api/public/model_enterprise_osd_cluster_payload.go
machi1990: That line makes sense with the current state as OCM is the only supported provider.
But in the likely event when there is another provider implementation hooked in, it'll be revisited - either the documentation is updated or the field (this one or other sibling fields) is completely made optional depending on the underlying providers.

internal/kafka/internal/clusters/provider.go
machi1990: Thanks @pawelpaszki, just one nit: I think we could simply return `types.ClusterSpec` other than a pointer to `types.ClusterSpec`.

internal/kafka/internal/clusters/types/types.go
```diff
@@ -88,6 +88,12 @@ type ClusterSpec struct {
 	StatusDetails string `json:"status_details"`
 	// additional information related to the cluster, can vary depending on the provider
 	AdditionalInfo api.JSON `json:"additional_info"`
+	// cloud provider region the cluster
+	Region string `json:"region"`
```
miguelsorianod, line 92: If new attributes are added here, all consumers of that type might make use of them. Are we setting that properly for those elements?
Quickly reviewing it seems that might not be the case

internal/kafka/internal/clusters/provider.go
ziccardi: > If you are referring to keeping an interface conforming to v1.Cluster and return that interface, I think that:
> 
> 1. Would still have details/methods specific to OCM which we don't want at the cluster provider level
> 2. I don't think it would scale over time. v1.Cluster is a big type with several methods.

The interface doesn't need to have all the methods of v1.Cluster. In the interface we can put only the methods we need.

miguelsorianod: > The interface doesn't need to have all the methods of v1.Cluster. In the interface we can put only the methods we need

Yes. That's why I said creating its own interface might make sense as a possible alternative too, aside from the suggested approach before which is using types.ClusterSpec. But in that case then it does not conform to v1.Cluster

ziccardi: > Even if you make an interface to conform to v1.Cluster v1.Cluster should not be returned.
> 

@miguelsorianod Why? If we return the interface, the caller will deal only with the methods that we expose in the interface. He won't have knowledge of what is the original object (unless he uses the reflection, but I don't think we should care of that).

ziccardi: > Yes. That's why I said creating its own interface might make sense as a possible alternative too, aside from the suggested approach before which is using types.ClusterSpec. But in that case then the interface does not conform to v1.Cluster.

Yes, I got what you mean. With `conform to v1.Cluster` I didn't mean that we need to publish all the `v1.Cluster` methods, but I meant that if the method we publish conforms to methods present in v1.Cluster, we can keep on returning it.

miguelsorianod: v1.Cluster is an OCM type. We shouldn't return v1.Cluste as it is an OCM type
If we create our own interface then we don't return v1.Cluster. We would be returning another type

ziccardi: > v1.Cluster is an OCM type. We shouldn't return v1.Cluster as it is an OCM type. If we create our own interface then we by definition we don't return v1.Cluster. We would be returning another type

I think we are talking about 2 different things:
in the signature, we will return the interface, but the `return` statement can return any object compatible with the interface declared in the signature.
If v1.Cluster is compatible, we can return that too.

miguelsorianod: > > Yes. That's why I said creating its own interface might make sense as a possible alternative too, aside from the suggested approach before which is using types.ClusterSpec. But in that case then the interface does not conform to v1.Cluster.
> 
> Yes, I got what you mean. With `conform to v1.Cluster` I didn't mean that we need to publish all the `v1.Cluster` methods, but I meant that if the method we publish conforms to methods present in v1.Cluster, we can keep on returning it.

By `conform to v1.Cluster` I understood you meant all the methods of it, as that's the terminology usually used when using interfaces, hence my response around that.
Assuming you mean having an interface with its own methods, as I mentioned that it might make sense inhttps://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1530#discussion_r1093471645.
But, I don't think it should conform to specific methods of v1.Cluster as the methods in v1.Cluster contain OCM-specific concepts. If your interface ends up providing methods that only apply to OCM you are defeating the purpose of the Provider concept which is being cloud provider agnostic

ziccardi: > By `conform to v1.Cluster` I understood you meant all the methods of it, as that's the terminology usually used when using interfaces, hence my response around that.

You are correct. But that usually means that `an object conforms to an interface`. This time I was talking about creating an interface with a `subset` of the methods of `v1.Cluster` (that's why I said `interface conforms to the object`). If we do that, we can keep returning `v1.Cluster` since it will, obviously, conform to our interface.

At the same time:

1. we won't have to make almost any change to our code since the methods will be the same
2. even if in the `return` statement we return `v1.Cluster`, our code will be decoupled from OCM because we will deal only with our interface, and our interface will expose only methods we are interested in

I don't see any advantage in returning a different object.

miguelsorianod: See my previous comment around why I think we shouldn't blindly imitate the methods from the v1.Cluster

internal/kafka/internal/clusters/types/types.go
machi1990: +1. I'll see what can be done here.

internal/kafka/internal/handlers/cluster.go
machi1990: @miguelsorianod this has been addressed, correct?

internal/kafka/internal/clusters/provider.go
machi1990: We'll go with `types.ClusterSpec` and update the existing code that makes references of that type as suggested in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1530#discussion_r1093461285

Thanks for the input

internal/kafka/internal/handlers/cluster.go
miguelsorianod: Yes it's been addressed :+1:

machi1990: Thanks

internal/kafka/internal/handlers/cluster.go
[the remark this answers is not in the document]
machi1990: Created https://issues.redhat.com/browse/MGDSTRM-10583 for this

internal/kafka/internal/api/public/api/openapi.yaml
```diff
@@ -1869,7 +1869,6 @@ components:
     EnterpriseOsdClusterPayloadExample:
       value:
         cluster_id: 1234abcd1234abcd1234abcd1234abcd
-        cluster_external_id: 69d631de-9b7f-4bc2-bf4f-4d3295a7b25e
         cluster_ingress_dns_name: apps.enterprise-aws.awdk.s1.devshift.org
```
pb82, line 1872: can we get that from ocm too?

internal/kafka/internal/handlers/cluster.go
```diff
@@ -108,6 +126,15 @@ func (h clusterHandler) RegisterEnterpriseCluster(w http.ResponseWriter, r *http
 	handlers.Handle(w, r, cfg, http.StatusOK)
 }
 
+func (h clusterHandler) validateOCMProviderAvailable(provider clusters.Provider, err error) handlers.Validate {
+	return func() *errors.ServiceError {
+		if err != nil || shared.IsNil(provider) {
```
pb82, line 131: I'm curious about what shared.IsNil provides over `== nil`?

internal/kafka/internal/api/public/api/openapi.yaml
machi1990: Yes, we can get the cluster dns. But the current logic in `GetClusterDNS` doesn't handle for the private case.
The CLI already does that and is providing us with the right value. We could keep it for now.
There is also future planned work where we might not need to provide this value at all as KFM won't need it: we could use that opportunity to get rid of cluster dns.

internal/kafka/internal/handlers/cluster.go
machi1990: @ziccardi might know more haha

internal/kafka/internal/api/public/api/openapi.yaml
pb82: Ok, we can handle that in a separate PR (the private cluster case) and let the cli handle it for the moment.

internal/kafka/internal/handlers/cluster.go
ziccardi: The function was born mainly to be used by the array's predicate (IsNilPredicate).
There it was mandatory because we didn't know if the generic `T` would have been a pointer or not.

Then we decided it was useful as a shared public function.

The main advantage over `== nil` should be that `IsNil` does not suffer the `nil is not nil` issue.

internal/kafka/internal/api/public/api/openapi.yaml
machi1990: @ziccardi that's more or less what we are doing at the moment with the `GetClusterDNS` method. But that doesn't work if the cluster is a private cluster - in this case the `ingresses` resources returns an empty list.

PR #1531, opened 2023-02-01
feat(certs-management): automatic certificate management

machi1990:
## Description

1. Add the capabability to automaticaly generate a wildcard certificate per standard kafka instance. Existing Kafkas will continue to share the wildcard certificate but it'll be automatically managed. Developer Kafkas will share one wildcard certificate under the `*.trial.<kafka-domain-name>`.

2. Implement memory and Vault storage based on AWS Secret Manager. The in memory is useful for testing. The vault storage or file storage based are useful when we need persistent storage. The vault one should be used when we want to securely store the secrets.

Closes https://issues.redhat.com/browse/MGDSTRM-10444, https://issues.redhat.com/browse/MGDSTRM-10445 and https://issues.redhat.com/browse/MGDSTRM-10443

This is highly based on the ADR and AP opened in this PR - https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/84
Supersedes https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1504

## Verification Steps

1. For the manual mode, integrations tests should pass
2. For the automatic mode, sync up with me and I'll guide you through the verification steps / share a demo with you.

- [ ] Code Review completed
- [x] Required Standard Operating Procedure (SOP) is added.


machi1990:
This PR can be reviewed already but let's not hold on merging it until other needed changes have been progressed.

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -232,3 +240,24 @@ func (k *KafkaRequest) CanBeAutomaticallySuspended() bool {
 func (k *KafkaRequest) DesiredBillingModelIsEnterprise() bool {
 	return shared.StringEqualsIgnoreCase(k.DesiredKafkaBillingModel, constants.BillingModelEnterprise.String())
 }
+
+// HasCertificateInfo returns true when the tls certificate info for this Kafka have been set
+func (k *KafkaRequest) HasCertificateInfo() bool {
+	return !shared.StringEmpty(k.KafkasRoutesBaseDomainName) &&
+		!shared.StringEmpty(k.KafkasRoutesBaseDomainTLSCrtRef) &&
+		!shared.StringEmpty(k.KafkasRoutesBaseDomainTLSKeyRef)
```
ziccardi, line 248:
```suggestion
	return !(shared.StringEmpty(k.KafkasRoutesBaseDomainName) ||
		shared.StringEmpty(k.KafkasRoutesBaseDomainTLSCrtRef) ||
		shared.StringEmpty(k.KafkasRoutesBaseDomainTLSKeyRef))
```

miguelsorianod:
> Implement memory and Vault storage based on AWS Secret Manager.

Is AWS Secret Manager related to Hashicorp's Vault? I assume that it isn't.
I find using the word "Vault" is confusing, as it might refer to Hashicorp's Vault.
If "Vault" is being used as a word to describe the general category I think "Secret Management Service" could be a term used to describe it. Or "Secret and Encryption Management Service"

internal/kafka/internal/config/aws.go
```diff
@@ -19,15 +19,26 @@ type AWSConfig struct {
 	Route53AccessKeyFile       string
 	Route53SecretAccessKey     string
 	Route53SecretAccessKeyFile string
+
+	// Used for AWS Secret Manager
+	SecretManagerAccessKey           string
```
miguelsorianod, line 24: Could we put this in its own type? For example, AWSSecretsManagerConfig.
In this way we avoid having the AWSConfig struct with lots of attributes and we can have contained validation and logic for the type

internal/kafka/internal/config/aws.go
```diff
@@ -57,5 +72,16 @@ func (c *AWSConfig) ReadFiles() error {
...
+	}
+	err = shared.ReadFileValueString(c.SecretManagerAccessKeyFile, &c.SecretManagerAccessKey)
+	if err != nil {
+		return err
+	}
+	err = shared.ReadFileValueString(c.SecretManagerSecretAccessKeyFile, &c.SecretManagerSecretAccessKey)
+	if err != nil {
+		return err
+	}
 	return err
```
ziccardi, line 86: unrelated to this PR, but here err is always `nil`. It would be more readable to return `nil`

internal/kafka/internal/services/kafka.go
```diff
@@ -886,13 +910,32 @@ func (k *kafkaService) GetManagedKafkaByClusterID(clusterID string) ([]managedka
...
+
+			var err error
+			certificate, err = k.kafkaTLSCertificateManagementService.GetCertificate(context.Background(), certRequest)
+			if err != nil {
+				return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to find kafka %q certificates", kafkaRequest.ID)
+			}
+
+		}
+
+		mk, err := buildManagedKafkaCR(kafkaRequest, k.kafkaConfig, k.keycloakService, certificate, enableKafkaExternalCertificate)
```
ziccardi, line 934: If `enableKafkaExternalCertificate` is false, we reuse the last certificate we got on line 927. Is this on purpose?

ziccardi: nvm. ignore this

machi1990: I think you are right. I'll move over the certificate variable inside the loop.

internal/kafka/internal/config/aws.go
machi1990: I'll do so.

internal/kafka/internal/config/aws.go
machi1990: I like the suggestion, I'll see what can be done consistently for all the other credentials as well. If not, then I'll defer this change to other PRs

machi1990:
> > Implement memory and Vault storage based on AWS Secret Manager.
> 
> Is AWS Secret Manager related to Hashicorp's Vault? I assume that it isn't. I find using the word "Vault" is confusing, as it might refer to Hashicorp's Vault. If "Vault" is being used as a word to describe the general category I think "Secret Management Service" could be a term used to describe it. Or "Secret and Encryption Management Service". 

Vault here is not specific to any secure storage implementation. By vault I simply intent to mean a secure compartment where we store stuff out of reach of bad actors.

> It would also be a good idea to scope it under the context where it is used (certificate-related?) if it's not code designed to be able to be used for other purposes

At the moment, the storage is contextualised for certificate. It is under the relevant package. We could easily extract it to shared package so that it can be used for other areas as well where we store sensitive content in DB. That's for a separate PR, for now I'll live the storage there.

internal/kafka/internal/config/aws.go
machi1990: done, thanks!

internal/kafka/internal/services/kafka.go
ziccardi: Yeah, that was the same that happened to me. Then I noticed that the value of `enableKafkaExternalCertificate` never changes (and thus the issue does not exists). I agree that it would be more readable with a non global variable.

miguelsorianod:
> Vault here is not specific to any secure storage implementation. By vault I simply intent to mean a secure compartment where we store stuff out of reach of bad actors. The underlying implementation is backed by aws secret manager.

Could you consider the alternative suggestions I made? it seems to me it is some kind of secret management service then. I find Vault quite confusing with the Hashicorp's service

machi1990:
> > Vault here is not specific to any secure storage implementation. By vault I simply intent to mean a secure compartment where we store stuff out of reach of bad actors. The underlying implementation is backed by aws secret manager.
> 
> Could you consider the alternative suggestions I made? it seems to me it is some kind of secret management service then. It also falls under the category where AWS Secrets Manager and similar services fall. I find Vault quite confusing with the Hashicorp's service, which also falls under the category of secret management service.

Yes. Let's rename it to avoid confusion.

internal/kafka/internal/config/aws.go
machi1990: done, thanks for the suggestion.

internal/kafka/internal/config/aws_config.go
```diff
@@ -0,0 +1,101 @@
...
+			SecretPrefix:            "kas-fleet-manager",
+		},
+	}
+}
+
+func (c *AWSConfig) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&c.ConfigForOSDClusterCreation.AccountIDFilePath, "aws-account-id-file", c.ConfigForOSDClusterCreation.AccountIDFilePath, "File containing AWS account id")
+	fs.StringVar(&c.ConfigForOSDClusterCreation.AccessKeyFilePath, "aws-access-key-file", c.ConfigForOSDClusterCreation.AccessKeyFilePath, "File containing AWS access key")
+	fs.StringVar(&c.ConfigForOSDClusterCreation.SecretAccessKeyFilePath, "aws-secret-access-key-file", c.ConfigForOSDClusterCreation.SecretAccessKeyFilePath, "File containing AWS secret access key")
+	fs.StringVar(&c.Route53.AccessKeyFilePath, "aws-route53-access-key-file", c.Route53.AccessKeyFilePath, "File containing AWS access key for route53")
```
miguelsorianod, line 63: I see there are 4 different flags around secrets manager. I also see the access key and secret access key are in two different new files.
Has it been considered making the different settings of the secret manager into its own config secret file?

internal/kafka/internal/config/aws_config.go
```diff
@@ -0,0 +1,101 @@
...
+			AccountIDFilePath:       "secrets/aws.accountid",
+			AccessKeyFilePath:       "secrets/aws.accesskey",
+			SecretAccessKeyFilePath: "secrets/aws.secretaccesskey",
+		},
+		Route53: awsRoute53Config{
+			AccessKeyFilePath:       "secrets/aws.route53accesskey",
+			SecretAccessKeyFilePath: "secrets/aws.route53secretaccesskey",
+		},
+		SecretManager: awsSecretManagerConfig{
+			AccessKeyFilePath:       "secrets/aws-secret-manager/aws_access_key_id",
```
miguelsorianod, line 51: Do the file paths need to be exported? Can we make them unexported?
In that way we hide this from consumers of the types outside the package

internal/kafka/internal/services/kafka_tls_certificate_management/certificate_revocation_reason.go
```diff
@@ -0,0 +1,46 @@
+package kafka_tls_certificate_management
```
miguelsorianod, line 1: Having a package with underscores is quite uncommon and not idiomatic.
Can we rename the package name? for example to certmgmt or kafkacertmgmt. This package is a subdirectory of  a `kafka` directory

internal/kafka/internal/config/aws_config.go
machi1990: > I also see the access key and secret access key are in two different new files.

Yes - AppInterface basically gives you a secret par file when you ask for aws secret manager.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
```diff
@@ -0,0 +1,131 @@
+package kafka_tls_certificate_management
```
miguelsorianod, line 1: Can we put this into the config package?
This package is inside services which I'd expect belong to the domain logic

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
```diff
@@ -0,0 +1,131 @@
...
+	CertificateManagementStrategy        string
+	StorageType                          string
+	EnableKafkaExternalCertificate       bool
+	ManualCertificateManagementConfig    ManualCertificateManagementConfig
+	AutomaticCertificateManagementConfig AutomaticCertificateManagementConfig
+}
+
+type ManualCertificateManagementConfig struct {
+	KafkaTLSCert     string
+	KafkaTLSCertFile string `validate:"required"`
```
miguelsorianod, line 37: Could we make the file paths unexported so they are hidden to consumers of other packages?

miguelsorianod, line 37: I find the "file" suffix confusing. Is it the content of the file? or is it a path?
Could we rename them to contain the "FilePath" suffix to make clear what it is?

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
```diff
@@ -0,0 +1,131 @@
...
+
+type AutomaticCertificateManagementConfig struct {
+	RenewalWindowRatio        float64 `validate:"gte=0,lte=1"`
+	EmailToSendNotificationTo string  `validate:"required,email"`
+	AcmeIssuerAccountKeyFile  string  `validate:"required"`
+	AcmeIssuerAccountKey      string
+	MustStaple                bool
+}
+
+func NewCertificateManagementConfig() *KafkaTLSCertificateManagementConfig {
```
miguelsorianod, line 50: Same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1531/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#r1102826227

machi1990: Sorry, the link doesn't get me to the previous comment. Do you mind giving me a gist of the previous comment again? Thanks

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
machi1990: Unfortunately I think they need to be exported - due to the validation annotation.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
machi1990: It can be moved there for consistency reason but I find that the structure we've is a bit technical - we could revisit this structure to be more functional oriented.

miguelsorianod: I am not sure I understand what do you mean with "a bit technical". The reasoning to put in in the other package is due to it belongs to a different layer than the domain logic

internal/kafka/internal/services/kafka_tls_certificate_management/certificate_revocation_reason.go
machi1990: I am okay to renaming it but 

> Can we rename the package name? for example to certmgmt or tlscertmgmt, or kafkatlscertmgmt. This package is a subdirectory of a kafka directory

I'll go with `kafkatlscertmgmt` though I am not that much of a big fan of acronyms / shortening names.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_service.go
```diff
@@ -0,0 +1,223 @@
...
+// GetCertificateRequest is the certificate request object.
+type GetCertificateRequest struct {
+	TLSCertRef string
+	TLSKeyRef  string
+}
+
+// areCertRefsDefined returns true if all certificate references are defined.
+// otherwise returns false
+func (req GetCertificateRequest) areCertRefsDefined() bool {
+	return !(shared.StringEmpty(req.TLSCertRef) || shared.StringEmpty(req.TLSKeyRef))
```
miguelsorianod, line 28: This can be rewritten as return `!shared.StringEmpty(req.TLSCertRef) & !shared.StringEmpty(req.TLSKeyRef)`.
I am fine with both approaches

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_service.go
```diff
@@ -0,0 +1,223 @@
...
+	RevokeCertificate(ctx context.Context, domain string, reason CertificateRevocationReason) error
+
+	// IsKafkaExternalCertificateEnabled returns whether kafka external certificate is enabled
+	IsKafkaExternalCertificateEnabled() bool
+
+	// IsAutomaticCertificateManagementEnabled returns whether automatic certificate management is enabled
+	IsAutomaticCertificateManagementEnabled() bool
+}
+
+// certificateManagementClientWrapper wrapps the certmagic.Config struct so that we can easily unit test the
```
miguelsorianod, line 61: What is certMagic? Can we get some information around it documented?

internal/kafka/internal/services/kafka_tls_certificate_management/secure_storage.go
```diff
@@ -0,0 +1,172 @@
...
+	return certmagic.KeyInfo{
+		Key:        key,
+		Modified:   *describeOutput.LastChangedDate,
+		Size:       int64(len(result)),
+		IsTerminal: strings.HasSuffix(key, "/"),
+	}, nil
+}
+
+func (storage *secureStorage) String() string {
+	return "VaultStorage"
```
miguelsorianod, line 167: I think this might be outdated

machi1990: good catch! it should be `SecureStorage`

internal/kafka/internal/services/kafka_tls_certificate_management/memory_storage.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/caddyserver/certmagic"
+)
+
+type inMemoryStorageItem struct {
+	value []byte
+	mu    *sync.Mutex
+	sync.Locker
+	lastModified time.Time
+}
+
```
miguelsorianod: It took me a while to figure it out and I am guess I would not be the only one:

This implements the `certmagic.Storage` Go interface.
Can we add interface guard clauses here to make it evident? The same should be done to the other implementations.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_service.go
machi1990: The library choosen for auto cert generation - github.com/caddyserver/certmagic 
The library is in the import block - more documentation on it can be found on their github repo.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
miguelsorianod: This depends on how you interpret this. As being part of the literal configuration or not.
It could be not validated through the struct tag and then validate it through the code

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_service.go
machi1990: I'll keep it to give @ziccardi a smile ;-)

internal/kafka/internal/services/kafka_tls_certificate_management/certificate_revocation_reason.go
miguelsorianod: Thinking about it. It might be better to have it with underscore. If it makes it hard to read or not evident then I think it is acceptable to keep it long / with underscores.

This is further reinforced by Go authors themselves https://go.dev/blog/package-names :
`On the other hand, if abbreviating a package name makes it ambiguous or unclear, don’t do it.`

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_service.go
miguelsorianod: Could we add some basic documentation around the type?

machi1990: Yes. Let me me see what can be added.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
machi1990: Ah right. Thanks. No, I've not considered it. It's possible to do it but I just ("this being an enhancement of an old PR") took inspiration on the already existing PRs. 
I am open to have this changed / revisited in follow up enhancements PRs.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
machi1990: Sorry I should have been clear.
We've a technical oriented layer approach where we put together modules that are technically similar: configs go into a config package, endpoints handlers goes onto the handler packages etc
 
Another approach is a functional approach where all modules for a particular function/feature are together. For example:
```
kafka_create_use_case/
  kafka_create_endpoint.go
  kafka_create_service.go
  kafka_create_config.go
  kafka_create._foo_bar_for_create_functionilty.go
  ....etc.go
```

internal/kafka/internal/services/kafka_tls_certificate_management/certificate_revocation_reason.go
machi1990: I do not know what happened and only seeing this now after changing to `kafkatlscertmgmt`.
I'll keep the suggested package name for now. We can always rename if it become confusing. 
Thanks.

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_service.go
machi1990: I've added something, can you check?

internal/kafka/internal/services/kafka_tls_certificate_management/kafka_tls_certificate_management_config.go
miguelsorianod: Changing this in future PR will prove difficult as once introduced in stage/prod it won't be trivial to migrate

.gitignore
```diff
@@ -45,9 +45,10 @@ cmd/kas-fleet-manager/kas-fleet-manager
 # Ignore secret files
 /secrets/db.ca_cert
 /secrets/aws.*
-/secrets/vault.*
```
miguelsorianod, line 48: Is this correct? I think vault was something that was added by connectors

miguelsorianod: See:

```

msoriano@localhost:~/go/src/github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager (...)$ git log secrets/vault/.gitignore
commit 6ff911ee7a5048045af6bd7749dee8bb7bf46ad8
Author: Dhiraj Bokde <dhirajsb@gmail.com>
Date:   Tue Nov 1 00:07:13 2022 +0000

    feat(connectors): related to APPSRE-4651, moved vault credentials config files
``

File: .gitignore [hunk 16]
```diff
@@ -45,9 +45,10 @@ cmd/kas-fleet-manager/kas-fleet-manager
...
+/secrets/aws-secret-manager.*
```
miguelsorianod, line 48: What would happen if we were to deploy in GCP?

docs/deploying-kas-fleet-manager-to-openshift.md
```diff
@@ -110,6 +111,9 @@ make deploy/secrets <OPTIONAL_PARAMETERS>
 - `OBSERVABILITY_RHSSO_METRICS_CLIENT_ID`: The client id for a RHSSO service account that has read metrics permission. Defaults to `''`
 - `OBSERVABILITY_RHSSO_METRICS_SECRET`: The client secret for a RHSSO service account that has read metrics permission. Defaults to `''`
 - `JWKS_VERIFY_INSECURE`: Skip TLS insecure verification for the connection for fetching jwks certificate. Defaults to value false.
+- `ACME_ISSUER_ACCOUNT_KEY`: The ACME Issuer account key used for the automatic management of certificate. This is required when certificate management mode is `automatic`. Defaults to `''`
+- `AWS_SECRET_MANAGER_SECRET_ACCESS_KEY`: AWS secret manager secret access key: Defaults to `''`. This is required when certificate management mode is `automatic`.
+- `AWS_SECRET_MANAGER_ACCESS_KEY`: AWS secret manager access key: Defaults to `''`. This is required when certificate management mode is `automatic`.
```
miguelsorianod, line 116: How do you control the certificate management mode? As a user, reading this line I end up with that question

miguelsorianod: same applies to other flags/params

docs/deploying-kas-fleet-manager-to-openshift.md
```diff
@@ -198,7 +202,11 @@ make deploy/service IMAGE_TAG=<your-image-tag-here> <OPTIONAL_PARAMETERS>
 - `ADMIN_API_SSO_BASE_URL`: Base URL of admin API endpints SSO. Defaults to `"https://auth.redhat.com"`
 - `ADMIN_API_SSO_ENDPOINT_URI`: admin API SSO endpoint URI. defaults to `"/auth/realms/EmployeeIDP"`
 - `ADMIN_API_SSO_REALM`: admin API SSO realm. Defaults to `"EmployeeIDP"`
-
+- `KAFKA_TLS_CERTIFICATE_MANAGEMENT_MUST_STAPLE`: The tls certificate management must staple. Adds the must staple TLS extension to the certificate signing request. The default value is `false`
+- `KAFKA_TLS_CERTIFICATE_MANAGEMENT_STRATEGY`: The tls certificate management strategy. Possible options are manual and automaitic. The default value is `manual`
```
miguelsorianod, line 206: What is manual and what is automatic?

.gitignore
machi1990: No, clearly not. I think a rename went wrong and this should stay as well.

openapi/kas-fleet-manager-private-admin.yaml
```diff
@@ -199,6 +199,54 @@ paths:
             application/json:
               schema:
                 $ref: 'kas-fleet-manager.yaml#/components/schemas/Error'
+  '/api/kafkas_mgmt/v1/admin/kafkas/{id}/revoke_certificate':
+    post:
+      description: Revokes the certificate for the Kafka instance by id
```
miguelsorianod, line 204: When we talk about "certificate" what are we talking about?

.gitignore
machi1990: You mean create a Kafka cluster in gcp?
It should be fine - it is basically just like what we do for route53. The tls certs will be stored securely in aws secret manager.

docs/deploying-kas-fleet-manager-to-openshift.md
machi1990: I think related configurations are documented right below e.g KAFKA_TLS_CERTIFICATE_MANAGEMENT_STRATEGY

openapi/kas-fleet-manager-private-admin.yaml
machi1990: Kafka tls certificate of course ;-)

openapi/kas-fleet-manager-private-admin.yaml
```diff
@@ -325,6 +373,15 @@ components:
           type: boolean
     SupportedKafkaSizeBytesValueItem:
       $ref: 'kas-fleet-manager.yaml#/components/schemas/SupportedKafkaSizeBytesValueItem'
+    KafkacertificateRevocationRequest:
+      type: object
+      properties:
+        revocation_reason:
+          type: integer
+          description: The certificate revocation reason. See https://www.rfc-editor.org/rfc/rfc5280#section-5.3.1 for the available reasons
```
miguelsorianod, line 381: Is the idea to explicitly couple this field to the x509 certificate revocation RFC intentional? Will the dataplane side need to understand that RFC?

docs/deploying-kas-fleet-manager-to-openshift.md
machi1990: I can a phrase or two for each mode but the names should be descriptive enough :-)

secrets/aws-secret-manager/aws_secret_access_key copy
```diff
@@ -0,0 +1 @@
+quOEpzR8mh/yp2Nfjhwi7YBamTZVEaSJbtmTblIQ
```
miguelsorianod, line 1: I see secrets commited here:
Some seem copies so I guess it is not intentional.
Other seem real secrets that were in the local repository. This is a security risk. Have the secrets been rotated?

openapi/kas-fleet-manager-private-admin.yaml
machi1990: > Will the dataplane side need to understand that RFC?

This is not for the data plane. 
It is an admin management API - different from the agent-clusters API (kas-fleet-manager-private.yaml openapi file).

> Is the idea to explicitly couple this field to the x509 certificate revocation RFC intentional?

Yes - it was intentional.

secrets/aws-secret-manager/aws_secret_access_key copy
machi1990: These are dummy secrets. Actually, this file should be ignored - I wonder why it is there.

miguelsorianod: Are you sure those are dummy secrets?

openapi/kas-fleet-manager-private-admin.yaml
miguelsorianod: What is a kafka tls certificate?

secrets/aws-secret-manager/aws_secret_access_key copy
machi1990: Yes.

internal/kafka/internal/config/kafka_tls_certificate_management_config.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if !arrays.Contains(validCertificateManagementStrategies, c.CertificateManagementStrategy) {
+		return fmt.Errorf("invalid certificate management strategy %q supplied. Valid strategies are %v", c.CertificateManagementStrategy, validCertificateManagementStrategies)
+	}
+
+	if c.CertificateManagementStrategy == AutomaticCertificateManagement && c.EnableKafkaExternalCertificate {
+		err := validator.New().Struct(c.AutomaticCertificateManagementConfig)
+		if err != nil {
+			return errors.Wrap(err, "error validating the automatic kafka tls  certificate management configuration")
```
pb82, line 123: nitpick: extra space between `tls  certificate`

internal/kafka/internal/services/kafka.go
```diff
@@ -881,13 +905,33 @@ func (k *kafkaService) GetManagedKafkaByClusterID(clusterID string) ([]managedka
...
+			certRequest := kafkatlscertmgmt.GetCertificateRequest{
+				TLSCertRef: kafkaRequest.KafkasRoutesBaseDomainTLSCrtRef,
+				TLSKeyRef:  kafkaRequest.KafkasRoutesBaseDomainTLSKeyRef,
+			}
+
+			var err error
+			certificate, err = k.kafkaTLSCertificateManagementService.GetCertificate(context.Background(), certRequest)
+			// TODO - gracefully handle errors so that we do not block reconciliation of others Kafkas
+			if err != nil {
+				return nil, errors.NewWithCause(errors.ErrorGeneral, err, "failed to find kafka %q certificates", kafkaRequest.ID)
```
pb82, line 925: should we log the error and continue here instead of returning? Currently, other Kafka requests don't get reconciled if an error happens.

machi1990: I'll prefer if we've a way to signal to the FSO that this MK is in erroneous state as part of the MK CR Spec. And the FSO will decide - basically skip the reconciliation on their end if they want to etc.

At the moment, if we do not include it in the list and the Kafka is already in the data plane - it'll cause a bunch of warnings in the Fleetshard side. But there could be other side effects that I do not know about. 

Also worth noting, if the error happens, it's likely going to be temporary (due to the backing storage service e.g AWS outage) and chances are that not one Kafka is going to be affect but all Kafkas.

internal/kafka/internal/services/kafka.go
```diff
@@ -1337,20 +1383,63 @@ func (k *kafkaService) AssignBootstrapServerHost(kafkaRequest *dbapi.KafkaReques
...
+	// When the strategy is automatic, there are three cases:
+	// 1. We use the Kafka domain given in the Kafka config as the base domain for already prepared Kafkas
+	// 2. The base domain name is constructed by concatenating the kafka id and the kafka domain name given in the Kafka config for all standard Kafkas in preparing state i.e Kafkas that do not have the bootstrap server host already set
+	// 3. The base domain name is constructed by concatenating the "trial" and the kafka domain name given in the Kafka config for all developer Kafkas in preparing state i.e Kafkas that do not have the bootstrap server host already set
+	if !kafkaRequest.HasCertificateInfo() {
+		if !k.kafkaTLSCertificateManagementService.IsAutomaticCertificateManagementEnabled() {
+			kafkaRequest.KafkasRoutesBaseDomainName = k.kafkaConfig.KafkaDomainName
+		} else if kafkaRequest.Status == constants.KafkaRequestStatusAccepted.String() || kafkaRequest.Status == constants.KafkaRequestStatusPreparing.String() {
+			if kafkaRequest.IsADeveloperInstance() {
+				kafkaRequest.KafkasRoutesBaseDomainName = fmt.Sprintf("%s.%s", constants.TrialKafkasDomainShard, k.kafkaConfig.KafkaDomainName)
```
pb82, line 1415: why are developer instances handled differently?

machi1990: To keep our letsencrypt usage limits sane by reducing the consumption of whatever quota we've due to dev Kafkas being spinned up - during workshop etc
The reasoning is also capture in the ADR-90:   https://github.com/bf2fc6cc711aee1a0c2a/architecture/pull/84/files#diff-9ab93b157069c0c5414cc56ec41ae2ba8e6307f5ef2dd034ac269c8258b548f9

internal/kafka/internal/services/kafkatlscertmgmt/memory_storage.go
```diff
@@ -0,0 +1,117 @@
...
+	sync.Locker
+	lastModified time.Time
+}
+
+func (item inMemoryStorageItem) Lock() {
+	item.mu.Lock()
+}
+
+func (item inMemoryStorageItem) Unlock() {
+	item.mu.Lock()
```
pb82, line 27: should this be `item.mu.Unlock`?

machi1990: good catch! Pushed the update there.

internal/kafka/internal/services/kafka.go
pb82: Right, i knew there was a reason, but couldn't remember what it was. Thanks that makes sense. No individual revokation for developer instances then.

machi1990: > No individual revokation for developer instances then.

Exactly. It is considered as a shared cert.

pb82 (approved):
LGTM

PR #1535, opened 2023-02-03
test: add integration tests for testing kafka expiration logic

JameelB:
## Description
Integration tests for testing the Kafka expiration logic from setting the `expires_at` value to deleting the Kafka when expired. 

Follows on from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1499.

## Verification Steps
- [ ] CI checks passes

- [ ] Code Review completed


machi1990:
ping @miguelsorianod @ziccardi can you've a look on this one? Thanks

internal/kafka/test/integration/kafka_expiration_test.go
```diff
@@ -0,0 +1,513 @@
...
+	mockKasfFleetshardSync.Start()
+	defer mockKasfFleetshardSync.Stop()
+
+	clusterID, getClusterErr := common.GetRunningOsdClusterID(h, t)
+	if getClusterErr != nil {
+		t.Fatalf("Failed to retrieve cluster details: %v", getClusterErr)
+	}
+	if clusterID == "" {
+		panic("No cluster found")
+	}
```
machi1990, line 189: Do we need a running OSD? Alternatively we can directly seed a cluster in the database.

internal/kafka/test/integration/kafka_expiration_test.go
```diff
@@ -0,0 +1,513 @@
...
+func TestKafka_ExpirationBasedOnLifespanSeconds(t *testing.T) {
+	g := gomega.NewWithT(t)
+
+	ocmServer := mocks.NewMockConfigurableServerBuilder().Build()
+	defer ocmServer.Close()
+
+	longLifespanSeconds := 178000
+	shortLifespanSeconds := 2
+
+	h, client, tearDown := test.NewKafkaHelperWithHooks(t, ocmServer, func(kafkaConfig *config.KafkaConfig, quotaManagementListConfig *quota_management.QuotaManagementListConfig) {
```
machi1990, line 407: do we need to set `quotaManagementListConfig.EnableInstanceLimitControl` to `true`?

internal/kafka/test/integration/kafka_expiration_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+	dummyOrgID := "dummy-org-1"
+	dummyOrgUser := "org-1-user"
+	dummyUser1 := "dummy-user-2"
+
+	defaultInstanceSize := mocksupportedinstancetypes.BuildKafkaInstanceSize(
+		mocksupportedinstancetypes.WithLifespanSeconds(nil),
+	)
+
+	h, client, tearDown := test.NewKafkaHelperWithHooks(t, ocmServer, func(kafkaConfig *config.KafkaConfig, quotaManagementListConfig *quota_management.QuotaManagementListConfig) {
```
machi1990, line 50: same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1535/files#r1105472756

internal/kafka/test/integration/kafka_expiration_test.go
```diff
@@ -0,0 +1,513 @@
...
+	ctx := h.NewAuthenticatedContext(account, nil)
+	result, resp, err := client.DefaultApi.CreateKafka(ctx, true, public.KafkaRequestPayload{
+		Name: "kafka-with-long-lifespan",
+		Plan: fmt.Sprintf("%s.x1", types.STANDARD), // has a long lifespan of 2d
+	})
+	if resp != nil {
+		_ = resp.Body.Close()
+	}
+	g.Expect(err).ToNot(gomega.HaveOccurred(), "failed to create kafka with long lifespan")
+	expectedExpirationDate := result.CreatedAt.Add(time.Duration(longLifespanSeconds) * time.Second).Format(time.RFC1123Z)
```
miguelsorianod, line 489: Why this timestamp format?

miguelsorianod: Can't the times be directly compared?

internal/kafka/test/integration/kafka_expiration_test.go
```diff
@@ -0,0 +1,513 @@
...
+	g.Expect(result.ExpiresAt.Format(time.RFC1123Z)).To(gomega.Equal(expectedExpirationDate), "expected expires_at for kafka with long lifespan does not match actual")
+
+	// wait for kafka to be ready and ensure the expiration date does not get updated
+	kafka, err := common.WaitForKafkaToReachStatus(ctx, test.TestServices.DBFactory, client, result.Id, constants.KafkaRequestStatusReady)
+	g.Expect(err).ToNot(gomega.HaveOccurred(), "failed to wait for kafka with long lifespan to reach status %q", constants.KafkaRequestStatusReady)
+	g.Expect(kafka.ExpiresAt.Format(time.RFC1123Z)).To(gomega.Equal(expectedExpirationDate), "expected expires_at for kafka with long lifespan does not match actual")
```
miguelsorianod, line 495: Why is this compared again?

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: for some reason, the time format outputted in the tests is different from my local one. So I formatted it to ensure that both dates are the same. The tests were failing the pr checks otherwise.

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: don't really need to, Kafka expiration isn't really based on instance limit with quota management list, rather it's based on if the user has the quota to create that Kafka and if that quota is still active/expired (configured by expiration_date in the kafka supported types config).

machi1990: I pointed this out as there were recently failures in nightly (in the original TestKafka_KafkaExpiration fixed by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1555) related to expiration since instance limit control is turned off: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e63f0cf8637886adf0df7d29dac13c34106f4857/internal/kafka/internal/environments/development.go#L20

Due to that, quota was always considered active https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e63f0cf8637886adf0df7d29dac13c34106f4857/internal/kafka/internal/services/quota/quota_management_list_service.go#L236-L238 

quota being active or not is used in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e63f0cf8637886adf0df7d29dac13c34106f4857/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L418 to reset the `expires_at` to nil in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e63f0cf8637886adf0df7d29dac13c34106f4857/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L430-L433

This behavior might have some on impacts on the Kafka that we expect to be deleted since they've expired https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e63f0cf8637886adf0df7d29dac13c34106f4857/internal/kafka/test/integration/kafka_expiration_test.go#L284-L332

machi1990: TL;DR:

If the test passes when OCM_ENV is set to `development` then we are good. 
If it fails, please let's consider the suggestion I raised above for both of these tests.

JameelB: Thanks @machi1990, good catch on this, I completely missed that

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/e63f0cf8637886adf0df7d29dac13c34106f4857/internal/kafka/internal/services/quota/quota_management_list_service.go#L236-L238

It makes sense to set that parameter to true in this case. 👍

miguelsorianod: This makes me wonder: Why are we considering the quota being active when instance limit control is disabled?

JameelB: @miguelsorianod that's true. Not sure if it makes sense. no matter if there's a limit control, the quota being active should really depend on whether the quota is given to the org/user and if the expiration date hasn't been reached when it's defined. 

I assume this was added to mimic the behaviour of the ReserveQuota where it skips the quota allowance check (whether org/user is given quota) + instance limit checks (whether org/user has reached their quota limit) when this flag is set to false. 

The name of the flag is misleading if this is the expected behaviour. cc @ziccardi

miguelsorianod: +1

miguelsorianod:
Hi @JameelB, what's the current state of this? I see some feedback comments there.

Thanks

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: I think this needs further discussion to figure out what the expected behaviour is. Let's not include this change with this PR. If we decide that this behaviour/parameter name needs to be changed, I can create a follow on PR with those changes.

JameelB:
@miguelsorianod the comments have been addressed. This is ready for review.

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: It's to ensure that the expiration date never gets updated. We expect the expiration date to be set on creation. KFM should not set it again during reconcile, this verifies that.

miguelsorianod: :+1:

internal/kafka/test/integration/kafka_expiration_test.go
miguelsorianod: I think this is fixable. But it can be investigated in a future change

miguelsorianod (approved):

internal/kafka/test/integration/kafka_expiration_test.go
```diff
@@ -0,0 +1,513 @@
...
+			mockkafka.WithExpiresAt(sql.NullTime{}),
+		),
+	}
+
+	// create dummy kafkas
+	var kafkas []*dbapi.KafkaRequest
+	kafkas = append(kafkas, kafkasToRemain...)
+	kafkas = append(kafkas, kafkasToExpire...)
+	if err := db.Create(&kafkas).Error; err != nil {
+		g.Expect(err).NotTo(gomega.HaveOccurred())
```
ziccardi, line 327: This will always fail

internal/kafka/test/integration/kafka_expiration_test.go
```diff
@@ -0,0 +1,513 @@
...
+	// test cases for verifying expiration logic for users registered as service account
+	// create dummy kafkas for user registered as a service account in the quota management list
+	for i, k := range kafkas {
+		k.ID = fmt.Sprintf("dummy-kafka-for-user-%d", i)
+		k.OrganisationId = "unregistered-org"
+		k.Owner = dummyUser1
+	}
+
+	if err := db.Create(&kafkas).Error; err != nil {
+		g.Expect(err).NotTo(gomega.HaveOccurred())
```
ziccardi, line 367: This will always fail

internal/kafka/test/integration/kafka_expiration_test.go
ziccardi: @miguelsorianod @JameelB the reason is that if the `EnableInstanceLimitControl` is `false`, then the user has an unlimited quota. With the previous version of the quota list, the user could create unlimited instances (disregarding the billing model)  if it was false, so the same is done here.

ziccardi (changes requested):
Some `Expect` can't succeed

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: good catch!

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: good catch!

internal/kafka/test/integration/kafka_expiration_test.go
miguelsorianod: I guess this might be actually a strange way to intentionally make it fail because of the `return` at the end.
The way it should be rewritten is:
``
err := db.Create(&kafkas).Error;
``

internal/kafka/test/integration/kafka_expiration_test.go
miguelsorianod: I think we should reevaluate this and the implications.
Conceptually it seems it might be independent. Maybe is the name that is confusing as Jameel mentioned?

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: updated there, thanks!

internal/kafka/test/integration/kafka_expiration_test.go
JameelB: updated :)

PR #1536, opened 2023-02-03
feat: add optional cluster_id field to create kafka on enterprise cluster

pawelpaszki:
## Description
feat: add optional cluster_id field to create kafka on enterprise cluster

## Verification Steps
tests passing

- [ ] Code Review completed


internal/kafka/internal/handlers/validation.go
```diff
@@ -37,9 +38,16 @@ const minimunNumberOfNodesForTheKafkaMachinePool = 3
 
 func validateKafkaBillingModel(ctx context.Context, kafkaService services.KafkaService, kafkaConfig *config.KafkaConfig, kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
 	return func() *errors.ServiceError {
+		billingModel := shared.SafeString(kafkaRequestPayload.BillingModel)
+		// enterprise kafkas billing model validation
+
+		if !shared.StringEmpty(kafkaRequestPayload.ClusterId) && !shared.StringEqualsIgnoreCase(billingModel, string(v1.BillingModelStandard)) {
```
machi1990, line 44: the billing model has to be `enterpise`: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/config/kafka-instance-types-configuration.yaml#L82

internal/kafka/internal/handlers/validation.go
```diff
@@ -37,9 +38,16 @@ const minimunNumberOfNodesForTheKafkaMachinePool = 3
...
+			return errors.InvalidBillingAccount("invalid billing model: %s, only %v is allowed", billingModel,
+				v1.BillingModelStandard)
```
machi1990, line 46: the billing mode has to be standard https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/config/kafka-instance-types-configuration.yaml#L82

internal/kafka/internal/handlers/validation.go
```diff
@@ -37,9 +38,16 @@ const minimunNumberOfNodesForTheKafkaMachinePool = 3
...
+		}
 		// No explicitly set kafka billing mode is allowed for now, in which case
 		// an implementation-defined default is chosen
-		if shared.StringEmpty(kafkaRequestPayload.BillingModel) {
+		if billingModel == "" {
```
machi1990, line 50:
```suggestion
		if shared.StringEmpty(kbillingModel) {
```

internal/kafka/internal/services/kafka.go
```diff
@@ -400,20 +402,63 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
...
+			if err != nil {
+				return errors.GeneralError("failed to get kafka instance size: %s", kafkaRequest.SizeId)
+			}
+			capacityUsed = capacityUsed + int32(instSize.CapacityConsumed)*sizeCount.Count
+		}
+		clusterCapacityUsed := int(capacityInfo[kafkaRequest.InstanceType].MaxNodes - capacityUsed)
+		if clusterCapacityUsed < instanceSize.CapacityConsumed {
+			return errors.TooManyKafkaInstancesReached("no available space left on cluster with id: %s for kafka of size: %s", kafkaRequest.ClusterID, kafkaRequest.SizeId)
+		}
+		kafkaRequest.DesiredKafkaBillingModel = "enterprise"
```
machi1990, line 445: let's not use a hardcoded billing mode here.

it'll be good to extract somewhere

internal/kafka/internal/services/kafka.go
```diff
@@ -1376,3 +1421,21 @@ func (k *kafkaService) IsQuotaEntitlementActive(kafkaRequest *dbapi.KafkaRequest
...
+	Count  int32
+}
+
+func (k *kafkaService) GetKafkaSizeCountOfEnterpriseCluster(clusterID string) ([]*ClusterSizeCount, error) {
+	dbConn := k.connectionFactory.New()
+	var sizeCountPerCluster []*ClusterSizeCount
+	if err := dbConn.Model(&dbapi.KafkaRequest{}).
+		Select("size_id, count(1) as Count").
+		Group("size_id").
+		Where("cluster_id = ?", clusterID).
```
machi1990, line 1436: We can also ignore kafka in `deleting` state.

openapi/kas-fleet-manager.yaml
```diff
@@ -1563,6 +1563,10 @@ components:
           description: billing model to use
           type: string
           nullable: true
+        cluster_id:
+          description: enterprise OSD cluster ID to be used for kafka creation
+          type: string
+          nullable: true
```
machi1990, line 1569: there is a ask to also return the cluster_id info for enterprise kafkas. 

The UI/CLI and other API consumer will base their logic off it.

internal/kafka/internal/services/kafka.go
pawelpaszki: I am aware of this. I see it as bigger refactor though to create constants for all possible billing models. currently we use hardcoded strings everywhere. would you agree to create a follow up Jira to make billing models constant and use them across the codebase?

internal/kafka/internal/handlers/kafka.go
```diff
@@ -65,6 +67,11 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
 
 			convKafka.CloudProvider, convKafka.Region, _ = getCloudProviderAndRegion(ctx, h.service, &kafkaRequestPayload, h.providerConfig)
 
+			// enterprise kafkas should be assigned to specified cluster, if its ID is provided
+			if !shared.StringEmpty(kafkaRequestPayload.ClusterId) {
+				convKafka.ClusterID = *kafkaRequestPayload.ClusterId
```
machi1990, line 72: @pawelpaszki 

What I am about to say is something that's not noted in the JIRA ask but it is something we've to do.

At the moment, we've the kafka unassignment logic https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/internal/services/data_plane_kafka.go#L396 

The intention was that, if a Kafka is rejected in the data plane due to insufficient resources, we reassign it onto a different data plane. 
This logic has to be limited only to managed kafkas and not applied for enterprise kafkas.
My proposal is then to add a check on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/internal/services/data_plane_kafka.go#L396 that if the kafka billing model is `enterprise` then it is not re-assigned i.e we keep it on the existing cluster.

Now the question becomes, what should we do if a kafka is rejected due to insufficient resources?
1. We could keep them to provisioning state and log an error message to sentry. This will be preferably, until we  know what we do under this case. 
2. change the state of the kafka to `failed` and add `cluster full` as failed_reason. This could be good for user as they see the failures reason

internal/kafka/internal/handlers/validation.go
machi1990: I also think we should validate that, once the billing model is `enterprise` - then a cluster_id has to be given.

internal/kafka/internal/services/kafka.go
```diff
@@ -400,20 +402,63 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
...
-		cluster, e := k.clusterPlacementStrategy.FindCluster(kafkaRequest)
-		if e != nil || cluster == nil {
-			msg := fmt.Sprintf("no available cluster found for Kafka instance type '%s' in region '%s'", kafkaRequest.InstanceType, kafkaRequest.Region)
-			if e != nil {
-				logger.Logger.Error(fmt.Errorf("%s:%w", msg, e))
-			} else {
-				logger.Logger.Infof(msg)
-			}
+	// enterprise kafkas have their cluster ID assigned when a request is validated in the handler
+	if kafkaRequest.ClusterID != "" {
```
machi1990, line 406: When a Kafka is an enterprise kafka (cluster_id set and billing_model is `enterprise`), then we shouldn't verify: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/internal/services/kafka.go#L391-L403

machi1990, line 406: I think this logic merits its place in the different placement strategies as it is one of the placement strategies.

I propose that this logic being moved onto https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/internal/services/cluster_placement_strategy.go 

and for each of the existing placement strategies (which are current scaling type dependent)

1. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/internal/services/cluster_placement_strategy.go#L35
2. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/internal/services/cluster_placement_strategy.go#L60
3. https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/internal/services/cluster_placement_strategy.go#L142

We've the check, "if the Kafka is enterprise (billing model is `enterprise` and cluster_id set)"  then apply this logic.
The advantage is that, the placement strategy is co-located in one place.

internal/kafka/internal/services/kafka.go
```diff
@@ -1376,3 +1421,21 @@ func (k *kafkaService) IsQuotaEntitlementActive(kafkaRequest *dbapi.KafkaRequest
...
+type ClusterSizeCount struct {
+	SizeId string
+	Count  int32
+}
+
+func (k *kafkaService) GetKafkaSizeCountOfEnterpriseCluster(clusterID string) ([]*ClusterSizeCount, error) {
+	dbConn := k.connectionFactory.New()
+	var sizeCountPerCluster []*ClusterSizeCount
+	if err := dbConn.Model(&dbapi.KafkaRequest{}).
+		Select("size_id, count(1) as Count").
```
machi1990, line 1434: might be good to also return the kafka instance type as the streaming unit notion is dependent on this

machi1990, line 1434:
```suggestion
		Select("size_id, instance_type, count(1) as Count").
```

internal/kafka/internal/services/kafka.go
```diff
@@ -1376,3 +1421,21 @@ func (k *kafkaService) IsQuotaEntitlementActive(kafkaRequest *dbapi.KafkaRequest
...
+		Group("size_id").
```
machi1990, line 1435:
```suggestion
		Group("size_id, instance_type").
```

internal/kafka/internal/services/kafka.go
```diff
@@ -400,20 +402,63 @@ func (k *kafkaService) RegisterKafkaJob(kafkaRequest *dbapi.KafkaRequest) *error
...
+		sizeCountOfCluster, e := k.GetKafkaSizeCountOfEnterpriseCluster(kafkaRequest.ClusterID)
+
+		if e != nil {
+			return errors.GeneralError("failed to establish remaining capacity of cluster with id: %s", kafkaRequest.ClusterID)
+		}
+
+		capacityUsed := int32(0)
+
+		for _, sizeCount := range sizeCountOfCluster {
+			instSize, err := k.kafkaConfig.GetKafkaInstanceSize(types.STANDARD.String(), sizeCount.SizeId)
```
machi1990, line 435:
```suggestion
			instSize, err := k.kafkaConfig.GetKafkaInstanceSize(kafkaRequest.InstanceType, sizeCount.SizeId)
```

internal/kafka/internal/services/kafka.go
machi1990, line 1436: don't we need to also check if the cluster_id is an enterprise cluster?

machi1990: Asking this because the method name `GetKafkaSizeCountOfEnterpriseCluster` has enterprise on it, and there is no where we are checking if it is an enterprise cluster

internal/kafka/internal/handlers/validation.go
pawelpaszki: `the billing mode has to be standard` - you mean enterprise? 

`I also think we should validate that, once the billing model is enterprise - then a cluster_id has to be given.` - isn't it already done here `if !shared.StringEmpty(kafkaRequestPayload.ClusterId)`?

internal/kafka/internal/services/kafka.go
```diff
@@ -1376,3 +1421,21 @@ func (k *kafkaService) IsQuotaEntitlementActive(kafkaRequest *dbapi.KafkaRequest
 
 	return quotaService.IsQuotaEntitlementActive(kafkaRequest)
 }
+
+type ClusterSizeCount struct {
+	SizeId string
+	Count  int32
+}
+
+func (k *kafkaService) GetKafkaSizeCountOfEnterpriseCluster(clusterID string) ([]*ClusterSizeCount, error) {
```
machi1990, line 1430: I think we could enhance this method to only return the streaming unit count per instance type. 
```suggestion
func (k *kafkaService) ComputeUsedStreamingUnitCountOfCluster(clusterID string) (StreamingUnitCountPerInstanceType, error) {
```

and StreamingUnitCountPerInstanceType  

```
type StreamingUnitCountPerInstanceType map[KafkaInstanceType]int
```

It also makes sense for this method to be moved onto cluster_service.go file.

internal/kafka/test/integration/kafka_create_test.go
```diff
@@ -605,3 +606,106 @@ func TestKafkaCreate_ValidatePlanParam(t *testing.T) {
 	}
 	g.Expect(err).To(gomega.HaveOccurred(), "error should have occurred when attempting to create kafka unsupported size_id")
 }
+
+func TestKafkaCreate_EnterpriseKafkas(t *testing.T) {
```
machi1990, line 610: <3

internal/kafka/internal/handlers/validation.go
machi1990: > `the billing mode has to be standard` - you mean enterprise?

Sorry, yes. It has to be enterprise.
 
> `I also think we should validate that, once the billing model is enterprise - then a cluster_id has to be given.` - isn't it already done here `if !shared.StringEmpty(kafkaRequestPayload.ClusterId)`?

I can see the condition but I do not think if it covers the case where: someone attempts to create a Kafka with the enterprise billing but they've not supplied a clusterID.
 
What I am proposing is to add a check if `billingModel == "enterprise" && clusterID == "" then error("You must supply a cluster id")`

internal/kafka/internal/services/kafka.go
pawelpaszki: added constant

machi1990: The PR alone will be using the "enterprise" constant at least twice. 
We could just start by moving it onto the https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ba3f5f14f64443a12a20a2df6d38e045b3637054/internal/kafka/constants/kafka.go constants file and we'll see what can be done about the other billing modes in separate PRs

internal/kafka/internal/handlers/validation.go
pawelpaszki: changed

machi1990: Thanks

internal/kafka/internal/handlers/kafka.go
machi1990: I will go with option (1)

internal/kafka/internal/handlers/kafka.go
```diff
@@ -65,6 +67,11 @@ func (h kafkaHandler) Create(w http.ResponseWriter, r *http.Request) {
 
 			convKafka.CloudProvider, convKafka.Region, _ = getCloudProviderAndRegion(ctx, h.service, &kafkaRequestPayload, h.providerConfig)
 
+			// enterprise kafkas should be assigned to specified cluster, if its ID is provided
+			if !shared.StringEmpty(kafkaRequestPayload.ClusterId) {
```
machi1990, line 71: This needs to be moved to `ConvertKafkaRequest`

ziccardi, line 71: Should we use 
```suggestion
			if !shared.StringEmpty(kafkaRequestPayload.ClusterId, true) {
```
so that even `all spaces` is considered `empty` ?

machi1990: No need for this. The `ValidateNotEmptyClusterId` validator already takes care of this.

machi1990:
@pb82 can you've a look at this?

internal/kafka/internal/handlers/validation.go
```diff
@@ -37,9 +37,22 @@ const minimunNumberOfNodesForTheKafkaMachinePool = 3
 
 func validateKafkaBillingModel(ctx context.Context, kafkaService services.KafkaService, kafkaConfig *config.KafkaConfig, kafkaRequestPayload *public.KafkaRequestPayload) handlers.Validate {
 	return func() *errors.ServiceError {
+		billingModel := shared.SafeString(kafkaRequestPayload.BillingModel)
+		// enterprise kafkas billing model validation
+
+		if !shared.StringEmpty(kafkaRequestPayload.ClusterId) && !shared.StringEqualsIgnoreCase(billingModel, constants.BillingModelEnterprise.String()) {
+			return errors.InvalidBillingAccount("invalid billing model: %s, only %v is allowed", billingModel,
```
pb82, line 44: can be `%s` since `.String()` is called on the billing model. Or simply always use `%v`.

machi1990: +1, I'll change it.

internal/kafka/internal/handlers/validation.go
```diff
@@ -199,12 +212,14 @@ func getCloudProviderAndRegion(
 	providerName := arrays.FirstNonEmptyOrDefault(defaultProvider.Name, kafkaRequest.CloudProvider)
 	// Validation for Cloud Provider
 	provider, providerSupported := supportedProviders.GetByName(providerName)
-	if !providerSupported {
+
+	// We only return a validation error if the provider is not supported and Kafka doesn't have a dedicated data plane cluster assigned to it
+	if !providerSupported && shared.StringEmpty(kafkaRequest.ClusterId) {
```
pb82, line 217: should we add an additional check for instance type == enterprise here? Cluster id is (currently at least) only supported for enterprise, in all other cases the normal validation should take place.

pb82: On second thought, that makes the condition somewhat hard to reason about. What do you think?

machi1990: That's already done here in the `validateKafkaBillingModel` validator

internal/kafka/internal/services/cluster_placement_strategy.go
```diff
@@ -22,17 +24,79 @@ func NewClusterPlacementStrategy(clusterService ClusterService, dataplaneCluster
...
 	default:
-		clusterSelection = &FirstReadyCluster{clusterService}
+		clusterSelection = &FirstReadyCluster{clusterService, kafkaConfig}
 	}
 	return clusterSelection
 }
 
+// findDesiredDataPlaneClusterIfItHasCapacityAvailable finds and returns the desired data plane cluster if
+// if it has capacity available.
+type findDesiredDataPlaneClusterIfItHasCapacityAvailable struct {
```
pb82, line 34: Is this the name of the placement strategy? Would something like `findDataPlaneClusterByIdIfItHasCapacityAvailable` be more clear?

machi1990: Yes, that's the name of the placement strategy. 
Notice how it has been called from the other placement strategy as well.

I am okay with changing the name to what you are suggesting.

internal/kafka/internal/handlers/validation.go
machi1990: done

internal/kafka/internal/services/cluster_placement_strategy.go
machi1990: done

internal/kafka/internal/services/clusters.go
```diff
@@ -881,3 +889,46 @@ func (c *clusterService) FindStreamingUnitCountByClusterAndInstanceType() (Kafka
...
+		instSize, err := c.kafkaConfig.GetKafkaInstanceSize(sizeCount.InstanceType, sizeCount.SizeId)
+		if err != nil {
+			return nil, err
+		}
+
+		streamingUnitCount := int64(instSize.CapacityConsumed) * sizeCount.Count
+		instanceType := kafkaTypes.KafkaInstanceType(sizeCount.InstanceType)
+		existingCount, ok := streamingUnitCountPerInstanceType[instanceType]
+
+		if !ok {
```
pb82, line 926: The map is already initialized with the instance types and a count of zero. We could shorten this to just:
```
streamingUnitCountPerInstanceType[instanceType] += streamingUnitCount
```

machi1990:
Thanks @pb82

internal/kafka/constants/kafka.go
```diff
@@ -130,3 +130,13 @@ func GetSuspendedStatuses() []string {
 func GetDeletingStatuses() []string {
 	return []string{KafkaRequestStatusDeprovision.String(), KafkaRequestStatusDeleting.String()}
 }
+
+type BillingModel string
```
miguelsorianod, line 134: Can we call it `KafkaBillingModel`?
We currently have several billing model levels (kafka billing model, ams billing model, quota management lis billing model, ...) and this would help us differentiate them

machi1990: > I am assuming this refers to a Kafka billing model and not to an AMS billing model. Could you tell me which one is the case?

The assumption is correct. This is a kafka billing model not ams billing model.

> Assuming it is the former, can we call it KafkaBillingModel?

There is already a `KafkaBillingModel`, maybe a rename to `KafkaBillingModelID` might be best suitable, what do you think?

> We currently have several billing model levels (kafka billing model, ams billing model, quota management lis billing model, ...) and this would help us differentiate them

I agree.

miguelsorianod: If there is already a `KafkaBillingModel` why are we not using that?

machi1990: It's two different concepts. This code wants the ID of the billing model.

miguelsorianod: The ID is taken from the configuration

PR #1538, opened 2023-02-05
fix: removed call to reserved_resources

ziccardi:
## Description
we are getting a 403 error when we try to access the reserved_resources of certain subscriptions. After some discussion, fixing that permission error is not trivial, so we discussed an alternative implementation that won't involve reserved_resources.
This commit implements that solution: take the billing model from the `cluster_billing_model` of the subscription.

Another change is the way we detect if a subscription for a cluster already exists: after some discussion with the AMS team, we discovered that we couldn't have more than one subscription with the same `cluster_id` value. That means we no longer need to check the reserved resources: if a subscription for a given cluster_id exists, then it is the subscription we are searching for.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/quota/ams_quota_service.go
+++ b/internal/kafka/internal/services/quota/ams_quota_service.go
@@ -2,10 +2,6 @@ package quota
 
 import (
 	"fmt"
-	"net/http"
-
-	v1 "github.com/openshift-online/ocm-sdk-go/authorizations/v1"
-
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/cloudproviders"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
@@ -307,7 +303,17 @@ func (q amsQuotaService) ReserveQuota(kafka *dbapi.KafkaRequest) (string, *error
 	return resp.Subscription().ID(), nil
 }
 
+// ReserveQuotaIfNotAlreadyReserved reserves quota for the received KafkaRequest only if no quota has already been assigned for
+// the `KafkaRequest.cluster_id`
+// Returns the ID of the subscription associated to this cluster (whether it is a new one or not)
 func (q amsQuotaService) ReserveQuotaIfNotAlreadyReserved(kafka *dbapi.KafkaRequest) (string, *errors.ServiceError) {
+	getClusterID := func(kafka *dbapi.KafkaRequest) string {
+		if kafka.DesiredKafkaBillingModel == "eval" {
+			return kafka.ID + "-eval"
+		}
+		return kafka.ID
+	}
+
 	instanceType, err := q.kafkaConfig.SupportedInstanceTypes.Configuration.GetKafkaInstanceTypeByID(kafka.InstanceType)
 	if err != nil {
 		return "", errors.GeneralError("failed checking current quota: %v", err)
@@ -317,25 +323,28 @@ func (q amsQuotaService) ReserveQuotaIfNotAlreadyReserved(kafka *dbapi.KafkaRequ
 		return "", errors.GeneralError("failed checking current quota: %v", err)
 	}
 
-	subscriptions, err := q.amsClient.FindSubscriptions(fmt.Sprintf("cluster_id='%s'", kafka.ClusterID))
+	subscriptions, err := q.amsClient.FindSubscriptions(fmt.Sprintf("cluster_id='%s'", getClusterID(kafka)))
 	if err != nil {
 		return "", errors.GeneralError("failed checking current quota: %v", err)
 	}
-	for _, subscription := range subscriptions {
-		// get the reserved resources
-		if subscription.Plan().ID() == kafkaBillingModel.AMSProduct && subscription.Status() == string(v1.SubscriptionStatusActive) {
-			reservedResources, err := q.amsClient.GetReservedResourcesBySubscriptionID(subscription.ID())
-			if err != nil {
-				return "", errors.GeneralError("failed checking current quota: %v", err)
-			}
 
-			if arrays.AnyMatch(reservedResources, func(rr *amsv1.ReservedResource) bool { return rr.ResourceName() == kafkaBillingModel.AMSResource }) {
-				return subscription.ID(), nil
-			}
+	if len(subscriptions) > 0 {
+		if len(subscriptions) > 1 {
+			logger.Logger.V(10).Errorf("found more than one subscription for cluster id %q", getClusterID(kafka))
 		}
+
+		// at most one subscription can exist for one cluster_id. This is enforced by AMS
+		subscription := subscriptions[0]
+
+		// should never enter this `if`. We check it to be extremely sure
+		if !kafkaBillingModel.HasSupportForAMSBillingModel(string(subscription.ClusterBillingModel())) {
+			return "", errors.GeneralError("failed checking current quota. Expected billing model to be one of %v, but found %q", kafkaBillingModel.AMSBillingModels, string(subscription.ClusterBillingModel()))
+		}
+
+		return subscription.ID(), nil
 	}
 
-	// quota is not defined. We need to reserve a new one
+	// no subscriptions found. We need to reserve new quota.
 	return q.ReserveQuota(kafka)
 }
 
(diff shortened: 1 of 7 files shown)


internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -388,27 +383,55 @@ func (q amsQuotaService) DeleteQuotaForBillingModel(subscriptionID string, kafka
...
 
-	amsRelatedResourceBillingModel, err := q.getAMSRelatedResourceBillingModel(kafka.SubscriptionId, kafkaBillingModel)
-	if err != nil || shared.StringEmpty(amsRelatedResourceBillingModel) {
-		return false, err
+	// Using a new `err1` otherwise I would get `nil is not nil` error here since GetSubscriptionByID returns *ServiceError
+	subscription, ok, err1 := q.GetSubscriptionByID(kafka.SubscriptionId)
+	if err1 != nil || !ok {
+		return false, err1
+	}
+	amsRelatedResourceBillingModel, ok := q.amsBillingModelToAMSBillingModelCategory(string(subscription.ClusterBillingModel()), kafkaBillingModel)
```
miguelsorianod, line 430: This variable name is outdated.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -388,27 +383,55 @@ func (q amsQuotaService) DeleteQuotaForBillingModel(subscriptionID string, kafka
...
 func (q amsQuotaService) IsQuotaEntitlementActive(kafka *dbapi.KafkaRequest) (bool, error) {
 	kafkaBillingModel, err := q.kafkaConfig.GetBillingModelByID(kafka.InstanceType, kafka.ActualKafkaBillingModel)
 	if err != nil {
 		return false, err
 	}
 
-	amsRelatedResourceBillingModel, err := q.getAMSRelatedResourceBillingModel(kafka.SubscriptionId, kafkaBillingModel)
-	if err != nil || shared.StringEmpty(amsRelatedResourceBillingModel) {
-		return false, err
+	// Using a new `err1` otherwise I would get `nil is not nil` error here since GetSubscriptionByID returns *ServiceError
```
miguelsorianod, line 425: When I have those scenarios I usually name the error variables:
`svcErr` for ServiceError(s)
`err` for standard go errors

miguelsorianod:
Side note: It is a good practice to write commits in present-tense imperative style and not in past tense. See https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181 imperative-mood section. When writing commits you can think as: "If applied, this commit will ...." .

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -317,25 +323,24 @@ func (q amsQuotaService) ReserveQuotaIfNotAlreadyReserved(kafka *dbapi.KafkaRequ
...
-			if err != nil {
-				return "", errors.GeneralError("failed checking current quota: %v", err)
-			}
 
-			if arrays.AnyMatch(reservedResources, func(rr *amsv1.ReservedResource) bool { return rr.ResourceName() == kafkaBillingModel.AMSResource }) {
-				return subscription.ID(), nil
-			}
+	if len(subscriptions) > 0 {
+		// at most one subscription can exist for one cluster_id. This is enforced by AMS
+		subscription := subscriptions[0]
```
JameelB, line 333: It shouldn't be possible but can we add a log here or an error just in case we get more than 1 subscription? then at least we'd be able to know.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -388,27 +383,55 @@ func (q amsQuotaService) DeleteQuotaForBillingModel(subscriptionID string, kafka
...
 	if err != nil {
 		return nil, false, errors.NewWithCause(errors.ErrorGeneral, err, "error getting the subscriptions for id '%s'", subscriptionID)
 	}
 
-	return resp.Body(), true, nil
+	return subscription, ok, nil
 }
 
-// Checks if an organisation has a SKU entitlement and that entitlement is still active in AMS
+func (q amsQuotaService) amsBillingModelToAMSBillingModelCategory(amsBillingModel string, kafkaBillingModel config.KafkaBillingModel) (string, bool) {
```
JameelB, line 394: This method name can be confusing, can we add a comment describing this method explaining what is 'ams billing model' and 'ams billing model category'

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -388,27 +383,55 @@ func (q amsQuotaService) DeleteQuotaForBillingModel(subscriptionID string, kafka
...
+	// Using a new `err1` otherwise I would get `nil is not nil` error here since GetSubscriptionByID returns *ServiceError
+	subscription, ok, err1 := q.GetSubscriptionByID(kafka.SubscriptionId)
+	if err1 != nil || !ok {
+		return false, err1
+	}
+	amsRelatedResourceBillingModel, ok := q.amsBillingModelToAMSBillingModelCategory(string(subscription.ClusterBillingModel()), kafkaBillingModel)
+	if !ok {
+		logger.Logger.V(10).Infof("incompatible ams billing model found (%q) for kafka billing model %q in subscription %q",
+			string(subscription.ClusterBillingModel()), kafka.SubscriptionId, kafkaBillingModel.ID)
+		return false, nil
```
JameelB, line 434: If the ams billing model can't be found, i would've thought this should return an error?

internal/kafka/internal/services/quota/ams_quota_service_test.go
```diff
@@ -2390,130 +2331,3 @@ func Test_amsQuotaService_IsQuotaEntitlementActive(t *testing.T) {
 		})
 	}
 }
-
-func Test_amsQuotaService_getAMSRelatedResourceBillingModel(t *testing.T) {
```
JameelB, line 2394: I think we should still have a unit test for this refactored method to ensure we get the correct billing model.

internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -388,27 +383,55 @@ func (q amsQuotaService) DeleteQuotaForBillingModel(subscriptionID string, kafka
...
+		if kafkaBillingModel.HasSupportForAMSBillingModel(amsBillingModel) {
+			return ocm.AMSRelatedResourceBillingModelMarketplace.String(), true
+		}
+	case shared.StringEqualsIgnoreCase(amsBillingModel, ocm.AMSRelatedResourceBillingModelStandard.String()):
+		if kafkaBillingModel.HasSupportForStandard() {
+			return ocm.AMSRelatedResourceBillingModelStandard.String(), true
+		}
+	}
+
+	//logger.Logger.V(10).Infof("ams related resource billing model not found for subscription %q and kafka billing model %q", subscriptionID, kafkaBillingModel.ID)
```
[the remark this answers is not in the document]
ziccardi: No, it is not logged below (line 432). Deleting

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: It wasn't in the previous implementation, so I kept the same behaviour

internal/kafka/internal/services/quota/ams_quota_service_test.go
ziccardi: The method has been deleted. We don't call `reserved resources` anymore.

JameelB: but it's been refactored to `amsBillingModelToAMSBillingModelCategory` which should have its own unit test. I think this can be refactored to accommodate for the changes there or create a new one.

miguelsorianod:
> Side note: It is a good practice to write commits in present-tense imperative style and not in past tense. See https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181 imperative-mood section. When writing commits you can think as: "If applied, this commit will ...." .

This means the commit in this PR can be renamed to:
`fix: remove call to reserved_resources`

internal/kafka/internal/services/quota/ams_quota_service_test.go
ziccardi: It has not been refactored in that method. It is a totally different method.
`amsBillingModelToAMSBillingModelCategory` just returns `marketplace` for any marketplace ams billing model and `standard` for `standard`.

ziccardi:
> > Side note: It is a good practice to write commits in present-tense imperative style and not in past tense. See https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181 imperative-mood section. When writing commits you can think as: "If applied, this commit will ...." .
> 
> This means the commit in this PR can be renamed to: `fix: remove call to reserved_resources`

Yup, I know. Sorry, I forgot to update the commit message. Done now.

internal/kafka/internal/services/quota/ams_quota_service_test.go
miguelsorianod: What Jameel is trying to say is that we should have unit tests for it. It is a new method.

miguelsorianod (approved):
Looks good to me.

To me this can be merged once the tests pass

internal/kafka/internal/services/quota/ams_quota_service_test.go
ziccardi: I see. I'm not a fan of tests on private interface, other than that the method is as simple as a switch on 3 values. I don't thin it worths a test. In any case, if we want, I agree with @JameelB in adding it in #1549

miguelsorianod: > I see. I'm not a fan of tests on private interface, other than that the method is as simple as a switch on 3 values. I don't thin it worths a test. In any case, if we want, I agree with @JameelB in adding it in #1549

As a side comment, I think the answer to whether testing private interface is good or not is "it depends". It is not as simple as "private tests shouldn't be tested".
Sometimes the public API of a type is very broad and then it is reasonable to test private methods. Of course, if the public API of a type is very broad this is a potential symptom of a design issue where the class "doing too much". For that the alternative is having smaller types that do more specific things.
But, as we know that's not the code we have / can control /change in a reasonable time in some cases, and even the small types might be package-private so the same question would apply.

PR #1550, opened 2023-02-07
fix: replace sed separator with '#'

ziccardi:
## Description

New secrets contain the '/' char, causing sed to fail.

- [ ] Code Review completed

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -43,18 +43,18 @@ if [[ -z "${OCM_ENV}" ]] || [[ "${OCM_ENV}" == "${INTEGRATION_ENV}" ]];
 then
   cp docker/Dockerfile_template_mocked Dockerfile_integration_tests
 else
-  if [[ -z "${OCM_ENV}" ]] || [[ -z "${AWS_ACCESS_KEY}" ]] || [[ -z "${AWS_ACCOUNT_ID}" ]] || 
-    [[ -z "${AWS_SECRET_ACCESS_KEY}" ]] || [[ -z "${OCM_OFFLINE_TOKEN}" ]] || [[ -z "${OBSERVATORIUM_CONFIG_ACCESS_TOKEN}" ]] || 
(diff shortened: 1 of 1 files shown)


miguelsorianod (commented):
I see more than the aws secret access key related code has been changed.

This assumes `#` is a character that cannot appear in any of the elements that have been changed.

Has that  been verified?

ziccardi:
> I see more than the aws secret access key related code has been changed.
> 
> This assumes `#` is a character that cannot appear in any of the elements that have been changed.
> 
> Has that been verified?

AWS account ID: composed by 12 digit 
AWS access Key ID: only characters. See [here](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetAccessKeyInfo.html)
AWS secret access key: (?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{40}(?![A-Za-z0-9/+=]). (see [here](https://aws.amazon.com/it/blogs/security/a-safer-way-to-distribute-aws-credentials-to-ec2/))
OCM_OFFLINE_TOKEN: concatenated base64 strings - no hash
OBSERVATORIUM= \w{3}_base 64 - no hash

machi1990 (approved):

miguelsorianod:
:+1: . A note about the AWS access key and private access key is that they offer no guarantee about the set of characters that they will use. They can change it at any moment. But there's not much we can do and it does not seem very likely so we can go ahead with it

ziccardi:
> 👍 . A note about the AWS access key and private access key is that they offer no guarantee about the set of characters that they will use. They can change it at any moment. But there's not much we can do and it does not seem very likely to change so we can go ahead with it

They can of course change the format in the future. Currently, however they say the format is [\w]*:
image

miguelsorianod:
> > +1 . A note about the AWS access key and private access key is that they offer no guarantee about the set of characters that they will use. They can change it at any moment. But there's not much we can do and it does not seem very likely to change so we can go ahead with it
> 
> They can of course change the format in the future. Currently, however they say the format is [\w]*: image

Yes. That's for access key.
For secret access key they don't seem to document it in the API reference

PR #1554, opened 2023-02-08
feat: add database dump to integration tests

ziccardi:
## Description

Add database dump (kafka_requests) to integration test

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/test/common/kafka_pollers.go
+++ b/internal/kafka/test/common/kafka_pollers.go
@@ -24,10 +24,10 @@ const (
 )
 
 // WaitForNumberOfKafkaToBeGivenCount - Awaits for the number of kafkas to be exactly X
-func WaitForNumberOfKafkaToBeGivenCount(ctx context.Context, db *db.ConnectionFactory, client *public.APIClient, count int32) error {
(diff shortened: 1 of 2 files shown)


internal/kafka/test/common/kafka_pollers.go
```diff
@@ -46,8 +46,13 @@ func WaitForNumberOfKafkaToBeGivenCount(ctx context.Context, db *db.ConnectionFa
 				currentCount = list.Size
 				return currentCount == count, err
 			}
-		}).
-		Build().Poll()
+		})
+
+	if len(customize) > 0 {
+		pb = customize[0](pb)
```
JameelB, line 52: why do this only execute the first function? should it iterate and execute all of the customize functions passed?

ziccardi: No. I used the `var args` only to make the parameter optional and avoid breaking all the tests.
Only the first one is used.

JameelB: since there's no stopping people from passing multiple functions here, i think it would be great to iterate and apply all but since we don't have a use case for that at the moment, i'm fine with this and we can revisit at a later point if needed 👍

JameelB (approved):

PR #1560, opened 2023-02-10
feat: add service account and image pull secrets to observability nam…

pb82:
This PR adds two additional secrets to the observability namespace:

1. The image pull secret required to pull the image for the observability resources from quay
2. A secret containing the data plane sso credentials. This is used by Prometheus to obtain a token to authenticate with the Observatorium proxy.

These changes are in preparation for the Observability updates for Enterprise, where we will remove all shared credentials from the data plane cluster.

### DIFF
--- a/internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
+++ b/internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
@@ -38,26 +38,25 @@ import (
 )
 
 const (
-	observabilityNamespace                     = constants.ObservabilityOperatorNamespace
-	observabilityOperatorGroupName             = "observability-operator-group-name"
-	observabilityCatalogSourceName             = "observability-operator-manifests"
-	observabilitySubscriptionName              = "observability-operator"
-	observatoriumDexSecretName                 = "observatorium-configuration-dex"
-	observatoriumSSOSecretName                 = "observatorium-configuration-red-hat-sso"
-	syncsetName                                = "ext-managedservice-cluster-mgr"
-	strimziAddonNamespace                      = constants.StrimziOperatorNamespace
-	strimziQEAddonNamespace                    = "redhat-managed-kafka-operator-qe"
-	kasFleetshardAddonNamespace                = constants.KASFleetShardOperatorNamespace
-	kasFleetshardQEAddonNamespace              = "redhat-kas-fleetshard-operator-qe"
-	openIDIdentityProviderName                 = "Kafka_SRE"
-	mkReadOnlyGroupName                        = "mk-readonly-access"
-	mkSREGroupName                             = "kafka-sre"
-	mkReadOnlyRoleBindingName                  = "mk-dedicated-readers"
-	mkSRERoleBindingName                       = "kafka-sre-cluster-admin"
-	dedicatedReadersRoleBindingName            = "dedicated-readers"
-	clusterAdminRoleName                       = "cluster-admin"
-	kafkaInstanceProfileType                   = "bf2.org/kafkaInstanceProfileType"
-	observabilityProxyServiceAccountSecretName = "observability-proxy-credentials"
+	observabilityNamespace          = constants.ObservabilityOperatorNamespace
+	observabilityOperatorGroupName  = "observability-operator-group-name"
+	observabilityCatalogSourceName  = "observability-operator-manifests"
+	observabilitySubscriptionName   = "observability-operator"
+	observatoriumDexSecretName      = "observatorium-configuration-dex"
+	observatoriumSSOSecretName      = "observatorium-configuration-red-hat-sso"
+	syncsetName                     = "ext-managedservice-cluster-mgr"
+	strimziAddonNamespace           = constants.StrimziOperatorNamespace
+	strimziQEAddonNamespace         = "redhat-managed-kafka-operator-qe"
+	kasFleetshardAddonNamespace     = constants.KASFleetShardOperatorNamespace
+	kasFleetshardQEAddonNamespace   = "redhat-kas-fleetshard-operator-qe"
+	openIDIdentityProviderName      = "Kafka_SRE"
+	mkReadOnlyGroupName             = "mk-readonly-access"
+	mkSREGroupName                  = "kafka-sre"
+	mkReadOnlyRoleBindingName       = "mk-dedicated-readers"
+	mkSRERoleBindingName            = "kafka-sre-cluster-admin"
+	dedicatedReadersRoleBindingName = "dedicated-readers"
+	clusterAdminRoleName            = "cluster-admin"
+	kafkaInstanceProfileType        = "bf2.org/kafkaInstanceProfileType"
 )
 
 var clusterMetricsStatuses = []api.ClusterStatus{
@@ -112,6 +111,7 @@ type ClusterManagerOptions struct {
 	ClusterService             services.ClusterService
 	CloudProvidersService      services.CloudProvidersService
 	KasFleetshardOperatorAddon services.KasFleetshardOperatorAddon
+	SsoService                 sso.KafkaKeycloakService
 	OsdIdpKeycloakService      sso.OsdKeycloakService
 	ProviderFactory            clusters.ProviderFactory
 }
@@ -771,7 +771,7 @@ func (c *ClusterManager) buildResourceSet(cluster api.Cluster) types.ResourceSet
 		c.buildObservabilityCatalogSourceResource(),
 		c.buildObservabilityOperatorGroupResource(),
 		c.buildObservabilitySubscriptionResource(),
-		c.buildDataplaneServiceaccountSecretResource(&cluster),
+		c.buildObservabilityRemoteWriteServiceAccountCredential(&cluster),
 	)
 
 	strimziNamespace := strimziAddonNamespace
@@ -879,15 +879,16 @@ func (c *ClusterManager) buildObservatoriumDexSecretResource() *k8sCoreV1.Secret
 	}
 }
 
-func (c *ClusterManager) buildDataplaneServiceaccountSecretResource(cluster *api.Cluster) *k8sCoreV1.Secret {
+func (c *ClusterManager) buildObservabilityRemoteWriteServiceAccountCredential(cluster *api.Cluster) *k8sCoreV1.Secret {
 	return &k8sCoreV1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      observabilityProxyServiceAccountSecretName,
+			Name:      "observability-proxy-credentials",
 			Namespace: observabilityNamespace,
 		},
 		StringData: map[string]string{
 			"client_id":     cluster.ClientID,
 			"client_secret": cluster.ClientSecret,
+			"issuer_url":    c.SsoService.GetRealmConfig().ValidIssuerURI,
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -874,6 +879,19 @@ func (c *ClusterManager) buildObservatoriumDexSecretResource() *k8sCoreV1.Secret
 	}
 }
 
+func (c *ClusterManager) buildDataplaneServiceaccountSecretResource(cluster *api.Cluster) *k8sCoreV1.Secret {
```
miguelsorianod, line 882: I find this name might be too generic:
* What is a ServiceAccount in the context of clusters_mgr?
* Is this "service account" used for multiple functionalities in the data plane cluster?

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -812,6 +814,9 @@ func (c *ClusterManager) buildResourceSet(cluster api.Cluster) types.ResourceSet
 	if s := c.buildImagePullSecret(kasFleetshardNamespace); s != nil {
 		r = append(r, s)
 	}
+	if s := c.buildImagePullSecret(observabilityNamespace); s != nil {
```
miguelsorianod, line 817: Is the image pull secret content valid to pull images from the repository where the observability image proxy will be located? The `buildImagePullSecret` method always reads from the same file and therefore from the same set of image pull credentials

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -38,25 +38,26 @@ import (
...
+	kasFleetshardQEAddonNamespace              = "redhat-kas-fleetshard-operator-qe"
+	openIDIdentityProviderName                 = "Kafka_SRE"
+	mkReadOnlyGroupName                        = "mk-readonly-access"
+	mkSREGroupName                             = "kafka-sre"
+	mkReadOnlyRoleBindingName                  = "mk-dedicated-readers"
+	mkSRERoleBindingName                       = "kafka-sre-cluster-admin"
+	dedicatedReadersRoleBindingName            = "dedicated-readers"
+	clusterAdminRoleName                       = "cluster-admin"
+	kafkaInstanceProfileType                   = "bf2.org/kafkaInstanceProfileType"
+	observabilityProxyServiceAccountSecretName = "observability-proxy-credentials"
```
miguelsorianod, line 60: Is the `observabilityProxyServiceAccountSecretName` value going to be used outside of the  `buildDataplaneServiceaccountSecretResource` method? If not, I'd prefer having it inside the method. The scope that should be located is the minimal scope where it will be needed

miguelsorianod:
> These changes are in preparation for the Observability updates for Enterprise, where we will remove all shared credentials from the data plane cluster.

Could you expand on this a little bit? Can you describe the current scenario we are and the scenario we are going to? In this PR I see credentials on the data planes being deployed too

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -874,6 +879,19 @@ func (c *ClusterManager) buildObservatoriumDexSecretResource() *k8sCoreV1.Secret
...
+func (c *ClusterManager) buildDataplaneServiceaccountSecretResource(cluster *api.Cluster) *k8sCoreV1.Secret {
+	return &k8sCoreV1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      observabilityProxyServiceAccountSecretName,
+			Namespace: observabilityNamespace,
+		},
+		StringData: map[string]string{
+			"client_id":     cluster.ClientID,
+			"client_secret": cluster.ClientSecret,
+		},
```
machi1990, line 891: might be good to add the sso url where the account resides

machi1990: something akin to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/25ee923c7994f5422d9652e98547f71ef4f5b586/internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go#L901 but the config would need to come from keycloak config.

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
pb82: It is the FSO service account, so yes it is used for multiple purposes. I'll try to find a better name for it.

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
pb82: The image is in the rhoas org in quay. We have tested the image pull secret and were able to pull with it. @vmanley when you tested it, did you copy the image pull secret from one of the other namespaces?

pb82:
@miguelsorianod 

`Could you expand on this a little bit? Can you describe the current scenario we are and the scenario we are going to? In this PR I see credentials on the data planes being deployed too`

Sure. The current status is:

1. Observability Operator pulls configuration from Github using a Github access token that is deployed on the cluster.
2. Observabiltiy Operator deploys a token refresher instance on the dataplane which is configured with our Observatorium write credentials.
3. Prometheus now sends remote write requests through this token refresher to Observatorium

This is a problem because it requires two credentials (Github access token, Observatorium credentials) on the data plane cluster, which in the case of Enterprise means they are exposed to the customer.

The desired status is:

1. Observability Operator deploys an additional image containing the Observability Resources.
2. Observability Operator fetches configuration from the resources container. This removes the need for the Github access token.
3. Observability Operator does *not* deploy a data plane token refresher any more and instead configures Prometheus to send remote write requests to the observatorium proxy running in the data plane.
4. Observability Operator also configures Prometheus to fetch a token from RHSSO for every remote write request. The token is generated using the data plane rhsso credentials provided in this PR. This is the same credentials that are also used by the FSO to talk to the Fleet Manager.
5. The Observatorium proxy verifies the token on the remote write requests against the Fleet Manager.
6. This eliminates the need to expose the Observatorium credentials.

pb82:
@machi1990 @miguelsorianod I have addressed the comments, would you mind having another look?

miguelsorianod:
> @miguelsorianod
> 
> `Could you expand on this a little bit? Can you describe the current scenario we are and the scenario we are going to? In this PR I see credentials on the data planes being deployed too`
> 
> Sure. The current status is:
> 
>     1. Observability Operator pulls configuration from Github using a Github access token that is deployed on the cluster.
> 
>     2. Observabiltiy Operator deploys a token refresher instance on the dataplane which is configured with our Observatorium write credentials.
> 
>     3. Prometheus now sends remote write requests through this token refresher to Observatorium
> 
> 
> This is a problem because it requires two credentials (Github access token, Observatorium credentials) on the data plane cluster, which in the case of Enterprise means they are exposed to the customer.
> 
> The desired status is:
> 
>     1. Observability Operator deploys an additional image containing the Observability Resources.
> 
>     2. Observability Operator fetches configuration from the resources container. This removes the need for the Github access token.
> 
>     3. Observability Operator does _not_ deploy a data plane token refresher any more and instead configures Prometheus to send remote write requests to the observatorium proxy running in the data plane.
> 
>     4. Observability Operator also configures Prometheus to fetch a token from RHSSO for every remote write request. The token is generated using the data plane rhsso credentials provided in this PR. This is the same credentials that are also used by the FSO to talk to the Fleet Manager.
> 
>     5. The Observatorium proxy verifies the token on the remote write requests against the Fleet Manager.
> 
>     6. This eliminates the need to expose the Observatorium credentials.

:+1: thanks. This helps considerably to understand the changes. It would be helpful to add this in the description of the PR.

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -874,6 +878,20 @@ func (c *ClusterManager) buildObservatoriumDexSecretResource() *k8sCoreV1.Secret
...
+func (c *ClusterManager) buildDataplaneRhssoCredentialsSecretResource(cluster *api.Cluster) *k8sCoreV1.Secret {
+	return &k8sCoreV1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "observability-proxy-credentials",
+			Namespace: observabilityNamespace,
+		},
+		StringData: map[string]string{
+			"client_id":             cluster.ClientID,
+			"client_secret":         cluster.ClientSecret,
+			"rhsso_auth_server_url": c.ObservabilityConfiguration.RedHatSsoAuthServerUrl,
```
machi1990, line 890: Thanks for adding this.
However the value `c.ObservabilityConfiguration.RedHatSsoAuthServerUrl` only works if the sso provider type is `redhat_sso`. We could make this a bit generic and do something like what's done here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/36c3940df8b8d5deb312bbeadee8dd6c22ccaf5b/internal/kafka/internal/services/kas_fleetshard_operator_addon.go#L167
i.e get this value from the keycloak config.

pb82: That's done now.

machi1990 (approved):
lgtm

@pb82 Let's note that, Obs operator needs to construct a valid token url from the issuer_url. 
suffixing the issuer_url with `"/protocol/openid-connect/token"` should be enough.

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -770,6 +771,7 @@ func (c *ClusterManager) buildResourceSet(cluster api.Cluster) types.ResourceSet
 		c.buildObservabilityCatalogSourceResource(),
 		c.buildObservabilityOperatorGroupResource(),
 		c.buildObservabilitySubscriptionResource(),
+		c.buildDataplaneRhssoCredentialsSecretResource(&cluster),
```
miguelsorianod, line 774: I feel like this could be even more concrete. Ok, dataplane uses Red Hat SSO (is that always the case? if ssoservice has not been created with the redhat sso setting that would not be the case), but what is it used for / who uses it?

pb82: It is used by the Observability Operator to configure the Prometheus to fetch a token and send it along with all remote write requests (more specifically, this section in the Prom config: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#oauth2).

So how about:

`buildRemoteWriteRhssoCredentialsSecret` ?

machi1990: > buildRemoteWriteRhssoCredentialsSecret

Maybe `buildRemoteWriteServiceAccountCredential` as it removes the RH SSO from the method name completely.

miguelsorianod: I think I would associate the name to observability operator

miguelsorianod: Yes, or something similar. 
The important part is that we associate the name to who uses it and what for if it is only for some specific task.
If it is used for multiple tasks for the same element then I would keep out of the name one of the specific tasks.

pb82: thanks, i've changed the name to `buildObservabilityRemoteWriteServiceAccountCredential`

PR #1561, opened 2023-02-10
feat: whitelist cluster_id in search param of kafka list endpoint

pb82:
Allows users to search for `cluster_id` when listing kafkas.

# Why?

This is useful for example to find all instances on enterprise clusters and will be used by the UI and CLI for this purpose.
Searching for a specific cluster_id is safe because we still validate the requester's organization, so you won't be able to list instances you don't have access to.


ziccardi (approved):
lgtm

internal/kafka/internal/api/admin/private/api/openapi.yaml
```diff
@@ -95,7 +95,7 @@ paths:
           Search criteria.
 
           The syntax of this parameter is similar to the syntax of the `where` clause of an
-          SQL statement. Allowed fields in the search are `cloud_provider`, `name`, `owner`, `region`, and `status`. Allowed comparators are `<>`, `=`, `LIKE`, or `ILIKE`.
+          SQL statement. Allowed fields in the search are `cloud_provider`, `name`, `owner`, `region`, `status` and `cluster_id`. Allowed comparators are `<>`, `=`, `LIKE`, or `ILIKE`.
```
machi1990, line 98: Not related to this PR, but I think it would be class to allow the `IN` operator as well - to do things like cluster_id IN ('cluster-id-1', 'cluster-id-2').

@ziccardi @pb82 what do you think?

pb82: That would be useful for example if you have more than one Enterprise cluster. I'll have a look if that can be supported.

ziccardi: Agree. Supporting it should be pretty easy as well.

machi1990: Thanks. 

> Agree. Supporting it should be pretty easy as well.

I'l be happy to review a followup PR my fratello, grazie :-)

machi1990 (approved):

PR #1562, opened 2023-02-10
feat: add support for IN and NOT in search clauses

ziccardi:
## Description
This PR adds support for the `IN` and the `NOT` SQL clauses.

Example 1
```sql
name IN ('value1', 'value2')
```

Example 2
```sql
((cloud_provider = Value and name = value1) and (owner <> value2 or region=b ) or owner in ('owner1', 'owner2', 'owner3')) or owner=c or name=e and region LIKE '%test%'
```
Example 3
```sql
((cloud_provider = Value and name = value1) and (owner <> value2 or region=b ) or owner not in ('owner1', 'owner2', 'owner3')) or owner=c or name=e and region LIKE '%test%'
```

## Verification Steps

Try to search for clusters with the new `IN` and `NOT` clauses

- [ ] Code Review completed


machi1990 (approved):

machi1990 (changes requested):
The change looks good. Can we updated https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/38fc814295cb7daa42ae27386b8f06ee46acf680/openapi/kas-fleet-manager.yaml#L2196 to say that IN is supported?

ziccardi:
Thinking again about this implementation, I think the `not in` could be pretty dangerous if used together with `cluster_id`.
A bit less the `in`.

I changed this to `draft` because I want to ensure that there is no risk that we return clusters the user is not entitled to see.

I was thinking about removing the `not`. WDYT?

@pb82 @machi1990 ^

miguelsorianod:
The clusters that should be returned should be only the ones where the user is the owner, its org if it's an org admin, etc.

machi1990:
@ziccardi IIRC, we also have a filter that applies the logic that @miguelsorianod said.

Might be good to add a test cases here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/38fc814295cb7daa42ae27386b8f06ee46acf680/internal/kafka/test/integration/kafka_list_search_order_test.go so that  this is tested end to end that might help us clear any doubts.

ziccardi:
@valdar @dhirajsb This change will also add the `in` and `not in` filters to the connector `list` endpoint.

internal/kafka/test/integration/kafka_list_search_order_test.go
```diff
@@ -158,6 +176,22 @@ func Test_KafkaListSearchAndOrderBy(t *testing.T) {
...
 			notContains:   []string{nonExistentKafkaName},
 		},
+		{
+			name:          "Filter cluster_id not in ('')",
+			searchOpts:    &public.GetKafkasOpts{Search: optional.NewString("cluster_id not in ('')")},
+			wantErr:       false,
+			expectedSize:  3,
+			expectedTotal: 3,
+			notContains:   []string{nonExistentKafkaName, mockKafkaName4},
+		},
```
machi1990, line 186: @ziccardi thanks for adding this.

Is it possible to add the inverse test case? e.g `cluster_id IN (cluster_id_1, cluster_id_2)` and ensure only Kafka within the organization are returned? Thanks

machi1990 (changes requested):
Left one test case suggestion.

CI is unhappy since the openapi files needs to be regenerated. Can you have a look? Thanks

machi1990:
> @valdar @dhirajsb This change will also add the `in` and `not in` filters to the connector `list` endpoint.

@valdar @dhirajsb can you verify if this doesn't impact connectors? Thanks

internal/kafka/test/integration/kafka_list_search_order_test.go
ziccardi: done!

valdar (approved):
This impacts connector but we should be fine since all the listable resources should be guarded against listing resources for which the user has no rights: 

- namespaces: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/57344b33501049ccb00de9fb79b35eb3d1801b92/internal/connector/internal/handlers/connector_namespace.go#L223-L237
- clusters: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ddda7d09fbc9dd0f2d3114431361b7337f93d80e/internal/connector/internal/services/connector_cluster.go#L252
- connectors: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/7cdf417ab9960d4adabf277b6ee086160460509c/internal/connector/internal/services/connectors.go#L234
- deployments: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ddda7d09fbc9dd0f2d3114431361b7337f93d80e/internal/connector/internal/services/connector_cluster.go#L444

machi1990 (changes requested):
Thanks @ziccardi, the change looks good from KFM side.

As @valdar also confirmed that it'll impact connectors but we should be fine since everything will remain org scope, let's enlist the support of the IN operator here as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/openapi/connector_mgmt.yaml#L1643 well

machi1990 (approved):

PR #1582, opened 2023-02-21
build(docker): update base image to ubi-minimal 9.1

machi1990:
## Description

Update the KFM base image to 9.1 

## Verification Steps

build and deploy KFM image and verify that it is functional

- [ ] Code Review completed
- ~~[ ] All PR comments are resolved either by addressing them or creating follow up tasks~~


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9-minimal:9.0.0
```
machi1990, line 1: @pawelpaszki I've also renamed this file. Let me know if you see any issues with it.
I've also renamed the place where it is used.

pawelpaszki: @machi1990 - I think there should be no issue

machi1990: Thanks @pawelpaszki

pawelpaszki (approved):
/lgtm

machi1990:
Thanks @pawelpaszki for the review

.github/dependabot.yaml
```diff
@@ -9,3 +9,15 @@ updates:
     directory: '/'
     schedule:
       interval: 'weekly'
+  - package-ecosystem: "docker"
```
miguelsorianod, line 12: Could you explain what are these changes?

docker/Dockerfile
miguelsorianod: Is it referenced from any jenkins pipeline / ci pipeline?

machi1990: Yes & No.

Yes - because this file is used to push changes to the version stream pipeline for auto deployment to Stage.
No-because it is not referenced directly but via the build_deploy.sh file which has been updated as well.

.github/dependabot.yaml
machi1990: The changes in this file are to enable dependabot upgrades for dockerfiles and github actions.

miguelsorianod: :+1: .
My assumption is that the way it work in those cases is that for Dockerfiles it checks from URLs in the FROM field and for github actions it checks for the action names in the defined actions. Let me know if that's not accurate.

I would have liked these changes to be independent from this PR as they are totally unrelated.

machi1990: > My assumption is that the way it work in those cases is that for Dockerfiles it checks from URLs in the FROM field and for github actions it checks for the action names in the defined actions. Let me know if that's not accurate.

That's correct. There are few PRs already opened. 

> I would have liked these changes to be independent from this PR as they are totally unrelated.

Ideally yes, but I preferred an option of having two atomic commits which works equally well.

PR #1590, opened 2023-02-21
chore: update envoy base image

machi1990:
## Description

The changelog are available at:
- https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.21/v1.21.0
- https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.22/v1.22.0
- https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.23/v1.23.0
- https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.24/v1.24.0
- https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.25/v1.25.0

NOTE: Please go through them and point out anything that we need to look at as far as our usage of envoy is concerned.

## Verification Steps

1. `make deploy/project` 
2. Deploy limitador and verify that rate limiting works.
To deploy limitador, follow the steps https://github.com/Kuadrant/limitador/tree/main/limitador-server/kubernetes#k8s-deployment 
The following are to be noted:
- Update the content of `limitador-config-configmap.yaml` so that the `data` part of the configmap looks like:
```
  limits.yaml: |
    ---
    -
      namespace: development:kas_fleet_manager
      max_value: 10
      seconds: 60
      variables:
        - path
      conditions:
        - path == kas_fleet_manager_v1
```
With this you can only make 10 requests in a minute to KFM public endpoints.

You do not need to deploy kuadrant. You can limit yourself only to deploying limitador.

3. Update the envoy config rate limit address so that it points to the deployed limitador: https://github.com/machi1990/kas-fleet-manager/blob/5384ef1adb9b05b4db8d044dd124e5f9bcf0191b/templates/envoy-config-configmap.yml#L79 e.g `limitador.<the-namespace-from-1>.svc.cluster.local`  
4. Deploy KFM on OpenShift and verify that you can interact with KFM and that rate limiting is applied. 
5. Go through and changelog to double check if this update has missed on anything we are using

- [ ] Code Review completed

### DIFF
--- a/templates/envoy-config-configmap.yml
+++ b/templates/envoy-config-configmap.yml
@@ -114,6 +114,8 @@ data:
                       prefix_rewrite: /stats/prometheus
               http_filters:
               - name: envoy.filters.http.router
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
 
       # This listener is used to accept inbound API requests.
       - name: ingress
@@ -208,7 +210,9 @@ data:
 
               # This is mandatory in order to have the HTTP routes above.
               - name: envoy.filters.http.router
-
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+              
               # We need this in order to generate JSON responses according to
               # our API guidelines, otherwise Envoy will generate plain text
               # responses.
--- a/templates/service-template.yml
+++ b/templates/service-template.yml
@@ -514,7 +514,7 @@ parameters:
 
 - name: ENVOY_IMAGE
   description: Envoy image
-  value: envoyproxy/envoy:v1.20.2
+  value: envoyproxy/envoy:v1.25.1
 
 - name: KAFKA_DOMAIN_NAME
   description: The domain name to use for Kafka instances


machi1990:
@miguelsorianod I would like to get your review on this one since you've done the initial envoy setup. 

I'll followup updating our live environments separately after this PR is merged.

miguelsorianod:
Let's keep in mind that the envoy config in the bf2 kas fleet manager repository is different than the ones in app-interface used by stage and prod. Additionally, let's also keep in mind that the envoy config between stage and prod in app-interface is also different

machi1990:
> Let's keep in mind that the envoy config in the bf2 kas fleet manager repository is different than the ones in app-interface used by stage and prod. Additionally, let's also keep in mind that the envoy config between stage and prod in app-interface is also different

Yes, there are separate MRs opened already in app-interface. Links are on the parent JIRA.

miguelsorianod:
The changes in the PR look good to me. I haven't verified the deployment.
I am going to review the changelogs.

miguelsorianod:
I've reviewed the changelogs. Summary of identified potential changes that might have some impact:
* 1.20.x: No relevant changes
* 1.21.x: No relevant changes
* 1.22.x:
  * 1.22.0 - http: removing support for long-deprecated old style filter names, e.g. envoy.router, envoy.lua.
    * This is the change that made our config break and that motivated the changes in this PR
  * 1.22.0 - tls: set TLS v1.2 as the default minimal version for servers. Users can still explicitly opt-in to 1.0 and 1.1 using tls_minimum_protocol_version.
    * This might have some impact if clients connecting to it don't support TLS v1.2. It seems unlikely but we should keep it in mind in case some issue is detected
   * 1.22.0 - http: changed the http status code to 504 from 408 if the request times out after the request is completed. This behavior can be temporarily reverted by setting the runtime guard envoy.reloadable_features.override_request_timeout_by_gateway_timeout to false.
     * This is a change in behavior but it seems like an edge case and it also seems unlikely that someone is relying on it in our case
* 1.23.x: no relevant changes
* 1.24.x: No relevant changes
* 1.25.x:
  * 1.25.0 - tls: added support for intermediate CA as trusted CA. The peer certificate issued by an intermediate CA will be trusted by building valid partial chain. Before, it could not be verified without trusting its ancestor root CA and building a full chain. trust_ca. This change can be reverted via the runtime flag envoy.reloadable_features.enable_intermediate_ca.
    * We make use of trusted_ca but I think we don't set an intermediate CA so it should be safe

Given this I think we should be good upgrading.

miguelsorianod:
I see the needs-app-interface-changes has been removed from the PR. I am re-adding it again. Even if the changes have been made I think keeping the label is good so we can go back and see what PRs implied changes in app interface

machi1990:
Thanks for the breakdown of the changes and for double checking the impact this change might have. @miguelsorianod

miguelsorianod (approved):
I am approving this. It looks good to me.

However, let's wait to merge until we get verification from an additional person

JameelB (approved):
Verified changes on an OpenShift 4.11.27 instance. Works as expected 👍

PR #1595, opened 2023-02-23
fix: sets the actual_kafka_billing_model and the desired_kafka_billing del for old kafkas where billing_model was empty

ziccardi:
## Description
Some very old kafka had billing_model set to the empty string. That became an issue now that we make computations using that value

- [ ] Code Review completed


internal/kafka/internal/services/quota/ams_quota_service.go
```diff
@@ -16,12 +16,18 @@ import (
 	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 )
 
+type AMSQuotaService interface {
```
miguelsorianod, line 19: I understand this interface is temporary and also the passing of it on the other types right?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -377,6 +393,55 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
 	metrics.UpdateClusterStatusCapacityMaxCount(c.CloudProvider, c.Region, c.InstanceType, c.ClusterId, float64(c.MaxUnits))
 }
 
+// TODO: this is a temporary reconciler added to value the `actual_kafka_billing_model` and `desired_kafka_billing_model` for
+// kafkas where billing_model has always been empty. See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1115#discussion_r902415419 for details
+func (k *KafkaManager) reconcileBillingModel(kafkas dbapi.KafkaList) serviceErr.ErrorList {
```
miguelsorianod, line 398: Can we add a more  migration specific name to this method? `reconcileBillingModel` is very generic

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -377,6 +393,55 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+	var svcErrors serviceErr.ErrorList
+
+	for _, kafka := range kafkas {
+		if kafka.ActualKafkaBillingModel != "" {
```
machi1990, line 402: we could also ignore Kafkas in other deprovision or deleting states

internal/kafka/internal/services/quota/ams_quota_service.go
ziccardi: yes. It is just needed by the temporary reconciler to get the subscription. It will be removed together with the reconciler

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: True, but I don't think it is a usefult optimisation for this temporary code.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
ziccardi: Any suggestion?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
machi1990: It could happen that the first time this reconciler runs, if the problematic Kafkas are in deleting state, we could run onto a situation where a bunch of errors are sent to reported https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/f4bbc3a5b5d0048b5444bf5741fe8f98771459ca/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go#L422-L428 

The check could save us (the Eng on primary) from investigation time.

ziccardi: > It could happen that the first time this reconciler runs, if the problematic Kafkas are in deleting state, we could run onto a situation where a bunch of errors are sent to reported

Why? Even if deleted the subscription is still there in `deleted` status.

machi1990: Okay. I was not aware of that, I thought delete completely removes the subscription from AMS. Thanks for the clarification. In that case the errors won't happen.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
miguelsorianod: for example tempMigrateEmptyBillingModels

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -261,3 +263,46 @@ func (k *KafkaRequest) IsUsingSharedTLSCertificate(kafkaConfig *config.KafkaConf
 func (k *KafkaRequest) IsADeveloperInstance() bool {
 	return k.InstanceType == types.DEVELOPER.String()
 }
+
+//////////////////////////////////
+/// functional builder
+
```
miguelsorianod, line 269: This seems to have been done for testing purposes.

Could we scope the names and the file where it belongs more to be more concrete?

internal/kafka/internal/api/dbapi/kafka_request_types.go
```diff
@@ -261,3 +263,46 @@ func (k *KafkaRequest) IsUsingSharedTLSCertificate(kafkaConfig *config.KafkaConf
...
+func makeFunc[T any](dest string) func(value T) kafkaRequestFunctionalParam {
+	return func(value T) kafkaRequestFunctionalParam {
+		return func(request *KafkaRequest) {
+			r := reflect.ValueOf(request)
+			reflect.Indirect(r).FieldByName(dest).Set(reflect.ValueOf(value))
+		}
+	}
+}
+
+var WithID = makeFunc[string]("ID")
```
miguelsorianod, line 282: We should consider avoiding global variables where possible.
Could we move this to specific unexported test related methods?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -377,6 +393,55 @@ func (k *KafkaManager) updateClusterStatusCapacityMaxMetric(c services.KafkaStre
...
+			continue
+		}
+		// infer the billing model from the subscription
+		quotaService, factoryErr := k.quotaServiceFactory.GetQuotaService(api.QuotaType(k.kafkaConfig.Quota.Type))
```
miguelsorianod, line 406: This and the retrieval of the specific quota service implementation can be removed outside of the loop, as it is shared for all kafka instances.

internal/kafka/internal/api/dbapi/kafka_request_types.go
ziccardi: No. It is a function and it is used in a different package.

miguelsorianod: The returned value is a function but it is stored in a global variable

ziccardi: It is stored in a global variable of type `function` and is meant to be used as `function` in tests outside of this package. I could move it to a 'test` package, but it needs to be visible outside the package.

miguelsorianod: moving it to a test package makes sense. Or renaming it in a way that it is clear that it is used for testing purposes.

> It is stored in a global variable of type function and is meant to be used as function in tests outside of this package.

That still doesn't require us to put it in a global variable. It could be in a function itself

ziccardi: > As a side note, have we considered to create it on the package where the tests are located?

Yes. There are many tests where we create KafkaRequests, I believe the functional builder can be useful in more than just one test.

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
```diff
@@ -3,11 +3,14 @@ package kafka_mgrs
...
 	"database/sql"
 	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/services/quota"
+	amsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	"testing"
 	"time"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/constants"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi"
+	. "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/api/dbapi/testutils"
```
miguelsorianod, line 13: Would it be possible to explicitly use the package name? if `testutils` is not specific enough we can add a package alias.
Even though it is a little bit more verbose, it find it easier for readers to identify where a given method/function is defined

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
```diff
@@ -195,6 +189,95 @@ func TestKafkaManager_Reconcile(t *testing.T) {
 	}
 }
 
+func TestKafkaManager_ReconcileBillingModels(t *testing.T) {
```
miguelsorianod, line 192: The test function name should be updated to match the method to be tested

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr_test.go
```diff
@@ -195,6 +189,95 @@ func TestKafkaManager_Reconcile(t *testing.T) {
...
+							GetSubscriptionByIDFunc: func(subscriptionID string) (*amsv1.Subscription, bool, *errors.ServiceError) {
+								subscription, _ := amsv1.NewSubscription().ClusterBillingModel(tt.fields.billingModel).Build()
+								return subscription, true, nil
+							},
+						}, nil
+					},
+				},
+			}
+
+			k.Reconcile()
```
miguelsorianod, line 271: Here Reconcile is being called. Is it possible to just call the function being tested?

miguelsorianod (changes requested):
Looks good to me. I left a couple of minor comments

miguelsorianod (approved):

machi1990:
The fix should have landed in Prod by now. I am going to revert this.

PR #1596, opened 2023-02-23
feat: return all sizes regardless of capacity for enterprise

pb82:
Return all instance sizes for Enterprise clusters, regardless of capacity.

### DIFF
--- a/internal/kafka/internal/presenters/cluster.go
+++ b/internal/kafka/internal/presenters/cluster.go
@@ -11,7 +11,6 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 )
 
 func PresentEnterpriseClusterListItem(cluster api.Cluster) public.EnterpriseClusterListItem {
@@ -75,7 +74,7 @@ func PresentEnterpriseCluster(cluster api.Cluster, consumedStreamingUnitsInTheCl
 	storedCapacityInfo, ok := cluster.RetrieveDynamicCapacityInfo()[types.STANDARD.String()]
 	if ok {
 		presentedCluster.CapacityInformation = presentEnterpriseClusterCapacityInfo(consumedStreamingUnitsInTheCluster, storedCapacityInfo)
-		supportedInstanceTypes, err := presentEnterpriseClusterSupportedInstanceTypes(presentedCluster.CapacityInformation.RemainingKafkaStreamingUnits, kafkaConfig)
+		supportedInstanceTypes, err := presentEnterpriseClusterSupportedInstanceTypes(kafkaConfig)
 		if err != nil {
 			return public.EnterpriseCluster{}, err
 		}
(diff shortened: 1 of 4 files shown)


machi1990 (changes requested):
thanks @pb82 the code change looks good to me.

1. We need to update the integration test - https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6524c0c8ee7217cd4b5e70e2e410b376a242795d/internal/kafka/test/integration/cluster_get_enterprise_test.go#L253 and ensure that the size remains 2. Same as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6524c0c8ee7217cd4b5e70e2e410b376a242795d/internal/kafka/test/integration/cluster_get_enterprise_test.go#L215-L218

2. We need to update the unit tests for the handler https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6524c0c8ee7217cd4b5e70e2e410b376a242795d/internal/kafka/internal/handlers/cluster_test.go#L1037-L1145

internal/kafka/internal/presenters/cluster.go
```diff
@@ -118,15 +117,7 @@ func presentEnterpriseClusterSupportedInstanceTypes(remainingCapacity int32, kaf
...
-	availableSizes := arrays.Filter(standardInstanceType.Sizes, func(size config.KafkaInstanceSize) bool {
-		return int32(size.CapacityConsumed) <= remainingCapacity
-	})
-
-	presentedSizes := []public.SupportedKafkaSize{}
-
-	if len(availableSizes) > 0 {
-		presentedSizes = GetSupportedSizes(&config.KafkaInstanceType{Sizes: availableSizes})
-	}
+	presentedSizes := GetSupportedSizes(&config.KafkaInstanceType{Sizes: standardInstanceType.Sizes})
```
machi1990, line 120: let's also remove the `remainingCapacity` argument as it is not used anywhere in this function https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/6524c0c8ee7217cd4b5e70e2e410b376a242795d/internal/kafka/internal/presenters/cluster.go#L101

miguelsorianod:
Could you expand on the motivation of this change and also on how capacity calculation is involved in enterprise clusters?

pb82:
@miguelsorianod UI needs to know the supported sizes that can be installed on an enterprise cluster. We do have capacity information from the data plane and previously we would only return sizes that can still fit. However, after discussion with @riccardo-forina , they prefer to display all the sizes all the time.

miguelsorianod:
> @miguelsorianod UI needs to know the supported sizes that can be installed on an enterprise cluster. We do have capacity information from the data plane and previously we would only return sizes that can still fit. However, after discussion with @riccardo-forina , they prefer to display all the sizes all the time.

Thanks for the context, it is helpful :+1: .
I see the change is done for the enterprise type only in this PR. Do the comments about the size and the UI also apply to non enterprise clusters?

miguelsorianod:
The changes look good to me.
After solving the issues mentioned by Manyanda it should be ok to approve

internal/kafka/internal/presenters/cluster.go
```diff
@@ -70,22 +68,12 @@ func PresentEnterpriseCluster(cluster api.Cluster, consumedStreamingUnitsInTheCl
 		SupportedInstanceTypes:        public.SupportedKafkaInstanceTypesList{},
 	}
 
-	// enterprise clusters only supports standard instance type for now. It is safe to hardcode this.
-	storedCapacityInfo, ok := cluster.RetrieveDynamicCapacityInfo()[types.STANDARD.String()]
-	if ok {
-		presentedCluster.CapacityInformation = presentEnterpriseClusterCapacityInfo(consumedStreamingUnitsInTheCluster, storedCapacityInfo)
```
miguelsorianod, line 77: This logic has been removed but the type still presents the capacity information as an attribute from the response right?
Is this what we want?

machi1990: Why has this line `presentedCluster.CapacityInformation = presentEnterpriseClusterCapacityInfo(consumedStreamingUnitsInTheCluster, storedCapacityInfo)` been removed?

openapi/kas-fleet-manager.yaml
```diff
@@ -1527,32 +1527,6 @@ components:
             supported_instance_types:
               type: object
               $ref: "#/components/schemas/SupportedKafkaInstanceTypesList"
-            capacity_information:
```
machi1990, line 1530: @pb82 this information is needed by the UI and CLI, we've to keep it.

machi1990 (approved):
thanks @pb82 @miguelsorianod

PR #1600, opened 2023-02-27
feat: add observability proxy deployment

miguelsorianod:
## Description
Related to https://issues.redhat.com/browse/MGDSTRM-10333.

This PR adds the OCP templates needed to deploy [Observability Remote Write Proxy](https://github.com/bf2fc6cc711aee1a0c2a/observability-remote-write-proxy).

The aim of Observability Remote Write Proxy in KAS Fleet Manager context is for the managed kafka related Prometheus metrics in the Data Planes to be sent to Observatorium indirectly through the proxy.
The motivation of sending the credentials indirectly through the proxy is to avoid needing to store Observatorium credentials in the Data Planes themselves.

This PR adds the following OCP templates:
* `templates/observability-remote-write-proxy-oidc-secret.yml`: Contains the needed K8s secret when OIDC token retrieval is enabled in Observability Remote Write Proxy.
* `templates/observability-remote-write-proxy-route.yml`: Contains an OCP Route used to access the Observability Remote Write Proxy API
* `templates/observability-remote-write-proxy.yml`: Contains the Observability Remote Write Proxy K8s Deployment and related K8s objects like K8s Service, PodDisruptionBudget, ...

The templates have been intentionally been implemented separatedly from templates/service-template.yml, templates/secret-template.yml and templates/route-template.yml. The reason for that is that the Observability Remote Write Proxy is considered independent of the KAS Fleet Manager deployment and its related elements themselves.

Aside from the templates, other changes made have been:
* Add Makefile targets to allow deploying Observability Remote Write Proxy more easily during development
* Add documentation around deploying Observability Remote Write Proxy

Documentation in the "populating configuration" document has intentionally not been included in this PR. The reason for that is that there are some needed Observability Operator Changes needed on it to support Observability Remote Write Proxy. Another separate PR will be opened around documentation in the "populating configuration" document and it will only be merged when the functionality in Observability Operator is implemented and released.

As a side note, the configuration and deployment of Observability Remote Write Proxy is optional in the development stage. The implication of not configuring Observability Remote Write Proxy is that the Data Plane metrics will not be sent anywhere.

Once this is merged some additional tasks will be performed on AppInterface side to be able to use Observability Remote Write Proxy in the stage and production environments.

## Verification Steps

Deploy Observability Remote Write Proxy and configure KAS Fleet Manager to use it.
To do that:
1. Deploy Observability Remote Write Proxy in OCP. Read the documentation of this PR on how to do it. Additionally, temporarily edit the OCP template to use the `8ed2bb9` container image tag to get the latest functionality. Observatorium Stage can be used when setting the Observatorium related information. If there are any doubts about what values to set on some steps please let me know.
2. Configure KAS Fleet Manager to use Observability Remote Write Proxy. To do so, run KAS Fleet Manager with the `--observability-red-hat-sso-observatorium-gateway` flag set to the URL provided by the OCP Route created for Observability Remote Write Proxy. The URL scheme has to be included as part of the URL. An Observability Operator version that includes the latest functionality that allows it to use Observability Remote Write Proxy has to be used
3. Verify that metrics are being sent to Observatorium through Observability Remote Write Proxy

- [ ] Code Review completed

### DIFF
--- a/templates/observability-remote-write-proxy-oidc-secret.yml
+++ b/templates/observability-remote-write-proxy-oidc-secret.yml
@@ -1,0 +1,23 @@
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: observability-remote-write-proxy-oidc-secret
+  annotations:
+    description: "Observability Remote Write Proxy OIDC secret"
+objects:
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: observability-remote-write-proxy-oidc-config
+  data:
+    oidc-config: "${OBSERVABILITY_REMOTE_WRITE_PROXY_OIDC_CONFIG}"
+
+parameters:
+- name: OBSERVABILITY_REMOTE_WRITE_PROXY_OIDC_CONFIG
+  description: >-
+    Base64 encoded Observability Remote Write Proxy OIDC configuration.
+    It should contain the YAML contents. See
+    https://github.com/bf2fc6cc711aee1a0c2a/observability-remote-write-proxy
+    for details on the configuration format.
+  required: true
--- a/templates/observability-remote-write-proxy-route.yml
+++ b/templates/observability-remote-write-proxy-route.yml
@@ -1,0 +1,23 @@
+# This file contains an OpenShift template that creates a route helping to
+# directly access the Observability Remote Write Proxy in the
+# development environment.
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+name: observability-remote-write-proxy-route
+metadata:
+  name: observability-remote-write-proxy
+objects:
+- apiVersion: route.openshift.io/v1
+  kind: Route
+  metadata:
+    name: observability-rw-proxy
+    labels:
+      app.kubernetes.io/name: observability-remote-write-proxy
+  spec:
+    to:
+      kind: Service
+      name: observability-rw-proxy
+    tls:
+      termination: edge
+      insecureEdgeTerminationPolicy: Redirect
--- a/templates/observability-remote-write-proxy.yml
+++ b/templates/observability-remote-write-proxy.yml
@@ -141,39 +141,6 @@ objects:
       selector:
         matchLabels:
           app.kubernetes.io/name: observability-remote-write-proxy
-  - kind: NetworkPolicy
-    apiVersion: networking.k8s.io/v1
-    metadata:
-      labels:
-        app.kubernetes.io/component: prometheus-remote-write-proxy
-        app.kubernetes.io/name: observability-remote-write-proxy
-      name: observability-remote-write-proxy
-    spec:
-      podSelector:
-        matchLabels:
-          app.kubernetes.io/component: prometheus-remote-write-proxy
-          app.kubernetes.io/name: observability-remote-write-proxy
-      policyTypes:
-      - Ingress
-      ingress:
-      - from:
-          # Allow access from the default OpenShift Ingress Controller pods.
-          # The default OpenShift Ingress Controller pods are located in a K8s
-          # Namespace that contains the 'network.openshift.io/policy-group' K8s
-          # label key with the value 'ingress' and the 'name' label with the
-          # value 'openshift-ingress'. Additionally, those pods have the label
-          # 'ingresscontroller.operator.openshift.io/deployment-ingresscontroller'
-          # with the value 'default'.
-        - namespaceSelector:
-            matchLabels:
-              network.openshift.io/policy-group: ingress
-              name: openshift-ingress
-          podSelector:
-            matchLabels:
-              ingresscontroller.operator.openshift.io/deployment-ingresscontroller: default
-        ports:
-        - port: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_PORT}}
-          protocol: TCP
 
 parameters:
 - name: OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_NAME
--- a/templates/observability-remote-write-proxy.yml
+++ b/templates/observability-remote-write-proxy.yml
@@ -1,0 +1,335 @@
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: observability-remote-write-proxy
+  annotations:
+    description: "Prometheus Remote Write Proxy for Observability"
+objects:
+  - kind: Deployment
+    apiVersion: apps/v1
+    metadata:
+      labels:
+        app.kubernetes.io/component: prometheus-remote-write-proxy
+        app.kubernetes.io/name: observability-remote-write-proxy
+        app.kubernetes.io/version: ${OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_TAG}
+      name: observability-remote-write-proxy
+    spec:
+      replicas: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_REPLICAS}}
+      strategy:
+        rollingUpdate:
+          maxSurge: 25%
+          maxUnavailable: 25%
+        type: RollingUpdate
+      selector:
+        matchLabels:
+          app.kubernetes.io/component: prometheus-remote-write-proxy
+          app.kubernetes.io/name: observability-remote-write-proxy
+      template:
+        metadata:
+          labels:
+            app.kubernetes.io/component: prometheus-remote-write-proxy
+            app.kubernetes.io/name: observability-remote-write-proxy
+        spec:
+          securityContext:
+            runAsNonRoot: true
+            seccompProfile:
+              type: RuntimeDefault
+          affinity:
+            # Prefer not co-locating Observability Remote Write Proxy pods in the
+            # same zone. This means that K8s scheduler will try to spread
+            # the different Observability Remote Write Proxy pods among the
+            # available zones (logical failure domains)
+            podAntiAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+              - podAffinityTerm:
+                  labelSelector:
+                    matchExpressions:
+                    - key: app.kubernetes.io/name
+                      operator: In
+                      values:
+                      - observability-remote-write-proxy
+                  topologyKey: topology.kubernetes.io/zone
+                weight: 100
+          containers:
+          - name: observability-remote-write-proxy
+            image: ${OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_NAME}:${OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_TAG}
+            imagePullPolicy: ${OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_PULL_POLICY}
+            ports:
+            - containerPort: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_PORT}}
+              name: http
+            - containerPort: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_METRICS_PORT}}
+              name: metrics
+            command:
+              - /proxy
+              - --proxy.listen.port=${OBSERVABILITY_REMOTE_WRITE_PROXY_PORT}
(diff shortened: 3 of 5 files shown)


templates/observability-remote-write-proxy-route.yml
```diff
@@ -0,0 +1,23 @@
...
+# development environment.
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+name: observability-remote-write-proxy-route
+metadata:
+  name: observability-remote-write-proxy
+objects:
+- apiVersion: route.openshift.io/v1
+  kind: Route
```
miguelsorianod, line 12: The addition of a separate route than the one used by KFM is intentional. Details around the reason and motivations why this has been done are available in the https://issues.redhat.com/browse/MGDSTRM-10333 Jira issue

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,337 @@
...
+    spec:
+      ports:
+        - name: http
+          port: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_PORT}}
+          targetPort: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_PORT}}
+          protocol: TCP
+      selector:
+        app.kubernetes.io/component: prometheus-remote-write-proxy
+        app.kubernetes.io/name: observability-remote-write-proxy
+  - kind: Service
```
pb82, line 121: why an extra service for the metrics and not another port in the `observability-rw-proxy` service?

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,337 @@
...
+      ingress:
+      - from:
+          # Allow access from the default OpenShift Ingress Controller pods.
+          # The default OpenShift Ingress Controller pods are located in a K8s
+          # Namespace that contains the 'network.openshift.io/policy-group' K8s
+          # label key with the value 'ingress' and the 'name' label with the
+          # value 'openshift-ingress'. Additionally, those pods have the label
+          # 'ingresscontroller.operator.openshift.io/deployment-ingresscontroller'
+          # with the value 'default'.
+        - namespaceSelector:
```
miguelsorianod, line 169: This assumes that in stage / prod cluster there is a namespace called openshift-ingress with the label `network.openshift.io/policy-group: ingress`. It also assumes that there are some pods in that namespace with the label `ingresscontroller.operator.openshift.io/deployment-ingresscontroller: default`.
This is the case with clusters created from OCM, but it might be different in the case of AppSRE clusters. In case we detect any issue this should be readapted if we decide to keep the NetworkPolicy

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,337 @@
...
+          labels:
+            app.kubernetes.io/component: prometheus-remote-write-proxy
+            app.kubernetes.io/name: observability-remote-write-proxy
+        spec:
+          securityContext:
+            runAsNonRoot: true
+            seccompProfile:
+              type: RuntimeDefault
+          affinity:
+            # Prefer not co-locating Observability Remote Write Proxy pods in the
```
miguelsorianod, line 39: We try to spread the 1..N possible replicas among different AZs

templates/observability-remote-write-proxy.yml
miguelsorianod: It imitates what's been done for KFM.
It is also useful to keep it separated to have independent management.

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,335 @@
...
+
+- name: OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_TAG
+  description: Observability Remote Write Proxy container image tag
+  value: "latest"
+
+- name: OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_PULL_POLICY
+  description: >-
+    Observability Remote Write Proxy container image pull policy.
+    Accepted values: ['Always', 'IfNotPresent', 'Never']"
+  value: "IfNotPresent"
```
miguelsorianod, line 191: Two options here:
Set the default as `Always` and in AppInterface set it explicitly to IfNotPresent. Or leave it as IfNotPresent in which case during development in some cases it will be needed to be changed (for example if a `latest` tag is used)

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,335 @@
...
+    spec:
+      ports:
+        - name: metrics
+          port: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_METRICS_PORT}}
+          targetPort: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_METRICS_PORT}}
+          protocol: TCP
+      selector:
+        app.kubernetes.io/component: prometheus-remote-write-proxy
+        app.kubernetes.io/name: observability-remote-write-proxy
+  - kind: PodDisruptionBudget
```
miguelsorianod, line 135: a PodDisruptionBudget has been added to ensure minimum availability when pod evictions occur

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,335 @@
...
+    metadata:
+      labels:
+        app.kubernetes.io/component: prometheus-remote-write-proxy
+        app.kubernetes.io/name: observability-remote-write-proxy
+        app.kubernetes.io/version: ${OBSERVABILITY_REMOTE_WRITE_PROXY_IMAGE_TAG}
+      name: observability-remote-write-proxy
+    spec:
+      replicas: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_REPLICAS}}
+      strategy:
+        rollingUpdate:
```
miguelsorianod, line 20: Progressive rollout of newer Observability Remote Write Proxy images on the Deployment

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,335 @@
...
+              - /proxy
+              - --proxy.listen.port=${OBSERVABILITY_REMOTE_WRITE_PROXY_PORT}
+              - --proxy.metrics.port=${OBSERVABILITY_REMOTE_WRITE_PROXY_METRICS_PORT}
+              - --proxy.forwardUrl=${OBSERVABILITY_REMOTE_WRITE_PROXY_FORWARD_URL}
+              - --oidc.enabled=${OBSERVABILITY_REMOTE_WRITE_PROXY_OIDC_ENABLED}
+              - --oidc.filename=${OBSERVABILITY_REMOTE_WRITE_PROXY_OIDC_CONFIG_FILEDIR}/${OBSERVABILITY_REMOTE_WRITE_PROXY_OIDC_CONFIG_FILENAME}
+              - --token.verification.enabled=${OBSERVABILITY_REMOTE_WRITE_PROXY_TOKEN_VERIFICATION_ENABLED}
+              - --token.verification.url=${OBSERVABILITY_REMOTE_WRITE_PROXY_TOKEN_VERIFICATION_URL}
+            resources:
+              requests:
```
miguelsorianod, line 73: Configurable resource requests and limits. If desired, we can configure different resource requests and limits depending on the environment.
For now we intentionally decided that:
* We will set both cpu and memory requests (see defaults in the template)
* We will set a limit of memory. We have set a default that we hope it is high enough that the limit is not reached
* We will not set a limit of cpu. We will track usage from the metrics and once we have observed the behavior for a while we will decide an appropriate cpu limit

templates/observability-remote-write-proxy-route.yml
```diff
@@ -0,0 +1,23 @@
+# This file contains an OpenShift template that creates a route helping to
+# directly access the Observability Remote Write Proxy in the
+# development environment.
```
miguelsorianod, line 3: This route is for development only. For our stage and production environment different routes will be created, from different OCP templates located in another repository

templates/observability-remote-write-proxy-oidc-secret.yml
```diff
@@ -0,0 +1,23 @@
+---
```
miguelsorianod, line 1: This secret is only used for development purposes.
For our stage and production environment different secrets will be created and configured

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,335 @@
+---
```
miguelsorianod, line 1: Additional work will be needed for stage and production environments to integrate this template into them

miguelsorianod:
As a side comment, I've been noticing that our integration tests fail from time to time due to some flaky test in connectors related code. Could someone from the connectors team review it? It's been happening in quite a few PRs and it delays the tests passing. We should avoid having flaky tests.
An example is a run for this PR: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/actions/runs/4282002151/jobs/7455725088.

On another note, could you create a team similar to `kas-fleet-manager-reviewers` where connector team members supposed to be involved on connectors related code is involved? This would avoid needing to ping specific people from the connectors team, remembering their names and also update it when people join/leave the connectors team.

cc @valdar @lburgazzoli @dhirajsb.

Thanks

miguelsorianod:
cc @pb82 around Observability Operator and Fleet Manager, will the changes related to Observability Remote Write Proxy have an impact on connectors code? From previous talks we had I seem to remember you mentioned Fleet Manager needing to use it due to specific Observability Operator reasons

templates/observability-remote-write-proxy.yml
machi1990: Not related to this PR, but worth noting that for this to be applied in app-interface we'll need https://issues.redhat.com/browse/MGDSTRM-8802 to be completed i.e `PodDisruptionBudget` added as a managed resource

miguelsorianod: I will take care of adding it when doing the changes on app-interface side

machi1990: Thank you.

machi1990 (commented):
The code change looks good to me. I've not verified the change as I do not have an OpenShift cluster at the moment, this can be merged if already verified by another individual @miguelsorianod @pb82 @valeriiashapoval 

Thanks for the quality PR as usual @miguelsorianod

templates/observability-remote-write-proxy.yml
```diff
@@ -0,0 +1,337 @@
...
+  - kind: PodDisruptionBudget
+    apiVersion: policy/v1
+    metadata:
+      name: observability-remote-write-proxy-pdb
+    spec:
+      minAvailable: ${{OBSERVABILITY_REMOTE_WRITE_PROXY_PDB_MIN_AVAILABLE}}
+      selector:
+        matchLabels:
+          app.kubernetes.io/name: observability-remote-write-proxy
+  - kind: NetworkPolicy
```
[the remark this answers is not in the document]
machi1990: @pb82 any thoughts on this?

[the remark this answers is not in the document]
pb82: I don't think it's needed and in the future we may want to remove the control plane token refresher in favour of the proxy (although that still needs to be thought out because of external access).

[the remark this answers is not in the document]
miguelsorianod: I am going to remove it from the PR then.
For this, I am going to add a new comment that removes it without squashing it. The reason for that is so we can see how it was configured in case we need to add it in the future.

templates/observability-remote-write-proxy.yml
pb82: +1 to set `Always` as the default, this allows us to use latest.

miguelsorianod: Ok. But then in Stage/Prod we don't want `Always` and we will need to override there. The reason for that is that there we will use a fixed tag and it is unnecessary to pull the image every time

pb82 (approved):
LGTM

pb82:
@miguelsorianod regarding RHOC: Observability Operator will configure Prometheus to send an auth token with every remote write request only *if* the new secret containing those credentials is there. Since they are using their own fleet manager which does not create that secret, they can continue to pass down the direct Observatorium URL.

PR #1602, opened 2023-02-28
fix: ensure image can be built on m1 macos

JameelB:
## Description
When running any of the image build make targets on an M1 MacOS, it fails with the following error: 
```
 => [builder 7/9] COPY . ./                                                                                                                                                                                                                                                         0.2s 
 => ERROR [builder 8/9] RUN go mod vendor                                                                                                                                                                                                                                           0.2s 
------                                                                                                                                                                                                                                                                                   
 > [builder 8/9] RUN go mod vendor:
#12 0.148 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
------
executor failed running [/bin/sh -c go mod vendor]: exit code: 255
```

More details on this error can be found in this [thread](https://stackoverflow.com/questions/71040681/qemu-x86-64-could-not-open-lib64-ld-linux-x86-64-so-2-no-such-file-or-direc).

Note that this is a temporary workaround. An issue has been created to look into how to build KFM in multiple platforms natively, see [MGDSTRM-10800](https://issues.redhat.com/browse/MGDSTRM-10800).

## Verification Steps
1. Run make image/build or make image/build/internal and ensure it builds successfully
    - Verify this on MacOS
    - Verity this on Linux

<!-- Please strikethrough options not relevant using two tildes ~~Text~~. Do not delete non relevant options -->
- [ ] ~~CI and all relevant tests are passing~~
- [ ] Code Review completed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,19 +159,12 @@ openapi/spec/validate: specinstall
 	${SPECTRAL} lint ../../openapi/kas-fleet-manager.yaml ../../openapi/kas-fleet-manager-private-admin.yaml ;\
 	}
 
-	
-
-ifeq ($(shell uname -s | tr A-Z a-z), darwin)
-        PGHOST:="127.0.0.1"
-else
-        PGHOST:="172.18.0.22"
-endif
-
 ### Environment-sourced variables with defaults
 # Can be overriden by setting environment var before running
 # Example:
 #   OCM_ENV=testing make run
 #   export OCM_ENV=testing; make run
+
 # Set the environment to development by default
 ifndef OCM_ENV
 	OCM_ENV:=integration
@@ -591,6 +584,12 @@ db/generate/insert/cluster:
 	echo -e "Run this command in your database:\n\nINSERT INTO clusters (id, created_at, updated_at, cloud_provider, cluster_id, external_id, multi_az, region, status, provider_type) VALUES ('"$$id"', current_timestamp, current_timestamp, '"$$provider"', '"$$id"', '"$$external_id"', "$$multi_az", '"$$region"', 'cluster_provisioned', 'ocm');";
 .PHONY: db/generate/insert/cluster
 
+## make targets for building and pushing images to a registry
+# Set var defaults based on os
+ifeq ($(shell uname -s | tr A-Z a-z), darwin)
+CONTAINER_IMAGE_BUILD_PLATFORM ?= --platform linux/amd64
+endif
+
 # Login to docker
 docker/login:
 	$(DOCKER) --config="${DOCKER_CONFIG}" login -u "${QUAY_USER}" -p "${QUAY_TOKEN}" quay.io
@@ -603,7 +602,7 @@ docker/login/internal:
 
 # Build the binary and image
 image/build:
-	$(DOCKER) --config="${DOCKER_CONFIG}" build --pull -t "$(external_image_registry)/$(image_repository):$(image_tag)" .
+	$(DOCKER) --config="${DOCKER_CONFIG}" build $(CONTAINER_IMAGE_BUILD_PLATFORM) --pull -t "$(external_image_registry)/$(image_repository):$(image_tag)" .
 .PHONY: image/build
 
 # Build and push the image
@@ -614,7 +613,7 @@ image/push: image/build
 # build binary and image for OpenShift deployment
 image/build/internal: IMAGE_TAG ?= $(image_tag)
 image/build/internal:
-	$(DOCKER) build -t "$(shell $(OC) get route default-route -n openshift-image-registry -o jsonpath="{.spec.host}")/$(image_repository):$(IMAGE_TAG)" .
+	$(DOCKER) build $(CONTAINER_IMAGE_BUILD_PLATFORM) -t "$(shell $(OC) get route default-route -n openshift-image-registry -o jsonpath="{.spec.host}")/$(image_repository):$(IMAGE_TAG)" .
 .PHONY: image/build/internal
 
 # push the image to the OpenShift internal registry
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+### Environment-sourced variables with defaults
+# Can be overriden by setting environment var before running
+# Example:
+#   OCM_ENV=testing make run
+#   export OCM_ENV=testing; make run
 MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 DOCS_DIR := $(PROJECT_PATH)/docs
@@ -57,6 +62,9 @@ else
 GOBIN=$(shell $(GO) env GOBIN)
 endif
 
+export GOPROXY=https://proxy.golang.org
+export GOPRIVATE=gitlab.cee.redhat.com
+
 LOCAL_BIN_PATH := ${PROJECT_PATH}/bin
 # Add the project-level bin directory into PATH. Needed in order
 # for `go generate` to use project-level bin directory binaries first
@@ -69,6 +77,9 @@ CURL ?= curl
 
 OC ?= oc
 OCM ?= ocm
+
+# Set the environment to integration by default
+OCM_ENV ?= integration
 
 GOLANGCI_LINT ?= $(LOCAL_BIN_PATH)/golangci-lint
 golangci-lint:
@@ -158,36 +169,6 @@ openapi/spec/validate: specinstall
 	cd ${LOCAL_BIN_PATH}/rhoasapi-installation ;\
 	${SPECTRAL} lint ../../openapi/kas-fleet-manager.yaml ../../openapi/kas-fleet-manager-private-admin.yaml ;\
 	}
-
-### Environment-sourced variables with defaults
-# Can be overriden by setting environment var before running
-# Example:
-#   OCM_ENV=testing make run
-#   export OCM_ENV=testing; make run
-
-# Set the environment to development by default
-ifndef OCM_ENV
-	OCM_ENV:=integration
-endif
-
-ifndef TEST_SUMMARY_FORMAT
-	TEST_SUMMARY_FORMAT=short-verbose
-endif
-
-export GOPROXY=https://proxy.golang.org
-export GOPRIVATE=gitlab.cee.redhat.com
-
-ifndef SERVER_URL
-	SERVER_URL:=http://localhost:8000
-endif
-
-ifndef TEST_TIMEOUT
-	ifeq ($(OCM_ENV), integration)
-		TEST_TIMEOUT=15m
-	else
-		TEST_TIMEOUT=5h
-	endif
-endif
 
 # Prints a list of useful targets.
 help:
@@ -300,6 +281,19 @@ binary:
 install: verify lint
 	$(GO) install ./cmd/kas-fleet-manager
 .PHONY: install
+
+# --------------------- Test Targets --------------------- #
+# Make targets for running tests related commands
+
+# Set var defaults for test targets
+TEST_SUMMARY_FORMAT ?= short-verbose
+ifndef TEST_TIMEOUT
+	ifeq ($(OCM_ENV), integration)
+		TEST_TIMEOUT=15m
+	else
+		TEST_TIMEOUT=5h
+	endif
+endif
 
 # Runs the unit tests.
 #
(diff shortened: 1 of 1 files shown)


pb82 (approved):
Verified on linux, `make image/build` succeeds

pawelpaszki (approved):
Verified on MAC:

```
make image/build

docker --config=""/Users/pawelpaszki/kafka/managed-services-api/.docker"" build --pull -t "default-route-openshift-image-registry.apps-crc.testing/kas-fleet-manager-pawelpaszki/kas-fleet-manager:1677585248" .
[+] Building 92.1s (15/15) FINISHED                                                                                                                                                                                                            
 => [internal] load build definition from Dockerfile                                                                                                                                                                                      0.0s
 => => transferring dockerfile: 816B                                                                                                                                                                                                      0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                         0.0s
 => => transferring context: 738B                                                                                                                                                                                                         0.0s
 => [internal] load metadata for registry.access.redhat.com/ubi9-minimal:9.1.0                                                                                                                                                            2.7s
 => [builder 1/9] FROM registry.access.redhat.com/ubi9-minimal:9.1.0@sha256:61925d31338b7b41bfd5b6b8cf45eaf80753d415b0269fc03613c5c5049b879e                                                                                              6.2s
 => => resolve registry.access.redhat.com/ubi9-minimal:9.1.0@sha256:61925d31338b7b41bfd5b6b8cf45eaf80753d415b0269fc03613c5c5049b879e                                                                                                      0.0s
 => => sha256:43246102cf37b0825786ca78e9a6e153c26d1a4241ddf53092e5b0ac18fe77bf 429B / 429B                                                                                                                                                0.0s
 => => sha256:3135bd90aad672f44305a68c1d7e1ab1b894ec6e0c6964d5ab40295af1920437 6.26kB / 6.26kB                                                                                                                                            0.0s
 => => sha256:ba958a445f00d91e4019b520c467e36e7f5bc07da02f2a87e9ccefbe4a70ff6d 37.85MB / 37.85MB                                                                                                                                          4.5s
 => => sha256:61925d31338b7b41bfd5b6b8cf45eaf80753d415b0269fc03613c5c5049b879e 1.47kB / 1.47kB                                                                                                                                            0.0s
 => => extracting sha256:ba958a445f00d91e4019b520c467e36e7f5bc07da02f2a87e9ccefbe4a70ff6d                                                                                                                                                 1.5s
 => [internal] load build context                                                                                                                                                                                                         3.7s
 => => transferring context: 138.86MB                                                                                                                                                                                                     3.7s
 => [builder 2/9] RUN microdnf install -y tar gzip make which git                                                                                                                                                                        10.7s
 => [builder 3/9] RUN curl -O -J https://dl.google.com/go/go1.19.6.linux-amd64.tar.gz                                                                                                                                                     5.3s
 => [builder 4/9] RUN tar -C /usr/local -xzf go1.19.6.linux-amd64.tar.gz                                                                                                                                                                  4.6s 
 => [builder 5/9] RUN ln -s /usr/local/go/bin/go /usr/local/bin/go                                                                                                                                                                        0.2s 
 => [builder 6/9] WORKDIR /workspace                                                                                                                                                                                                      0.0s 
 => [builder 7/9] COPY . ./                                                                                                                                                                                                               1.0s 
 => [builder 8/9] RUN go mod vendor                                                                                                                                                                                                      18.5s 
 => [builder 9/9] RUN make binary                                                                                                                                                                                                        41.7s
 => [stage-1 2/2] COPY --from=builder /workspace/kas-fleet-manager /usr/local/bin/                                                                                                                                                        0.2s 
 => exporting to image                                                                                                                                                                                                                    0.4s 
 => => exporting layers                                                                                                                                                                                                                   0.4s 
 => => writing image sha256:9c55c2841661c3960dcb7aa8a9ac012600b8114d06001678d77787de16b7d08a                                                                                                                                              0.0s 
 => => naming to default-route-openshift-image-registry.apps-crc.testing/kas-fleet-manager-pawelpaszki/kas-fleet-manager:1677585248                                                                                                       0.0s 

Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
```

miguelsorianod:
Maybe a cleaner way would be setting the envvar if the user running the Makefile is on a mac?

miguelsorianod:
This could be achieved by doing something like (pseudocode, needs to be reviewed):
```
OS=$(uname)
if [ "$OS" = 'Darwin' ]; then
  export DOCKER_DEFAULT_PLATFORM ?= linux/amd64
fi
```

miguelsorianod:
> This could be achieved by doing something like (pseudocode, needs to be reviewed):
> 
> ```
> OS=$(uname)
> if [ "$OS" = 'Darwin' ]; then
>   export DOCKER_DEFAULT_PLATFORM ?= linux/amd64
> fi
> ```
> 
> It should also be verified that when running in a non-mac OS that variable is not set on the docker build

As a side note, I wonder if the same issue happens if podman is used, and also whether podman makes use of that envvar too

JameelB:
> > This could be achieved by doing something like (pseudocode, needs to be reviewed):
> > ```
> > OS=$(uname)
> > if [ "$OS" = 'Darwin' ]; then
> >   export DOCKER_DEFAULT_PLATFORM ?= linux/amd64
> > fi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It should also be verified that when running in a non-mac OS that variable is not set on the docker build
> 
> As a side note, I wonder if the same issue happens if podman is used, and also whether podman makes use of that envvar too

I tested the suggestion with podman and docker, it works with docker but unfortunately it doesn't work with podman, seems like they don't use that env var 😞. 

However, the build cmd does have a `--platform` parameter for both which we could specify if the os is darwin as an alternative.  

```
ifeq ($(shell uname -s | tr A-Z a-z), darwin)
	DOCKER_BUILD_PLATFORM ?= --platform linux/amd64
endif

...
# example build cmd
image/build:
        $(DOCKER) build $(DOCKER_BUILD_PLATFORM) ...
```

miguelsorianod:
> > > This could be achieved by doing something like (pseudocode, needs to be reviewed):
> > > ```
> > > OS=$(uname)
> > > if [ "$OS" = 'Darwin' ]; then
> > >   export DOCKER_DEFAULT_PLATFORM ?= linux/amd64
> > > fi
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It should also be verified that when running in a non-mac OS that variable is not set on the docker build
> > 
> > 
> > As a side note, I wonder if the same issue happens if podman is used, and also whether podman makes use of that envvar too
> 
> I tested the suggestion with podman and docker, it works with docker but unfortunately it doesn't work with podman, seems like they don't use that env var disappointed.
> 
> However, the build cmd does have a `--platform` parameter for both which we could specify if the os is darwin as an alternative.
> 
> ```
> ifeq ($(shell uname -s | tr A-Z a-z), darwin)
> 	DOCKER_BUILD_PLATFORM ?= --platform linux/amd64
> endif
> 
> ...
> # example build cmd
> image/build:
>         $(DOCKER) build $(DOCKER_BUILD_PLATFORM) ...
> ```

Let's do that. But let's called it `CONTAINER_IMAGE_BUILD_PLATFORM` instead, to make the name agnostic of the container engine being used

Makefile
```diff
@@ -70,6 +70,14 @@ CURL ?= curl
 OC ?= oc
 OCM ?= ocm
 
+# Set os specific variables
+ifeq ($(shell uname -s | tr A-Z a-z), darwin)
+    PGHOST:="127.0.0.1"
```
machi1990, line 75: nitpick: this line is not aligned with others?

Makefile
```diff
@@ -70,6 +70,14 @@ CURL ?= curl
 OC ?= oc
 OCM ?= ocm
 
+# Set os specific variables
+ifeq ($(shell uname -s | tr A-Z a-z), darwin)
+	PGHOST:="127.0.0.1"
```
JameelB, line 75: Side note, i don't see this used anywhere. Can we remove it?

Makefile
JameelB: fixed this just now. I added a comment about removing this though in this thread: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1602#discussion_r1121711662

Makefile
```diff
@@ -159,19 +159,16 @@ openapi/spec/validate: specinstall
...
-endif
-
 ### Environment-sourced variables with defaults
 # Can be overriden by setting environment var before running
 # Example:
 #   OCM_ENV=testing make run
 #   export OCM_ENV=testing; make run
+
+ifeq ($(shell uname -s | tr A-Z a-z), darwin)
+    CONTAINER_IMAGE_BUILD_PLATFORM ?= --platform linux/amd64
```
miguelsorianod, line 169: I see some issues on the CI side.
Maybe this default value has to be quoted?

It contains dashes which mght be being interpreted as some shell flag

JameelB: i tried removing this assignment and i'm also getting the same issue. 

In main, as soon as I removed these lines, I get the same issue as what's shown in the CI
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/738b7b881aa293915f8374e1ae380549f75e5d8c/Makefile#L165-L169

These vars never get assigned which causes the error
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/738b7b881aa293915f8374e1ae380549f75e5d8c/Makefile#L177-L198

Going to look into this a bit further 👀

machi1990: @JameelB I'd suggest we defer removing this onto another PR.

miguelsorianod: I see the spaces inside the PGHOST ifelse  are whitespaces and not tabs. Might that be causing the issue?

Makefile
```diff
@@ -70,6 +70,39 @@ CURL ?= curl
...
+### Environment-sourced variables with defaults
+# Can be overriden by setting environment var before running
+# Example:
+#   OCM_ENV=testing make run
+#   export OCM_ENV=testing; make run
+export GOPROXY=https://proxy.golang.org
+export GOPRIVATE=gitlab.cee.redhat.com
+
+ifeq ($(shell uname -s | tr A-Z a-z), darwin)
+	CONTAINER_IMAGE_BUILD_PLATFORM ?= --platform linux/amd64
```
miguelsorianod, line 82: Could we double quote the value?

Makefile
miguelsorianod: > When testing this, it seems like Makefile doesn't like var definitions between target definitions.
> 
> I've moved all of the var definitions before the target definitions starts to ensure they're set correctly.

Even though this makes the tests pass and "fix the issue" it would be good to figure out what is happening.
A way to do this is to create a very small makefile where we try the combinations.

JameelB: Here's a gist with different combinations: https://gist.github.com/JameelB/4ecd37efb2cf4b88e7f75c6a1302e1c4

Makefile
JameelB: updated, note that I had to use echo for this in the image build commands to get rid of the quotes as docker build doesn't accept `"--platform linux/amd64"`

Makefile
miguelsorianod: I think I can identify where is the issue:

Inside `ifndef` ... `endif` blocks it seems that the characters used to indent matters.
For example, for the TEST_VAR_4 ifndef block:
if inside the body the TEST_VAR_4 initialization is not indented then it seems to work
if inside the body the TEST_VAR_4 initialization is indented with spaces it seems to work
if inside the body the TEST_VAR_4 initialization is indented with tabs it does not seem to work

miguelsorianod: This means that the order where var definitions appear shouldn't really matter. The issue is another and has to do with indentation inside Makefile conditional directives

Makefile
miguelsorianod: Ok then let's unquote it. It seems that Makefile doesn't really interpret single/double quotes and it considers them literally

Makefile
JameelB: good spot, that looks like it.

Vars like `OCM_ENV` which is used across multiple make targets should  probably stay at the top though. 

We can re-arrange some of the target specific ones like `TEST_SUMMARY_FORMAT` to be grouped with the test targets if preferred. Some of the vars here are only ever used for one target as well (i.e. SERVER_URL), in that case it should probably be defined as a target variable. 🤔 

There's probably more vars in there that needs to be reviewed to be removed or moved 🤔 I can create a follow on PR for this?

miguelsorianod: rearranging them is good for organization purposes so we can do that.
But in any case we have to fix the issues of tabs being used in conditionals that are not shell conditionals but makefile conditionals.
For this PR let's do the order reorganization and fixing the indentation for the makefile conditionals

Makefile
JameelB: quotes removed 👍

Makefile
JameelB: i've updated this to remove the indentation from the make if statements. 

I've also moved the ones in this block to group them with their related make targets but I haven't looked at any of the other var definitions outside of this.

miguelsorianod (approved):

PR #1605, opened 2023-03-01
set default value for service account to avoid creating them when they are not needed when running integration tests. 

machi1990:
## Description

1. set default value for service account to avoid creating them when they are not needed when running integration tests. 
2. Plus other minor cleanups and improvements  in the test helper

## Verification Steps

Integration tests should pass

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/converters/cluster.go
+++ b/internal/kafka/internal/converters/cluster.go
@@ -25,6 +25,8 @@ func ConvertCluster(cluster *api.Cluster) []map[string]interface{} {
 			"multi_az":                cluster.MultiAZ,
 			"status":                  cluster.Status,
 			"cluster_id":              cluster.ClusterID,
+			"client_id":               cluster.ClientID,
+			"client_secret":           cluster.ClientSecret,
 			"external_id":             cluster.ExternalID,
 			"created_at":              cluster.Meta.CreatedAt,
 			"updated_at":              cluster.Meta.UpdatedAt,
--- a/internal/kafka/test/helper.go
+++ b/internal/kafka/test/helper.go
@@ -188,7 +188,7 @@ func deleteClustersServiceAccountsFromSSOService(cluster *api.Cluster, keycloakC
 
 		err := kafkaSsoService.DeleteServiceAccountInternal(cluster.ClientID)
 		if err != nil {
-			glog.Warningf("Failed to delete Fleetshard client from SSO for cluster %q. The sso provider is %q", cluster.ClusterID, keycloakConfig.SelectSSOProvider)
+			glog.Warningf("Failed to delete Fleetshard client with id %q from SSO for cluster %q due to %q. The sso provider is %q", cluster.ClientID, cluster.ClusterID, err.Error(), keycloakConfig.SelectSSOProvider)
 		}
 	}
 
@@ -205,7 +205,7 @@ func deleteClustersServiceAccountsFromSSOService(cluster *api.Cluster, keycloakC
 	err := osdSSOService.DeRegisterClientInSSO(cluster.ID)
 
 	if err != nil {
-		glog.Warningf("Failed to delete IDP configuration client from SSO for cluster %q. The sso provider is %q", cluster.ClusterID, keycloakConfig.SelectSSOProvider)
+		glog.Warningf("Failed to delete IDP configuration client from SSO for cluster %q due to %q. The sso provider is %q", cluster.ClusterID, err.Error(), keycloakConfig.SelectSSOProvider)
 	}
 }
 
@@ -222,6 +222,6 @@ func deleteKafkasServiceAccountsFromSSOService(kafka *dbapi.KafkaRequest, keyclo
 	err := kafkaSsoService.DeleteServiceAccountInternal(kafka.CanaryServiceAccountClientID)
 
 	if err != nil {
-		glog.Warningf("Failed to delete Kafka canary service account from SSO service for kafka %q. The sso provider is %q", kafka.Name, keycloakConfig.SelectSSOProvider)
+		glog.Warningf("Failed to delete Kafka canary service account with id %q from SSO service for kafka %q due to %q. The sso provider is %q", kafka.CanaryServiceAccountClientID, kafka.Name, err.Error(), keycloakConfig.SelectSSOProvider)
 	}
 }
--- a/internal/kafka/test/helper.go
+++ b/internal/kafka/test/helper.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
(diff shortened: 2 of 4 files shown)


miguelsorianod:
Could you provide some more context around this?

What is "service account" in this context? What has recently changed that has made this needed?

machi1990:
> Could you provide some more context around this?
> 
> What is "service account" in this context? What has recently changed that has made this needed?

Nothing has changed. But there are midstream test failures related to service account limits being reached. See the midstream pipelines for more details. The reason the limits were reached was because some tests creating service accounts even when they were not needed.

miguelsorianod:
> Nothing has changed. But there are midstream test failures related to service account limits being reached. See the midstream pipelines for more details. The reason the limits were reached was because some tests creating service accounts even when they were not needed.

Wouldn't the solution to that be deleting the serviceaccounts after they are used?

By not letting the reconciler create the SAs some tests might not catch the same set of issues

machi1990:
> > Nothing has changed. But there are midstream test failures related to service account limits being reached. See the midstream pipelines for more details. The reason the limits were reached was because some tests creating service accounts even when they were not needed.
> 
> Wouldn't the solution to that be deleting the serviceaccounts after they are used?

The solution is two fold:
- First, proper test setup under the context of what is being tested E.g when the test want the service account to be created, it has to be setup in a way that it can do so.  This is the case for some of (cluster successfully reconciliation, kafka successfully reconciliation, service accounts creation..) tests already. Another proper test setup, if the test does not want a service account to be created, it has to avoid creating it all: this is the case for most of our tests that uses the builders
- Second, when the test setup requires the creation of the service accounts, then before the test ends, the test has to ensure that the service accounts are deleted. This is there in place.

Now one can ask, if two is there already why do we need (1) ? 

The answer is with (1), two is (sometimes a.k.a) executed before the service accounts are created by the reconciler and just before the test is teardown. The test setup could be improved a little bit to ensure that we only delete the service accounts after the reconcilers have been stopped: this way we only delete the service accounts that have been created. However, that requires a little more time to restructure the test setup than I was willing to invest and I preferred to tackle the root cause first of not creating the SAs and keep the current setup to delete the ones that we've wanted to be explicitly created by (2).

> By not letting the reconciler create the SAs some tests might not catch the same set of issues

hmm what other tests, the ones that needs the service accounts created?

pawelpaszki:
@machi1990 - is there a way to verify your changes?

machi1990:
> @machi1990 - is there a way to verify your changes?

Yes, thank you. You could run the tests against RH sso and verify that after the tests have completed - there are no left over service accounts that have been created.
For that: 
1. create a service account against rh sso
2. change https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/f3f61d25ad4d3fd80099fd848375acd789348c8c/internal/kafka/internal/environments/integration.go#L38 and point it to https://sso.redhat.com
3. change https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/f3f61d25ad4d3fd80099fd848375acd789348c8c/internal/kafka/internal/environments/integration.go#L40 and point it to `redhat_sso`
3. Run the tests with OCM_ENV=integration. e.g `OCM_ENV=integration TEST_SUMMARY_FORMAT=standard-verbose  make test/integration/kafka` 

The PR might not have covered all the cases so if you see anything left over please point it out. I could use the logs as well to debug further - once you've the logs from (4) and there are left over service accounts, please share the logs via some internal channel.

pawelpaszki:
@machi1990 - there are no left over service accounts after verifying using your instructions. I'll leave @miguelsorianod confirm that his comments were addressed

machi1990:
> @machi1990 - there are no left over service accounts after verifying using your instructions. I'll leave @miguelsorianod confirm that his comments were addressed

Thanks @pawelpaszki

miguelsorianod:
> > > Nothing has changed. But there are midstream test failures related to service account limits being reached. See the midstream pipelines for more details. The reason the limits were reached was because some tests creating service accounts even when they were not needed.
> > 
> > 
> > Wouldn't the solution to that be deleting the serviceaccounts after they are used?
> 
> The solution is two fold:
> 
>     * First, proper test setup under the context of what is being tested E.g when the test want the service account to be created, it has to be setup in a way that it can do so.  This is the case for some of (cluster successfully reconciliation, kafka successfully reconciliation, service accounts creation..) tests already. Another proper test setup, if the test does not want a service account to be created, it has to avoid creating it all: this is the case for most of our tests that uses the builders
> 
>     * Second, when the test setup requires the creation of the service accounts, then before the test ends, the test has to ensure that the service accounts are deleted. This is there in place.
> 
> 
> Now one can ask, if two is there already why do we need (1) ?
> 
> The answer is with (1), two is (sometimes a.k.a) executed before the service accounts are created by the reconciler and just before the test is teardown. The test setup could be improved a little bit to ensure that we only delete the service accounts after the reconcilers have been stopped: this way we only delete the service accounts that have been created. However, that requires a little more time to restructure the test setup than I was willing to invest and I preferred to tackle the root cause first of not creating the SAs and keep the current setup to delete the ones that we've wanted to be explicitly created by (2).
> 
> > By not letting the reconciler create the SAs some tests might not catch the same set of issues
> 
> hmm what other tests, the ones that needs the service accounts created?

The solution outlined here it is reasonable.
However, I think the main issue is that we don't want canary SAs being created right? In that case, a cleaner solution would seem to control that as a feature flag. Then, in the tests we disable canary SA by setting the appropriate feature flag

machi1990:
Sorry for the late reply. Only getting to this now.

> The solution outlined here it is reasonable. However, I think the main issue is that we don't want canary SAs being created right? 

Yes - the canary service account and the fleetshard service account.

> In that case, a cleaner solution would seem to control that as a feature flag. Then, in the tests we disable canary SA by setting the appropriate feature flag

I would agree. But this being test specific behavior, I think we can live off adding another feature flag(s) to control the creation or not of service accounts.

machi1990:
@pawelpaszki @miguelsorianod ping for approval

pawelpaszki (approved):

miguelsorianod:
> I would agree. But this being test specific behavior, I think we can live off adding another feature flag(s) to control the creation or not of service accounts.

I think having a feature flag to control it would be worth it even if it is test-specific. More and more tests will be added which will increase the complexity of them for where this is involved. I find it will be cleaner to directly just don't do it and enable it when needed instead of working it around adding dummy values. This can be done in a future PR

miguelsorianod (approved):
See my previous comment

machi1990:
Thanks @pawelpaszki @miguelsorianod for the review

machi1990:
> > I would agree. But this being test specific behavior, I think we can live off adding another feature flag(s) to control the creation or not of service accounts.
> 
> I think having a feature flag to control it would be worth it even if it is test-specific. More and more tests will be added which will increase the complexity of them for where this is involved. I find it will be cleaner to directly just don't do it and enable it when needed instead of working it around adding dummy values. This can be done in a future PR

Thanks, happy to review the enhancement PR.

PR #1609, opened 2023-03-03
fix: token check should not consider multiaz value of cluster

pb82:
token check should not consider multiaz value of cluster


internal/kafka/internal/handlers/observatorium_proxy.go
```diff
@@ -40,7 +40,7 @@ func (h observatoriumProxyHandler) ValidateTokenAndExternalClusterID(w http.Resp
 				return nil, errors.Unauthenticated("unable to get client ID from the token")
 			}
 
-			cluster, err := h.clusterService.FindCluster(services.FindClusterCriteria{ExternalID: external_id, MultiAZ: true})
+			cluster, err := h.clusterService.FindCluster(services.FindClusterCriteria{ExternalID: external_id})
```
machi1990, line 43:
```suggestion
			cluster, err := h.clusterService.FindCluster(services.FindClusterCriteria{ExternalID: externalID})
```

machi1990 (approved):
left a naming suggestion, otherwise lgtm

when the test passes, we can merge this.

miguelsorianod (approved):
Looks good to me.
It can be merged once the tests pass.

PR #1617, opened 2023-03-07
feat(enterprise): validate that the organization owns the cluster that's being registered

machi1990:
## Description

Validate that the organization owns the cluster that's being registered. Before this, any org admin could register the cluster even the ones that are not owned.

Closes https://issues.redhat.com/browse/MGDSTRM-10583

## Verification Steps

Added integration tests and unit test should pass

- [ ] Code Review completed
- ~~[ ] All PR comments are resolved either by addressing them or creating follow up tasks~~

### DIFF
--- a/internal/kafka/internal/clusters/ocm_provider.go
+++ b/internal/kafka/internal/clusters/ocm_provider.go
@@ -9,6 +9,7 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/clusters/types"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/client/ocm"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	svcErrors "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
@@ -533,4 +534,23 @@ func (o *OCMProvider) GetClusterResourceQuotaCosts() ([]types.QuotaCost, error) 
 		})
 	}
 	return quotaCostList, nil
+}
+
+func (o *OCMProvider) CheckIfOrganizationIsTheClusterOwner(externalOrganizationID, clusterID, clusterExternalID string) error {
+	organizationID, err := o.ocmClient.GetOrganisationIdFromExternalId(externalOrganizationID)
+	if err != nil || shared.StringEmpty(organizationID) {
+		return svcErrors.NewWithCause(svcErrors.ErrorGeneral, err, "failed to retrieve organization details")
+	}
+
+	query := fmt.Sprintf("cluster_id = '%s' AND external_cluster_id = '%s' AND organization_id = '%s'", clusterID, clusterExternalID, organizationID)
+	subcriptions, err := o.ocmClient.FindSubscriptions(query)
+	if err != nil {
+		return svcErrors.NewWithCause(svcErrors.ErrorGeneral, err, "failed to validate if the organization with id %q owns the cluster with id %q", externalOrganizationID, clusterID)
+	}
+
+	if arrays.IsEmpty(subcriptions) {
+		return svcErrors.Forbidden("organization with id %q does not own the cluster with id %q", externalOrganizationID, clusterID)
+	}
+
+	return nil
 }
--- a/internal/kafka/internal/clusters/ocm_provider_test.go
+++ b/internal/kafka/internal/clusters/ocm_provider_test.go
@@ -1622,3 +1622,130 @@ func TestOCMProvider_GetClusterResourceQuotaCosts(t *testing.T) {
 		})
 	}
 }
+
+func TestOCMProvider_CheckIfOrganizationIsTheClusterOwner(t *testing.T) {
+	g := gomega.NewWithT(t)
+	type fields struct {
+		ocmClient *ocm.ClientMock
+	}
+	type args struct {
+		externalOrganizationID string
+		clusterID              string
+		clusterExternalID      string
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "return an error when getting the ams organization id fails",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						return "", fmt.Errorf("error")
+					},
+				},
+			},
+			args: args{
+				externalOrganizationID: "org-id",
+				clusterID:              "cluster-id",
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: true,
+		},
+		{
+			name: "return an error when getting the organization is not found from AMS",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						return "", nil
+					},
+				},
+			},
+			args: args{
+				externalOrganizationID: "org-id",
+				clusterID:              "cluster-id",
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: true,
+		},
+		{
+			name: "return an error when getting the cluster subscription fails",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						return "some-ams-org-id", nil
+					},
+					FindSubscriptionsFunc: func(query string) ([]*accountsmgmtv1.Subscription, error) {
+						g.Expect(query).To(gomega.Equal(`cluster_id = 'cluster-id' AND external_cluster_id = 'cluster-external-id' AND organization_id = 'some-ams-org-id'`))
+						return nil, fmt.Errorf("some error")
+					},
+				},
+			},
+			args: args{
+				externalOrganizationID: "org-id",
+				clusterID:              "cluster-id",
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: true,
+		},
+		{
+			name: "return an error when no subscription is found for the given cluster information",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						return "some-ams-org-id", nil
+					},
+					FindSubscriptionsFunc: func(query string) ([]*accountsmgmtv1.Subscription, error) {
+						g.Expect(query).To(gomega.Equal(`cluster_id = 'cluster-id' AND external_cluster_id = 'cluster-external-id' AND organization_id = 'some-ams-org-id'`))
+						return []*accountsmgmtv1.Subscription{}, nil
+					},
+				},
+			},
+			args: args{
+				externalOrganizationID: "org-id",
+				clusterID:              "cluster-id",
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: true,
+		},
+		{
+			name: "do not return an error when subscription is found for the given cluster information",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						return "some-ams-org-id", nil
+					},
+					FindSubscriptionsFunc: func(query string) ([]*accountsmgmtv1.Subscription, error) {
+						g.Expect(query).To(gomega.Equal(`cluster_id = 'cluster-id' AND external_cluster_id = 'cluster-external-id' AND organization_id = 'some-ams-org-id'`))
+						return []*accountsmgmtv1.Subscription{
+							{},
+						}, nil
+					},
+				},
+			},
+			args: args{
+				externalOrganizationID: "org-id",
+				clusterID:              "cluster-id",
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		testcase := tt
+		t.Run(testcase.name, func(t *testing.T) {
+			t.Parallel()
+			g := gomega.NewWithT(t)
+			o := &OCMProvider{
+				ocmClient: testcase.fields.ocmClient,
+			}
+
+			err := o.CheckIfOrganizationIsTheClusterOwner(testcase.args.externalOrganizationID, testcase.args.clusterID, testcase.args.clusterExternalID)
+			g.Expect(err != nil).To(gomega.Equal(testcase.wantErr))
+			g.Expect(testcase.fields.ocmClient.GetOrganisationIdFromExternalIdCalls()[0].ExternalId).To(gomega.Equal(testcase.args.externalOrganizationID))
+		})
+	}
+}
--- a/internal/kafka/internal/clusters/provider.go
+++ b/internal/kafka/internal/clusters/provider.go
@@ -46,6 +46,8 @@ type Provider interface {
 	// GetClusterResourceQuotaCosts returns a list of quota cost information related to resources used for the provisioning and
 	// terraforming of data plane clusters for the authenticated user.
 	GetClusterResourceQuotaCosts() ([]types.QuotaCost, error)
+	// CheckIfOrganizationIsTheClusterOwner perform the check to verify if the organization (given by the external organization id) owns the cluster with the given cluster id and cluster external id
+	CheckIfOrganizationIsTheClusterOwner(externalOrganizationID, clusterID, clusterExternalID string) error
 }
 
 // ProviderFactory used to return an instance of Provider implementation
--- a/internal/kafka/internal/clusters/standalone_provider.go
+++ b/internal/kafka/internal/clusters/standalone_provider.go
@@ -583,3 +583,7 @@ func (s *StandaloneProvider) GetClusterResourceQuotaCosts() ([]types.QuotaCost, 
 	var quotaCostList []types.QuotaCost
 	return quotaCostList, nil
 }
+
+func (s *StandaloneProvider) CheckIfOrganizationIsTheClusterOwner(externalOrganizationID, clusterID, clusterExternalID string) error {
+	return nil
+}
--- a/internal/kafka/internal/handlers/cluster.go
+++ b/internal/kafka/internal/handlers/cluster.go
@@ -56,27 +56,16 @@ func (h clusterHandler) RegisterEnterpriseCluster(w http.ResponseWriter, r *http
 			validateKafkaMachinePoolNodeCount(&clusterPayload),
 		},
 		Action: func() (interface{}, *errors.ServiceError) {
-
 			clusterSpec, getClusterErr := provider.GetClusterSpec(clusterPayload.ClusterId)
 			if getClusterErr != nil && shared.IsNil(clusterSpec) {
 				return nil, errors.GeneralError("failed to get cluster by ID: %s", clusterPayload.ClusterId)
 			}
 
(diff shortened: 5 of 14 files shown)


internal/kafka/internal/clusters/ocm_provider_test.go
```diff
@@ -1622,3 +1622,140 @@ func TestOCMProvider_GetClusterResourceQuotaCosts(t *testing.T) {
...
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: true,
+		},
+		{
+			name: "return an error when getting the cluster subscription fails",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						g.Expect(externalID).To(gomega.Equal("org-id"))
```
ziccardi, line 1681: This can be done out of the mocked function:
```go
g.Expect(testcase.fields.ocmClient.GetOrganisationIdFromExternalIdCalls()[0].ExternalId).To(gomega.Equal("org-id"))
```

The only requirement is that `ocmClient` in the `fields` struct is declared as `*ocm.ClientMock`:
```go
	type fields struct {
		ocmClient *ocm.ClientMock
	}
```

internal/kafka/internal/clusters/ocm_provider_test.go
```diff
@@ -1622,3 +1622,140 @@ func TestOCMProvider_GetClusterResourceQuotaCosts(t *testing.T) {
...
+						return "some-ams-org-id", nil
+					},
+					GetSubscriptionsFunc: func(page, size int, search string) ([]*accountsmgmtv1.Subscription, error) {
+						g.Expect(page).To(gomega.Equal(0))
+						g.Expect(size).To(gomega.Equal(1))
+						g.Expect(search).To(gomega.Equal(`cluster_id = 'cluster-id' AND external_cluster_id = 'cluster-external-id' AND organization_id = 'some-ams-org-id'`))
```
ziccardi, line 1687: see previous comment

internal/kafka/internal/clusters/ocm_provider_test.go
```diff
@@ -1622,3 +1622,140 @@ func TestOCMProvider_GetClusterResourceQuotaCosts(t *testing.T) {
...
```
ziccardi, line 1710: see previous comment

internal/kafka/internal/clusters/ocm_provider_test.go
```diff
@@ -1622,3 +1622,140 @@ func TestOCMProvider_GetClusterResourceQuotaCosts(t *testing.T) {
...
```
ziccardi, line 1733: see previous comment

internal/kafka/internal/handlers/cluster.go
```diff
@@ -86,6 +75,22 @@ func (h clusterHandler) RegisterEnterpriseCluster(w http.ResponseWriter, r *http
 				return nil, errors.New(errors.ErrorUnauthorized, "non admin user not authorized to perform this action")
 			}
 
+			ownershipCheckErr := provider.CheckIfOrganizationIsTheClusterOwner(orgId, clusterPayload.ClusterId, clusterSpec.ExternalID)
+			if ownershipCheckErr != nil {
+				serviceErr, ok := ownershipCheckErr.(*errors.ServiceError)
+				if !ok {
+					return nil, errors.NewWithCause(errors.ErrorGeneral, ownershipCheckErr, ownershipCheckErr.Error())
+				}
+				return nil, serviceErr
```
ziccardi, line 84: I'd prefer the opposite (`if ok` instead of `if !ok`):

``` go
	serviceErr, ok := ownershipCheckErr.(*errors.ServiceError)
	if ok {
		return nil, serviceErr
	}
	return nil, errors.NewWithCause(errors.ErrorGeneral, ownershipCheckErr, ownershipCheckErr.Error())
```

or

```go
	if serviceErr, ok := ownershipCheckErr.(*errors.ServiceError); ok {
		return nil, serviceErr	
	}
	return nil, errors.NewWithCause(errors.ErrorGeneral, ownershipCheckErr, ownershipCheckErr.Error())
```

internal/kafka/internal/clusters/ocm_provider_test.go
ziccardi: > The reason why I prefer doing it inline is that it is scoped to a testcase and avoid the need to perform conditional checks in some scenario where a particular testcase shouldn't perform this call.

Make sense

internal/kafka/internal/handlers/cluster.go
machi1990: +1  I had forgotten to refactor it.

internal/kafka/internal/clusters/ocm_provider_test.go
machi1990: thanks, responded there https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1617#discussion_r1127988847

internal/kafka/internal/clusters/ocm_provider_test.go
machi1990: thanks, responded there https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1617#discussion_r1127988847

internal/kafka/internal/clusters/ocm_provider_test.go
```diff
@@ -1622,3 +1622,140 @@ func TestOCMProvider_GetClusterResourceQuotaCosts(t *testing.T) {
...
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: true,
+		},
+		{
+			name: "do not return an error when subscription is found for the given cluster information",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						g.Expect(externalID).To(gomega.Equal("org-id"))
```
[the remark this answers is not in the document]
machi1990: done

internal/kafka/internal/clusters/ocm_provider_test.go
machi1990: responded there https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1617#discussion_r1127988847

internal/kafka/internal/clusters/ocm_provider_test.go
```diff
@@ -1622,3 +1622,140 @@ func TestOCMProvider_GetClusterResourceQuotaCosts(t *testing.T) {
...
+				clusterExternalID:      "cluster-external-id",
+			},
+			wantErr: true,
+		},
+		{
+			name: "return an error when no subscription is found for the given cluster information",
+			fields: fields{
+				ocmClient: &ocm.ClientMock{
+					GetOrganisationIdFromExternalIdFunc: func(externalID string) (string, error) {
+						g.Expect(externalID).To(gomega.Equal("org-id"))
```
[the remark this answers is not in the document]
machi1990: done

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -534,3 +534,22 @@ func (o *OCMProvider) GetClusterResourceQuotaCosts() ([]types.QuotaCost, error)
...
+		return svcErrors.NewWithCause(svcErrors.ErrorGeneral, err, "failed to retrieve organization details")
+	}
+
+	search := fmt.Sprintf("cluster_id = '%s' AND external_cluster_id = '%s' AND organization_id = '%s'", clusterID, clusterExternalID, organizationID)
+	subcriptions, err := o.ocmClient.GetSubscriptions(0, 1, search)
+	if err != nil {
+		return svcErrors.NewWithCause(svcErrors.ErrorGeneral, err, "failed to validate if the organization with id %q owns the cluster with id %q", externalOrganizationID, clusterID)
+	}
+
+	if len(subcriptions) == 0 {
```
ziccardi, line 550: Side comment: I wonder if adding an `arrays.IsEmpty(ary []T) bool` function to the array utilities would make these tests more readable.

machi1990: +1 - I'd agree. Ever since there is `shared.StringEmpty` I've been using it almost everywhere.

test/mocks/api_server.go
```diff
@@ -720,6 +736,31 @@ func NewSubscriptionList() *subscriptionList {
...
 }
 
+type organizationList struct {
+	Page  int             `json:"page"`
+	Size  int             `json:"size"`
+	Total int             `json:"total"`
+	Items json.RawMessage `json:"items"`
+}
+
+func (l *organizationList) WithItems(items interface{}) (*organizationList, error) {
```
ziccardi, line 746:
```suggestion
func (l *organizationList) WithItems(items any) (*organizationList, error) {
```

ziccardi:
Code looks good to me. I just added a minor comment about using `any` instead of `interface{}` (`any` is just an alias for `interface{}` but, IMO, it is more readable).

pkg/shared/utils/arrays/generic_array_utils_test.go
```diff
@@ -403,3 +403,38 @@ func Test_ForEach(t *testing.T) {
...
+		args args
+		want bool
+	}{
+		{
+			name: "return true if array is empty",
+			args: args{
+				values: []int{},
+			},
+			want: true,
+		},
```
ziccardi, line 422: mind adding a test with `nil` ?

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -99,3 +99,8 @@ func ForEach[T any](values []T, consumer ConsumerFunc[T]) {
 		consumer(values[i])
 	}
 }
+
+// IsEmpty returns true if the array is empty, otherwise returns false.
```
miguelsorianod, line 103: Can we clarify that an array is considered "empty" in this context if the array is nil or if the array is not nil but has zero elements?

internal/kafka/internal/clusters/standalone_provider.go
```diff
@@ -583,3 +583,7 @@ func (s *StandaloneProvider) GetClusterResourceQuotaCosts() ([]types.QuotaCost,
 	var quotaCostList []types.QuotaCost
 	return quotaCostList, nil
 }
+
+func (s *StandaloneProvider) CheckIfOrganizationIsTheClusterOwner(externalOrganizationID, clusterID, clusterExternalID string) error {
+	return nil
```
miguelsorianod, line 588: Shouldn't this check the config file?

pkg/shared/utils/arrays/generic_array_utils.go
machi1990: Thanks, i'll add that.

internal/kafka/internal/clusters/standalone_provider.go
machi1990: Which config? The cluster list config or the quota list management configuration?

pkg/shared/utils/arrays/generic_array_utils_test.go
machi1990: done

pkg/shared/utils/arrays/generic_array_utils.go
machi1990: done

internal/kafka/internal/clusters/standalone_provider.go
miguelsorianod: I was thinking about the quota list management configuration

machi1990: Right. I think the list config is specific to a Kafka quota i.e we do not data plane cluster quota info in that config.

internal/kafka/internal/clusters/ocm_provider.go
```diff
@@ -534,3 +535,22 @@ func (o *OCMProvider) GetClusterResourceQuotaCosts() ([]types.QuotaCost, error)
...
 }
+
+func (o *OCMProvider) CheckIfOrganizationIsTheClusterOwner(externalOrganizationID, clusterID, clusterExternalID string) error {
+	organizationID, err := o.ocmClient.GetOrganisationIdFromExternalId(externalOrganizationID)
+	if err != nil || shared.StringEmpty(organizationID) {
+		return svcErrors.NewWithCause(svcErrors.ErrorGeneral, err, "failed to retrieve organization details")
+	}
+
+	search := fmt.Sprintf("cluster_id = '%s' AND external_cluster_id = '%s' AND organization_id = '%s'", clusterID, clusterExternalID, organizationID)
+	subcriptions, err := o.ocmClient.GetSubscriptions(0, 1, search)
```
miguelsorianod, line 546: Seeing `o.ocmClient.GetSubscriptions(0, 1, search)` is not a very good interface: What is 0? and 1?

I think we can create a data type for page and size. For example `PagingOptions`

pkg/client/ocm/client.go
```diff
@@ -56,6 +56,8 @@ type Client interface {
 	GetQuotaCostsForProduct(organizationID, resourceName, product string) ([]*amsv1.QuotaCost, error)
 	// GetCurrentAccount returns the account information of the current authenticated user
 	GetCurrentAccount() (*amsv1.Account, error)
+	// GetSubscriptions returns the list of subscriptions that matches the given search criteria
+	GetSubscriptions(page, size int, search string) ([]*amsv1.Subscription, error)
```
miguelsorianod, line 60: I think it might be more accurate to call it ListSubscriptions as that is how AMS calls it

miguelsorianod: There is also a FindSubscriptions method that I think might have overlapping functionality. Do we really need both methods?

ziccardi: Good spot. I believe we could enhance the `FindSubscriptions` to add page and pagesize and get rid of this method.
To avoid breaking the current code, we could use `functional options`:

```go
func Page(page int) FindSubscriptionsPageParams {
	return func(r *amsv1.SubscriptionsListRequest) {
		r.Page(page)
	}
}

func PageSize(pageSize int) FindSubscriptionsPageParams {
	return func(r *amsv1.SubscriptionsListRequest) {
		r.Size(pageSize)
	}
}

type FindSubscriptionsPageParams func(r *amsv1.SubscriptionsListRequest)

func (c client) FindSubscriptions(query string, params ...FindSubscriptionsPageParams) ([]*amsv1.Subscription, error) {
	req := c.connection.AccountsMgmt().V1().Subscriptions().List().Search(query)
	for _, param := range params {
		param(req)
	}

	r, err := req.Send()
	if err != nil {
		return nil, err
	}
	return r.Items().Slice(), nil
}
```

This way, we could call the method as:
```go
// no options
c.FindSubscriptions(query)

// specific page
c.FindSubscriptions(query, Page(3))

// specific page with page size
c.FindSubscriptions(query, Page(3), PageSize(50))
```

miguelsorianod: I am not concerned about breaking the code. It should not be too difficult to refactor.

I am unsure about whether we need to go as far as using the functional options patterns for paging of the method.
I don't expect the set of paging options to be very high or unpredictable so in that case I tend to prefer to keep it simple and just use a typed struct with the options. More options can be added in the future is needed as attributes.
It will keep the signature easy, there will be less abstraction and indirections and if we need to add a new paging option (which won't be frequent) it should be as easy as adding a new attribute.
It would also avoid having a variadic argument which carry the implication that there cannot be other additional arguments after it.

As explained, for a case like this I would prefer to keep it simple and have a solution that contains a balance between simplicity and practicality.

pkg/shared/utils/arrays/generic_array_utils.go
```diff
@@ -99,3 +99,9 @@ func ForEach[T any](values []T, consumer ConsumerFunc[T]) {
...
```
miguelsorianod, line 103: To be accurate, this is about a slice and not about an array

pkg/client/ocm/client.go
ziccardi: > I am not concerned about breaking the code. It should not be too difficult to refactor.

Yes, but we will need to pass the new parameter even when unnecessary.

> It would also avoid having a variadic argument which carry the implication that there cannot be other additional arguments after it.

True, but even `page, pagesize` are parameters passed typically as the first or last parameter, so passing other parameters after the `page` will be unusual.

Anyway, the implementation detail is debatable, and I'm open to any solution.

miguelsorianod: I see your points and the first point is certainly an implication (as using functional options has other implications)

:+1: to your comment, I think both options are reasonable too. I would tend to simplicity in this specific case for the reasons outlined (there are other cases where using the functional options pattern is a very appropriate pattern) but I am open to whatever is chosen.

machi1990: @ziccardi @miguelsorianod the existing `FindSubscriptions` method is sufficient. For this use case, size was added to ensure only one item is returned but that's not really needed since the constructed query is unique enough to return one element.

machi1990: done, thanks for catching this one

test/mocks/api_server.go
```diff
@@ -367,6 +369,10 @@ func (b *MockConfigurableServerBuilder) SetGetOrganizationQuotaCost(list *amsv1.
 	b.handlerRegister[EndpointOrganizationQuotaCostGet] = buildMockRequestHandler(list, err)
 }
 
+func (b *MockConfigurableServerBuilder) SetGetOrganizations(list *amsv1.OrganizationList, err *ocmErrors.ServiceError) {
```
pawelpaszki, line 372:
```suggestion
func (b *MockConfigurableServerBuilder) SetOrganizationsGetResponse(list *amsv1.OrganizationList, err *ocmErrors.ServiceError) {
```

pawelpaszki: can we also add a comment with the url for which the response is set?

machi1990: Thanks Pawel, doing it now

internal/kafka/internal/handlers/cluster.go
```diff
@@ -86,6 +75,23 @@ func (h clusterHandler) RegisterEnterpriseCluster(w http.ResponseWriter, r *http
 				return nil, errors.New(errors.ErrorUnauthorized, "non admin user not authorized to perform this action")
 			}
 
+			ownershipCheckErr := provider.CheckIfOrganizationIsTheClusterOwner(orgId, clusterPayload.ClusterId, clusterSpec.ExternalID)
+			if ownershipCheckErr != nil {
+				serviceErr, ok := ownershipCheckErr.(*errors.ServiceError)
```
pb82, line 80: `CheckIfOrganizationIsTheClusterOwner` only seems to return `ServiceError`, is this cast needed?

machi1990: `CheckIfOrganizationIsTheClusterOwner` returns the `error` interface. However, the ocm_provider implementation returns the `ServiceError` concrete implementation of that interface. Another implementation of the `CheckIfOrganizationIsTheClusterOwner` could return a different error implementation. 
 
So yes, the cast is needed since here we do not know about the `ServiceError` but the only `error` interface.

test/mocks/api_server.go
machi1990: I noticed that there are other methods that do not adhere to this as well. 
I change to only those that I can - i.e adjacent methods while leaving the rest for some other time.

internal/kafka/internal/clusters/provider.go
```diff
@@ -46,6 +46,8 @@ type Provider interface {
 	// GetClusterResourceQuotaCosts returns a list of quota cost information related to resources used for the provisioning and
 	// terraforming of data plane clusters for the authenticated user.
 	GetClusterResourceQuotaCosts() ([]types.QuotaCost, error)
+	// CheckIfOrganizationIsTheClusterOwner perform the check to verify if the organization (given by the exteranl organization id) owns the cluster with the given cluster id and cluster external id
```
pawelpaszki, line 49:
```suggestion
	// CheckIfOrganizationIsTheClusterOwner perform the check to verify if the organization (given by the external organization id) owns the cluster with the given cluster id and cluster external id
```

pawelpaszki (approved):
looks good to me. feel free to merge after CI passes

PR #1619, opened 2023-03-07
MGDCTRS-1085: PATCH returns wrong HTTP response if payload is incorrect

manstis:
## Description
See https://issues.redhat.com/browse/MGDCTRS-1085

This PR reports HTTP409 (conflict) should the User try to patch an immutable property.

## Verification Steps
1. Create a Connector.
2. Try to patch an immutable property.
3. `HTTP409` / `CONNECTOR-MGMT-21: An attempt was made to modify an immutable field. Immutable fields are: Name, Kafka and SchemaRegistry.` error should be received (it used to be `HTTP200`). 

- [x] Code Review completed
- [x] ~Required metrics/dashboards/alerts have been added (or PR created).~
- [x] ~Required Standard Operating Procedure (SOP) is added.~
- [x] ~JIRA has been created for changes required on the client side~

@kelvah @dhirajsb Do you know if this will require a UI change?

### DIFF
--- a/internal/connector/internal/handlers/connectors.go
+++ b/internal/connector/internal/handlers/connectors.go
@@ -12,22 +12,21 @@ import (
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/services/authz"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/services/phase"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/services/vault"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/handlers"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
+	coreServices "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/secrets"
+	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/golang/glog"
+	"github.com/gorilla/mux"
 	"github.com/spyzhov/ajson"
 	"io"
 	"net/http"
 	"reflect"
 	"strings"
-
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
-	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
-	coreServices "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/services"
-	jsonpatch "github.com/evanphx/json-patch"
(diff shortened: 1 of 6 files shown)


manstis:
`Lint & Test` failed due to:
```
[2023-03-07T15:58:43.962Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
    at main (/snapshot/repo/dist/src/index.js)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```
Seems unrelated to this PR. Any ideas anyone?

internal/connector/internal/handlers/connectors.go
```diff
@@ -389,6 +390,27 @@ func validateConnectorPatch(bytes []byte, ct *dbapi.ConnectorType) *errors.Servi
 	return nil
 }
 
+func validateConnectorImmutableProperties(resource public.Connector, originalResource public.Connector) handlers.Validate {
```
valdar, line 393: have you checked that these are the only immutable properties? I seems to remember there was some logic in the patch method implementation to allow only some specific parts to be patchable like https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/connector/internal/handlers/connectors.go#L228

valdar (changes requested):

internal/connector/internal/handlers/connectors.go
manstis: Thanks @valdar for making me double check.

I effectively had the validation inverted. It should be better now.

internal/connector/internal/handlers/connectors.go
```diff
@@ -272,6 +267,11 @@ func (h ConnectorsHandler) Patch(w http.ResponseWriter, r *http.Request) {
 				}
 			}
 
+			// If we didn't change anything, then just skip the update...
+			if reflect.DeepEqual(originalResource, resource) {
```
manstis, line 271: I moved the `DeepEqual` to after the validations had run so that I could validate changes to immutable fields with the same approach as the other validations; otherwise I had to treat it slightly different (i.e. either not as a `handlers.Validate` function or at least perform the new check before the `DeepEqual` comparison was made).

kelvah:
> @kelvah @dhirajsb Do you know if this will require a UI change?

@manstis IIUC no, it doesn't. 

Users are only able to change name, service account and connector (specific configuration) when editing connectors using the console.

cc @gashcrumb, please correct me if I'm wrong.

manstis:
@valdar Is this OK now?

internal/connector/internal/handlers/connectors.go
```diff
@@ -389,6 +389,29 @@ func validateConnectorPatch(bytes []byte, ct *dbapi.ConnectorType) *errors.Servi
...
+		}
+		if patch.NamespaceId != originalResource.NamespaceId {
+			valid = false
+		}
+		if patch.Channel != originalResource.Channel {
+			valid = false
+		}
+
+		if !valid {
+			err := errors.BadRequest("An attempt was made to modify an immutable field.")
```
lgarciaaco, line 407: Can you include a hint about the offending field? That way the user recognize the source of the conflict and can fix the problem.

lgarciaaco (approved):
In think this is solid with a minor comment. Maybe you want to wait for @dhirajsb input before merging.

Thanks!!

internal/connector/internal/handlers/connectors.go
manstis: Done.

lgarciaaco (approved):

valdar (approved):

PR #1620, opened 2023-03-07
fix(certs): wait for dns propagation when the DNS resolver challenge is being performed

machi1990:
## Description

Wait for dns propagation when the DNS resolver challenge is being performed to avoid errors like:
```
authorization failed: HTTP 400 urn:ietf:params:acme:error:dns - DNS problem: NXDOMAIN looking up TXT for _acme-challenge.cg3nrilq873kpludqfug.kafka.bf2.dev - check that a DNS record exists for this domain (ca=https://acme-v02.api.letsencrypt.org/directory)
```

which might occur if the dns are not propagated.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
+++ b/internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
@@ -197,8 +197,9 @@ func createCertMagicClient(awsConfig *config.AWSConfig,
 	kafkaTLSCertificateManagementConfig *config.KafkaTLSCertificateManagementConfig,
 	storage certmagic.Storage) *certmagic.Config {
 	provider := &route53.Provider{
-		AccessKeyId:     awsConfig.Route53.AccessKey,
-		SecretAccessKey: awsConfig.Route53.SecretAccessKey,
+		WaitForPropagation: true,
+		AccessKeyId:        awsConfig.Route53.AccessKey,
+		SecretAccessKey:    awsConfig.Route53.SecretAccessKey,
 	}
 
 	certmagic.Default.RenewalWindowRatio = kafkaTLSCertificateManagementConfig.AutomaticCertificateManagementConfig.RenewalWindowRatio


pb82 (approved):

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
```diff
@@ -197,8 +197,9 @@ func createCertMagicClient(awsConfig *config.AWSConfig,
 	kafkaTLSCertificateManagementConfig *config.KafkaTLSCertificateManagementConfig,
 	storage certmagic.Storage) *certmagic.Config {
 	provider := &route53.Provider{
-		AccessKeyId:     awsConfig.Route53.AccessKey,
-		SecretAccessKey: awsConfig.Route53.SecretAccessKey,
+		WaitForPropagation: true,
```
miguelsorianod, line 200: This will block the reconciler up to 60 seconds usually (https://docs.aws.amazon.com/Route53/latest/APIReference/API_ChangeResourceRecordSets.html).

Have we explored the alternative to just reconcile again until the desired state is reached?

machi1990: Yes. That's known and acceptable.

> Have we explored the alternative to just reconcile again until the desired state is reached?

This was what was there before this PR but it ended up being causing tons of error

miguelsorianod: > Yes. That's known and acceptable.

I have some concerns regarding that. Blocking 60 seconds a reconciler seems quite a high value to me.

> This was what was there before this PR but it ended up being causing tons of error

Can you expand this further? If there is a step that checks that state is insync that should cover it.

machi1990: > Can you expand this further? If there is a step that checks that state is insync that should cover it.

If you do not wait for dns propagation, there is no step in the resolver that waits for the insync. It'll just create the record and let the CA continue with the rest of the validation check. At this stage for the challenge to be successfully resolved, it depends on when the dns record becomes published. Adding it the wait for propagation ensures that the resolver waits for the state to be insync. 

This is the lib that's in use https://github.com/libdns/route53

miguelsorianod: > there is no step in the resolver that waits for the insync

Then it should wait for but asynchronously by repeating the reconciling. I don't have details of the library but it should at least be possible to check for the changeset record state

machi1990: > > Yes. That's known and acceptable.
> 
> I have some concerns regarding that. Blocking 60 seconds a reconciler seems quite a high value to me.

Thanks, I was wrong onto saying that this would have been acceptable because my initial thinking was that we were offloading reconciling of an individual Kafka onto a separate goroutine: that's not true and we process them in serial. 
This means that if we `N` Kafkas on the queue of preparing Kafka, the last Kafka will be processed in `N-1` minutes. That's a lot of time. 

I've opened a JIRA to manage certificate in a non blocking manner while ensuring that this doesn't affect `ManagedKafkaCR` reconciliation with the data plane.  
Additionally, we'll continue to think about how we can properly handle the errors that this PR aimed to fix. For now, let's revert this PR: I've opened https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1621 for this.

miguelsorianod: Thanks for it :+1: .

PR #1621, opened 2023-03-08
Revert "fix(certs): wait for dns propagation when the DNS resolver challenge is being performed"

machi1990:
Reverts bf2fc6cc711aee1a0c2a/kas-fleet-manager#1620

This after discussion on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1620#discussion_r1129183561 we would like to avoid potential blocking & long running operations. I prefer to revert this for now to avoid long reconciliation times especially if the queue of Kafkas is big.

The errors that were meant to be fixed by bf2fc6cc711aee1a0c2a/kas-fleet-manager#1620 didn't endup blocking the reconciliation of the Kafka as it was retried and succeeded on second try. 

Let's revert the PR bf2fc6cc711aee1a0c2a/kas-fleet-manager#1620 for now while we continue to think about better ways of handling the errors.


miguelsorianod (approved):
Can be merged once the tests pass

PR #1625, opened 2023-03-10
Ensure deferred functions are executed before Go runtime exits

manstis:
## Description
Fixes https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1314#issuecomment-1322325692

## Verification Steps
1. Run integration tests twice without tearing down the database.

- [ ] ~All acceptance criteria specified in JIRA have been completed~
- [ ] ~Documentation added for the feature~
- [x] Code Review completed
- [ ] ~Required metrics/dashboards/alerts have been added (or PR created).~
- [ ] ~Required Standard Operating Procedure (SOP) is added.~
- [ ] ~JIRA has been created for changes required on the client side~

### DIFF
--- a/internal/connector/test/integration/feature_test.go
+++ b/internal/connector/test/integration/feature_test.go
@@ -31,6 +31,7 @@ func init() {
 }
 
 func TestMain(m *testing.M) {
+	exitCode := 0
 	for _, arg := range os.Args[1:] {
 		if arg == "-test.v=true" || arg == "-test.v" || arg == "-v" { // go test transforms -v option
 			opts.Format = "pretty"
@@ -49,8 +50,6 @@ func TestMain(m *testing.M) {
 	t := &testing.T{}
 
 	ocmServer := mocks.NewMockConfigurableServerBuilder().Build()
-	defer ocmServer.Close()
-
 	helper, teardown = test.NewHelperWithHooksAndDBsetup(t, ocmServer,
 		[]string{"INSERT INTO connector_types (id, name, checksum) VALUES ('OldConnectorTypeId', 'Old Connector Type', 'fakeChecksum1')",
 			"INSERT INTO connector_type_labels (connector_type_id, label) VALUES ('OldConnectorTypeId', 'old_connector_type_label')",
@@ -82,9 +81,12 @@ func TestMain(m *testing.M) {
 		},
 		connector.ConfigProviders(false),
 	)
-	defer teardown()
 
-	os.Exit(m.Run())
+	exitCode = m.Run()
+
+	defer os.Exit(exitCode)
+	defer ocmServer.Close()
+	defer teardown()
 }
 
 func TestFeatures(t *testing.T) {
--- a/internal/kafka/test/integration/kafka/kafka_test.go
+++ b/internal/kafka/test/integration/kafka/kafka_test.go
@@ -22,6 +22,7 @@ func init() {
 }
 
 func TestMain(m *testing.M) {
+	exitCode := 0
 	for _, arg := range os.Args[1:] {
 		if arg == "-test.v=true" || arg == "-test.v" || arg == "-v" { // go test transforms -v option
 			opts.Format = "pretty"
@@ -38,12 +39,13 @@ func TestMain(m *testing.M) {
 	t := &testing.T{}
 
 	ocmServer := mocks.NewMockConfigurableServerBuilder().Build()
-	defer ocmServer.Close()
-
 	helper, _, teardown = ktest.NewKafkaHelper(t, ocmServer)
-	defer teardown()
 
-	os.Exit(m.Run())
+	exitCode = m.Run()
+
+	defer os.Exit(exitCode)
+	defer ocmServer.Close()
+	defer teardown()
 }
 
 func TestFeatures(t *testing.T) {
--- a/test/helper.go
+++ b/test/helper.go
@@ -141,7 +141,6 @@ func NewHelperWithHooksAndDBsetup(t *testing.T, httpServer *httptest.Server, set
 		glog.Fatalf("Unable to initialize testing environment: %s", err.Error())
 	}
 
-	h.CleanDB()
 	h.ResetDB()
 	h.SetupDB(setupDBsqlStatements)
 	env.Start()
--- a/internal/connector/internal/handlers/connector_agent.go
+++ b/internal/connector/internal/handlers/connector_agent.go
@@ -3,7 +3,6 @@ package handlers
 import (
 	"context"
 	"fmt"
-	"io"
 	"net/http"
 	"strconv"
 	"strings"
@@ -202,7 +201,7 @@ func (h *ConnectorClusterHandler) ListDeployments(w http.ResponseWriter, r *http
 									if waitForCancelOrTimeoutOrNotification(ctx, 30*time.Second, sub) {
 										// ctx was canceled... likely due to the http connection being closed by
 										// the client.  Signal the event stream is done.
-										return io.EOF, nil
+										return nil, nil
 									}
 
 									// get a new DB connection...


machi1990:
@miguelsorianod can you've a look at this? Seems related to #1314

internal/connector/test/integration/feature_test.go
```diff
@@ -82,9 +83,10 @@ func TestMain(m *testing.M) {
 		},
 		connector.ConfigProviders(false),
 	)
+	defer func() { os.Exit(exitCode) }()
 	defer teardown()
```
manstis, line 87: This deferred method was not being called following `os.Exit(..)`.

If you run `make db/login` after running the integration tests you can see the database tables and content are still there... they should have been removed by this `teardown()` method invocation.

Furthermore, you can also verify this by placing the following in `internal/connector/internal/migrations/migrations.go`
```
...
addOrgIDAnnotations("202212050000"),
	&gormigrate.Migration{
		ID: "test",
		Migrate: func(g *gorm.DB) error {
			fmt.Println("============== RUNNING MIGRATION")
			return nil
		},
		Rollback: func(g *gorm.DB) error {
			fmt.Println("============== RUNNING ROLLBACK")
			return nil
		},
...
```
You can observe that the rollback method is not called.

internal/kafka/test/integration/kafka/kafka_test.go
```diff
@@ -22,6 +22,7 @@ func init() {
 }
 
 func TestMain(m *testing.M) {
+	exitCode := 0
```
manstis, line 25: I made a similar change for `kas` as it has the same problem (but it does not manifest itself).

internal/connector/test/integration/feature_test.go
manstis: I followed the approach suggested [here](https://www.golinuxcloud.com/golang-handle-defer-calls-os-exit/) ("Method-2")

test/helper.go
```diff
@@ -141,16 +141,16 @@ func NewHelperWithHooksAndDBsetup(t *testing.T, httpServer *httptest.Server, set
 		glog.Fatalf("Unable to initialize testing environment: %s", err.Error())
 	}
 
-	h.CleanDB()
```
manstis, line 144: `CleanDB()` is [called by](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/test/helper.go#L334) `ResetDB()` (on L#145). No need to call it twice.

test/helper.go
```diff
@@ -141,16 +141,16 @@ func NewHelperWithHooksAndDBsetup(t *testing.T, httpServer *httptest.Server, set
...
 	h.SetupDB(setupDBsqlStatements)
 	env.Start()
 	return h, buildTeardownHelperFn(
-		env.Stop,
 		h.CleanDB,
 		metrics.Reset,
 		stopJWKMockServer,
-		env.Cleanup)
+		env.Cleanup,
+		env.Stop,
```
manstis, line 152: It made more sense to me to "stop" the environment (which asks each `BootService` to _stop_) last.

The code also asks each `BootService` to "clean up" which, IMO, should happen before they stop).

Happy to revert the order of these calls; if others disagree.

test/helper.go
miguelsorianod: :+1:

internal/kafka/test/integration/kafka/kafka_test.go
```diff
@@ -41,9 +42,10 @@ func TestMain(m *testing.M) {
 	defer ocmServer.Close()
 
 	helper, _, teardown = ktest.NewKafkaHelper(t, ocmServer)
+	defer func() { os.Exit(exitCode) }()
```
miguelsorianod, line 45: Shouldn't this be moved above the defer call to `ocmServer.Close()` ?

The order on which deferred calls are run is LIFO and I understand the Exit method call should be performed at the end.

internal/connector/test/integration/feature_test.go
miguelsorianod: Good catch :+1: .

The reason about why deferred was not being called is documented in the `os.Exit` function documentation available in https://pkg.go.dev/os#Exit:

Exit causes the current program to exit with the given status code. Conventionally, code zero indicates success, non-zero an error. The program terminates immediately; **deferred functions are not run**. 

The approach used looks good to me :+1:

test/helper.go
manstis: I've restored the clean-up order.. 

Thinking about it over the weekend `env.Cleanup` should probably be last; after everything has stopped.

internal/kafka/test/integration/kafka/kafka_test.go
manstis: Thanks. Good spot. I didn't see the other deferral. Fixed.

miguelsorianod (changes requested):
The changes look good to me :+1: 

Could you squash redundant commits to end up with a clean history?

After that and the tests pass it will be good to merge :+1:

manstis:
@miguelsorianod 

> Could you squash redundant commits to end up with a clean history?

Squashed.

miguelsorianod:
It seems the action times out @manstis . I see it is currently set to `14` minutes.
It is interesting that this is happening in this PR. The changes here should not have substantially increased the execution time. Maybe it was already at the limit and successfully running the teardown function ends up going over the thresold.

We can increase the timeout of the action.

manstis:
@miguelsorianod 

> We can increase the timeout of the action and see what happens.

Quite probably. Do you know where/who/how it can be done? I'm not an admin on this GitHub repo.

manstis:
@miguelsorianod Please see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1632

That will need merging before it affects this PR.. 

There are some "tricks" that can be used; but since it's only a "timeout" property merging it IMO should is OK.

manstis:
@miguelsorianod As @machi1990 pointed out there are 2 timeouts defined:
1. GitAction Workflow configuration
2. `gotestsum` configuration

Can we be certain which has caused the timeout?

miguelsorianod:
> @miguelsorianod As @machi1990 pointed out there are 2 timeouts defined:
> 
>     1. GitAction Workflow configuration
> 
>     2. `gotestsum` configuration
> 
> 
> Can we be certain which has caused the timeout?
> 
> I think we should align both at 20 minutes and, if it still fails, investigate again.

I understand the action itself caused the timeout, as shown in the logs:
```
Error: The action has timed out.
```

manstis:
If the _action_ was the integration test it'd have been timed out by `gotestsum` at 14 minutes..

So again, I propose the PR aligning `gotestsum` timeout to 20 minutes is merged and then we re-assess.

miguelsorianod:
> If the _action_ was the integration test it'd have been timed out by `gotestsum` at 14 minutes..
> 
> So again, I propose the PR aligning `gotestsum` timeout to 20 minutes is merged and then we re-assess.

That's not correct. The `gotestsum` timeout has a timeout, but `gotestsum` is run several times. One time for kafka tests and one time for connectors tests. So it can go over the action timeout.

In any case, if you really want to match the timeouts you can add a temporary commit in this PR.

manstis:
Hi @machi1990 @miguelsorianod I found an issue in COS where a web server connection was not being closed.

This lead to a failure to stop the API Server in integration tests and the timeout (now we try to _teadown_).

It looks like a [change](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/commit/2db1ce1bd2ecc9fb7bb558eb372474358917e94d) in KAS lead to the undesirable scenario.

All good now though.

Shall I rollback the CI timeout change?

manstis:
@lburgazzoli @valdar @dhirajsb [FYI](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1625#issuecomment-1467998563).

I was told support for "watch" was to be removed for Connectors (and hence not to be supported for Processors).

Was this based on there being issues with its operation (now fixed) or otherwise?

I'm just wondering whether support now needs to be removed.

miguelsorianod:
> Shall I rollback the CI timeout change?

If the increase of the observed running times was due to the described connectors issue then yes the timeout change should be rolled back

manstis:
> ...  the timeout change should be rolled back

See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1636

miguelsorianod:
> @lburgazzoli @valdar @dhirajsb [FYI](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1625#issuecomment-1467998563).
> 
> I was told support for "watch" was to be removed for Connectors (and hence not to be supported for Processors).
> 
> Was this based on there being issues with its operation (now fixed) or otherwise?
> 
> I'm just wondering whether support now needs to be removed.

@manstis do you want to wait to get an answer to this before proceeding with the PR merge or should we go ahead?

miguelsorianod (approved):

PR #1637, opened 2023-03-14
[WIP] update gorm and related libraries

miguelsorianod:
## Description
Continuation of https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1314. Full context there.

- [ ] Code Review completed


machi1990:
Can't wait for this update to land! Thanks @miguelsorianod @manstis for making this possible

miguelsorianod:
Some tests are still failing. Both kafka and connector ones.
I have been working on reviewing the kafka related ones. After investigating the issues I found that there are a couple of issues: The way the `Scan` method is implemented for `api.JSON` with some edge cases and the `api.MarshalJSON` with some edge cases too.
I got some local changes that made the tests pass but at the same time I found out there are newer gorm and db drivers versions. After locally updating them I found out that that makes the tests pass so it seems to be a bug introduced in the current versions of the libraries in the PR.
I would have updated and pushed the changes but when updating gorm and related libraries the way they are defined point to an arbitrary commit in gorm instead of a tag. Because of this I am going to wait until they update the dependencies to point to a git tag, to provide us with higher stability guarantees.

I observed the connector tests are still failing but let's wait until we update the gorm libraries to see if they still fail.

go.mod
```diff
@@ -55,8 +54,8 @@ require (
 	gopkg.in/resty.v1 v1.12.0
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	gorm.io/driver/postgres v1.0.8
-	gorm.io/gorm v1.21.7
+	gorm.io/driver/postgres v1.4.6
```
machi1990, line 57: I saw there is a new release: https://github.com/go-gorm/postgres/releases/tag/v1.5.0

go.mod
```diff
@@ -55,8 +54,8 @@ require (
...
+	gorm.io/gorm v1.24.3
```
machi1990, line 58: I saw there is a new release: https://github.com/go-gorm/gorm/releases/tag/v1.24.6

go.mod
miguelsorianod: Yes, but unfortunately it points to a non-tagged commit of gorm as mentioned in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1637#issuecomment-1471500564, which is the reason I am holding this for now

go.mod
miguelsorianod: See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1637#discussion_r1154569175

go.mod
machi1990: Ah right, thanks for the update. I didn't go as far as checking if that's changed with the new release. Thanks for checking.

internal/kafka/internal/services/clusters_test.go
```diff
@@ -136,7 +136,7 @@ func Test_Cluster_Create(t *testing.T) {
 				cluster: mocks.BuildCluster(nil),
 			},
 			setupFn: func() {
-				mocket.Catcher.Reset().NewMock().WithQuery("INSERT").WithExecException()
+				mocket.Catcher.Reset().NewMock().WithQuery("INSERT").WithQueryException()
```
ziccardi, line 139: Why this change?

ziccardi: > Because with recent gorm changes the Save method performs INSERTs returning data. The same happens for the Create method

Does this mean that ‘WithExcecException’ no longer catches INSERT? It may be necessary to carefully review each test that utilizes `mocket`.

miguelsorianod: The ones you see that have been changed in the PR are ones that were found because the tests started to fail with the old value

ziccardi: Since the INSERT is no longer caught by the WithExecException, a test that was supposed to fail in case of an INSERT may now pass. 

This could result in the test missing a future change that introduces a bug that triggers the INSERT.

ziccardi: Yep. Got that. We need to carefully review the tests to understand when it is the case and fix it. 
Even if the test is not failing, it doesn't mean it is working as intended, so just running them won't be enough.

miguelsorianod: Hi @ziccardi it seems with the new gorm version recently released (1.25.0) they reverted that change and the `Save` method is no longer returning errors again.
I will update the PR again to not include that change and make the tests pass.

I think we can close the opened jira issue then.

miguelsorianod:
I updated the gorm and gorm postgres dependencies to the versions show below:
```
  gorm.io/driver/postgres v1.5.0
  gorm.io/gorm v1.25.0
```

miguelsorianod:
After performing the upgrades mentioned in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1637#issuecomment-1503067641 the unit tests pass, the kafka integration tests also pass but the connectors tests fail

internal/kafka/internal/services/clusters_test.go
miguelsorianod: updated

miguelsorianod:
> After performing the upgrades mentioned in [#1637 (comment)](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1637#issuecomment-1503067641) the unit tests pass, the kafka integration tests also pass but the connectors tests fail.
> 
> It seems all connectors tests are failing:
> 
> ```
> 2023-04-11T10:08:23.3263810Z DONE 33 tests, 33 failures in 46.126s
> ```

cc @manstis could you take a look? I suspect it might have to do with the migrations again as all the tests are failing with "unknown" state.

manstis:
@miguelsorianod I suspect it is caused by https://github.com/go-gorm/playground/pull/571

Newer (latest) versions of `gorm`'s postgres driver fail to migrate `XXXserial` columns.

manstis:
@miguelsorianod I am doing some more investigations locally and will report back.

miguelsorianod:
> @miguelsorianod I am doing some more investigations locally and will report back.

Thanks :+1:

miguelsorianod:
> Hi @miguelsorianod
> 
> I've made the applicable changes to my fork: [manstis@57abdc7](https://github.com/manstis/kas-fleet-manager/commit/57abdc78eaed75396329545f14455a2b2e171682)
> 
> You could include them in this PR and it should be OK.

Can your changes be merged to main before this PR? or they depend on the changes of this PR being available? If not I will include the change as you said

manstis:
@miguelsorianod Since the changes are only required due to the `gorm` upgrade I think they should be part of this PR (for the same reason as to why there are `kafka` changes in this PR already too and they have not first been applied to `main`).

miguelsorianod:
> @miguelsorianod Since the changes are only required due to the `gorm` upgrade I think they should be part of this PR (for the same reason as to why there are `kafka` changes in this PR already too and they have not first been applied to `main`).

It seems reasonable. I will add it into the PR. The authorship will be lost though.

miguelsorianod:
> > @miguelsorianod Since the changes are only required due to the `gorm` upgrade I think they should be part of this PR (for the same reason as to why there are `kafka` changes in this PR already too and they have not first been applied to `main`).
> 
> It seems reasonable. I will add it into the PR. The authorship will be lost though.

I have been able to keep the authorship by getting your fork's branch and cherry-picking the commit

miguelsorianod:
All the checks are passing now. I will do a re-review and start planning the rollout of this changes for KAS Fleet Manager, as merning it will rollout those changes to KAS Fleet Manager Stage and Production.

machi1990 (approved):

manstis (approved):
LGTM :+1:

miguelsorianod:
I've made the following manual tests in a development environment:

* Performing Kafkas DB migration twice works correctly when using the new gorm version
* Performing Connectors DB migration twice works correctly when using the gorm version
* Performing Kafka creation before gorm update, then migrating to new gorm version including reapplying db migration works correctly. The kafka intsances remain ready and no errors are observed
* Creating Kafka instances after gorm update works correctly
* Kafka instances can be listed after gorm updates for both newly created instances and also for previously created instances
* Kafka instances can be deleted after form updates for both newly created instances and also for previously created instances
* Instances created before the gorm update and instances created after the gorm update look the same. This includes empty and null values
* OCP cluster entry pre and post migration looks the same. This includes empty and null values

I am currently waiting for some midstream pipelines issues to be fixed. Once that is done I will proceed to merge this and release it to stage

PR #1639, opened 2023-03-15
feat(certs): perform certificate management in async mannner

machi1990:
## Description

The sync management of certificate can take 20s and up to a minute for each domain. If we've a large que of Kafkas this might problematic as the last Kafkas will see its certificate being processed after `(20-60)s * numberOfKafkas`. Since the processing of each domain is independent, we cna make the generation of a certificate async.

- [ ] Code Review completed


internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
```diff
@@ -186,41 +200,51 @@ func NewKafkaTLSCertificateManagementService(
...
 	kafkaTLSCertificateManagementConfig *config.KafkaTLSCertificateManagementConfig,
 	storage certmagic.Storage) *certmagic.Config {
 	provider := &route53.Provider{
-		WaitForPropagation: false,
-		AccessKeyId:        awsConfig.Route53.AccessKey,
-		SecretAccessKey:    awsConfig.Route53.SecretAccessKey,
+		WaitForPropagation: true,
+		// wait up to 150 seconds for the temporary txt record to be propagated.
+		// this is a blocking operation but it is acceptable since the management of certificate for each domain is done async
+		MaxWaitDur:      150 * time.Second,
```
machi1990, line 215: We wait for the txt domain to be INSYNC for up to 150 seconds

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
```diff
@@ -186,41 +200,51 @@ func NewKafkaTLSCertificateManagementService(
...
+	magic.Logger = zap.NewNop()
+	magic.OnEvent = func(ctx context.Context, event string, data map[string]any) error {
+		if event == certFailedEvent {
+			logger.NewUHCLogger(ctx).Errorf("certificate management failed with the following event details: %v", data)
+			return nil
+		}
+
+		logger.NewUHCLogger(ctx).V(10).Infof("event %q received with data %v", event, data)
+		return nil
+	}
```
machi1990, line 247: We log to error to our monitoring system if there is cert management error

internal/kafka/internal/services/kafka.go
```diff
@@ -502,6 +505,22 @@ func (k *kafkaService) PrepareKafkaRequest(kafkaRequest *dbapi.KafkaRequest) *er
...
+
+		if err != nil {
+			if !coreErrors.Is(err, fs.ErrNotExist) {
+				return errors.NewWithCause(errors.ErrorGeneral, err, "error checking if certificate for kafka %q has been created", kafkaRequest.ID)
+			}
+
+			// do not continue the reconciliation of a preparing Kafka if the certificate has not been generated yet.
+			return nil
+		}
+	}
```
machi1990, line 522: If we are managing cert automatically, we wait for the cert to be generated before we can move the state of the Kafka from preparing to provisioning

machi1990:
@miguelsorianod this is based on our latest investigation. Can you've a look? Thanks

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
```diff
@@ -147,9 +152,18 @@ func (certManagementService *kafkaTLSCertificateManagementService) RevokeCertifi
 		return nil // the certificate is revoked manually in manual mode
 	}
 
+	wildcardDomain := fmt.Sprintf("*.%s", domain)
+	refs := certManagementService.certManagementClient.GetCerticateRefs(wildcardDomain)
+
+	// No need to revoke the certificate if it does not exists
+	if !certManagementService.storage.Exists(ctx, refs.TLSCertRef) {
+		logger.NewUHCLogger(ctx).V(10).Infof("certificate for domain %q does not exist in Storage. It is not going to be revoked", domain)
```
JameelB, line 160: do we need to create a new logger each time?

machi1990: Yes, as the `ctx Context` object is a different. 
That being said, we could use the singleton logger that we've.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -48,18 +51,37 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
...
+		secretCache:      secretCache,
+		storageLock:      sync.Mutex{},
+		storageItemLocks: map[string]*sync.Mutex{},
+		secretPrefix:     config.SecretManager.SecretPrefix,
 	}, nil
 }
 
 func (storage *secureStorage) Lock(ctx context.Context, key string) error {
-	return nil // NOOP as AWS Secret manager uses versioning mechanism to achieve optimistic locking
+	mu, ok := storage.storageItemLocks[key]
```
ziccardi, line 63: Now that we manage the certs asynchronously, can happen that 2 threads call this method at the same time? If yes, here we could have a race condition.

machi1990: This will always be called in serial.

ziccardi: Then what is the use of the lock?

machi1990: For a given `key`, the lock here helps guarding against processing of certificate management if the previous one for the given key hasn't finished. But for each given `key` there will be at least a `30s` interval between each cert management request.

ziccardi: If It is not a thread synchronisation issue, it could be done with a simple check instead of a lock.

I would add a comment to the `storage` class describing that the class is not thread-safe (seeing the lock mechanisms could be misleading )

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -48,18 +51,39 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
...
+	if !ok {
+		// initialize the storage item lock the first time if it does not exist
+		storage.storageLock.Lock()
+		defer storage.storageLock.Unlock()
+		mu = &sync.Mutex{}
+		storage.storageItemLocks[key] = mu
+		mu.Lock()
```
ziccardi, line 70: This piece of code confuses me a bit: in a previous comment, we said that this method is  guaranteed to be called synchronously, here however, we are calling 
```go
		storage.storageLock.Lock()
		defer storage.storageLock.Unlock()
```
Unless I miss something that is used to protect against race conditions.
These locks are useless if this is guaranteed to be called synchronously.
If this is not guaranteed to be called synchronously, we can incur the race issue discussed in the previous comment.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: Sorry, I misunderstood you comment about the `What is the use of the lock, then?`.
My answer on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1639#discussion_r1137193186 was applicable for a key specific lock - which was highlighted by the first comment.
The other global lock `storage.storageLock.Lock()` (which I should probably rename) is prevent simultaneous access to the key specific locks. Re-reading everything again, you are right (and thank you for this) that there could be a race condition which might not have been covered by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1639#discussion_r1137193186. 
This could typically happen for developer Kafkas that share the same certificate: this means that the key could be the same as well. Only standard kafkas have their separate certificate.
Now if we go to the developer Kafka case, it could happen that one developer kafka in `preparing` state and another in `ready` state have the certificate being reconciled at the same time: If this happens then we'll have a race condition.
The options: 
- Find unique certs and reconcile them in all our reconcilers
- Or make this thread safe

I'll go with the later option since we are at it and hopeful it should be quick (famous last word... :-)). The first option could be sort of covered by the second. And once we've the second option in, the storage could be used anywhere in a thread safe manner as long as `Lock` and `Unlock` are called.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: Thanks @ziccardi I've added a comment https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1639#discussion_r1137851529. 
I'll make sure that this is thread-safe.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: > The options:

> Find unique certs and reconcile them in all our reconcilers
> Or make this thread safe
> I'll go with the later option since we are at it and hopeful it should be quick (famous last word... :-)). The first option could be sort of covered by the second. And once we've the second option in, the storage could be used anywhere in a thread safe manner as long as Lock and Unlock are called.

Thinking about this, we might need the two options to be future proof. 
i.e we still need the storage to be thread safe.
and we'll need option one - which I'll refine as

machi1990: JIRA created https://issues.redhat.com/browse/MGDSTRM-10921

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
miguelsorianod: Do we have some insight or documented information about the approximate time range it takes to propagate?

miguelsorianod: What happens if that doesn't happen in that specified time? is an error returned?

machi1990: > Do we have some insight or documented information about the approximate time range it takes to propagate usually?

From https://aws.amazon.com/route53/faqs/ they say 60s under normal condition.

```
Q. How quickly will changes I make to my DNS settings on Amazon Route 53 propagate globally?

Amazon Route 53 is designed to propagate updates you make to your DNS records to its world-wide network of authoritative DNS servers within 60 seconds under normal conditions. A change is successfully propagated world-wide when the API call returns an INSYNC status listing.

Note that caching DNS resolvers are outside the control of the Amazon Route 53 service and will cache your resource record sets according to their time to live (TTL). The INSYNC or PENDING status of a change refers only to the state of Route 53’s authoritative DNS servers.
```

> What happens if that doesn't happen in that specified time? is an error returned?

Yes, an error will happen.

machi1990: > Yes, an error will happen.

Just to extend on what I said here - an error will happen but this won't mean a Kafka will fail to be reconciled. The reconciler will will output that error message and on the next reconciler run, we'll attempt to reconcile the cert again up until it succeeds.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -48,18 +51,36 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
...
-		secretPrefix: config.SecretManager.SecretPrefix,
+		secretClient:     secretClient,
+		secretCache:      secretCache,
+		lock:             sync.Mutex{},
+		storageItemLocks: map[string]*sync.Mutex{},
+		secretPrefix:     config.SecretManager.SecretPrefix,
 	}, nil
 }
 
 func (storage *secureStorage) Lock(ctx context.Context, key string) error {
```
miguelsorianod, line 62: From which thread(s) are the Lock and Unlock methods called? is it from the thread that ManageAsync triggers? is it from the reconciler thread or from both?

machi1990: These are called from the ManageAsync which in turn is called from our different reconciler for each Kafka in a given state (`preparing`, `ready` or `provisioning`).

miguelsorianod, line 62: We can document that Lock blocks until it can acquire the name-specific lock, and that the process of obtaining a name-specific lock is atomic, as required per the certmagic.Locker interface documentation

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -115,7 +136,10 @@ func (storage *secureStorage) Delete(ctx context.Context, key string) error {
 }
 
 func (storage *secureStorage) Exists(ctx context.Context, key string) bool {
-	_, err := storage.Load(ctx, key)
+	name := storage.constructSecretName(key)
```
miguelsorianod, line 139: What's the motivation of the change in the Exists method?

machi1990: Good catch on this: the change is desirable but not critical for this PR i.e it could be a separate PR.

Exists should check if the key truly exists in the backend storage. With the use of `.Load` which has caching (which needs to be revisited) built on it - we were in some some cases returning `true` since the content is in the cache but the certificate isn't in the storage backend.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: @ziccardi this is done already. I'll resolve the comment. Thanks for raising this one.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: done

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: taken out of the PR

machi1990:
@ziccardi @JameelB @miguelsorianod this should be ready for another round of reviews

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -48,18 +51,41 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
...
+		secretCache:      secretCache,
+		lock:             sync.Mutex{},
+		storageItemLocks: map[string]*sync.Mutex{},
+		secretPrefix:     config.SecretManager.SecretPrefix,
 	}, nil
 }
 
+// Lock blocks until it can acquire the name-specific lock.
+// The process of obtaining a name-specific lock is atomic, as required per the certmagic.Locker
+// interface documentation: https://github.com/caddyserver/certmagic/blob/master/storage.go#L87.
```
miguelsorianod, line 64: Instead of referring to github code let's refer to the interface documentation: https://pkg.go.dev/github.com/caddyserver/certmagic#Locker

internal/kafka/internal/services/kafka.go
```diff
@@ -502,6 +505,22 @@ func (k *kafkaService) PrepareKafkaRequest(kafkaRequest *dbapi.KafkaRequest) *er
...
+			TLSCertRef: kafkaRequest.KafkasRoutesBaseDomainTLSCrtRef,
+			TLSKeyRef:  kafkaRequest.KafkasRoutesBaseDomainTLSKeyRef,
+		})
+
+		if err != nil {
+			if !coreErrors.Is(err, fs.ErrNotExist) {
+				return errors.NewWithCause(errors.ErrorGeneral, err, "error checking if certificate for kafka %q has been created", kafkaRequest.ID)
+			}
+
+			// do not continue the reconciliation of a preparing Kafka if the certificate has not been generated yet.
```
miguelsorianod, line 519: We could add debug logging level at this point. Might be a little verbose though

internal/kafka/internal/services/kafka.go
```diff
@@ -1428,15 +1447,11 @@ func (k *kafkaService) ManagedKafkasRoutesTLSCertificate(kafkaRequest *dbapi.Kaf
...
+	kafkaRequest.KafkasRoutesBaseDomainTLSCrtRef = certManagementOutput.TLSCertRef
 
-	dbConn := k.connectionFactory.New().Model(kafkaRequest)
-	if err := dbConn.Updates(values).Error; err != nil {
-		return errors.NewWithCause(errors.ErrorGeneral, err, "failed to update kafka %q", kafkaRequest.ID)
+	if svcErr := k.Update(kafkaRequest); svcErr != nil {
+		return svcErr
 	}
 
 	logger.Logger.Infof("management of tls certificate for the domain %q of kafka with id %q successfully finished", kafkaRequest.KafkasRoutesBaseDomainName, kafkaRequest.ID)
```
miguelsorianod, line 1457: I think this log is not accurate as now it is async and it's not finished by that point

internal/kafka/internal/services/kafka.go
```diff
@@ -1428,15 +1447,11 @@ func (k *kafkaService) ManagedKafkasRoutesTLSCertificate(kafkaRequest *dbapi.Kaf
...
-		"kafkas_routes_base_domain_tls_crt_ref": certManagementOutput.TLSCertRef,
-		"kafkas_routes_base_domain_tls_key_ref": certManagementOutput.TLSKeyRef,
-	}
+	kafkaRequest.KafkasRoutesBaseDomainTLSKeyRef = certManagementOutput.TLSKeyRef
+	kafkaRequest.KafkasRoutesBaseDomainTLSCrtRef = certManagementOutput.TLSCertRef
 
-	dbConn := k.connectionFactory.New().Model(kafkaRequest)
-	if err := dbConn.Updates(values).Error; err != nil {
-		return errors.NewWithCause(errors.ErrorGeneral, err, "failed to update kafka %q", kafkaRequest.ID)
+	if svcErr := k.Update(kafkaRequest); svcErr != nil {
```
miguelsorianod, line 1453: Is there a possibility that the kafkaRequest contains non-zero valued fields that are different than what's in the DB aside from the ones being set in this method?

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
```diff
@@ -3,13 +3,18 @@ package kafkatlscertmgmt
...
 
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared"
 	"github.com/caddyserver/certmagic"
 	"github.com/libdns/route53"
+	"go.uber.org/zap"
 )
 
+const certFailedEvent = "cert_failed"
```
miguelsorianod, line 16: Can we call it certmagicCertFailedEvent?

internal/kafka/internal/services/kafka.go
machi1990: good point.

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service_test.go
```diff
@@ -137,44 +137,57 @@ func Test_kafkaTLSCertificateManagementService_RevokeCertificate(t *testing.T) {
...
+
 	tests := []struct {
 		name    string
 		fields  fields
 		args    args
 		wantErr bool
 	}{
 		{
-			name: "should return an error when revoking the certificate returns an error",
+			name: "should not revoke the certificae if running in manual mode",
```
miguelsorianod, line 160: Typo certificae -> certificate

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
```diff
@@ -147,9 +152,18 @@ func (certManagementService *kafkaTLSCertificateManagementService) RevokeCertifi
 		return nil // the certificate is revoked manually in manual mode
 	}
 
+	wildcardDomain := fmt.Sprintf("*.%s", domain)
+	refs := certManagementService.certManagementClient.GetCerticateRefs(wildcardDomain)
+
+	// No need to revoke the certificate if it does not exists
```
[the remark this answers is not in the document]
machi1990: It was something that was lacking but was made evident by this change. 
It could typically happen when a preparing kafka is deleted before the certificate are there.

internal/kafka/internal/services/kafka.go
machi1990: At the moment no, but it could be possible in the future. I wanted the convinience of using the `kafkaRequest` but I guess I'll need to change it back and make it more specific as it was initially.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: thanks.

internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
machi1990: thanks, changed!

internal/kafka/internal/services/kafka.go
machi1990: changed back to what it was before

miguelsorianod: Another alternative could be creating an empty kafkarequest, copying the desired values there and then use that one to perform the update. Although I haven't tried it

machi1990: Yes. That would also work. I've changed it back so I'll keep it as it.

internal/kafka/internal/services/kafka.go
machi1990: debug log added

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -48,18 +51,41 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
...
 func (storage *secureStorage) Lock(ctx context.Context, key string) error {
-	return nil // NOOP as AWS Secret manager uses versioning mechanism to achieve optimistic locking
+	storage.lock.Lock()
+	mu, ok := storage.storageItemLocks[key]
+	if !ok {
+		// initialize the storage item lock the first time if it does not exist yet
+		mu = &sync.Mutex{}
+		storage.storageItemLocks[key] = mu
+	}
+	storage.lock.Unlock()
```
ziccardi, line 75: This can be optimised to avoid locking when not needed (double-checked locking pattern):
```suggestion
	mu, ok := storage.storageItemLocks[key]
	if !ok {
		storage.lock.Lock()
		mu, ok = storage.storageItemLocks[key]
		if !ok {
			// initialize the storage item lock the first time if it does not exist yet
			mu = &sync.Mutex{}
			storage.storageItemLocks[key] = mu
		}
		storage.lock.Unlock()
	}
```

machi1990: Thanks. Yes, I considered this initially and decided to went with the current approach in the PR because the operation that happens in between the `storage.lock.Lock()` and `storage.lock.Unlock()` is not that time consuming. 
Definitely this pattern is useful when acquiring a key item lock becomes an expensive operation e.g when we want to implement a distributed lock

miguelsorianod:
One of the flaws of this implementation is that nothing prevents two pods in the KFM deployment to request the same certificate issue.
The locking mechanism implemented prevents the same certificate to be issued twice in the context of the KFM process. However, multiple pods run in different processes and therefore that locking mechanism does not apply between them.
It is true that KFM makes use of a leader election mechanism that prevents 2 pods from running simultaneously. However, there are other scenarios where multiple pods are involved, with only 1 being leader, where a simultanous certificate issuing might happen.

Example:
Sequence of events having 2 KFM Pods A, B:
1. Pod A reconciles is leader calls manage async and locks the name
2. We perform a deploy of a new KFM version
3. Pod B becomes leader
4. Pod B reconciles and calls manage async because even though pod A locked the name, that lock was stored in local Pod A storage

The way to solve that would be having some centralized synchronization point so we have a distribution lock mechanism that the pods make use of. For example, one option would be centralizing it at database level similarly to what is done with the leader election mechanism locks. 

cc @machi1990 to try to understand better the implications, what happens if the same certificate issuing request is executed twice simultaneously? would we end up with 2 certificates? or one would end up erroring, or both?

Depending on what are the implications we can decide to go forward with the local implementation assuming this might happen in some scenarios and implement later the distributed locking mechanism.

miguelsorianod:
> > to try to understand better the implications, what happens if the same certificate issuing request is executed twice simultaneously? would we end up with 2 certificates? or one would end up erroring, or both?
> 
> Thanks, without locking in place the implication is that there will be more than one certificate request for the same wildcard domain:
> 
>     * this means a few extra calls to the backend storage e.g when checking if the certificate exists in storage if it has not been cached for example
> 
>     * possibly, if the request are two many, we'll be seeing errors from LE due to rate limits: specifically the duplicate certificate limit https://letsencrypt.org/docs/duplicate-certificate-limit
> 
> 
> Will that affect the functionality? In my opinion no. We'll only received the error alerts. Kafkas won't be affect as the certificate from whichever call that was successful and made it to the storage will be the one that's going to be used for a Kafka in question.

I am still not sure I understand the implications. If we leave aside the potential issues related to rate limiting, is it possible that we end up with 2 certificate issuing requests succeeding for the same name? or is it always guaranteed that one of the calls will fail?
Wouldn't the pod B always be trying to get the certificate even when it failed?

> 
> > Depending on what are the implications we can decide to go forward with the local implementation assuming this might happen in some scenarios and implement later the distributed locking mechanism.
> 
> Ideally, the locking mechanism should be centralized as opposed to the current mechanism. I think we should continue to observe scenario for this and improve on top of this: if we notice that it is a recurring theme, and the load on the external systems becomes high.

Depending on the implications I agree with this reasoning. Before going forward with this I would like to understand the potential implications that I am trying to understand.

machi1990:
> I am still not sure I understand the implications. If we leave aside the potential issues related to rate limiting, is it possible that we end up with 2 certificate issuing requests succeeding for the same name? or is it always guaranteed that one of the calls will fail?

The two calls will fail only if we have not reached the limits. The two calls could succeed and certificate will be updated to the storage twice i.e we won't end up with the two certificate in the storage, but one: the last successful processed request.

> Wouldn't the pod B always be trying to get the certificate even when it failed?

The pod that failed will keep on trying to manage the certificate if and only if it remains as the leader pod and it hasn't received the newly generated certificate from pod A.

machi1990:
Thank you @miguelsorianod @ziccardi for the review.
I've opened https://issues.redhat.com/browse/MGDSTRM-10938 to improve locking further.

PR #1641, opened 2023-03-16
ci(image): add PR check that ensures that image can be built

machi1990:

- [ ] Code Review completed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,7 +146,10 @@ jobs:
           flags: unittests # optional
           name: codecov-kas-fleet-manager # optional
           fail_ci_if_error: true # optional (default = false)
-          verbose: true # optional (default = false)
+          verbose: true # optional (default = false)    
+      - name: Validate that kas-fleet-manager images can be built successful.
+        if: github.event_name == 'pull_request'
+        run: make image/build
       - name: Build and publish image to quay.io
         if: github.event_name == 'push'
         env:
@@ -156,4 +159,3 @@ jobs:
         run: |
           chmod +x ./build_deploy.sh
           ./build_deploy.sh
-
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -81,3 +81,9 @@ sed -i -e 's/mas_sso/redhat_sso/g' internal/connector/internal/environments/inte
 docker login -u "${QUAY_USER}" -p "${QUAY_TOKEN}" quay.io
 docker build -t "$IMAGE_NAME" -f Dockerfile_integration_tests .
 docker run -i "$IMAGE_NAME"
+
+# ensure that the kas-fleet-manager docker image can be built when PR check runs
+if [[ -z "${OCM_ENV}" ]] || [[ "${OCM_ENV}" == "${INTEGRATION_ENV}" ]];
+then
+  make image/build
+fi


pr_check.sh
```diff
@@ -81,3 +81,9 @@ sed -i -e 's/mas_sso/redhat_sso/g' internal/connector/internal/environments/inte
 docker login -u "${QUAY_USER}" -p "${QUAY_TOKEN}" quay.io
 docker build -t "$IMAGE_NAME" -f Dockerfile_integration_tests .
 docker run -i "$IMAGE_NAME"
+
+# ensure that the kas-fleet-manager docker image can be built when PR check runs
```
miguelsorianod, line 85: Do we need this? in midstream the pipeline already tries to buy the image right?
Maybe let's keep this simple

pawelpaszki: was thinking the same. in the midstream the image is built when github secrets are provided as env vars. does it mean that we will built an image twice?

.github/workflows/ci.yaml
```diff
@@ -147,6 +147,9 @@ jobs:
           name: codecov-kas-fleet-manager # optional
           fail_ci_if_error: true # optional (default = false)
           verbose: true # optional (default = false)
+      - name: Validate that kas-fleet-manager images build successfully.
+        if: github.event_name == 'pull_request'
+        run: make image/build
```
miguelsorianod, line 152: The set of steps is a little bit strange here. This step builds the image through the Makefile
but then in main we also build the image later.

pr_check.sh
machi1990: > Do we need this? in midstream the pipeline already tries to buy the image right?

The image is only built when something is merged on `main` for both Github and midstream pipelines. The usefulness of this is that we'll be sure that before we merge something in main, it can successful build the image.

.github/workflows/ci.yaml
machi1990: Yes. The later stage build and push the image when a change happens on the `main` branch. 
Here we only want to check that the image can be built.

pr_check.sh
miguelsorianod: but in midstream it won't be merged until the image is built no?

machi1990: Without this, if a PR introduce a breaking step as far as image build is concerned, that PR will be merged on the midstream KFM repo, which then kicks a midstream pipeline to build the image: it is at this step where it'll fail - long after the PR pipeline has finished.

.github/workflows/ci.yaml
miguelsorianod: We should revisit the organization of the CI workflows and jobs as the current organization does not seem ideal.
For example building the container image on PRs could be done in parallel without waiting to any task.

pr_check.sh
miguelsorianod: Ok so the the image won't be pushed to the repository but the content will be merged to midstream main.
What concerns me is that that file is becoming kind of a container for tasks in midstream. It might make sense to separate in different jobs for the midstream PR CI flow.

.github/workflows/ci.yaml
machi1990: +1 on that. I think most of the steps here can be separate jobs

pr_check.sh
machi1990: > Ok so the the image won't be pushed to the repository but the content will be merged to midstream main.

Yes. 

> What concerns me is that that file is becoming kind of a container for tasks in midstream. It might make sense to separate in different jobs for the midstream PR CI flow.

Yes. This is very valid concern. Do you mind creating a JIRA for this? I won't be able to get to that as part of this PR but it is something we should really revisit for both midstream and upstream pipelines as mentioned in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1641#discussion_r1138641797

miguelsorianod (approved):
:+1: let's go ahead with the changes in here keeping in mind we should revisit the structuring of our CI

pr_check.sh
miguelsorianod: I will create a jira for it

machi1990:
pushed a small change to reword the naming of a workflow step.

machi1990:
Thanks for the review @miguelsorianod

pr_check.sh
machi1990: Thanks

pawelpaszki (approved):

PR #1638, opened 2023-03-14
Minor reviews cert storage

machi1990:
## Description

1. use the in memory storage constructor
2. apply the prefix filter when constructing the filter value for listing keys in aws secret manager

## Checklist (Definition of Done)

- [ ] All acceptance criteria specified in JIRA have been completed
- [ ] Unit and integration tests added that prove the fix is effective or the feature works (tested against emulated and non-emulated OCM environment)
- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer
- [ ] All PR comments are resolved either by addressing them or creating follow up tasks
- [ ] Required metrics/dashboards/alerts have been added (or PR created).
- [ ] Required Standard Operating Procedure (SOP) is added.
- [ ] JIRA has been created for changes required on the client side

PR #1642, opened 2023-03-16
fix(secure-storage): only trim by the aws secret manager global prefix to get the resulting key

machi1990:
## Description

This follows up on
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1638/commits/64d56dd8ccb7899eca2c888f5e7b01286a89d702

which introduce an unintended bug of returning wrong keys as expected by certmagic since prefix was removed.

For example:

If the storage.secretPrefix is "kas-fleet-manager"

When we call `Store(ctx, "example/secret-one")` then the resulting secret name in the storage backend is going to be "kas-fleet-manager/example/secret-one". When we call `List(ctx, "example", true)` then the returned list of keys should be `["example/secret-one"]` and not `/secret-one` as it is current is. This was because we trimmed by the filterValue, an exising bug from the first version. We should only trim by the "kas-fleet-manager/" prefix.

- [ ] Code Review completed


machi1990:
@pb82 @miguelsorianod this builds on top of https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1638 to enhance the List implementation. Can you have a look please?

machi1990:
I'll work on adding unit test for the whole storage implementation on a separate PR tomorrow.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -12,19 +12,23 @@ import (
...
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
 	"github.com/aws/aws-secretsmanager-caching-go/secretcache"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
 	"github.com/caddyserver/certmagic"
 )
 
+//go:generate moq -out secretmanager_client_moq.go . SecretManagerClient
+type SecretManagerClient = secretsmanageriface.SecretsManagerAPI
```
miguelsorianod, line 22: This is a nice approach, and it makes me think that it should be possible to to the same for other external types that we are generating workarounds for. Specifically, for Gocloak and Route53 where we call the moq command referencing the vendor directories.

We could define type alias for them and generate them instead from our packages, removing the generation workarounds.
What do you think? If so, I can take a look a it.

machi1990: +1 on that. I think it'll also remove the need for vendoring. Thank you for having a look.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -194,5 +199,9 @@ func (storage *secureStorage) String() string {
 }
 
 func (storage *secureStorage) constructSecretName(key string) string {
-	return fmt.Sprintf("%s/%s", storage.secretPrefix, key)
+	return fmt.Sprintf("%s%s", storage.constructAWSSecretManagerKeysPrefix(), key)
```
pb82, line 202: why no slash here? Is this already part of what `constructAWSSecretManagerKeysPrefix` returns?

machi1990: Yes. That's correct, it is part of that method call.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
machi1990: @pb82 this is done https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/Makefile#L417-L418

miguelsorianod: I will work on generating those mocks locally

miguelsorianod:
Is the unit testing done? if so, it looks good to me

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage_test.go
```diff
@@ -0,0 +1,652 @@
...
+				},
+			},
+			args: args{
+				key:   "some-key-to-store",
+				value: []byte("some byte to store"),
+			},
+			wantErr: true,
+		},
+		{
+			name: "successful updates the secret when it exists than ResourceExistsException",
```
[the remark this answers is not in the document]
pb82: I think it should be `successfully`

pb82 (approved):

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage_test.go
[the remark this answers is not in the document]
pb82: successfully updates the secret when it exists rather than raising ResourceExistsException?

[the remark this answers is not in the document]
machi1990: Thanks. I'll update this.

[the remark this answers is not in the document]
machi1990: updated

machi1990:
> Is the unit testing done? if so, it looks good to me

Thanks for the review @miguelsorianod @pb82 just pushed an update to address https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1642#discussion_r1139933930 wording issue

miguelsorianod (approved):
If the unit testing has been finished and the tests pass it can be merged

PR #1647, opened 2023-03-17
WIP: structured dataplane observability config and credentials

pb82:
A structured way to support the different possible Observability configurations. This adds a yaml file in the secrets directory where the config is split between `proxy` (Obs RW Proxy) and `direct` (send directly to Observatorium).

Values relevant in both cases can be kept at the top level. This is a work in progress to propose a new configuration format for Observability.


secrets/dataplane-observability-config.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
miguelsorianod, line 1: Let's include a .sample file for the configuration documentation. But let's first discuss the configuration design

secrets/dataplane-observability-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+# the url points directly to an (authenticated) remote write endpoint
+# this is optional, either proxy or direct configurations must be provided
+# direct:
+#   remoteWriteUrl: https://observatorium-mst.api.stage.openshift.com
+#   metricsClientId: dummy
+#   metricsClientSecret: dummy
+
+# optional: a GitHub token to pull observability resources
+# if not provided, the resources have to be either a public GitHub repository, or
+# a public quay repository
```
machi1990, line 16: It would have been nice to also being able to be explicit about this but OO just infers that for now.

secrets/dataplane-observability-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+# this is optional, either proxy or direct configurations must be provided
+proxy:
+  remoteWriteUrl: https://observability-rw-proxy-managed-services-stage.apps.app-sre-stage-0.k3s7.p1.openshiftapps.com
+
+# the url points directly to an (authenticated) remote write endpoint
+# this is optional, either proxy or direct configurations must be provided
+# direct:
+#   remoteWriteUrl: https://observatorium-mst.api.stage.openshift.com
+#   metricsClientId: dummy
+#   metricsClientSecret: dummy
```
machi1990, line 12: What about the dex based support? We could also consider adding them in this file. Or we could scope that to a future improvement once this initial structure support is in

secrets/dataplane-observability-config.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# the url points to an auth proxy before the remote write endpoint
+# this is optional, either proxy or direct configurations must be provided
+proxy:
+  remoteWriteUrl: https://observability-rw-proxy-managed-services-stage.apps.app-sre-stage-0.k3s7.p1.openshiftapps.com
```
machi1990, line 5: We can make this empty

miguelsorianod:
As we discussed, let's redesign the configuration interface and the implementation.
We have been discussing the new design for through other channels.

secrets/dataplane-observability-config.yaml
pb82: If we want to continue supporting dex based Observatorium, we can consider moving the config into this file. Although it looks quite different, there is a long lived token instead of clientId and secret.

pkg/client/observatorium/observability_config.go
```diff
@@ -12,15 +12,19 @@ import (
 )
 
 const (
-	AuthTypeDex = "dex"
-	AuthTypeSso = "redhat"
+	AuthTypeDex                 = "dex"
+	AuthTypeSso                 = "redhat"
+	RemoteWriteConfigTypeProxy  = "proxy"
```
miguelsorianod, line 17: I understand this constant is not needed anymore

pkg/client/observatorium/observability_config.go
```diff
@@ -12,15 +12,19 @@ import (
...
+	RemoteWriteConfigTypeDirect = "direct"
```
miguelsorianod, line 18: Same as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1647/files#r1142040255

pkg/client/observatorium/observability_config.go
```diff
@@ -80,6 +70,74 @@ type ObservabilityCloudWatchLoggingConfig struct {
...
 
+type DataPlaneObservabilityOIDCCredentials struct {
+	ClientID     string `yaml:"client_id" validate:"required"`
+	ClientSecret string `yaml:"client_secret" validate:"required"`
+}
+
+type DataPlaneObservabilityOIDCConfiguration struct {
+	AuthorizationServer string                                 `yaml:"authorization_server" validate:"required"`
+	Realm               string                                 `yaml:"realm" validate:"required"`
+	Credentials         *DataPlaneObservabilityOIDCCredentials `yaml:"credentials" validate:"required,dive"`
```
miguelsorianod, line 81: Is there some reason this is a pointer? I think we might be able to directly use `DataPlaneObservabilityOIDCCredentials`.

pkg/client/observatorium/observability_config.go
```diff
@@ -80,6 +70,74 @@ type ObservabilityCloudWatchLoggingConfig struct {
...
+}
+
+type DataPlaneObservabilityRemoteWriteConfiguration struct {
+	RemoteWriteUrl    string                                   `yaml:"remote_write_url" validate:"required"`
```
miguelsorianod, line 85: Well known acronyms should be written in capitalized letters as long as it does not affect exportability of the variables / types. In this case, it should be called `RemoteWriteURL

pkg/client/observatorium/observability_config.go
```diff
@@ -205,11 +282,13 @@ func (c *ObservabilityConfiguration) AddFlags(fs *pflag.FlagSet) {
...
-	fs.StringVar(&c.ObservabilityConfigAccessTokenFile, "observability-config-access-token-file", c.ObservabilityConfigAccessTokenFile, "File contains the access token to the observability operator configuration repo")
 	fs.StringVar(&c.ObservabilityConfigTag, "observability-config-tag", c.ObservabilityConfigTag, "Tag or branch to use inside the observability configuration repo")
 
 	fs.StringVar(&c.ObservabilityCloudWatchLoggingConfig.configFilePath, "observability-cloudwatchlogging-config-file-path", "secrets/observability-cloudwatchlogs-config.yaml", "Path to a file containing the configuration for Observability related to AWS CloudWatch Logging in YAML format. Only takes effect when --observability-cloudwatchlogging-enable is set")
 	fs.BoolVar(&c.ObservabilityCloudWatchLoggingConfig.CloudwatchLoggingEnabled, "observability-cloudwatchlogging-enable", false, "Enable Observability to deliver data plane logs to AWS CloudWatch")
+
+	fs.StringVar(&c.DataPlaneObservabilityConfig.configFilePath, "dataplane-observability-config", "secrets/dataplane-observability-config.yaml", "Path to a file containing the observability configuration for the data plane")
```
miguelsorianod, line 290: Can we name it `--dataplane-observability-config-file-path` instead?
I find the `--dataplane-observability-config" name ambiguous: Is it the configuration content itself or the path of a file containing it?
For that reason I propose renaming it to the proposed name above

pkg/client/observatorium/observability_config.go
```diff
@@ -205,11 +282,13 @@ func (c *ObservabilityConfiguration) AddFlags(fs *pflag.FlagSet) {
...
+	fs.BoolVar(&c.DataPlaneObservabilityConfig.Enabled, "dataplane-observability-config-enabled", true, "Determines if the data plane observability config file must be provided")
```
miguelsorianod, line 291: Can we call it `--dataplane-observability-config-enable` instead?
Other flags use `*enable*` instead of `*enabled*`

pkg/client/observatorium/observability_config.go
```diff
@@ -80,6 +70,74 @@ type ObservabilityCloudWatchLoggingConfig struct {
...
+
+type DataPlaneObservabilityConfig struct {
+	configFilePath string
+	Enabled        bool
+
+	RemoteWriteConfiguration *DataPlaneObservabilityRemoteWriteConfiguration `yaml:"remote_write_configuration" validate:"required,dive"`
+	GithubResourcesAuthToken string
+}
+
+func (c *DataPlaneObservabilityConfig) GetOIDCClientID() string {
```
miguelsorianod, line 97: I see lots of checks where `c.Enabled` is checked spread around several methods.
I think this might be overly complicated. Couldn't we only use those methods if it is enabled? That is, DataPlaneObservabilityConfig is only used by the clients if it is enabled

miguelsorianod:
Could you summarize and describe the set of changes in the original description of the PR?
This will help reviewers to get some context and information about the intended changes. This should also include explanation of the removal of elements and what are the implications for them.

templates/service-template.yml
```diff
@@ -1085,14 +1080,9 @@ objects:
             - --observatorium-timeout=${OBSERVATORIUM_TIMEOUT}
             - --observatorium-auth-type=${OBSERVATORIUM_AUTH_TYPE}
             - --observatorium-token-file=/secrets/service/observatorium.token
-            - --observability-red-hat-sso-logs-client-id-file=/secrets/observatorium/logs.clientId
```
miguelsorianod, line 1088: If we go ahead with the removal of flags I understand some changes on the Makefile might be needed too. Has that been reviewed?
The same question applies to the secrets templates

templates/service-template.yml
```diff
@@ -1085,14 +1080,9 @@ objects:
             - --observatorium-timeout=${OBSERVATORIUM_TIMEOUT}
             - --observatorium-auth-type=${OBSERVATORIUM_AUTH_TYPE}
             - --observatorium-token-file=/secrets/service/observatorium.token
```
miguelsorianod, line 1082: Wouldn't the set of changes in this PR require the addition of new parameters and flag setting in the service template?

secrets/dataplane-observability-config.yaml
pb82: Ok, or should I put some example dummy URL in?

machi1990: Yes. A dummy URL is also fine.

pkg/client/observatorium/observability_config.go
pb82: we could, but I would prefer to keep the use of those methods simple with all the checks being opaque to the caller. For example, we use them to assign fields in a secret in cluster_mgr.go. We would have to move the enabled check there.

miguelsorianod: With what I propose we make the `DataPlaneObservabilityConfig` methods simpler instead of spreading them among all the methods of it.
Then on the client side the only thing you do is to check if it is enabled and if it is then you use it.
DataPlaneObservabilityConfig methods have the precondition that it will be enabled (other than the read and validate one)

machi1990: @pb82 @miguelsorianod there is this check https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3cb3aa93e8989c6654abc741ab883772fcf0c0b4/internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go#L783 that I already see, my thinking is that is enough to control the creation of the service account secret. And we do not need the extra `c.Enabled` checks in the methods in the config

miguelsorianod: > @pb82 @miguelsorianod there is this check
> 
> https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/3cb3aa93e8989c6654abc741ab883772fcf0c0b4/internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go#L783
> that I already see, my thinking is that is enough to control the creation of the service account secret. And we do not need the extra `c.Enabled` checks in the methods in the config

+1. Or combine with Enabled && HasOIDCConfiguration better

templates/secrets-template.yml
```diff
@@ -208,9 +196,5 @@ objects:
   metadata:
     name: kas-fleet-manager-observatorium-configuration-red-hat-sso
   stringData:
-    logs.clientId: ${OBSERVABILITY_RHSSO_LOGS_CLIENT_ID}
-    logs.clientSecret : ${OBSERVABILITY_RHSSO_LOGS_SECRET}
-    metrics.clientId: ${OBSERVABILITY_RHSSO_METRICS_CLIENT_ID}
-    metrics.clientSecret: ${OBSERVABILITY_RHSSO_METRICS_SECRET}
     grafana.clientId: ${OBSERVABILITY_RHSSO_GRAFANA_CLIENT_ID}
     grafana.clientSecret: ${OBSERVABILITY_RHSSO_GRAFANA_CLIENT_SECRET}
```
machi1990, line 200: these two are used for the control plane to read metrics using the token refresher, correct?

secrets/dataplane-observability-config.yaml
```diff
@@ -0,0 +1,13 @@
+---
```
machi1990, line 1: now that we've a sample file, can we ignore this?

templates/secrets-template.yml
pb82: Yes, we still need them.

pkg/client/observatorium/observability_config.go
machi1990: Yes @pb82 but only if you do not want the secret key to be present.
The current behaviour is that an `""` secret is going to be created which I think can still be achieved by removing the enabled check. To be clear, the nillness check has to stay. In that case, we'll continue to send `""` string.

machi1990: i.e send an `""` string if the underlying value is empty in the configuration or the configuration is `nil`. Which is what we intend to do.

templates/secrets-template.yml
machi1990: I am trying to determine if these can be removed (not necessarily within this PR but something to take note and create a JIRA for)

pkg/client/observatorium/observability_config.go
pb82: Ok, let me refactor this then.

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr_test.go
```diff
@@ -59,35 +59,39 @@ var (
...
+		AuthType:                   "redhat",
+		AuthToken:                  "dummy",
+		AuthTokenFile:              "dummy",
+		ObservabilityConfigTag:     "main",
+		ObservabilityConfigRepo:    "dummy",
+		ObservabilityConfigChannel: "resources",
+		DataPlaneObservabilityConfig: observatorium.DataPlaneObservabilityConfig{
+			Enabled: true,
+			RemoteWriteConfiguration: &observatorium.DataPlaneObservabilityRemoteWriteConfiguration{
+				RemoteWriteUrl: "https://dummy",
```
machi1990, line 83:
```suggestion
				RemoteWriteURL: "https://dummy",
```

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr_test.go
```diff
@@ -59,35 +59,39 @@ var (
...
+				OIDCConfiguration: &observatorium.DataPlaneObservabilityOIDCConfiguration{
+					AuthorizationServer: "https://dummy",
+					Realm:               "dummy",
+					Credentials: &observatorium.DataPlaneObservabilityOIDCCredentials{
```
machi1990, line 87:
```suggestion
					Credentials: observatorium.DataPlaneObservabilityOIDCCredentials{
```

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr_test.go
```diff
@@ -2379,14 +2391,12 @@ func buildResourceSet(observabilityConfig observatorium.ObservabilityConfigurati
 			Type: k8sCoreV1.SecretTypeOpaque,
 			StringData: map[string]string{
 				"authType":               observatorium.AuthTypeSso,
-				"gateway":                observabilityConfig.RedHatSsoGatewayUrl,
+				"gateway":                observabilityConfig.DataPlaneObservabilityConfig.GetRemoteWriteUrl(),
```
machi1990, line 2394:
```suggestion
				"gateway":                observabilityConfig.DataPlaneObservabilityConfig.GetRemoteWriteURL(),
```

docs/deploying-kas-fleet-manager-to-openshift.md
```diff
@@ -174,13 +174,13 @@ make deploy/service IMAGE_TAG=<your-image-tag-here> <OPTIONAL_PARAMETERS>
 - `OBSERVATORIUM_RHSSO_TENANT`: Tenant of an Observatorium using RHSSO as authentication. Defaults to `''`.
```
JameelB, line 174: Since the `OBSERVATORIUM_RHSSO_GATEWAY` is going to be removed, it can also be removed from this list.

secrets/dataplane-observability-config.yaml.sample
```diff
@@ -0,0 +1,40 @@
+---
+# [required]
+remote_write_configuration:
+  # [required] V2 URL endpoint of the Prometheus instance to be used by the data planes
+  # to send metrics. The Prometheus instance has to offer a Remote Write endpoint.
+  # If oidc_configuration is set that endpoint has to be offered under the
+  # /api/v1/metrics/<tenant>/api/v1/receive" URL path, where <tenant> is defined by the
+  # --observability-red-hat-sso-tenant Fleet Manager flag.
```
JameelB, line 8: Is there a reason why we still define the tenant as a flag rather than within this config?

templates/secrets-template.yml
JameelB: +1 these are still needed. Since the secret no longer has any of the dataplane secrets, this can be re-purposed for only the observatorium secrets required by kfm or it can be moved to the kas-fleet-manager secret file.

secrets/dataplane-observability-config.yaml.sample
```diff
@@ -0,0 +1,40 @@
...
+  # If oidc_coniguration is set that endpoint has to be offered under the
+  # "/" URL path.
```
JameelB, line 10: I find this description confusing. Does this mean that the endpoint has to be offered under both `/` and `/api/v1/metrics/<tenant>/api/v1/receive`?

secrets/dataplane-observability-config.yaml.sample
```diff
@@ -0,0 +1,40 @@
...
+  # The format of remote_write_url is `<url-scheme>:<hostname>:<port>` where
+  # <port> is optional.
+  remote_write_url: # fill here
+  # [optional] OIDC configuration used to retrieve OIDC tokens to authenticate against the
+  # Prometheus Remote Write endpoint URL. If not provided, the
+  # KAS Fleet Manager - Data Plane cross communication OIDC token will be used.
```
JameelB, line 16: `KAS Fleet Manager - Data Plane cross communication OIDC token`. What's this? 🤔 Is this talking about our current oidc config and credentials we use with the data plane token refreshers?

secrets/dataplane-observability-config.yaml.sample
pb82: I believe a `not` is missing in the second part.

templates/secrets-template.yml
machi1990: Thanks @JameelB @pb82 I am fine with the answers. Let's keep it that way for now and revisit what @JameelB said in follow up PRs

internal/kafka/internal/config/dataplane_cluster_config.go
```diff
@@ -92,8 +92,8 @@ const (
 
 // constants for Observability Operator installation via OpenShift Lifecycle Manager
 const (
-	defaultObservabilityOperatorIndexImage  = "quay.io/rhoas/observability-operator-index:v4.1.2"
-	defaultObservabilityOperatorStartingCSV = "observability-operator.v4.1.2"
+	defaultObservabilityOperatorIndexImage  = "quay.io/rhoas/observability-operator-index:v4.2.0"
```
miguelsorianod, line 95: Is there any reason this is being updated in this PR? is it related?

docs/populating-configuration.md
```diff
@@ -374,3 +374,11 @@ previously created.
 
 Additionally, make sure to set the Sentry URL endpoint and Sentry project when
 starting the Fleet Manager server. See [Sentry-related CLI flags in Fleet Manager](./feature-flags.md#sentry)
+
+## Configure data plane observability & remote write
```
miguelsorianod, line 378: Do you think this should belong in `Setup the Observability stack secrets` instead?

secrets/dataplane-observability-config.yaml.sample
miguelsorianod: Yes. a `not` is missing in the second part :+1:

secrets/dataplane-observability-config.yaml.sample
```diff
@@ -0,0 +1,40 @@
+---
+# [required]
+remote_write_configuration:
+  # [required] V2 URL endpoint of the Prometheus instance to be used by the data planes
```
miguelsorianod, line 4: This `V2` is not needed. It is a remain from when we were prototyping the configuration

secrets/dataplane-observability-config.yaml.sample
miguelsorianod: Aside from what peter's was mentioning, the reason we didn't move it already in the config is because we kind of reached to the conclusion that the issue is that KFM shouldn't be worried about the tenant at all. From the point of view of KFM the only thing that is needed is to have knowledge of a Prometheus Remote Write Endpoint. Whether Observatorium (a specific implementation of a remote write endpoint) has the concept of a tenant it is irrelevant. From the point of view of KFM it's just an URL.
Therefore, the long term solution for that wouldn't be moving the tenant to the config, but removing the concept of tenant altogether

internal/kafka/internal/config/dataplane_cluster_config.go
pb82: OO 4.2.0 is required for the proxy and since I needed to test this anyway, I've updated our default values too.

pb82: Ah sorry, you're referring to the quay resources. Similar reason though, with OO 4.2.0 this now becomes the default way of fetching observability resources. So we should also udpate our defaults.

pkg/client/observatorium/observability_config.go
```diff
@@ -205,11 +242,13 @@ func (c *ObservabilityConfiguration) AddFlags(fs *pflag.FlagSet) {
...
 	fs.StringVar(&c.ObservabilityConfigTag, "observability-config-tag", c.ObservabilityConfigTag, "Tag or branch to use inside the observability configuration repo")
 
 	fs.StringVar(&c.ObservabilityCloudWatchLoggingConfig.configFilePath, "observability-cloudwatchlogging-config-file-path", "secrets/observability-cloudwatchlogs-config.yaml", "Path to a file containing the configuration for Observability related to AWS CloudWatch Logging in YAML format. Only takes effect when --observability-cloudwatchlogging-enable is set")
 	fs.BoolVar(&c.ObservabilityCloudWatchLoggingConfig.CloudwatchLoggingEnabled, "observability-cloudwatchlogging-enable", false, "Enable Observability to deliver data plane logs to AWS CloudWatch")
+
+	fs.StringVar(&c.DataPlaneObservabilityConfig.configFilePath, "dataplane-observability-config-file-path", "secrets/dataplane-observability-config.yaml", "Path to a file configuring a remote write target. If provided, metrics from the data plane will be sent to the specified target.")
+	fs.BoolVar(&c.DataPlaneObservabilityConfig.Enabled, "dataplane-observability-config-enable", true, "Determines if the data plane observability config file must be provided")
```
miguelsorianod, line 251: I think this description is not helpful. More than focusing on whether this controls the file or not we should focus on what it enables/disable (i.e. sending metrics from the data plane etc...).

pkg/client/observatorium/observability_config.go
```diff
@@ -252,29 +284,10 @@ func (c *ObservabilityConfiguration) ReadFiles() error {
...
-	metricsSecretErr := shared.ReadFileValueString(c.MetricsSecretFile, &c.MetricsSecret)
-	if metricsSecretErr != nil {
-		return metricsSecretErr
+	err := c.ObservabilityCloudWatchLoggingConfig.validate()
+	if err != nil {
+		return err
 	}
 
-	return nil
+	return c.DataPlaneObservabilityConfig.validate()
```
miguelsorianod, line 292: I always prefer to do an explicit if err != nil check because although now that is the last thing that is validated in the future that might not be the case which will make you move this away from the return.
That is a personal preference though so I am good with either way.

internal/kafka/internal/services/data_plane_cluster.go
```diff
@@ -52,7 +52,7 @@ func (d *dataPlaneClusterService) GetDataPlaneClusterConfig(ctx context.Context,
 
 	return &dbapi.DataPlaneClusterConfig{
 		Observability: dbapi.DataPlaneClusterConfigObservability{
-			AccessToken: d.ObservabilityConfig.ObservabilityConfigAccessToken,
+			AccessToken: d.ObservabilityConfig.DataPlaneObservabilityConfig.GithubResourcesAuthToken,
```
miguelsorianod, line 55: If DataPlaneObservabilityConfig is not enabled then `AccessToken` will be set to an empty string? Is that what we want? If so, :+1:

pb82: Yes, an empty string is fine here. That's also consistent with the other secrets.

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -777,9 +777,15 @@ func (c *ClusterManager) buildResourceSet(cluster api.Cluster) types.ResourceSet
 		c.buildObservabilityCatalogSourceResource(),
 		c.buildObservabilityOperatorGroupResource(),
 		c.buildObservabilitySubscriptionResource(),
-		c.buildObservabilityRemoteWriteServiceAccountCredential(&cluster),
 	)
 
+	// if no oidc configuration is provided, or if it is disabled, use the data plane service account
+	if !(c.ObservabilityConfiguration.DataPlaneObservabilityConfig.Enabled && c.ObservabilityConfiguration.DataPlaneObservabilityConfig.HasOIDCConfiguration()) {
```
miguelsorianod, line 783: I find the way this condition is written is confusing.
The comment describes the more natural way that it is described:
if !enabled || !hasoidconfiguration would be more understandable.

templates/service-template.yml
```diff
@@ -480,7 +475,7 @@ parameters:
 - name: OBSERVABILITY_CONFIG_REPO
   displayName: Observability configuration repo URL
   description: URL of the observability configuration repo
-  value: "https://api.github.com/repos/bf2fc6cc711aee1a0c2a/observability-resources-mk/contents"
+  value: "quay.io/rhoas/observability-resources-mk"
```
miguelsorianod, line 478: What is the reason this has been changed? The new URL seems a quay container image repository while the previous one seems  GitHub content URL

templates/service-template.yml
```diff
@@ -489,8 +484,8 @@ parameters:
 
 - name: OBSERVABILITY_CONFIG_TAG
   displayName: Observability configuration tag
-  description: Tag or branch to use inside the configuration repo
-  value: "main"
+  description: Tag or branch in a git repository, or image tag in a container registry
+  value: "latest"
```
miguelsorianod, line 488: Why has this been changed from `main` to `latest?

templates/secrets-template.yml
```diff
@@ -165,6 +153,7 @@ objects:
     kubeconfig: ${KUBE_CONFIG}
     image-pull.dockerconfigjson: "${IMAGE_PULL_DOCKER_CONFIG}"
     gcp.api-credentials: "${GCP_API_CREDENTIALS}"
+    dataplane-observability-config.yaml: ${DATAPLANE_OBSERVABILITY_CONFIG}
```
miguelsorianod, line 156: I wonder if this should become its own K8s secret

templates/service-template.yml
pb82: Before it was referring to the `main` branch of the resources repository, now it's referring to the `latest` tag in the quay repo. So the change is related to the name of the tag where the latest upstream changes are available.

Makefile
```diff
@@ -548,6 +548,7 @@ secrets/setup/empty:
 	touch $(SECRETS_DIR)/kafka-tls.crt
 	touch $(SECRETS_DIR)/kafka-tls.key
 	touch $(SECRETS_DIR)/image-pull.dockerconfigjson
+	touch $(SECRETS_DIR)/dataplane-observability-config.yaml
```
miguelsorianod, line 551: Should the:

	touch $(SECRETS_DIR)/rhsso-logs.clientId
	touch $(SECRETS_DIR)/rhsso-logs.clientSecret
	touch $(SECRETS_DIR)/rhsso-metrics.clientId
	touch $(SECRETS_DIR)/rhsso-metrics.clientSecret

be removed?

templates/service-template.yml
```diff
@@ -1085,15 +1080,11 @@ objects:
...
-            - --observability-red-hat-sso-logs-client-id-file=/secrets/observatorium/logs.clientId
-            - --observability-red-hat-sso-logs-secret-file=/secrets/observatorium/logs.clientSecret
-            - --observability-red-hat-sso-metrics-client-id-file=/secrets/observatorium/metrics.clientId
-            - --observability-red-hat-sso-metrics-secret-file=/secrets/observatorium/metrics.clientSecret
             - --observability-red-hat-sso-auth-server-url=${OBSERVATORIUM_RHSSO_AUTH_SERVER_URL}
             - --observability-red-hat-sso-realm=${OBSERVATORIUM_RHSSO_REALM}
             - --observability-red-hat-sso-token-refresher-url=${OBSERVATORIUM_TOKEN_REFRESHER_URL}
-            - --observability-red-hat-sso-observatorium-gateway=${OBSERVATORIUM_RHSSO_GATEWAY}
             - --observability-red-hat-sso-tenant=${OBSERVATORIUM_RHSSO_TENANT}
+            - --dataplane-observability-config-file-path=/secrets/service/dataplane-observability-config.yaml
```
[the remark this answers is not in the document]
pb82: I omitted it because the default value is `true`.

pkg/client/observatorium/observability_config.go
```diff
@@ -31,19 +31,10 @@ type ObservabilityConfiguration struct {
 	DexPasswordFile string `json:"password_file" yaml:"password_file"`
 
 	// Red Hat SSO configuration
-	RedHatSsoGatewayUrl        string `json:"redhat_sso_gateway_url" yaml:"redhat_sso_gateway_url"`
 	RedHatSsoAuthServerUrl     string `json:"redhat_sso_auth_server_url" yaml:"redhat_sso_auth_server_url"`
 	RedHatSsoRealm             string `json:"redhat_sso_realm" yaml:"redhat_sso_realm"`
 	RedHatSsoTenant            string `json:"redhat_sso_tenant" yaml:"redhat_sso_tenant"`
 	RedHatSsoTokenRefresherUrl string `json:"redhat_sso_token_refresher_url" yaml:"redhat_sso_token_refresher_url"`
-	MetricsClientId            string `json:"redhat_sso_metrics_client_id" yaml:"redhat_sso_metrics_client_id"`
```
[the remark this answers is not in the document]
miguelsorianod: For the new config file isn't that only used for metrics?

templates/service-template.yml
miguelsorianod: This means that we changed the default to use a container image, am I correct?

templates/service-template.yml
```diff
@@ -489,8 +484,8 @@ parameters:
 
 - name: OBSERVABILITY_CONFIG_TAG
   displayName: Observability configuration tag
-  description: Tag or branch to use inside the configuration repo
-  value: "main"
+  description: Tag or branch in a git repository, or image tag in a container registry
```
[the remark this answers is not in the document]
miguelsorianod: Can we update the descriptions of both the OBSERVABILITY_CONFIG_TAG and OBSERVABILITY_CONFIG_REPO to reflect what is supported and expected?

templates/service-template.yml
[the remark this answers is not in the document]
miguelsorianod: But that's not the same. If we don't add it there's no configurability at template level that a user can do.

pkg/client/observatorium/observability_config.go
```diff
@@ -205,11 +242,13 @@ func (c *ObservabilityConfiguration) AddFlags(fs *pflag.FlagSet) {
...
 	fs.StringVar(&c.ObservabilityConfigTag, "observability-config-tag", c.ObservabilityConfigTag, "Tag or branch to use inside the observability configuration repo")
 
 	fs.StringVar(&c.ObservabilityCloudWatchLoggingConfig.configFilePath, "observability-cloudwatchlogging-config-file-path", "secrets/observability-cloudwatchlogs-config.yaml", "Path to a file containing the configuration for Observability related to AWS CloudWatch Logging in YAML format. Only takes effect when --observability-cloudwatchlogging-enable is set")
 	fs.BoolVar(&c.ObservabilityCloudWatchLoggingConfig.CloudwatchLoggingEnabled, "observability-cloudwatchlogging-enable", false, "Enable Observability to deliver data plane logs to AWS CloudWatch")
+
+	fs.StringVar(&c.DataPlaneObservabilityConfig.configFilePath, "dataplane-observability-config-file-path", "secrets/dataplane-observability-config.yaml", "Path to a file configuring a remote write target. If provided, metrics from the data plane will be sent to the specified target.")
```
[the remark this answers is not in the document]
pb82: Replacing with: `Path to a file containing the configuration for data plane observability, including remote write. If provided, data plane will send metrics to the provided remote write receiver`

secrets/dataplane-observability-config.yaml.sample
JameelB: makes sense now, thanks for the clarification!

secrets/dataplane-observability-config.yaml.sample
JameelB: I can't think of anything else apart from what is mentioned here at the top of my head 🤔 thanks for clarifying!

templates/service-template.yml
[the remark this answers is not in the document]
pb82: Changing the default to false, for development we don't need observatorium. We need to make sure this gets overridden with `true` in app-interface.

templates/secrets-template.yml
miguelsorianod: After talking through other channels we decided to keep it in the secret used in the PR for now.
In the future we might change it.

docs/populating-configuration.md
pb82: Yes, that makes sense. Looking at that section, it needs to be updated anyway because it doesn't yet describe how to set up the proxy. I'd prefer to do this in a different PR.

templates/service-template.yml
pb82: Yes, that's correct.

docs/populating-configuration.md
miguelsorianod: > Yes, that makes sense. Looking at that section, it needs to be updated anyway because it doesn't yet describe how to set up the proxy. I'd prefer to do this in a different PR.

I already have some local branch describing how to setup the proxy. You can add it in here and I will join everything

docs/deploying-kas-fleet-manager-to-openshift.md
miguelsorianod, line 1: The dataplane observability config enable template parameter is missing from here

docs/populating-configuration.md
```diff
@@ -262,33 +262,13 @@ related to the Data Planes and its Kafka instances Fleet Manager needs to:
...
 
-The following command is used to setup the various secrets needed by
-the Observability stack:
-```
-OBSERVATORIUM_CONFIG_ACCESS_TOKEN="<observatorium-config-access-token> \
-RHSSO_LOGS_CLIENT_ID=<rhsso-logs-client-id> \
-RHSSO_LOGS_CLIENT_SECRET=<rhsso-logs-client-secret> \
-RHSSO_METRICS_CLIENT_ID=<rhsso-metrics-client-id> \
-RHSSO_METRICS_CLIENT_SECRET=<rhsso-metrics-client-secret> \
-make observatorium/setup
```
miguelsorianod, line 285: Should we fully remove then the `observatorium/setup` Makefile target completely?

pb82: I'll check if it does anything useful. If not we can remote it.

pb82: I've removed it, it was only setting up those secrets.

miguelsorianod: The tests are failing due to the removal. This is due to several Dockerfiles and/or CI scripts were referencing it. Could you review it?

thanks

miguelsorianod:
I think this might impact kas installer as this PR contains breaking changes from the point of view of CLI flags and configuration files / secrets.

cc @MikeEdgar could you tell us if this might have an impact on it? We can talk about this if needed to evaluate the affected areas

docs/deploying-kas-fleet-manager-to-openshift.md
```diff
@@ -170,17 +170,17 @@ make deploy/service IMAGE_TAG=<your-image-tag-here> <OPTIONAL_PARAMETERS>
...
+- `DATAPLANE_OBSERVABILITY_CONFIG_ENABLE`: Enable sending metrics to the remote write receiver which is configured in the file referenced from `--dataplane-observability-config-file-path`.
 - `ENABLE_TERMS_ACCEPTANCE`: Enables terms acceptance through AMS. Defaults to `false`.
 - `ALLOW_DEVELOPER_INSTANCE`: Enables creation of developer Kafka instances. Defaults to `true`.
 - `QUOTA_TYPE`: Quota management service to be used. Options: `quota-management-list` and `ams`, Default: `quota-management-list`.
 - `KAS_FLEETSHARD_OLM_INDEX_IMAGE`: KAS Fleetshard operator OLM index image. Defaults to `quay.io/osd-addons/kas-fleetshard-operator:production-82b42db`.
 - `STRIMZI_OLM_INDEX_IMAGE`: Strimzi operator OLM index image. Defaults to `quay.io/osd-addons/managed-kafka:production-82b42db`.
-- `OBSERVABILITY_OPERATOR_INDEX_IMAGE`: Observability Operator index image. Defaults to `quay.io/rhoas/observability-operator-index:v4.1.2`.
+- `OBSERVABILITY_OPERATOR_INDEX_IMAGE`: Observability Operator index image. Defaults to `quay.io/rhoas/observability-operator-index:v42.0`.
```
vmanley, line 183:
```suggestion
- `OBSERVABILITY_OPERATOR_INDEX_IMAGE`: Observability Operator index image. Defaults to `quay.io/rhoas/observability-operator-index:v4.2.0`.
```

pb82: good catch!

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -777,9 +777,15 @@ func (c *ClusterManager) buildResourceSet(cluster api.Cluster) types.ResourceSet
 		c.buildObservabilityCatalogSourceResource(),
 		c.buildObservabilityOperatorGroupResource(),
 		c.buildObservabilitySubscriptionResource(),
-		c.buildObservabilityRemoteWriteServiceAccountCredential(&cluster),
```
MikeEdgar, line 780: Are `buildObservatoriumDexSecretResource` and `buildObservatoriumSSOSecretResource` still needed with the proxy?

pb82: The dex secret will be removed in a separate PR. The RHSSO secret is still created, but fields containing credentials are set to empty strings. Can be further improved in a separate PR.

pkg/client/observatorium/observability_config.go
```diff
@@ -94,6 +151,46 @@ func (c *ObservabilityCloudWatchLoggingConfig) validate() error {
...
 
+func (c *DataPlaneObservabilityConfig) validate() error {
+	if !c.Enabled {
+		return nil
+	}
+
+	validate := validator.New()
+
+	oidcConfig := c.RemoteWriteConfiguration.Authentication.OIDC
+	if oidcConfig != nil && oidcConfig.Static != nil && oidcConfig.Autogenerated != nil {
```
pb82, line 162: I intended to use a custom type validator, but couldn't make it work with Go playground validator and the docs for it are lacking.

miguelsorianod: This can be achieved by using a struct level validator with the `RegisterStructValidation` method.
See an example in https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go#L80

secrets/dataplane-observability-config.yaml.sample
```diff
@@ -0,0 +1,53 @@
...
+  # If oidc_coniguration is not set that endpoint has to be offered under the
+  # "/" URL path.
+  # The format of remote_write_url is `<url-scheme>:<hostname>:<port>` where
+  # <port> is optional.
+  remote_write_url: # fill here
+  # [required] Authentication method to be used to authenticate against
+  # the Remote Write endpoint. One and only one authentication method
+  # is required.
+  authentication:
+    # [optional] Use OIDC authentication method. One and only one oidc
```
miguelsorianod, line 18: We can change this to:
```
[optional] Use OIDC authentication method. If set, one and only one OIDC configuration type must be set
```

secrets/dataplane-observability-config.yaml.sample
```diff
@@ -0,0 +1,53 @@
...
+  # If oidc_configuration is set that endpoint has to be offered under the
+  # /api/v1/metrics/<tenant>/api/v1/receive" URL path, where <tenant> is defined by the
+  # --observability-red-hat-sso-tenant Fleet Manager flag.
+  # If oidc_coniguration is not set that endpoint has to be offered under the
+  # "/" URL path.
+  # The format of remote_write_url is `<url-scheme>:<hostname>:<port>` where
+  # <port> is optional.
+  remote_write_url: # fill here
+  # [required] Authentication method to be used to authenticate against
+  # the Remote Write endpoint. One and only one authentication method
```
miguelsorianod, line 15: We can change this to:
```
... . One and only one authentication method must be set
```

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr.go
```diff
@@ -934,15 +940,24 @@ func (c *ClusterManager) buildObservatoriumSSOSecretResource() *k8sCoreV1.Secret
...
+		"gateway":                "",
+		"redHatSsoAuthServerUrl": "",
+		"redHatSsoRealm":         "",
+		"metricsClientId":        "",
+		"metricsSecret":          "",
+	}
+
+	if observabilityConfig.DataPlaneObservabilityConfig.Enabled {
+		stringDataMap["gateway"] = observabilityConfig.DataPlaneObservabilityConfig.RemoteWriteConfiguration.RemoteWriteURL
+		if observabilityConfig.DataPlaneObservabilityConfig.HasStaticOIDCConfiguration() {
```
miguelsorianod, line 953: This is possibly not the best approach. We should have some type that simply gets the desired OIDC configuration elements without asking on the specific type of configuration.
Let's leave it like this for now but let's try to improve it in a future PR.

pkg/client/observatorium/observability_config.go
```diff
@@ -80,6 +70,73 @@ type ObservabilityCloudWatchLoggingConfig struct {
...
+
+func (c *DataPlaneObservabilityConfig) HasStaticOIDCConfiguration() bool {
+	if c.RemoteWriteConfiguration.Authentication.OIDC != nil {
+		return c.RemoteWriteConfiguration.Authentication.OIDC.Static != nil
+	}
+
+	return false
+}
+
+func (c *DataPlaneObservabilityRemoteWriteAuthTypeOIDC) GetStaticRealm() string {
```
miguelsorianod, line 124: Why are this getters being configured?

pb82: I can remove them. I think initially I had very long names for the struct fields, but using the fields directly instead of the getters is almost identical now.

openapi/kas-fleet-manager.yaml
```diff
@@ -2223,7 +2223,7 @@ components:
         Search criteria.
 
         The syntax of this parameter is similar to the syntax of the `where` clause of an
-        SQL statement. Allowed fields in the search are `cloud_provider`, `name`, `owner`, `region`, `status` and `cluster_id`. Allowed comparators are `<>`, `=`, `IN`, `NOT IN`, `LIKE`, or `ILIKE`.
+        SQL statement. Allowed fields in the search are `cloud_provider`, `name`, `owner`, `region`, `status`, `instance_type`, and `cluster_id`. Allowed comparators are `<>`, `=`, `IN`, `NOT IN`, `LIKE`, or `ILIKE`.
```
machi1990, line 2226: @pb82 this change shouldn't be included. It's already in `main` - you might need to rebase

machi1990 (approved):
the change lgtm, thanks @pb82 I've however left a comment regarding a rebase issue

I'll let @miguelsorianod have a look if his comments have been addressed.
Also let's make sure that kas-installer is updated and there is a new release that's aware of these changes. /cc @MikeEdgar @pb82

pkg/client/observatorium/observability_config.go
```diff
@@ -94,6 +135,52 @@ func (c *ObservabilityCloudWatchLoggingConfig) validate() error {
...
+	}
+}
+
+func (c *DataPlaneObservabilityConfig) validate() error {
+	if !c.Enabled {
+		return nil
+	}
+
+	validate := validator.New()
+	validate.RegisterStructValidation(validateExactlyOneOIDCTypeProvided, DataPlaneObservabilityRemoteWriteAuthTypeOIDC{})
```
miguelsorianod, line 157: We also need to validate that exactly one Authentication type is provided

pkg/client/observatorium/observability_config.go
```diff
@@ -94,6 +135,52 @@ func (c *ObservabilityCloudWatchLoggingConfig) validate() error {
 	return nil
 }
 
+func validateExactlyOneOIDCTypeProvided(sl validator.StructLevel) {
+	config := sl.Current().Interface().(DataPlaneObservabilityRemoteWriteAuthTypeOIDC)
```
miguelsorianod, line 139: This validation should only be performed if authentication is of the OIDC type

miguelsorianod: This is related to https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1647/files#r1149416320

pkg/client/observatorium/observability_config.go
miguelsorianod: Maybe the `required` validate struct tag already covers this. Could you verify it?

miguelsorianod: > Isn't that already the case? We're failing the validation if both are nil or if both are not nil. That only allows cases where exactly one is defined. If we later add a third option, we will have to update the validator anyway.

You are describing the validation around the possible OIDC configurations. What I am describing is the validation around the possible Authentications

internal/kafka/internal/workers/cluster_mgrs/clusters_mgr_test.go
```diff
@@ -780,7 +787,7 @@ func TestClusterManager_processProvisionedClusters(t *testing.T) {
 						return nil
 					},
 					ApplyResourcesFunc: func(cluster *api.Cluster, resources types.ResourceSet) *apiErrors.ServiceError {
-						// the resource set must contain 8 items: the observability resources plus two image pull secrets
+						// the resource set must contain 10 items: the observability resources plus two image pull secrets
 						if len(resources.Resources) != 10 {
 							return apiErrors.GeneralError(fmt.Sprintf("expected 8 items in the resource set but got %v", len(resources.Resources)))
```
miguelsorianod, line 792: The number here is outdated

MikeEdgar (approved):
lgtm from a kas-installer perspective. I've noted some additional changes needed in https://github.com/bf2fc6cc711aee1a0c2a/kas-installer/pull/279 once this change is merged.

miguelsorianod (approved):
Looks good to me. Great work :+1: .

PR #1663, opened 2023-03-24
fix(data_plane_kafka): only allow transition to deleting state if the kafka was in deprovision state

machi1990:
## Description

Only allow transition to deleting state if the kafka was in deprovision state. 
This prevents erroneous attempts to delete a Kafka that has not been marked for deprovisioning

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/data_plane_kafka.go
+++ b/internal/kafka/internal/services/data_plane_kafka.go
@@ -14,6 +14,7 @@ import (
 	serviceError "github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/metrics"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 )
@@ -194,6 +195,11 @@ func (d *dataPlaneKafkaService) processRealKafkaDeployment(ks *dbapi.DataPlaneKa
 			log.Errorf("kafka %q with status %q received errors from data plane: %q", kafka.ID, kafka.Status, readyCondition.Message)
 		}
 	case statusDeleted:
+		if !arrays.Contains(constants.GetDeletingStatuses(), kafka.Status) {
+			log.Errorf("kafka %q with status %q is not in deleting state and thus it cannot be deleted from %q data plane", kafka.ID, kafka.Status, kafka.ClusterID)
+			return
+		}
+
 		e = d.setKafkaClusterDeleting(kafka)
 	case statusRejected:
 		e = d.reassignKafkaCluster(kafka)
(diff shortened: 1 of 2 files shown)


internal/kafka/internal/services/data_plane_kafka.go
```diff
@@ -194,6 +195,11 @@ func (d *dataPlaneKafkaService) processRealKafkaDeployment(ks *dbapi.DataPlaneKa
 			log.Errorf("kafka %q with status %q received errors from data plane: %q", kafka.ID, kafka.Status, readyCondition.Message)
 		}
 	case statusDeleted:
+		if !arrays.Contains(constants.GetDeletingStatuses(), kafka.Status) {
+			log.Errorf("kafka %q with status %q is not in deleting state and thus it cannot be deleted from %q data plane", kafka.ID, kafka.Status, kafka.ClusterID)
+			return
+		}
+
```
JameelB, line 202: Do we know what causes FSO to send us a deleted status for a Kafka instance even though it wasn't deleted in KFM? 

Is it possible for a Kafka instance be deleted on the data plane even if it wasn't deleted in kfm? what happens in this case? 🤔

machi1990: > Do we know what causes FSO to send us a deleted status for a Kafka instance even though it wasn't deleted in KFM?

No, the fleetshard doesn't do this. But someone, if they maliciously grab the service account they can craft the call which might end up deleting the Kafka from KFM DB. This guard is to protect against that.

> Is it possible for a Kafka instance be deleted on the data plane even if it wasn't deleted in kfm? what happens in this case?

This shouldn't be possible. Unless otherwise, the MK CR is manually modified in the data plane.

machi1990: > Ideally this should never happen right? 

yes, if it happens then the system has a 15s to revert everything but that's a lot of time and the Kafka would've been deleted b then.

> Is there a way to prevent this atm? With the new hybrid model, cluster admins or anyone who has access/right permissions could do this.

I can't think of any. Maybe alerting but I am not sure if work has been done on this side. For now, if the user modifies it then they are messing with their own workload - basically they want to be and fly on their own. So it is best to lean on the trust we've with them.

JameelB: In that case, with this change, the Kafka instance remains to in a `ready` state. KFM will continue to send the Kafka CR to fleetshard sync. Do you know if the Kafka deployments will get re-created or will it error out on the data plane side? 🤔 Trying to wrap my head around the whole flow to see when this sort of case happens, what actions we'd need to take, if any :)

machi1990: As far as I am aware, KFM will continue to send the CR and the deployments will be recreated. 

To be fair, if the unintentional deletion is initiated from the data plane then we've to figure out the actor and have a word. Our system (the Fleetshard) cannot do it - unless otherwise told by KFM. 

It could well be initiated by just someone making a direct call to the API - IF they somehow grabbed the service accounts needed by FSO. It is more this case that this PR attempts to prevent, as if it is the first case an attempt initiated from the data plane by a malicious actor then there is a bigger issue.

In all the cases, the error will be seen in Sentry and we'll get the after and the primary engineer will need to take followup. Not sure if there is much we can do rather than ensuring that we're covered from credentials leaks etc

JameelB: We may need to make it clear that if that scenario happens, the deployments will be recreated. To the user, it may not be obvious that the Kafka instance is re-created and they may suddenly wonder why all of their data is gone 🤔 Are we ok with this?

JameelB: Also rather than it being logged, is it possible to have some sort of alert on this? That would make it more visible to sre as i don't think they check glitchtip atm.

PR #1664, opened 2023-03-24
Updated to app-services-sdk-go in Core SDK

jackdelahunt:
## Description
This PR updates the location and version of the `app-services-sdk-go` used. The old package used [here](https://github.com/redhat-developer/app-services-sdk-go) is now deprecated and is being replaced by the [Core SDK](https://github.com/redhat-developer/app-services-sdk-core).

## Verification Steps
- Check if all unit tests and linting has passed

- [ ] Code Review completed


go.mod
```diff
@@ -42,7 +42,7 @@ require (
 	github.com/prometheus/client_golang v1.14.0
 	github.com/prometheus/client_model v0.3.0
 	github.com/prometheus/common v0.42.0
-	github.com/redhat-developer/app-services-sdk-go/serviceaccounts v0.4.0
+	github.com/redhat-developer/app-services-sdk-core/app-services-sdk-go/serviceaccountmgmt v0.0.0-20230323122535-49460b57cc45
```
machi1990, line 45: Thanks for the PR. Quick question, there is no tag?

jackdelahunt: Because we are now publishing all packages in the SDK in one release the tag no longer points to the version (as that is the version of the parent project) but the commit/tag of the latest service account version.

Does that make sense?

jackdelahunt: The latest release is https://github.com/redhat-developer/app-services-sdk-core/releases

machi1990: I am trying to understand more. Does it mean that the tag 1.0.1 can't be used?

jackdelahunt: Yes, but it is pointing to `v1.0.1` just the tag is not it

machi1990:
@jackdelahunt can you rebase?

machi1990 (approved):
thanks for the PR, it needs a rebase @jackdelahunt

jackdelahunt:
> it needs a rebase

Done 👍

miguelsorianod (changes requested):
See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1664/files#r1149430528

go.mod
jackdelahunt: Hey sorry for late response, this pr adds the "core" version of where all SDKs are released under one tag, in our latest case `v1.0.1`.  Because the package of `serviceaccountmgmt` is two sub-directories deep into the repo it points to the commit not the tag.

miguelsorianod: Hi,

Usually in Go for the compiled binary the compiler only includes the functions and types referenced so even if the you import a massive package they shouldn't all be included.
Have you checked the differences in the compiled binary size by importing the whole package and a single subdirectory?
The other consideration is about the size of the source code to download. What's the size of it? is it big? If it were to be that massive that would be an indication that some refactoring would be a candidate to be considered, but it seems unlikely.
For example take the AWS SDK https://github.com/aws/aws-sdk-go which is quite big. You can see it also has different services within and the size is considerable and it is imported fully in lots of projects.

jackdelahunt: The reason why we include the subpackage and not everything is not due to binary size. Including the sub-package with the version of the sdk for example `serviceaccountmgmt v1.0.1` simply doesn't work. To get it to work we would need to create a tag in the repo `app-services-sdk-go/v1.0.1`. But we want to keep all releases and tags consistent

miguelsorianod: What I was proposing is not to include the subpackage, but the whole package.
But I see you don't perform tagging at the module level. I wonder if what you are doing is in line with how Go module version management should be performed. The way it is being done users will never be able to use a specific module version via tag.
Maybe what you want are submodules, or not having tags per package (ocm sdk go doesn't do that for example)?

jackdelahunt: Yes they won't be able to set a specific tag, I am not sure what you mean by the rest of the comment.

> What I was proposing is not to include the subpackage, but the whole package

I think this would be ideal but the rest of teh SDKs are structured like this and this is how the generator generates the code. There may be a way to change this but I will need to look into it. As for including only the top level module, I believe this still wouldn't fix the issue of not being able to specify the tag as the top level module is still a sub directory of the repo

jackdelahunt: > What are the "rest of the SDKs"?

The TS, Java and Python SDKs

> ocm sdk go doesn't do that for example

Yes looking at `ocm-sdk-go` it just contains subdirectories for each service, this is different to use as they are different modules

jackdelahunt:
@machi1990 done 👍

PR #1670, opened 2023-03-27
fix(cert): configure TestCA so that it points to Letsencrypt Prod instance as well

machi1990:
## Description

The LE prod instance is what we've configured by default.

If TestCA is not configure, the default letsencrypt Stage env was used which was causing an invalid account errors, like below:
```
https://acme-staging-v02.api.letsencrypt.org/acme/new-order: HTTP 400 urn:ietf:params:acme:error:malformed - KeyID header contained an invalid account URL: "https://acme-v02.api.letsencrypt.org/acme/acct/XXXXX"
```

## Verification Steps

Code review is enough

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
+++ b/internal/kafka/internal/services/kafkatlscertmgmt/kafka_tls_certificate_management_service.go
@@ -230,6 +230,7 @@ func createCertMagicClient(awsConfig *config.AWSConfig,
 		DisableTLSALPNChallenge: true,
 		DNS01Solver:             &certmagic.DNS01Solver{DNSProvider: provider},
 		CA:                      kafkaTLSCertificateManagementConfig.CertificateAuthorityEndpoint,
+		TestCA:                  kafkaTLSCertificateManagementConfig.CertificateAuthorityEndpoint,
 		AccountKeyPEM:           kafkaTLSCertificateManagementConfig.AutomaticCertificateManagementConfig.AcmeIssuerAccountKey,
 		Email:                   kafkaTLSCertificateManagementConfig.AutomaticCertificateManagementConfig.EmailToSendNotificationTo,
 	})


miguelsorianod:
Are we sure the changes of this PR are the correct approach?

Reading the documentation `TestCA` seems to be:
```
// TestCA is the endpoint of the directory for
	// an ACME CA to use to test domain validation,
	// but any certs obtained from this CA are
	// discarded
```

So it seems the idea is that Let's encrypt uses TestCA to perform domain validation, without really issuing any real certificate and then it uses `CA` to perform the real issuing.

machi1990:
> Are we sure the changes of this PR are the correct approach?

Yes.
 
> Reading the documentation `TestCA` seems to be:
> 
> ```
> // TestCA is the endpoint of the directory for
> 	// an ACME CA to use to test domain validation,
> 	// but any certs obtained from this CA are
> 	// discarded
> ```
> 
> So it seems the idea is that Let's encrypt uses TestCA to perform domain validation, without really issuing any real certificate and then it uses `CA` to perform the real issuing.

The TestCA is only used when a retry needs to be performed as far as I can tell from the code. It's a means to perform a sanity check end to end - domain validation among other things, a cert will be issued but it'll be dropped (not used) if  and only if the CA and TestCA are different. 
In all the cases, domain validation is performed against the CA.

miguelsorianod:
> > Are we sure the changes of this PR are the correct approach?
> 
> Yes.
> 
> > Reading the documentation `TestCA` seems to be:
> > ```
> > // TestCA is the endpoint of the directory for
> > 	// an ACME CA to use to test domain validation,
> > 	// but any certs obtained from this CA are
> > 	// discarded
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So it seems the idea is that Let's encrypt uses TestCA to perform domain validation, without really issuing any real certificate and then it uses `CA` to perform the real issuing.
> 
> The TestCA is only used when a retry needs to be performed as far as I can tell from the code. It's a means to perform a sanity check end to end - domain validation among other things, a cert will be issued but it'll be dropped (not used) if and only if the CA and TestCA are different. In all the cases, domain validation is performed against the CA.

Why can't we use stage for TestCA then?

machi1990:
> > > Are we sure the changes of this PR are the correct approach?
> > 
> > 
> > Yes.
> > > Reading the documentation `TestCA` seems to be:
> > > ```
> > > // TestCA is the endpoint of the directory for
> > > 	// an ACME CA to use to test domain validation,
> > > 	// but any certs obtained from this CA are
> > > 	// discarded
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So it seems the idea is that Let's encrypt uses TestCA to perform domain validation, without really issuing any real certificate and then it uses `CA` to perform the real issuing.
> > 
> > 
> > The TestCA is only used when a retry needs to be performed as far as I can tell from the code. It's a means to perform a sanity check end to end - domain validation among other things, a cert will be issued but it'll be dropped (not used) if and only if the CA and TestCA are different. In all the cases, domain validation is performed against the CA.
> 
> Why can't we use stage (for the sanity check) for TestCA then?

It's because the account we use doesn't exist there: this is the error that we are seeing.

miguelsorianod:
> > > > > Are we sure the changes of this PR are the correct approach?
> > > > 
> > > > 
> > > > Yes.
> > > > > Reading the documentation `TestCA` seems to be:
> > > > > ```
> > > > > // TestCA is the endpoint of the directory for
> > > > > 	// an ACME CA to use to test domain validation,
> > > > > 	// but any certs obtained from this CA are
> > > > > 	// discarded
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > So it seems the idea is that Let's encrypt uses TestCA to perform domain validation, without really issuing any real certificate and then it uses `CA` to perform the real issuing.
> > > > 
> > > > 
> > > > The TestCA is only used when a retry needs to be performed as far as I can tell from the code. It's a means to perform a sanity check end to end - domain validation among other things, a cert will be issued but it'll be dropped (not used) if and only if the CA and TestCA are different. In all the cases, domain validation is performed against the CA.
> > > 
> > > 
> > > Why can't we use stage (for the sanity check) for TestCA then?
> > 
> > 
> > It's because the account we use doesn't exist there: this is the error that we are seeing.
> 
> What account are you referring to? Is there a way that we can have it there? By default TestCA uses stage. So I would expect people don't change it that often to perform the sanity checks.

See this

machi1990:
>  What account are you referring to? 

The LE account.
 
 > Is there a way that we can have it there? 
 
 No we can't, the LE account we've is only for Prod
 
 > By default TestCA uses stage. So I would expect people don't change it that often to perform the sanity checks.

Stage can only be used if you've an account there or you generate accounts on the fly, which we don't.

miguelsorianod (approved):

machi1990:
> With the additional context I have after talking about it through other channels I see what's being done here. It is not ideal as we lose the capability of doing the process in the let's encrypt stage server to discard potential issues with the main one, but it is good enough for now. As we talked it might be nice to investigate in the future the possibility to use the same "account" for both let's encrypt stage and production or explore other possibilities.

+1 thanks for the review and additional feedback on potential improvement.

PR #1680, opened 2023-03-30
feat(certs): indicate that a Kafka can be paused for reconciliation in the data plane via the ManagedKafka CR annotation if we fail to fetch the certificate.

machi1990:
## Description

Indicate that a Kafka can be paused for reconciliation in the data plane via the ManagedKafka CR annotation if we fail to fetch the certificate.

Previously when we fail to fetch the certificate content of a given Kafka we'll return an error and fail the process of generating ManagedKafka of a given data plane. This means that we'd stop the reconciliation of all the Kafkas in that data plane as the Fleetshard sync was receiving an error each time. This PR with the help of
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/883 (an enhancement in the data plane) now makes it possible to indicate that a given ManagedKafka that has encountered an error can be paused paused from reconciliation. What this effectively means to the data plane is that:
- The local ManagedKafka in the data plane will continue to be seen as the source of truth
- The desired ManagedKafka that has been paused for reconciliation will be skipped and won't update/overwrite the local ManagedKafka. Unless otherwise it is marked for deletion, in which case the Kafka is going to be deleted.

In order to do this, we needed a way to indicate to the data plane that the MK can be paused for reconciliation: "managedkafka.bf2.org/pause-reconciliation".

We do this when an error occurs, which for now is when we fail to get the certificate of the Kafka. The data plane then won't reconcile the local state of the ManagedKafka. This ensures that at any time, we are able to reconcile all the Kafkas in a given cluster and not error when we fail to build the Managedkafka CR of a given Kafka - which can potential impact the reconciliation of others Kafkas.

Closes https://issues.redhat.com/browse/MGDSTRM-10919

~~This has to wait for https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/883 to be released.~~ it is now in Prod

## Verification Steps

Added unit test should pass
 
- [ ] Code Review completed


miguelsorianod:
What happens with the reported state of the managed kafka from the data plane side when the kafka is "paused for reconciliation"?

miguelsorianod:
If we mark the kafka as "paused for reconciliation"  who and when is that pause removed?

machi1990:
> What happens with the reported state of the managed kafka from the data plane side when the kafka is "paused for reconciliation"? What does KFM do in that case?

The state of the managed kafka is the one from the local MK in the data plane. If it was ready, then it'll remain ready. KFM continues to see this state as reported by the fleetshard sync.

> If we mark the kafka as "paused for reconciliation" who and when is that pause removed?

For case of certificates: it'll be up until the certificate can be fetched from the storage e.g in case of a partial AWS Secret Manager outage, it'll be up until the outage ends. In case the certificate doesn't already exist in the storage (e.g deleted due to triggered revocation or manually: this shouldn't happen but just so it happens), it'll be up until the next certificate reconcilation run: a max wait time of 3minutes i.e 30s reconciliation interval + up to 2 minutes for new cert to be generated)

miguelsorianod:
> > What happens with the reported state of the managed kafka from the data plane side when the kafka is "paused for reconciliation"? What does KFM do in that case?
> 
> The state of the managed kafka is the one from the local MK in the data plane. If it was ready, then it'll remain ready. KFM continues to see this state as reported by the fleetshard sync.
> 
> > If we mark the kafka as "paused for reconciliation" who and when is that pause removed?
> 
> For case of certificates: it'll be up until the certificate can be fetched from the storage e.g in case of a partial AWS Secret Manager outage, it'll be up until the outage ends. In case the certificate doesn't already exist in the storage (e.g deleted due to triggered revocation or manually: this shouldn't happen but just so it happens), it'll be up until the next certificate reconcilation run: a max wait time of 3minutes i.e 30s reconciliation interval + up to 2 minutes for new cert to be generated)

But who will remove the pause? On the control plane side I see if it is retrieve then we directly don't set the annotation. Does not setting the annotation equals to setting it explicitly to false? will data plane not seeing the annotation be interpreted as it should remove the pause?

machi1990:
> But who will remove the pause? 

Ah right. I see the question now. 

> On the control plane side I see if it is retrieve then we directly don't set the annotation. Does not setting the annotation equals to setting it explicitly to false? will data plane not seeing the annotation be interpreted as it should remove the pause?

Yes, not setting the annotation explicit means that it is not paused; It is the behavior implemented and honored in the data plane as well. The MK CR will only be paused if the annotation is set to `true`. Otherwise, it'll be reconciled.

miguelsorianod (approved):
:+1:  looks good to me

PR #1682, opened 2023-03-30
chore: remove unused Observatorium Red Hat SSO CLI flags

miguelsorianod:
## Description

With the changes from https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1647 the following CLI flags are not used anywhere anymore, as they are now set in the dataplane observability configuration file (in case it is enabled):
* observability-red-hat-sso-auth-server-url
* observability-red-hat-sso-realm

This PR takes care of removing them. This simplifies and makes less confusing the area around observatorium configuration.

It also takes care of removing all the related code, OCP templates and Makefile contents that are associated/related to those flags.

Additional remarks:
* This will need a post-cleanup of the KFM Stage and Production environments OCP template parameters
* This needs some changes in the managed-kafka-versions repository. I think they can be done after merging this PR so we should be good merging it once verified.
* There is an additional CLI flag --observatorium-auth-type that with the removal of the dex authentication support only accepts the redhat value (for Red Hat SSO based auth). Changes around that will be dealt in another PR
* KAS Installer does not need any changes related to this

## Verification Steps
* KAS Fleet Manager can run correctly
* Tests pass

- [ ] Code Review completed


pkg/client/observatorium/observability_config.go
```diff
@@ -22,9 +22,13 @@ const (
 
 type ObservabilityConfiguration struct {
 	// Red Hat SSO configuration
-	RedHatSsoAuthServerUrl     string `json:"redhat_sso_auth_server_url" yaml:"redhat_sso_auth_server_url"`
-	RedHatSsoRealm             string `json:"redhat_sso_realm" yaml:"redhat_sso_realm"`
-	RedHatSsoTenant            string `json:"redhat_sso_tenant" yaml:"redhat_sso_tenant"`
+	//// RedHatSsoTenant is used to specify the Observatorium tenant for
+	//// the Data Plane Remote Write configuration. It is not used for
+	//// the control plane connection against Observatorium
+	RedHatSsoTenant string `json:"redhat_sso_tenant" yaml:"redhat_sso_tenant"`
```
miguelsorianod, line 28: I also added a commit to clarify the usage of the two remaining parameters.
As a future improvement I think refactoring this parameters might be a good idea: putting them into its own struct, renaming them or rethinking how we expose them, etc.

machi1990: +1 makes sense

templates/service-template.yml
```diff
@@ -231,21 +231,11 @@ parameters:
   description: This is the amount of time before a leader lease expires.
   value: "1m"
 
-- name: OBSERVATORIUM_RHSSO_REALM
```
machi1990, line 234: My assumption and hope is that this is not used anywhere (in kas-installer and its dependent) I'll approve based on that.

templates/service-template.yml
```diff
@@ -231,21 +231,11 @@ parameters:
...
-  displayName: Observatorium Red Hat SSO realm
-  description: Observatorium Red Hat SSO realm for observability stack.
-  value: ""
-
 - name: OBSERVATORIUM_RHSSO_TENANT
   displayName: Observatorium Red Hat SSO tenant
   description: Observatorium Red Hat SSO tenant for observability stack.
   value: ""
 
-- name: OBSERVATORIUM_RHSSO_AUTH_SERVER_URL
```
machi1990, line 244: same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1682/files#r1153062704

machi1990 (approved):

templates/service-template.yml
miguelsorianod: See the additional remarks in the original description

templates/service-template.yml
miguelsorianod: See the additional remarks in the original description

PR #1686, opened 2023-03-30
MGDCTRS-2024: Support Smart Event Processors in the Control Plane

manstis:
See [https://issues.redhat.com/browse/MGDCTRS-2024](https://issues.redhat.com/browse/MGDCTRS-2024)

## Description
Add support for Smart Event `Processors` to `cos-fleet-manager`

## Verification Steps

1. Fetch this PR
2. Fetch `cos-fleet-manager`
3. Patch `cos-fleet-manager` to use this version of `kas-fleet-manager` by adding this to its `go.mod` file:
`replace github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager v0.0.0-20230120171214-d0c418df2eee => ../kas-fleet-manager`
4. Run `cos-fleet-manager`
5. Make a Cluster
6. Make a Namespace
7. Run `make run/docs` in `kas-fleet-manager`
8. See the new `Processor` endpoints. `Get`, `List`, `Create` and `Delete` are supported up to and including the database.

- [ ] ~Documentation added for the feature~
- [ ] Code Review completed
- [ ] ~Required metrics/dashboards/alerts have been added (or PR created)~
- [ ] ~Required Standard Operating Procedure (SOP) is added~


manstis:
@robpblake @danielezonca @lucamolteni @vaibhavjainwiz This is pretty ready for review now.

@lburgazzoli @valdar IDK whether somebody familiar with RHOC CP should review?

vaibhavjainwiz:
AFAIK [kas-fleet-manager](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager) is shared codebase between RHOSAK and RHOC. The processor functionality is required just for `RHOC`. I think Processor related code should be in [cos-fleet-manager](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleet-manager) repo

lucamolteni:
> AFAIK [kas-fleet-manager](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager) is shared codebase between RHOSAK and RHOC. The processor functionality is required just for `RHOC`. I think Processor related code should be in [cos-fleet-manager](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleet-manager) repo

A proper separation between the two would be nice to have but currently we don't have it, all RHOC code is still in `kas-fleet-manager`. I wouldn't start from this to separate the twos 

@lburgazzoli  see ^ as well

internal/connector/internal/api/dbapi/processor.go
```diff
@@ -0,0 +1,117 @@
...
+	string(ProcessorStatusPhaseFailed),
+	string(ProcessorStatusPhaseDeleted),
+}
+
+type Processor struct {
+	db.Model
+	Name            string
+	NamespaceId     string
+	ProcessorTypeId string
+	Channel         string
```
robpblake, line 48: @manstis Apologies if we have already discussed this, but do you know what is the status of support for `channel` on Connector and if it's something that is being actively used?

internal/connector/internal/api/dbapi/processor.go
```diff
@@ -0,0 +1,117 @@
...
+	DesiredState    ProcessorDesiredState
+	Owner           string
+	CloudProvider   string
```
robpblake, line 51: I'm not sure I understand how `CloudProvider`, `Region` and `MultiAZ` work for Processors 🤔  The `namespaceId` will reference a Connector Cluster that is pre-deployed by the user, so at least the `CloudProvider` and `Region` are already set and essentially inherited from the underlying OSD on which the namespace is "hosted".

`MultiAZ` will again depend on the capabilities of the underlying OSD hosting the namespace and FWICS, we don't actually have that information about the Connector Cluster (I could be completely missing something as I'm just getting used to Go and this codebase).

internal/connector/internal/api/private/api_connector_clusters_agent.go
```diff
@@ -499,6 +499,240 @@ func (a *ConnectorClustersAgentApiService) GetClusterAsignedConnectorNamespaces(
...
+/*
+GetClusterAssignedProcessorDeploymentById Returns a processor deployment assigned to the cluster.
+Returns a processor deployment assigned to the cluster.
+  - @param ctx _context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+  - @param connectorClusterId The id of the connector cluster
+  - @param processorDeploymentId The id of the processor deployment
+
+@return ProcessorDeployment
+*/
+func (a *ConnectorClustersAgentApiService) GetClusterAssignedProcessorDeploymentById(ctx _context.Context, connectorClusterId string, processorDeploymentId string) (ProcessorDeployment, *_nethttp.Response, error) {
```
robpblake, line 511: @manstis What's the use-case for retrieving a specific `ProcessorDeployment` by `connectorClusterId` and `processorDeploymentId`?

internal/connector/internal/handlers/processor_secrets.go
```diff
@@ -0,0 +1,36 @@
+package handlers
+
+import (
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/api/dbapi"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/connector/internal/services/vault"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/api"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/errors"
+)
+
+const OwningProcessorResourcePrefix = "/v1/processor/"
```
robpblake, line 10: @manstis Should this be `v2Alpha1` to align with the path under which we are exposing the Processor API?

internal/connector/internal/api/dbapi/processor.go
manstis: Support for different Connector `channel`'s was added recently [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/commit/6e3c3d817dc4b4dab4a95937b877722bc2c637ae). I don't know how actively used it is. We could choose to remove this field for `Processor`'s (for now) but I suspect it may follow.

internal/connector/internal/api/dbapi/processor.go
manstis: I can remove them. I will wait for all feedback to be assembled before making changes.

internal/connector/internal/handlers/processors.go
```diff
@@ -0,0 +1,421 @@
...
+				handlers.Validation("name", &resource.Name, handlers.WithDefault("New Processor"), handlers.MinLen(1), handlers.MaxLen(100)),
+				handlers.Validation("service_account.client_id", &resource.ServiceAccount.ClientId, handlers.MinLen(1)),
+				handlers.Validation("desired_state", (*string)(&resource.DesiredState), handlers.WithDefault("ready"), handlers.IsOneOf(dbapi.ValidProcessorDesiredStates...)),
+				validateProcessorImmutableProperties(patch, originalResource),
+				validatePatchAnnotations(resource.Annotations, originalResource.Annotations),
+				validateProcessor(&resource),
+			}
+
+			// Don't validate user's tenancy in admin api calls
+			// TODO {manstis} Processors are not exposed under the admin API at the moment
```
robpblake, line 265: @manstis Out of curiosity, how much work would it be to expose this under the admin API? That would give us the ability to stop/delete Processors that are behaving badly.

internal/connector/internal/api/private/api_connector_clusters_agent.go
manstis: `Processors` are deployed to the same `namespace`/`cluster` as the `Connector`'s to which they can integrate.

There is an API to retrieve `Connector`'s by `clusterId` so I thought it *might* be useful for `Processor`'s too.

I can remove every such method if the generated (app-services) API is out of scope (leaving only the methods needed for the Data Plane/Control Plane sync). Happy to do so just let me know your thoughts.

internal/connector/internal/handlers/processor_secrets.go
manstis: Yes it should. Good spot! I will change.

internal/connector/internal/api/private/api_connector_clusters_agent.go
robpblake:
```
Processors are deployed to the same namespace/cluster as the Connector's to which they can integrate.
```
Do you mean "can be" deployed to the same namespace/cluster? Kafka is the intermediary, so there is nothing from stopping a user from creating a `namespace`  on a completely different Connector Cluster and deploying their `Processors` there. As long as the Processor produces/consumes to the correct Kafka topics, Connectors running in completely different Connector clusters can also consume/produce to the same topics and things should "just work"™. 

```
Happy to do so just let me know your thoughts.
```

I'm sure there must be a use-case for the corresponding Connector method, but I'm not sure what it is yet. Perhaps lets check the synchronizer code to see what it's used for.

internal/connector/internal/config/processors.go
```diff
@@ -0,0 +1,27 @@
...
+		ProcessorsEnabled: false,
+	}
+}
+
+func (c *ProcessorsConfig) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&c.ProcessorsEnabled, "processors-enabled", c.ProcessorsEnabled, "Enable support for (Smart Event) Processors")
+}
+
+func (c *ProcessorsConfig) ReadFiles() error {
+	// There ae no external configuration files for Processors
```
robpblake, line 25: @manstis We'll need some form of configuration for Processors in terms of revision, container image and perhaps channels. We've spoken briefly in the chat about this, but I think exact structure will depend on whether we see value in supporting `channel` for Processors from the start. Happy to sync offline about the structure so that we can make a final decision 👍

internal/connector/internal/api/private/api_connector_clusters_agent.go
manstis: > I'm sure there must be a use-case for the corresponding Connector method, but I'm not sure what it is yet. 

The generated Connectors method is not used in the RHOC codebase.

IDK if there's another repository with a CLI client?

manstis: It's not used by {{cos-tools}} (the repository for `rhoc cli`) or `cos-fleetshard`... 

I think it's safe to say it might be redundant.

internal/connector/internal/api/dbapi/processor.go
danielezonca: I agree that `CloudProvier`, `Region` and `MultiAZ` are tied to the namespace/OSD so we can remove them from here

internal/connector/internal/handlers/processors.go
manstis: I'd be tempted to avoid scope creep to be honest.

These PRs are large enough already. Create a JIRA to consider a _consolidated_ Admin API might be a better option.

internal/connector/internal/api/dbapi/processor.go
manstis: Removed.

internal/connector/internal/api/private/api_connector_clusters_agent.go
lburgazzoli: I'm nut 100% sure as naming is not the same but for connectors, there is a rhoc subcommand to get a deployment by id: https://github.com/bf2fc6cc711aee1a0c2a/cos-tools/blob/main/rhoc/pkg/cmd/deployments/get/get.go

internal/connector/internal/config/processors.go
lburgazzoli: channels, images and co will come from a catalog at some point right ?

internal/connector/internal/api/dbapi/processor.go
lburgazzoli: yes they are used now, primary as a gate for filtering out connectors depending on the maturity level so i.e. come connectors won't be available in prod but they will be in dev/staging. Same thinking can be applied to preview conenctors at some point

internal/connector/internal/config/processors.go
robpblake: @lburgazzoli Yep, that's the plan 👍 We're trying to be as closely aligned to the Connector catalog concept as possible, even though we only have a single Processor type right now.

internal/connector/internal/handlers/processors.go
robpblake: Sure thing 👍

internal/connector/internal/handlers/processor_secrets.go
```diff
@@ -0,0 +1,36 @@
...
+		resource.ServiceAccount.ClientSecretRef = keyId
+	}
+	return nil
+}
+func getProcessorSecretRefs(resource *dbapi.Processor) (result []string, err error) {
+	if resource.ServiceAccount.ClientSecretRef != "" {
+		result = append(result, resource.ServiceAccount.ClientSecretRef)
+	}
+	return
+}
```
ziccardi, line 36: This function never returns errors

internal/connector/internal/services/processor_deployment.go
```diff
@@ -0,0 +1,249 @@
...
+		dbConn = dbConn.Where("\"Status\".\"upgrade_available\"")
+	}
+	if includeDanglingDeploymentsOnly {
+		dbConn = dbConn.Where("\"Processor\".\"deleted_at\" IS NOT NULL")
+	} else {
+		dbConn = dbConn.Where("\"Processor\".\"deleted_at\" IS NULL")
+	}
+
+	// Apply search query
+	if len(listArgs.Search) > 0 {
```
ziccardi, line 153:
```suggestion
	if !shared.StringEmpty(listArgs.Search) {
```

internal/connector/internal/api/private/api_connector_clusters_agent.go
manstis: See https://issues.redhat.com/browse/MGDCTRS-2269

internal/connector/internal/handlers/annotations.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		return nil
+	}
+}
+
+// validateAnnotations returns an error for invalid k8s annotations, or reserved domains
+func validateAnnotations(annotations map[string]string) *errors.ServiceError {
+	for k, v := range annotations {
+		errs := validation.IsQualifiedName(k)
+		if len(errs) != 0 {
```
ziccardi, line 37:
```suggestion
		if !arrays.IsEmpty(errs) {
```

internal/connector/internal/handlers/annotations.go
```diff
@@ -0,0 +1,68 @@
...
+			return errors.BadRequest("invalid annotation key %s: %s", k, strings.Join(errs, "; "))
+		}
+		errs = validation.IsValidLabelValue(v)
+		if len(errs) != 0 {
```
ziccardi, line 41:
```suggestion
		if !arrays.IsEmpty(errs) {
```

internal/connector/internal/handlers/annotations.go
```diff
@@ -0,0 +1,68 @@
...
+			return errors.BadRequest("invalid annotation value %s: %s", v, strings.Join(errs, "; "))
+		}
+		for _, d := range reservedDomains {
+			if strings.Contains(k, d) {
+				return errors.BadRequest("cannot use reserved annotation %s from domain %s", k, d)
+			}
+		}
```
ziccardi, line 48:
```suggestion
		if arrays.AnyMatch(reservedDomains, func(d string) bool { return strings.Contains(k, d)}) {
			return errors.BadRequest("cannot use reserved annotation %s from domain %s", k, d)
		}
```

internal/connector/internal/handlers/processor_secrets.go
manstis: Yes, this is true (and also true of the method [`stripProcessorSecretReferences(..)`](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1686/files#diff-b835efb128f9125006ba8a2c048a61d629283d3f0727169a30574dbc0fdb3bd5R12)).

At this point in time the definition of a `Processor` does not contain secret information (or better said we don't know if it will need to contain secret information) and therefore the two methods here are very basic. However if we need to make the secret handling more complicated it is probable these methods could need to return errors in the future.

I am therefore "future proofing" the code. Less so in this file itself; but in the places where these methods are invoked and we need to handle errors if/should they arise (in the future).

internal/connector/internal/handlers/processor_secrets.go
```diff
@@ -0,0 +1,36 @@
...
+func stripProcessorSecretReferences(resource *dbapi.Processor) *errors.ServiceError {
+	// clear out secrets..
+	resource.ServiceAccount.ClientSecret = ""
+	resource.ServiceAccount.ClientSecretRef = ""
+	return nil
+}
+
+func moveProcessorSecretsToVault(resource *dbapi.Processor, vault vault.VaultService) *errors.ServiceError {
+	// move secrets to a vault.
+	if resource.ServiceAccount.ClientSecret != "" {
```
ziccardi, line 21:
```suggestion
	if !shared.StringEmpty(resource.ServiceAccount.ClientSecret) {
```

internal/connector/internal/handlers/processor_secrets_test.go
```diff
@@ -0,0 +1,70 @@
...
+			ClientSecret:    "client-secret",
+			ClientSecretRef: "client-secret-ref",
+		},
+	}
+
+	_ = stripProcessorSecretReferences(&resource)
+
+	g.Expect(resource.ServiceAccount.ClientId).To(gomega.Equal("client-id"))
+	g.Expect(resource.ServiceAccount.ClientSecret).To(gomega.Equal(""))
+	g.Expect(resource.ServiceAccount.ClientSecretRef).To(gomega.Equal(""))
```
ziccardi, line 24:
```suggestion
	g.Expect(resource.ServiceAccount.ClientSecret).To(gomega.BeEmpty())
	g.Expect(resource.ServiceAccount.ClientSecretRef).To(gomega.BeEmpty())
```

internal/connector/internal/handlers/processor_secrets.go
ziccardi: In this case I would avoid a `naked return` and would use an explicit one to make it clear that we are always returning `nil`:
```go
return result, nil
```

internal/connector/internal/handlers/annotations.go
manstis: This does not work as we loose visibility of the `d` parameter for the error message.

In your proposal `d` is scoped to the anonymous function.

ziccardi: We could use this in this case:
```go
		if idx, d := arrays.FindFirst(reservedDomains, func(d string) bool { return strings.Contains(k, d) }); idx != -1 {
			return errors.BadRequest("cannot use reserved annotation %s from domain %s", k, d)
		}
```

manstis: We could do; however I'm reluctant to change the way we iterate a collection when there appears to be no real value added? Both ultimately use `strings.Contains(k, d)` and the implementation of `arrays.FindFirst(..)` uses `range` anyway too. Is there a compelling reason to make this change?

ziccardi: The only reason I see is that `FindFirst` makes the loop's intent clear at first sight.
But I'm happy to stick with the range loop.

manstis:
@lburgazzoli Another `token` related issue:
```
[2023-04-13T15:24:49.104Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
[2023-04-13T15:24:49.105Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
    at main (/snapshot/repo/dist/src/index.js)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
[2023-04-13T15:24:49.105Z] ['verbose'] End of uploader: 750 milliseconds
Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255
```

sutaakar (approved):
Looks good to me (was checking mostly tests)

internal/connector/internal/config/processors.go
manstis: Changes made to import a "Processor Catalog" using the same approach as taken by Connectors.

manstis:
@danielezonca @ziccardi @robpblake @lburgazzoli 

All changes emanating from your first reviews are complete.

The PR has as good as (if not better) test coverage than what exists for Connectors.

Could you please take another look so we can consider this for merging soon?

test/cucumber/http_request.go
```diff
@@ -251,7 +252,7 @@ func (s *TestScenario) iWaitUpToSecondsForAGETOnPathResponseSelectionToMatch(tim
 
 		select {
 		case <-ctx.Done():
-			return nil
+			return errors.NewTimeoutError(fmt.Sprintf("Wait timed-out looking for '%s' to match '%s'.", selection, expected), 0)
```
manstis, line 255: If an assertion timed out it used to pass.. it will now fail as timed out.

internal/connector/internal/handlers/processors.go
manstis: See https://issues.redhat.com/browse/MGDCTRS-2291

manstis:
RHOC has been cancelled.

PR #1687, opened 2023-03-30
feat: add distributed lock utility

ziccardi:
## Description
Currently, we already have a distributed lock mechanism based on the “SELECT ... FOR UPDATE” statement in the leader election code. 
However, this approach had one limitation: the lock table had to be prepopulated with the list of locks we wanted to support.

This new utility addresses this limitation by using a different approach. Instead of the “SELECT ... FOR UPDATE” statement, it uses the “INSERT INTO ... ON CONFLICT DO NOTHING” statement. This allows us to release the lock by simply sending a rollback to close the transaction and discard the row.

This new approach has several advantages:
1. There is no need to prepopulate the lock table.
2. The lock table does not grow indefinitely.
3. The lock ID can be generated dynamically at runtime.

While it is possible to achieve dynamic lock IDs with the “SELECT ... FOR UPDATE” approach, it would require multiple statements. With our new approach, everything is done with a single statement.

## WHAT
This update includes the addition of a new utility for managing distributed locks using the database. Additionally, the secure storage of certificates has been updated to utilize this new distributed lock mechanism.

## HOW
This pull request introduces two new objects: `DistributedLock` and `DistributedLockMgr`.

The `DistributedLock` object functions similarly to a `sync.Mutex`, allowing for locking and unlocking. The `DistributedLockMgr` object manages multiple locks with different IDs, which is useful when dealing with dynamically generated lock IDs.

The `DistributedLock` object has two primary methods: `Lock` and `Unlock`. The `Lock` method performs an `INSERT INTO distributed_locks (ID) VALUES (?) ON CONFLICT DO NOTHING` statement, which creates a row if it does not already exist or locks it if it does. 
The `Unlock `method simply rolls back the transaction, discarding the row and removing the lock.

## Verification Steps

Provided tests and integration tests should pass.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
+++ b/internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
@@ -3,10 +3,6 @@ package kafkatlscertmgmt
 import (
 	"context"
 	"fmt"
-	"io/fs"
-	"strings"
-	"sync"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -15,7 +11,11 @@ import (
 	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
 	"github.com/aws/aws-secretsmanager-caching-go/secretcache"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/internal/kafka/internal/config"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/sync"
 	"github.com/caddyserver/certmagic"
+	"io/fs"
+	"strings"
 )
 
 //go:generate moq -out secretmanager_client_moq.go . SecretManagerClient
@@ -24,14 +24,13 @@ type SecretManagerClient = secretsmanageriface.SecretsManagerAPI
 var _ certmagic.Storage = &secureStorage{}
 
 type secureStorage struct {
-	secretPrefix     string
-	lock             sync.Mutex
-	secretCache      *secretcache.Cache
-	storageItemLocks map[string]*sync.Mutex
-	secretClient     SecretManagerClient
+	secretPrefix string
+	secretCache  *secretcache.Cache
+	secretClient SecretManagerClient
+	lock         sync.DistributedLockMgr
 }
 
-func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementConfig config.AutomaticCertificateManagementConfig) (*secureStorage, error) {
+func newSecureStorage(connectionFactory *db.ConnectionFactory, config *config.AWSConfig, automaticCertificateManagementConfig config.AutomaticCertificateManagementConfig) (*secureStorage, error) {
 	awsConfig := &aws.Config{
 		Credentials: credentials.NewStaticCredentials(
 			config.SecretManager.AccessKey,
@@ -55,41 +54,23 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
 		return nil, err
 	}
 	return &secureStorage{
-		secretClient:     secretClient,
-		secretCache:      secretCache,
-		lock:             sync.Mutex{},
-		storageItemLocks: map[string]*sync.Mutex{},
-		secretPrefix:     config.SecretManager.SecretPrefix,
+		secretClient: secretClient,
+		secretCache:  secretCache,
+		secretPrefix: config.SecretManager.SecretPrefix,
+		lock:         sync.NewDistributedLockMgr(connectionFactory.New()),
 	}, nil
 }
 
 // Lock blocks until it can acquire the name-specific lock.
 // The process of obtaining a name-specific lock is atomic, as required per the certmagic.Locker
 // interface documentation: https://pkg.go.dev/github.com/caddyserver/certmagic#Locker.
-// NOTE: The intent here is to not implement a distributed locking mechanism but an in-process locking mechanism good enough for
-// our certificate management logic which happens within one reconciliater leader.
+// NOTE: A distributed lock is implemented by leveraging the database lock
 func (storage *secureStorage) Lock(ctx context.Context, key string) error {
-	storage.lock.Lock()
-	mu, ok := storage.storageItemLocks[key]
-	if !ok {
-		// initialize the storage item lock the first time if it does not exist yet
-		mu = &sync.Mutex{}
-		storage.storageItemLocks[key] = mu
-	}
-	storage.lock.Unlock()
-	mu.Lock()
-	return nil
-
+	return storage.lock.Lock(key)
 }
 
 func (storage *secureStorage) Unlock(ctx context.Context, key string) error {
-	mu, ok := storage.storageItemLocks[key]
-	if !ok {
-		return fs.ErrNotExist
-	}
-
-	mu.Unlock()
-	return nil
+	return storage.lock.Unlock(key)
 }
 
 func (storage *secureStorage) Store(ctx context.Context, key string, value []byte) error {
--- a/pkg/shared/utils/sync/distributed_lock.go
+++ b/pkg/shared/utils/sync/distributed_lock.go
@@ -1,0 +1,61 @@
+package sync
+
+import (
+	"github.com/pkg/errors"
+	"gorm.io/gorm"
+)
+
+// DistributedLock manages locks on the database through the usage of the `distributed_lock` table.
+// WARNING: a lock can only be released by the instance that initially created it.
+type DistributedLock interface {
+	Lock() error
+	Unlock() error
+	isLocked() bool
+}
+
+var _ DistributedLock = &distributedLock{}
+
+// NewDistributedLock returns a new instance of DistributedLock
+// db - the connection to the database
+// lockID - the ID of the lock. When multiple threads request to use the Lock function with the same lock ID, only the first thread is granted access while the remaining threads must wait until the lock is released before proceeding.
+func NewDistributedLock(db *gorm.DB, lockID string) DistributedLock {
+	return &distributedLock{db: db, id: lockID}
+}
+
+type distributedLock struct {
+	db     *gorm.DB
+	tx     *gorm.DB
+	id     string
+	locked bool
+}
+
+// Lock locks dl. If the lock is already in use, the caller blocks until the lock is available.
+func (dl *distributedLock) Lock() error {
+	dl.tx = dl.db.Begin()
+	result := dl.tx.Exec("INSERT INTO distributed_locks (ID) VALUES (?) ON CONFLICT DO NOTHING", dl.id)
+	if result.Error != nil {
+		dl.tx.Rollback() // close the transaction
+		dl.tx = nil
+		return errors.Wrapf(result.Error, "unable to acquire lock: %v", result.Error)
+	}
+	dl.locked = true
+	return nil
+}
+
+// Unlock unlock dl.
+func (dl *distributedLock) Unlock() error {
+	if dl.tx == nil {
+		return errors.Errorf("lock not acquired")
+	}
+
+	result := dl.tx.Rollback()
+	if result.Error != nil {
+		return errors.Wrapf(result.Error, "unable to release lock: %v", result.Error)
+	}
+	dl.locked = false
+	return nil
+}
+
+func (dl *distributedLock) isLocked() bool {
+	return dl.locked
+}
--- a/pkg/shared/utils/sync/distributed_lock.go
+++ b/pkg/shared/utils/sync/distributed_lock.go
@@ -6,6 +6,8 @@ import (
 )
 
 // DistributedLock manages locks on the database through the usage of the `distributed_lock` table.
+// It works similarly to a sync.Mutex, allowing for locking and unlocking.
+// The lock is acquired by executing an "INSERT INTO ... ON CONFLICT DO NOTHING" command on the database.
 // WARNING: a lock can only be released by the instance that initially created it.
 type DistributedLock interface {
 	Lock() error
--- a/pkg/shared/utils/sync/distributed_lock_mgr.go
+++ b/pkg/shared/utils/sync/distributed_lock_mgr.go
@@ -1,0 +1,69 @@
+package sync
+
+import (
+	"fmt"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/arrays"
+	"gorm.io/gorm"
+	"sync"
+)
+
+var mgrLock sync.Mutex
+
+var _ DistributedLockMgr = &distributedLockMgr{}
+
+func NewDistributedLockMgr(db *gorm.DB) DistributedLockMgr {
+	return &distributedLockMgr{
+		db:           db,
+		locks:        map[string][]DistributedLock{},
+		createLockFn: NewDistributedLock,
+	}
+}
+
+// DistributedLockMgr manages a list of DistributedLock objects. The DistributedLock object is automatically removed
+// from the list when the Unlock method is called.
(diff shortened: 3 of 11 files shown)


pkg/shared/utils/sync/distributed_lock_mock.go
```diff
@@ -0,0 +1,35 @@
...
+	db *gorm.DB
+	id string
+}
+
+func (dl *distributedLockMock) Lock() error {
+	return dl.dl.Lock()
+}
+
+func (dl *distributedLockMock) Unlock() error {
+	dl.dl.Unlock()
```
machi1990, line 29: this is causing a linting issue https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/actions/runs/4566392609/jobs/8058829565

pkg/shared/utils/sync/distributed_lock.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/pkg/errors"
+	"gorm.io/gorm"
+)
+
+// DistributedLock manages locks on the database through the usage of the `distributed_lock` table.
+// WARNING: a lock can be released only by the same instance that created it.
+type DistributedLock interface {
+	Lock() error
+	Unlock() error
+}
```
machi1990, line 13: Can we extend of this PR onto using the lock over here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/ed0c44234c8ff98dbe2be35e3bd42b5b52aa69bc/internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go#L66-L93 ? Thanks

pkg/shared/utils/sync/distributed_lock.go
```diff
@@ -0,0 +1,54 @@
+package sync
+
+import (
+	"github.com/pkg/errors"
+	"gorm.io/gorm"
+)
+
+// DistributedLock manages locks on the database through the usage of the `distributed_lock` table.
+// WARNING: a lock can be released only by the same instance that created it.
```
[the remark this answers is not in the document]
ziccardi: When you call the `lock` it starts a new transaction and `locks the row`.
When you call the `unlock` the lock is removed by ending the transaction (rollback).

If we perform a new version deployment, the pod is stopped and all its connections and transactions are dropped.

pkg/shared/utils/sync/distributed_lock.go
machi1990: The integration could be a separate commit if we prefer

pkg/shared/utils/sync/distributed_lock.go
```diff
@@ -0,0 +1,54 @@
...
+type distributedLock struct {
+	db *gorm.DB
+	tx *gorm.DB
+	id string
+}
+
+// Lock create the lock. If this lock is already locked, waits for the lock to be released
+func (dl *distributedLock) Lock() error {
+	dl.tx = dl.db.Begin()
+	result := dl.tx.Exec("INSERT INTO distributed_locks (ID) VALUES (?) ON CONFLICT DO NOTHING", dl.id)
```
miguelsorianod, line 33: What kind of lock is acquired with the insert into?
Have we checked that other transactions can read the table without blocking when looking for another ID 'y', at the same time that a transaction is currently locking ID 'x'?
Using the pg_lock table might help to figure out what kind of lock is explicitly being used.
I am aware that there's a unit test where the intention is to test that case but it does not really use the database so the real behavior is not being tested. For that an integration test might be more appropriate.

ziccardi: There are both unit and integration tests. An integration test has already been added to test on a real database.

> Have we checked that other transactions can read the table without blocking when looking for another ID 'y', at the same time that a transaction is currently locking ID 'x'?

Yes. No read is ever done into the table. Only `INSERT` is done and is rollbacked when the lock is unlocked.
Only a write lock is created for the newly inserted row.

pkg/shared/utils/sync/distributed_lock.go
ziccardi: Done!

pkg/shared/utils/sync/distributed_lock.go
ziccardi: Added a new object `DistributedLockMgr`. Integration tests are [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1687/files#diff-4ccf384f6c30940554c1750d115d0684af071e71dfec7578b8e7c1d25557a763).

This new utility allows to easily manage different locks with different lock IDs.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -55,41 +53,22 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
...
-	if !ok {
-		// initialize the storage item lock the first time if it does not exist yet
-		mu = &sync.Mutex{}
-		storage.storageItemLocks[key] = mu
-	}
-	storage.lock.Unlock()
-	mu.Lock()
-	return nil
-
+	return storage.lock.Lock(key)
```
machi1990, line 67: `storage.lock` is never initialized, this line will panic.

internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go
```diff
@@ -55,41 +53,22 @@ func newSecureStorage(config *config.AWSConfig, automaticCertificateManagementCo
...
 
 func (storage *secureStorage) Unlock(ctx context.Context, key string) error {
-	mu, ok := storage.storageItemLocks[key]
-	if !ok {
-		return fs.ErrNotExist
-	}
-
-	mu.Unlock()
-	return nil
+	return storage.lock.Unlock(key)
```
machi1990, line 71: same comment as https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1687/files#r1155607649

pkg/shared/utils/sync/distributed_lock.go
machi1990: Thanks, current checking it out

machi1990 (changes requested):
I've verified that Lock and Unlock works when obtaining/renewing certificates.
Lock only blocks the lock key of a given certificate if it was already locked: which is what we want.
Unlock release the lock as expected as well.

Thank you for the great work @ziccardi. However, to perform my verification I had to fix a panic error which I've pointed out in:
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1687/files#r1155607649

I'll hold on my approval until the fix for that is pushed.

pkg/shared/utils/sync/distributed_lock.go
```diff
@@ -0,0 +1,61 @@
...
+// WARNING: a lock can only be released by the instance that initially created it.
+type DistributedLock interface {
+	Lock() error
+	Unlock() error
+	isLocked() bool
+}
+
+var _ DistributedLock = &distributedLock{}
+
+// NewDistributedLock returns a new instance of DistributedLock
```
miguelsorianod, line 18: Can we add more details on the behavior of this as explained in the description of this PR?

pkg/shared/utils/sync/distributed_lock_mock.go
```diff
@@ -0,0 +1,39 @@
...
+
+// It appears that go-mocket does not offer a straightforward method to intercept the Rollback call. To make it easily
+// interceptable, this mock substitutes the rollback call in the Unlock method with a DELETE.
+
+func newDistributedLockMock(db *gorm.DB, lockID string) DistributedLock {
+	dl := NewDistributedLock(db, lockID)
+	return &distributedLockMock{dl: dl, db: db, id: lockID}
+}
+
+type distributedLockMock struct {
```
miguelsorianod, line 18: Does this work with a mocked database? My understanding is that that's the case
Maybe it would be good to explicitly mention that for potential users of the type. What do you think?

pkg/shared/utils/sync/distributed_lock_mock.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/pkg/errors"
+	"gorm.io/gorm"
+)
+
+var _ DistributedLock = &distributedLockMock{}
+
+// It appears that go-mocket does not offer a straightforward method to intercept the Rollback call. To make it easily
+// interceptable, this mock substitutes the rollback call in the Unlock method with a DELETE.
```
miguelsorianod, line 11: https://github.com/Selvatico/go-mocket/blob/master/stmt.go#L195 does this help?

pkg/shared/utils/sync/distributed_lock_mgr.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+var mgrLock sync.Mutex
+
+var _ DistributedLockMgr = &distributedLockMgr{}
+
+func NewDistributedLockMgr(db *gorm.DB) DistributedLockMgr {
+	return &distributedLockMgr{
+		db:           db,
+		locks:        map[string][]DistributedLock{},
```
miguelsorianod, line 17: Why is there a list of DistributedLock for each LockID?

pkg/shared/utils/sync/distributed_lock_mgr.go
```diff
@@ -0,0 +1,69 @@
...
+		createLockFn: NewDistributedLock,
+	}
+}
+
+// DistributedLockMgr manages a list of DistributedLock objects. The DistributedLock object is automatically removed
+// from the list when the Unlock method is called.
+type DistributedLockMgr interface {
```
miguelsorianod, line 24: What is the advantage of using this over the DistributedLock type?
That type already blocks when trying to get a specific lockID

pkg/shared/utils/sync/distributed_lock_mock.go
ziccardi: Not sure. The comment says it is used to simulate broken connections

pkg/shared/utils/sync/distributed_lock_mgr.go
ziccardi: Each thread must use a distinct `DistributedLock` to hold a separate transaction that waits for the lock’s release. Using a single `DistributedLock` object would result in an `Unlock` call unblocking all threads.

pkg/shared/utils/sync/distributed_lock_mgr.go
ziccardi: As discussed previously, locks must be released by the `DistributedLock` instance that created them.

In the `secure storage`, two methods handle `lock` and `unlock` operations for a specific `lockID`. That means that without `DistributedLockMgr`, all the logic for managing the `DistributedLock` list would need to be implemented in the `secure lock`. 

Since this is a common scenario, I think using `DistributedLockMgr` is a better solution.

pkg/shared/utils/sync/distributed_lock_mgr.go
miguelsorianod: > Each thread must use a distinct `DistributedLock` to hold a separate transaction that waits for the lock’s release. Using a single `DistributedLock` object would result in an `Unlock` call unblocking all threads.

I multiple threads are locked then when unlock is called shouldn't all the other threads be unlocked?

ziccardi: No. We don't have ONE thread performing the lock and a number of threads waiting for it to be released.
We have a number of threads trying to perform the lock.
Suppose we have 3 threads TH1, TH2, TH3:
```
- instant T1: TH1, TH2 and TH3 are calling the Lock method
- instant T2: TH1 obtains the lock. TH2 and TH3 are blocked, waiting to acquire the lock
- instant T3 : TH1 releases the lock. TH2 is now able to perform the lock. TH3 now waits for TH2 to release the lock
- instant T4: TH2 releases the lock. TH3 can now acquire the lock.
```

machi1990: @ziccardi I still struggling to see why we need to track the lists. Maybe, I am missing something.

I see here that we find only one lock that's current locked: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/c2559b7bc0edcda2f98475a841a98af496dd2666/pkg/shared/utils/sync/distributed_lock_mgr.go#L56 can't we just track this instead of keeping track of all the lists?

ziccardi: Nope, because in that case when the `unlock` removes the object from the `map` it would remove the lock for all the threads that are currently waiting on that thread.
Moreover, each `DistributedLock` object holds a reference to its transaction.

machi1990 (approved):
Thanks @ziccardi

pkg/shared/utils/sync/distributed_lock_mgr.go
machi1990: Thanks @ziccardi I'll share here what's been discussed as potential future improvement;

We tried tracking only the lock that's acquired but the integration tests where failing. We aren't sure where exactly the race condition was coming from to cause the failure: was it test related or implementation related, that's what we didn't get to nailing down and we decided to defer that improvement and consider it a future improvement in a separate followup PR.

pkg/shared/utils/sync/distributed_lock_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	testLock, ok := mutexMap[lockID]
+	if !ok {
+		testLock = &sync.Mutex{}
+		mutexMap[lockID] = testLock
+	}
+	return testLock
+}
+
+type __distributedLockMock struct {
```
miguelsorianod, line 37: Why is this created and not the `distributedLockMock`?

pkg/shared/utils/sync/distributed_lock_mock.go
```diff
@@ -0,0 +1,40 @@
...
+	return dl.dl.isLocked()
+}
+
+func (dl *distributedLockMock) Lock() error {
+	return dl.dl.Lock()
+}
+
+func (dl *distributedLockMock) Unlock() error {
+	_ = dl.dl.Unlock()
+	result := dl.db.Exec("DELETE FROM distributed_locks where ID = ?", dl.id)
```
miguelsorianod, line 35: As a future PR we should try to see if we can simulate rollback, because for test implementers / reviewer this behavior will be a unexpected

pkg/shared/utils/sync/distributed_lock_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+	current                *int32
+	dest                   *int32
+	g                      *gomega.WithT
+}
+
+func getNext(p *testRoutineParams) {
+	p.g.Expect(p.distributedLockManager.Lock(p.lockID)).To(gomega.Succeed())
+
+	*p.dest = *p.current + 1
+	time.Sleep(p.delay)
```
miguelsorianod, line 79: Do you think there might be a way to perform the tests without depending on sleeping?
The sleep times end up adding between several tests so I would prefer avoiding them if possible. We can create a follow-up jira issue to investigate that if at the moment we are not sure whether it is possible or not

ziccardi: We could consider alternative testing methods. Currently, I have used the sleep function to simulate load and ensure that the routines do not finish too quickly before the next one begins.

pkg/shared/utils/sync/distributed_lock_mgr_test.go
ziccardi: Because the `distributedLockMock` is based on a mocked database, while this one based on mutex.
It made creating unit tests for the `distributed_lock_mgr` easier without the need to mock the database.

pkg/shared/utils/sync/distributed_lock_mgr_test.go
miguelsorianod: :+1: let's investigate that if possible

pkg/shared/utils/sync/distributed_lock_mgr_test.go
miguelsorianod: Could we name it inMemoryDistributedLock or something like that then?

PR #1705, opened 2023-04-14
fix: create or update in-memory storage for both Lock and Store ops

MikeEdgar:
## Description

Attempts to `Lock` a key in the in-memory Let's Encrypt/certmagic storage result in a "file not found" error when `Store` for the same key has not yet been called. This change will create an entry for the key for either the `Lock` or `Store` operations and adds logging to allow users in development to obtain the generated LE account key.

## Verification Steps

1. Run KFM with these service template parameter
   ```
   KAFKA_TLS_CERTIFICATE_MANAGEMENT_STRATEGY='true'
   KAFKA_TLS_CERTIFICATE_MANAGEMENT_STORAGE_TYPE='in-memory'
   KAFKA_TLS_CERTIFICATE_MANAGEMENT_EMAIL=your email
   ENABLE_KAFKA_EXTERNAL_CERTIFICATE='true'
   ENABLE_KAFKA_CNAME_REGISTRATION='true'
   KAFKA_DOMAIN_NAME='kafka.bf2.dev'
   ```
   and these secret template parameters
   ```
   ROUTE53_ACCESS_KEY='..dev access key..'
   ROUTE53_SECRET_ACCESS_KEY='..dev secret key..'
   ```
2. Create a new Kafka instance, confirming that the certificate is provisioned from Let's Encrypt and provided to the data plane. The generated account ID key will be logged.

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/services/kafkatlscertmgmt/memory_storage.go
+++ b/internal/kafka/internal/services/kafkatlscertmgmt/memory_storage.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
 	"github.com/caddyserver/certmagic"
 )
 
@@ -40,7 +41,11 @@ func newInMemoryStorage() *inMemoryStorage {
 func (storage *inMemoryStorage) Lock(ctx context.Context, key string) error {
 	mu, ok := storage.store[key]
 	if !ok {
-		return fs.ErrNotExist
+		mu = inMemoryStorageItem{
+			lastModified: time.Now(),
+			mu:           &sync.Mutex{},
+		}
+		storage.store[key] = mu
 	}
 
 	mu.Lock()
@@ -58,11 +63,18 @@ func (storage *inMemoryStorage) Unlock(ctx context.Context, key string) error {
 }
 
 func (storage *inMemoryStorage) Store(ctx context.Context, key string, value []byte) error {
-	storage.store[key] = inMemoryStorageItem{
-		value:        value,
-		lastModified: time.Now(),
-		mu:           &sync.Mutex{},
+	mu, ok := storage.store[key]
+	if !ok {
+		mu = inMemoryStorageItem{
+			mu: &sync.Mutex{},
+		}
 	}
+	mu.value = value
+	mu.lastModified = time.Now()
+	if strings.HasPrefix(key, "acme/") {
+		logger.Logger.Infof("storing key '%s' with value %v", key, string(value))
+	}
+	storage.store[key] = mu
 	return nil
 }
 
--- a/internal/kafka/internal/services/kafkatlscertmgmt/memory_storage.go
+++ b/internal/kafka/internal/services/kafkatlscertmgmt/memory_storage.go
@@ -4,70 +4,45 @@ import (
 	"context"
 	"io/fs"
 	"strings"
-	"sync"
 	"time"
 
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/db"
 	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/logger"
+	"github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pkg/shared/utils/sync"
 	"github.com/caddyserver/certmagic"
 )
 
 var _ certmagic.Storage = &inMemoryStorage{}
 
 type inMemoryStorageItem struct {
-	value []byte
-	mu    *sync.Mutex
-	sync.Locker
+	value        []byte
 	lastModified time.Time
-}
-
-func (item inMemoryStorageItem) Lock() {
-	item.mu.Lock()
-}
-
-func (item inMemoryStorageItem) Unlock() {
-	item.mu.Unlock()
 }
 
 type inMemoryStorage struct {
 	store map[string]inMemoryStorageItem
+	lock  sync.DistributedLockMgr
 }
 
-func newInMemoryStorage() *inMemoryStorage {
+func newInMemoryStorage(connectionFactory *db.ConnectionFactory) *inMemoryStorage {
 	return &inMemoryStorage{
 		store: map[string]inMemoryStorageItem{},
+		lock:  sync.NewDistributedLockMgr(connectionFactory.New()),
 	}
 }
 
 func (storage *inMemoryStorage) Lock(ctx context.Context, key string) error {
-	mu, ok := storage.store[key]
-	if !ok {
-		mu = inMemoryStorageItem{
-			lastModified: time.Now(),
-			mu:           &sync.Mutex{},
-		}
-		storage.store[key] = mu
-	}
-
-	mu.Lock()
-	return nil
+	return storage.lock.Lock(key)
 }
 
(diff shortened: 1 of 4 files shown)


machi1990:
Hey @MikeEdgar thanks for the PR. It is a good catch. I've few remarks:
- I think we can re-use the distributed storage implementation, just like the one we've for the secure storage https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/9d14b767dc88eebaee00867bf57f44f8ad4873e0/internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go#L64-L74 and remove any locking details to the stored item. 
- The `in-memory` storage is not meant to be used for anything other than testing; If you restart the pod/process the certs are lost and are forced to be regenerated. I'll recommend consider using the file storage or the secure-storage. 
- Not related to the PR: I think we'll need to make the CA configurable so that for testing e.g with kas-installer someone can point to LE Stage env which doesn't have stricter rate limitings as the Prod env: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/9d14b767dc88eebaee00867bf57f44f8ad4873e0/internal/kafka/internal/config/kafka_tls_certificate_management_config.go#L54 What do you think? @MikeEdgar @miguelsorianod  

Other if we take my first point onto account about using an in memory storage, frequent asking for a certificate after process has been restarted will likely cause an issue on the 6th restart since some of the limits will start to kick in.

internal/kafka/internal/services/kafkatlscertmgmt/memory_storage.go
```diff
@@ -40,7 +41,11 @@ func newInMemoryStorage() *inMemoryStorage {
 func (storage *inMemoryStorage) Lock(ctx context.Context, key string) error {
```
machi1990, line 41: my suggestion will be to re-use what we've here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/9d14b767dc88eebaee00867bf57f44f8ad4873e0/internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go#L64-L74 

and remove any locking related details from the stored item.

machi1990 (changes requested):
I've added a suggestion

miguelsorianod:
> I think we can re-use the distributed storage implementation, just like the one we've for the secure storage 

Can you elaborate what do you mean with this? In memory storage is by definition not distributed. And that's not an issue for development purposes

miguelsorianod:
> Not related to the PR: I think we'll need to make the CA configurable so that for testing e.g with kas-installer someone can point to LE Stage env which doesn't have stricter rate limitings as the Prod env: 

That seems reasonable to me

miguelsorianod:
> The point I was trying to get across is that, we already have a locking strategy (and it is distributed) we can re-use that utility, instead of implementing (or rather continue implementing and maintaining) another locking strategy. So what I am really proposing is that the in-memory and secure-storage reuse the same locking mechanism current used for secure-storage.

I am still not sure I am understanding the proposal. Local in-memory storage can't be distributed

machi1990:
> > The point I was trying to get across is that, we already have a locking strategy (and it is distributed) we can re-use that utility, instead of implementing (or rather continue implementing and maintaining) another locking strategy. So what I am really proposing is that the in-memory and secure-storage reuse the same locking mechanism current used for secure-storage.
> 
> I am still not sure I am understanding the proposal. Local in-memory storage can't be distributed

Yes, the in-memory storage can't be distributed and we should be fine by just implementing in memory locking. 
But since we've locking utility in place already, we can just re-use it: this locking strategy is distributed so it can handle the locking needs for the in-memory strategy.

miguelsorianod:
>  But since we've locking utility in place already, we can just re-use it: this locking strategy is distributed so it can handle the locking needs for the in-memory strategy.

Ok I think I understand now. So what you are proposing is to use the database-based distributed locking for the in-memory implementation of certmanager. You are not proposing doing an in-memory implementation of the locking mechanism itself. Am I understanding correctly?

machi1990:
>  So what you are proposing is to use the database-based distributed locking for the in-memory implementation of certmanager's storage backend. 

Yes. That's a correct understanding.

> You are not proposing doing an in-memory implementation of the locking mechanism itself.

This is what is done in the PR but I think we might find ourselves maintaining two locking strategies - the database backed one and the in-memory one, while the former is enough in my opinion for both certmanager's storage types that we've.

miguelsorianod:
> > So what you are proposing is to use the database-based distributed locking for the in-memory implementation of certmanager's storage backend.
> 
> Yes. That's a correct understanding.
> 
> > You are not proposing doing an in-memory implementation of the locking mechanism itself.
> 
> This is what is done in the PR but I think we might find ourselves maintaining two locking strategies - the database backed one and the in-memory one, while the former is enough in my opinion for both certmanager's storage types that we've.

:+1: . That makes sense to me. As long as KAS Installer team is aware that this means that this implies a dependency on the KFM DB for the locking and are ok with it.

machi1990:
Yeah, let's hear what @MikeEdgar thinks as well.

machi1990:
> > Not related to the PR: I think we'll need to make the CA configurable so that for testing e.g with kas-installer someone can point to LE Stage env which doesn't have stricter rate limitings as the Prod env:
> 
> That seems reasonable to me

I've created https://issues.redhat.com/browse/MGDSTRM-11141 to review the configuration and structure so that it aligns with how we want to expose configuration going forward in KFM.
 
@MikeEdgar let us know if you'll need the CA to be configurable as well and the timelines for that as the JIRA I opened above can take longer to be progressed, so can seek for an alternative quick implementation.

MikeEdgar:
> let us know if you'll need the CA to be configurable as well and the timelines for that as the JIRA I opened above can take longer to be progressed, so can seek for an alternative quick implementation.

@machi1990, this would be a nice to have feature. I've actually been using the staging CA for my testing of this change. However, what I am really trying to do is get a cert issued by the production CA for use in development and test pipelines to use with managed connectors. 

I'll take a look at the other change to use the database locking mechanism for in-memory.

machi1990:
> I'll take a look at the other change to use the database locking mechanism for in-memory.

Thanks @MikeEdgar 

The database locking locking mechanism is already used for secure-storage: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/9d14b767dc88eebaee00867bf57f44f8ad4873e0/internal/kafka/internal/services/kafkatlscertmgmt/secure_storage.go#L30

the same usage pattern can be taken for the in-memory storage.
 
Additionally, we can use the same opportunity to remove: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/9d14b767dc88eebaee00867bf57f44f8ad4873e0/internal/kafka/internal/services/kafkatlscertmgmt/memory_storage.go#L18-L19 from the stored in memory item 

> However, what I am really trying to do is get a cert issued by the production CA for use in development and test pipelines to use with managed connectors.

It makes sense. If the pipeline runs frequently, I think we'll quickly reach one of the rate limits from LE production env. We could separate this conversation from this PR and if the problem arises (once the pipelines are there), we know exactly what to do with the alternatives options we've: to which I'll add - asking for limits override (https://letsencrypt.org/docs/rate-limits/#a-id-overrides-a-overrides) being another short-term alternative even though.

machi1990 (approved):
thanks @MikeEdgar

PR #1716, opened 2023-04-18
feat: add extra time that prevents from instantly suspending out of quota kafkas

pawelpaszki:
feat: add extra time that prevents from instantly suspending out of quota kafkas

- [ ] Code Review completed

### DIFF
--- a/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
+++ b/internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
@@ -440,7 +440,7 @@ func (k *KafkaManager) updateExpiresAtBasedOnQuotaEntitlement(kafka *dbapi.Kafka
 		}
 
 		// set expires_at to now + grace period days
-		expiresAtTime := time.Now().AddDate(0, 0, billingModel.GracePeriodDays)
+		expiresAtTime := time.Now().AddDate(0, 0, billingModel.GracePeriodDays).Add(time.Hour * 12)
 		logger.Logger.Infof("quota entitlement for kafka instance %q is no longer active, updating expires_at to %q", kafka.ID, expiresAtTime.Format(time.RFC1123Z))
 		return k.updateKafkaExpirationDate(kafka, &expiresAtTime)
 	}


internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -440,7 +440,7 @@ func (k *KafkaManager) updateExpiresAtBasedOnQuotaEntitlement(kafka *dbapi.Kafka
 		}
 
 		// set expires_at to now + grace period days
```
machi1990, line 442: the comment needs to be updated as well?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
```diff
@@ -440,7 +440,7 @@ func (k *KafkaManager) updateExpiresAtBasedOnQuotaEntitlement(kafka *dbapi.Kafka
...
-		expiresAtTime := time.Now().AddDate(0, 0, billingModel.GracePeriodDays)
+		expiresAtTime := time.Now().AddDate(0, 0, billingModel.GracePeriodDays).Add(time.Hour * 12)
```
machi1990, line 443: Just a question for my understanding: How was the 12 hours period choosen? And how is it different to grace period?

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
pawelpaszki: yes. I will update it. I also need to add code to remove expiration if quota is now > 0

internal/kafka/internal/workers/kafka_mgrs/kafkas_mgr.go
pawelpaszki: it was jus an arbitrary number of hours, so kafkas won't be instantly suspended if allowed quota intermittently shows 0

machi1990: Thanks for the clarification. Might be good to extract the "magic" number onto a const/variable with a proper name and comment if possible.

pawelpaszki: makes sense. will do that

